PR #18, opened 2021-03-01
Test timer integration, add worker shutdown

bergundy:
Note that the CI does not run integration tests yet.
The tests pass on my laptop.
I have another PR (#15) which is supposed to make them run but I'm having some trouble waiting for the Temporal server to start.


packages/worker/native/src/lib.rs
```diff
@@ -146,10 +146,14 @@ fn worker_complete_task(mut cx: FunctionContext) -> JsResult<JsUndefined> {
 }
 
 fn worker_shutdown(mut cx: FunctionContext) -> JsResult<JsUndefined> {
```
Sushisource, line 148: Doesn't need to be this PR but you'll probably want to have some mechanism to drain any remaining activations from `poll_task` until it returns `CoreError::ShuttingDown`. Remaining activations may exist due to replaying a workflow to catch up to the end of history which was already fetched from the server

Sushisource (approved):
Looking good! Glad this came together as expected.

packages/worker/native/src/lib.rs
bergundy: That's a really good point (see #19).

PR #37, opened 2021-03-16
Poll on demand

### LINKED ISSUE #33: Don't buffer poll results in core-lang bridge (lib.rs)
At the moment we poll in a separate thread and push results into a queue to callback into the JS thread.
We should change this to either have a bounded blocking queue or only poll when JS is ready for the next result.

The reason for this is that buffering events could have negative effects. For example, activities are automatically considered started when a poll request returns and if the worker is busy it might not actually start them in time.

bergundy:
Closes #33


packages/test/src/test-worker-lifecycle.ts
```diff
@@ -0,0 +1,73 @@
...
+    t.is(worker.getState(), 'RUNNING');
+    process.emit('SIGINT', 'SIGINT');
+    t.is(worker.getState(), 'STOPPING');
+    await p;
+    t.is(worker.getState(), 'STOPPED');
+    await t.throwsAsync(worker.run('shutdown-test'), { message: 'Poller was aleady started' });
+  });
+}
+
+test.serial('Mocked run shuts down gracefully', async (t) => {
```
Sushisource, line 21: Why do the mocked ones need to be run serially?

packages/worker/native/src/lib.rs
```diff
@@ -13,16 +16,20 @@ use temporal_sdk_core::{
...
-    condition: Condvar,
-    suspended: Mutex<bool>,
+    sender: SyncSender<PollItem>,
 }
 
 impl Finalize for Worker {}
 
 impl Worker {
-    pub fn new() -> Self {
+    pub fn new(sender: SyncSender<PollItem>) -> Self {
```
Sushisource, line 32: I find it a little easier to work with constructors like this where they create the channel inside of themselves, and return both `Self` and the receiving half. That way it can control it's own capacity etc

packages/worker/native/src/lib.rs
```diff
@@ -13,16 +16,20 @@ use temporal_sdk_core::{
...
+pub struct PollItem {
+    queue_name: String,
+    callback: Root<JsFunction>,
+}
+
 pub struct Worker {
     core: Box<dyn Core + Send + Sync>,
-    condition: Condvar,
-    suspended: Mutex<bool>,
+    sender: SyncSender<PollItem>,
```
Sushisource, line 26: A docstring about what this is for would be nice

In fact, generally I think we could use a lot more docstrings throughout. These libs are complex and a little explanation goes a long way.

packages/worker/native/src/lib.rs
```diff
@@ -36,92 +43,82 @@ impl Worker {
...
-    Ok(cx.boxed(worker))
-}
-
-fn worker_poll(mut cx: FunctionContext) -> JsResult<JsUndefined> {
-    let worker = cx.argument::<BoxedWorker>(0)?;
-    let queue_name = cx.argument::<JsString>(1)?.value(&mut cx);
-    let callback = cx.argument::<JsFunction>(2)?.root(&mut cx);
-    let arc_worker = Arc::clone(&**worker); // deref Handle and JsBox
-    let arc_callback = Arc::new(callback);
+    // TODO: is capacity of 1 enough here?
```
Sushisource, line 56: As long as you never ask the worker to poll without handling the result of the last poll, which makes sense in a single threaded context, so keeping the bound is probably a nice check.

packages/worker/native/src/lib.rs
```diff
@@ -36,92 +43,82 @@ impl Worker {
...
+                }
             }
         }
     });
-    Ok(cx.undefined())
+
+    Ok(cx.boxed(worker))
+}
+
+fn worker_poll(mut cx: FunctionContext) -> JsResult<JsUndefined> {
```
Sushisource, line 107: The difference between `worker_poll` and `Worker::poll` isn't super clear here.

packages/worker/src/worker.ts
```diff
@@ -120,29 +142,19 @@ export interface WorkerConstructor {
 
 export class NativeWorker implements NativeWorkerLike {
   protected readonly native: native.Worker;
+  protected readonly pollFn: (worker: native.Worker, queueName: string) => Promise<ArrayBuffer>;
 
   public constructor() {
     this.native = native.newWorker();
+    this.pollFn = promisify(native.workerPoll) as any; // Typescript isn't smart enough to deduce the type here
```
Sushisource, line 149: Can you reify it to a more specific type than `any` ?

packages/worker/src/worker.ts
```diff
@@ -233,81 +246,79 @@ class BaseWorker {
...
-        }
-      });
-
-      return function unsubscribe() {
-        // NOTE: We don't expose this observable directly so we don't have to shutdown here
-        deregisterSignalHandlers();
-      };
-    });
+    return merge(
+      this.stateSubject.pipe(
```
Sushisource, line 282: I find this whole big chain of RxJS operators a bit difficult to follow... but I'm really not sure I have any suggestions on how to make it a little more clear since I've not used it. 

Maybe factoring out some of the inner pieces into their own functions so the top level piece reads a bit more clearly?

bergundy (commented):
Thanks!

packages/worker/src/worker.ts
bergundy: Ended up changing the callback type, apparently the types of callbacks in the nodejs apis do not have an optional result argument.

Sushisource (approved):
Awesome. Definitely a bit easier to follow.

PR #56, opened 2021-03-30
Use async core interface

bergundy:
## What was changed:
Use the async core interface introduced in https://github.com/temporalio/sdk-core/pull/77

Worker is now created with the static async `Worker.create` method.

## How has this been tested?
Ran the existing tests which needed some modifications.

## Any docs updates needed?
Docs were updated.

### DIFF
--- a/packages/worker/native/src/lib.rs
+++ b/packages/worker/native/src/lib.rs
@@ -176,17 +176,25 @@ fn worker_new(mut cx: FunctionContext) -> JsResult<JsUndefined> {
                                             }
                                         }
                                     }
-                                    Request::CompleteWorkflowActivation {
-                                        completion,
+                                    Request::PollActivityTask {
+                                        queue_name,
                                         callback,
                                     } => {
-                                        match core.complete_workflow_task(completion).await {
-                                            Ok(()) => {
+                                        match core.poll_activity_task(&queue_name).await {
+                                            Ok(task) => {
                                                 queue.send(move |mut cx| {
                                                     let callback = callback.into_inner(&mut cx);
                                                     let this = cx.undefined();
                                                     let error = cx.undefined();
-                                                    let result = cx.undefined();
+                                                    let len = task.encoded_len();
+                                                    let mut result =
+                                                        JsArrayBuffer::new(&mut cx, len as u32)?;
+                                                    cx.borrow_mut(&mut result, |data| {
+                                                        let mut slice = data.as_mut_slice::<u8>();
+                                                        if let Err(_) = task.encode(&mut slice) {
+                                                            panic!("Failed to encode task")
+                                                        };
+                                                    });
                                                     let args: Vec<Handle<JsValue>> =
                                                         vec![error.upcast(), result.upcast()];
                                                     callback.call(&mut cx, this, args)?;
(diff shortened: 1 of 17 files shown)


packages/worker/native/src/lib.rs
```diff
@@ -1,64 +1,124 @@
...
+        Ok(())
+    });
+}
 
-        (worker, receiver)
-    }
+/// Send an error to JS via callback using an [EventQueue]
+fn send_error<T>(queue: Arc<EventQueue>, callback: Root<JsFunction>, error: T)
+where
+    T: ::std::fmt::Display + Send + 'static,
```
Sushisource, line 85: You don't need the leading `::` for these.

packages/worker/native/src/lib.rs
```diff
@@ -81,118 +141,287 @@ fn worker_new(mut cx: FunctionContext) -> JsResult<BoxedWorker> {
...
-                if should_break {
-                    break;
+    // TODO: make this configurable
+    let (sender, mut receiver) = channel::<Request>(1000);
+    let worker = Worker {
+        sender: sender.clone(),
+    };
+    let queue = Arc::new(cx.queue());
+
+    std::thread::spawn(move || {
```
Sushisource, line 151: Would be great to break up this big closure into smaller functions

packages/worker/native/src/lib.rs
```diff
@@ -81,118 +141,287 @@ fn worker_new(mut cx: FunctionContext) -> JsResult<BoxedWorker> {
...
+                match init(CoreInitOptions { gateway_opts }).await {
+                    Ok(result) => {
+                        send_result(queue.clone(), callback, |cx| Ok(cx.boxed(worker)));
+                        let core = Arc::new(result);
+                        loop {
+                            // TODO: handle this error
+                            let request = receiver.recv().await.unwrap();
+                            if matches!(request, Request::BreakPoller) {
+                                break;
+                            } else if matches!(request, Request::Shutdown) {
```
Sushisource, line 166: This reads a bit strange compared to just a `match` expression

packages/worker/native/src/lib.rs
```diff
@@ -81,118 +141,287 @@ fn worker_new(mut cx: FunctionContext) -> JsResult<BoxedWorker> {
...
+                                        Ok(()) => {
+                                            send_result(queue, callback, |cx| Ok(cx.undefined()));
+                                        }
+                                        Err(err) => {
+                                            send_error(queue, callback, err);
+                                        }
+                                    },
+                                    // Ignore BreakPoller and Shutdown, they're handled above
+                                    Request::BreakPoller => {}
+                                    Request::Shutdown => {}
```
Sushisource, line 278: Can be `Request::BreakPoller | Request::Shutdown`

packages/worker/native/src/lib.rs
```diff
@@ -81,118 +141,287 @@ fn worker_new(mut cx: FunctionContext) -> JsResult<BoxedWorker> {
...
+    }
+    Ok(cx.undefined())
+}
+
+/// Initiate a single activity task poll request.
+/// There should be only one concurrent poll request for this type.
+fn worker_poll_activity_task(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let worker = cx.argument::<BoxedWorker>(0)?;
+    let queue_name = cx.argument::<JsString>(1)?.value(&mut cx);
+    let callback = cx.argument::<JsFunction>(2)?;
```
Sushisource, line 314: You could impl some kind of `unpack` function on `FunctionContext` that would de-dupe these preambles and make things a bit more readable

bergundy (commented):
Thanks

packages/worker/native/src/lib.rs
Sushisource: @bergundy Something like this
```rust
trait FnContextExt {
  fn unpack(self) -> (BoxedWorkerType, QueueNameType, CallbackType);
}

impl FnContextExt for FunctionContext {
    /// Implement the three lines here and return 3-tuple (or a real struct)
}
```

packages/worker/native/src/lib.rs
Sushisource: Yup, very nice. :raised_hands:

PR #71, opened 2021-04-21
Pass in max concurrent activity/workflow tasks

Sushisource:
## What was changed:
Pass through options for task limits, also update to handle task_queue being a per-instance option instead of per-poll

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?


Sushisource:
@bergundy The way I did the option passing is just what came together easiest but I'm not sure if it's a great way. Let me know if it could be more idiomatic.

bergundy (commented):
Thanks @Sushisource,
I pushed some minor fixes and I'll merge if CI passes.

PR #75, opened 2021-04-27
Implement signals

bergundy:
## What was changed:
Implemented signals

## Why?
Obviously an important feature for the SDK, it will come in handy for testing workflows and activities.

## Checklist

1. Closes issue: 

2. How was this tested:
Added relevant tests

3. Any docs updates needed?
Updated the workflow package API reference


Sushisource (commented):
LGTM -- It's actually pretty darn slick how you can just stick globals at the top of the workflow file and things work. Feels a little dirty but I guess that's JS for you.

bergundy:
The global variables is a benefit we get from running each wf in a separate isolate.

PR #76, opened 2021-04-27
Add Activity Info to context, test cancellation with signals

bergundy:
## Why?
- Activity Info was missing
- Cancellation tests are less flakey with signals than with timeouts
-
## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?
Documented the Activity Info interface

### DIFF
--- a/packages/test-workflows/src/cancel-http-request.ts
+++ b/packages/test-workflows/src/cancel-http-request.ts
@@ -1,6 +1,7 @@
 import { Context, CancellationError, cancel, sleep } from '@temporalio/workflow';
 import { CancellableHTTPRequest } from '@interfaces';
 import { cancellableFetch } from '@activities';
+import { ResolvablePromise } from './resolvable-promise';
 
 const fetch = Context.configure(cancellableFetch, {
   type: 'remote',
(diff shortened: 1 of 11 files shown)


packages/test-workflows/src/cancel-http-request.ts
```diff
@@ -1,29 +1,46 @@
...
+const activityCancelled = new ResolvablePromise<void>();
+
+const signals = {
+  activityStarted(): void {
+    activityStarted.resolve(undefined);
+  },
+  activityCancelled(): void {
+    activityCancelled.resolve(undefined);
+  },
+};
```
Sushisource, line 22: Yeah, this is cool.

packages/test-workflows/src/resolvable-promise.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+
+/**
+ * Helper for creating promises which can be manually resolved
+ * Copied from workflow/common
```
Sushisource, line 5: Why does it need to be copied?

Sushisource (approved):

packages/test-workflows/src/resolvable-promise.ts
bergundy: It's a limitation of the module loader, that module isn't injected into the workflow runtime.
I should probably add a comment.

PR #77, opened 2021-04-29
Handle errors returned from core correctly

### LINKED ISSUE #51: Handle errors returned from core correctly
There are different types of errors that might be returned from the different core methods, define a clear strategy of handling these errors.

bergundy:
## What was changed:
- Heartbeat details is now an optional parameter
- Use typed errors for errors returned from Core
- Add error handling for errors returned from Core
- Add worker State types: `DRAINING | DRAINED`
- Fix isolates not being disposed on `removeFromCache` activations
- Always delete index.node before building rust code

## Checklist

1. Closes #51 

2. How was this tested:
Added tests for the added error handling.

3. Any docs updates needed?
Only docstrings

### DIFF
--- a/packages/worker/native/src/errors.rs
+++ b/packages/worker/native/src/errors.rs
@@ -1,0 +1,92 @@
+use neon::prelude::*;
+use once_cell::sync::OnceCell;
+
+/// An unhandled error while communicating with the server, considered fatal
+pub static TRANSPORT_ERROR: OnceCell<Root<JsFunction>> = OnceCell::new();
+/// Thrown after shutdown was requested as a response to a poll function, JS should stop polling
+/// once this error is encountered
+pub static SHUTDOWN_ERROR: OnceCell<Root<JsFunction>> = OnceCell::new();
+/// Workflow did something Core did not expect, it should be immediately deleted from the cache
+pub static WORKFLOW_ERROR: OnceCell<Root<JsFunction>> = OnceCell::new();
+/// Something unexpected happened, considered fatal
+pub static UNEXPECTED_ERROR: OnceCell<Root<JsFunction>> = OnceCell::new();
+
+/// This is one of the ways to implement custom errors in neon.
+/// Taken from the answer in GitHub issues: https://github.com/neon-bindings/neon/issues/714
+pub trait CustomError {
+    fn construct<'a, C>(&self, cx: &mut C, args: Vec<Handle<JsValue>>) -> JsResult<'a, JsObject>
+    where
+        C: Context<'a>;
+
+    fn from_string<'a, C>(&self, cx: &mut C, message: String) -> JsResult<'a, JsObject>
+    where
+        C: Context<'a>;
+
+    fn from_error<'a, C, E>(&self, cx: &mut C, err: E) -> JsResult<'a, JsObject>
+    where
+        C: Context<'a>,
+        E: std::fmt::Display;
+}
+
+// Implement `CustomError` for ALL errors in a `OnceCell`. This only needs to be
+// done _once_ even if other errors are added.
+impl CustomError for OnceCell<Root<JsFunction>> {
+    fn construct<'a, C>(&self, cx: &mut C, args: Vec<Handle<JsValue>>) -> JsResult<'a, JsObject>
+    where
+        C: Context<'a>,
+    {
+        let error = self
+            .get()
+            .expect("Expected module to be initialized")
+            .to_inner(cx);
+
+        // Use `.construct` to call this as a constructor instead of a normal function
+        error.construct(cx, args)
+    }
+
+    fn from_string<'a, C>(&self, cx: &mut C, message: String) -> JsResult<'a, JsObject>
+    where
+        C: Context<'a>,
+    {
+        let args = vec![cx.string(message).upcast()];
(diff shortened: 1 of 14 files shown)


packages/worker/native/src/errors.rs
```diff
@@ -0,0 +1,99 @@
...
+        E: std::fmt::Display,
+    {
+        self.from_string(cx, format!("{}", err))
+    }
+}
+
+/// This method should be manually called _once_ from JavaScript to initialize the module
+/// It expects a single argument, an object with the various Error constructors.
+/// This is a very common pattern in Neon modules.
+pub fn register_errors(mut cx: FunctionContext) -> JsResult<JsUndefined> {
```
Sushisource, line 69: What a bunch of weird ceremony. Hopefully Neon improves here, or maybe adds a macro or something. We could do that if we end up needing much more of this, but probably we won't.

packages/worker/native/src/lib.rs
```diff
@@ -340,7 +438,7 @@ fn worker_break_loop(mut cx: FunctionContext) -> JsResult<JsUndefined> {
         callback: callback.root(&mut cx),
     };
     if let Err(err) = worker.sender.blocking_send(request) {
-        callback_with_error(&mut cx, callback, err)?;
+        callback_with_error(&mut cx, callback, |cx| UNEXPECTED_ERROR.from_error(cx, err))?;
```
Sushisource, line 441: Maybe make a convenience function `callback_with_unexpected_error` to avoid repeating this closure a bunch.

packages/worker/native/src/lib.rs
```diff
@@ -79,28 +83,30 @@ where
 }
 
 /// Send an error to JS via callback using an [EventQueue]
-fn send_error<T>(queue: Arc<EventQueue>, callback: Root<JsFunction>, error: T)
+fn send_error<E, F>(queue: Arc<EventQueue>, callback: Root<JsFunction>, error_ctor: F)
 where
-    T: Display + Send + 'static,
+    E: Object,
+    F: for<'a> FnOnce(&mut TaskContext<'a>) -> JsResult<'a, E> + Send + 'static,
```
Sushisource, line 89: Wow, you had to use the https://doc.rust-lang.org/nomicon/hrtb.html . Rust expert already :D

packages/worker/src/worker.ts
```diff
@@ -733,21 +859,37 @@ export class Worker {
...
 
-    if (this.options.taskQueue === undefined) {
-      throw new Error('Worker taskQueue not defined');
-    }
-    const queueName = this.options.taskQueue;
+  protected takeUntilState<T>(state: State): MonoTypeOperatorFunction<T> {
+    return takeUntil(this.stateSubject.pipe(filter((value) => value === state)));
+  }
+
+  protected takeUntilIdle<T>(numInFlightTasksSubject: Observable<number>): MonoTypeOperatorFunction<T> {
```
Sushisource, line 874: Says tasks here but seems to only apply to activations? Would be good to be consistently either generic or specific.

packages/worker/src/worker.ts
```diff
@@ -679,51 +721,135 @@ export class Worker {
...
+  protected workflow$(
+    workflowCompletionFeedbackSubject = new Subject<coresdk.workflow_activation.WFActivation>(),
+    numInFlightActivationsSubject = new BehaviorSubject(0),
+    numRunningWorkflowInstancesSubject = new BehaviorSubject(0)
+  ): Observable<void> {
+    if (this.options.taskQueue === undefined) {
+      throw new TypeError('Worker taskQueue not defined');
+    }
+
+    // Comsume activations from Core and the feedback subject
```
Sushisource, line 822: Consume

packages/worker/src/worker.ts
```diff
@@ -679,51 +721,135 @@ export class Worker {
...
-      return task;
+      try {
+        const buffer = await this.nativeWorker.pollWorkflowActivation();
+        const task = coresdk.workflow_activation.WFActivation.decode(new Uint8Array(buffer));
+        const { taskToken, ...rest } = task;
+        this.log.debug('Got workflow activation', { taskToken: formatTaskToken(taskToken), ...rest });
+        return task;
+      } catch (err) {
+        // Transform a Workflow error into an activation with a single removeFromCache job
+        if (err instanceof errors.WorkflowError) {
```
Sushisource, line 793: It may make sense to have a way to distinguish these evictions from "real" ones

packages/worker/src/worker.ts
```diff
@@ -733,21 +860,37 @@ export class Worker {
...
 
-    if (this.options.taskQueue === undefined) {
-      throw new Error('Worker taskQueue not defined');
-    }
-    const queueName = this.options.taskQueue;
+  protected takeUntilState<T>(state: State): MonoTypeOperatorFunction<T> {
+    return takeUntil(this.stateSubject.pipe(filter((value) => value === state)));
+  }
+
+  protected takeUntilIdle<T>(numInFlightActivitiesSubject: Observable<number>): MonoTypeOperatorFunction<T> {
```
bergundy, line 875: Oops, it was supposed to say activations not activities.
I'll fix in the next PR.

PR #83, opened 2021-05-05
Prepare for alpha release

bergundy:
- Update README
- Minor API reference changes

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -165,13 +165,13 @@ export interface WorkerOptions {
   /**
    * Maximum number of Activities to execute concurrently.
    * Adjust this to improve Worker resource consumption.
-   * @default 10
+   * @default 100
    */
   maxConcurrentActivityExecutions?: number;
   /**
    * Maximum number of Workflow tasks to execute concurrently.
    * Adjust this to improve Worker resource consumption.
-   * @default 200
+   * @default 10
    */
   maxConcurrentWorkflowTaskExecutions?: number;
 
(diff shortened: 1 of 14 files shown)


sw-yx (approved):
LGTM. lets :shipit:

packages/worker/src/worker.ts
```diff
@@ -165,7 +165,7 @@ export interface WorkerOptions {
   /**
    * Maximum number of Activities to execute concurrently.
    * Adjust this to improve Worker resource consumption.
-   * @default 200
+   * @default 10
```
Sushisource, line 168: :sunglasses:

packages/worker/src/worker.ts
```diff
@@ -176,10 +176,10 @@ export interface WorkerOptions {
   maxConcurrentWorkflowTaskExecutions?: number;
```
Sushisource, line 176: Should this still be so high? Might make sense to crank the default down here a bit too.

packages/worker/src/worker.ts
sw-yx: wait why is this 😎 i thought it would be 😢

Sushisource: It's a sardonic :sunglasses:, lol

bergundy: It bumped it up to 100, I got it wrong, it's 10 WFTs and 100 ATs.
After we merge the bundler branch I'll bump it back up to 100.

packages/worker/src/worker.ts
bergundy: See comment above, thanks for pointing this out.

PR #84, opened 2021-05-05
Avoid needing git submodules by just using git ref in cargo.

Sushisource:
## What was changed:
Remove core submodule, use cargo refs instead

## Why?
Submodules are not fun

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?


vitarb (approved):

Sushisource:
Ah, crap, the protobufs present a bit of an issue here.

Sushisource:
Yeah, nevermind. Compiling the JS protos is unfortunately just too much of a hassle if we only use cargo.

bergundy:
There are tradeoffs for using cargo vs. a submodule.
For local development a submodule is more convenient IMO because it's easy to make changes to core or checkout some random branch.
For distribution and CI cargo is preferable because it shrinks the bundle side and simplifies CI configuration.
At this stage I'd rather keep the submodule.

Sushisource:
> There are tradeoffs for using cargo vs. a submodule.
> For local development a submodule is more convenient IMO because it's easy to make changes to core or checkout some random branch.
> For distribution and CI cargo is preferable because it shrinks the bundle side and simplifies CI configuration.
> At this stage I'd rather keep the submodule.

The protobuf thing is a real issue, but keep in mind with cargo you can always just use the `path` option to point it at your local version, that's what I do and is even easier than a submodule.

bergundy:
That's what I did initially, I just had `path` point to a separate clone of core.
If you figure out the protobuf issue let's try using cargo and see if it's convenient.

Issue #78, opened 2021-04-30
[Bug] Path aliases easily confused with package scopes


joebowbeer:
**Describe the bug**
The use of `@` aliases in the auto-generated packages is unnecessarily confusing and may increase the chance of unintended aliasing with 3rd party packages:

```
      paths: {
        '@workflows': ['../workflows'],
        '@workflows/*': ['../workflows/*'],
        '@activities': ['../activities'],
        '@activities/*': ['../activities/*'],
        '@interfaces': ['../interfaces'],
        '@interfaces/*': ['../interfaces/*'],
      },
```

Source: https://github.com/temporalio/sdk-node/blob/main/packages/create-project/src/index.ts#L128

For many developers and code reviewers, the `@` prefix indicates a package scope, such as `@temporalio` and not an alias prefix.

At the least, I suggest changing to a different prefix that won't be confused with similarly-spelled package scopes. 

However, because of the many problems and surprises with path aliases (see below), my advice is to _not_ add them to auto-generated code. Rather, let the devs add them at their discretion.

**Additional context**
In my experience, Typescript's path aliases have been a continual source of problems. I do use them on large projects, but I've made certain compromises in doing so, and always with the fear that there is yet another issue to be uncovered.

For example:

https://mitchellsimoens.com/2019/08/07/why-typescript-paths-failed-me/

bergundy:
Hi,

I'm totally fine with changing the prefix from `@`, do you have a suggestion for a better one?
I've only seen `@` used for aliases before and FWIU that's what is commonly used.

The article has some really good points but they don't apply to us since we only use the aliases for type information meaning that the aliases are omitted from the generated code.
There's one thing I would change in the in package initializer and that's not to add aliases to the worker package.
The only aliases we should have is `@activities` and `@interfaces` in the workflow project and `@interfaces` in the activities package.
WDYT?

joebowbeer:
`~` is another common alias character. I'd prefer it to `@`.

The problems I have path aliases stem from the fact that they are designed for loaders that allow remapping, not for rewriting paths:

https://github.com/microsoft/TypeScript/issues/10866#issuecomment-246789510

bergundy:
I'm fine with changing to `~` (even though it's also an alias to a user's home directory).

We have a custom module loader for the isolate runtime so that qualifies as a valid use case for path aliases.
[...]

PR #86, opened 2021-05-06
Use webpack to bundle Workflow code and load it into a startup snapshot

bergundy:
## What was changed:
- Use webpack to bundle Workflow code and load it into a startup snapshot
- Use commonjs instead of ES modules everywhere
- Get rid of worker type aliases, they were broken (See #78, in the future I'll remove the interfaces alias and change the prefix character `@`) 
- Increase the default concurrent WF activation limit to 100

## Why?
- Increase WF activation processing time twentyfold
- Better support for loading external packages
- Simplify the project structure by removing ES module build target

## Checklist

1. Closes issue: 

2. How was this tested:
Used the standard test suite

3. Any docs updates needed?
I'll open a PR to update the docs site.

### Before
before

### After
after


packages/worker/src/isolate-builder.ts
```diff
@@ -0,0 +1,151 @@
+import path from 'path';
+import os from 'os';
+import fs from 'fs-extra';
+import dedent from 'dedent';
+import ivm from 'isolated-vm';
+import webpack from 'webpack';
+import { ActivityOptions, validateActivityOptions } from '@temporalio/workflow';
+import { Logger } from './logger';
+
+export class WorkflowIsolateBuilder {
```
Sushisource, line 10: Docstrings would be great -- can be in another PR tho

packages/worker/src/isolate-builder.ts
```diff
@@ -0,0 +1,151 @@
...
+      const internals = require('@temporalio/workflow/lib/internals');
+      internals.state.activityDefaults = ${JSON.stringify(this.activityDefaults)};
+
+      const workflows = {};
+      export { init, internals, workflows };
+    `;
+    for (const wf of workflows) {
+      code += `workflows[${JSON.stringify(wf)}] = require(${JSON.stringify(path.join(this.workflowsPath, wf))});\n`;
+    }
+    await fs.writeFile(target, code);
```
Sushisource, line 55: Might be nice to avoid writing files to disk, but not sure if you can.

https://webpack.js.org/configuration/entry-context/#entry

Is unclear if you can pass file contents or some already-in-memory thing or if it always has to be on disk

packages/worker/src/isolate-builder.ts
```diff
@@ -0,0 +1,151 @@
...
+      )
+    );
+  }
+
+  public async registerActivities(prebuildDir: string): Promise<void> {
+    validateActivityOptions(this.activityDefaults);
+    const serializedOptions = JSON.stringify(this.activityDefaults);
+    for (const [specifier, module] of this.activities.entries()) {
+      let code = dedent`
+        import { scheduleActivity } from '@temporalio/workflow';
```
Sushisource, line 99: Not super important, but maybe worth considering using https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#user-content-creating-and-printing-a-typescript-ast instead of constructing JS by creating strings. A more type safe way to codegen.

packages/worker/src/isolate-builder.ts
```diff
@@ -0,0 +1,151 @@
...
+            ${k}.type = ${type};
+            ${k}.options = ${serializedOptions};
+          `;
+        }
+      }
+      const modulePath = path.resolve(prebuildDir, `${specifier}.js`);
+      try {
+        await fs.mkdir(path.dirname(modulePath), { recursive: true });
+      } catch (err) {
+        if (err.code !== 'EEXIST') throw err;
```
Sushisource, line 120: This might be platform sensitive. Alternatively, is there a constant that can be imported for this code? I can't even find docs about how the error is supposed to work...https://nodejs.org/api/fs.html#fs_fspromises_mkdir_path_options

Sushisource (approved):
This is awesome!

bergundy (commented):
Thanks @Sushisource

PR #97, opened 2021-05-13
TLS fixes

bergundy:
## What
See commit messages
## Why?
- Uniformity with other SDKs
- Some recent features were broken


packages/client/src/index.ts
```diff
@@ -359,11 +362,27 @@ export function compileWorkflowOptions({
 }
 
 /**
- * Convert {@link ConnectionOptions.tls} to {@link grpc.ChannelCredentials} and
- * add the grpc.ssl_target_name_override GRPC {@link ConnectionOptions.channelArgs}
+ * Normalize {@link ConnectionOptions.tls} by turning false and null to undefined and true to and empty object
+ * NOTE: this function is duplicated in `packages/worker/src/worker.ts` for lack of a shared library
+ */
+function normalizeTlsConfig(tls?: ConnectionOptions['tls']): TLSConfig | undefined {
+  return typeof tls === 'object' ? (tls === null ? undefined : tls) : tls ? {} : undefined;
```
Sushisource, line 369: This is so painfully javascript :laughing:

packages/worker/native/src/lib.rs
```diff
@@ -442,46 +439,47 @@ fn worker_new(mut cx: FunctionContext) -> JsResult<JsUndefined> {
         target_url: Url::parse(&url).unwrap(),
         namespace: server_options
             .get(&mut cx, "namespace")?
-            .downcast_or_throw::<JsString, FunctionContext>(&mut cx)?
+            .downcast_or_throw::<JsString, _>(&mut cx)?
             .value(&mut cx),
```
Sushisource, line 443: This 3 line preamble is a bit visually noisy. What do you think about replacing it with this (written off my head, not guaranteed to compile :smile: )

```rust
macro_rules! js_value {
    ($js_obj:ident, $js_cx:ident, $prop_name:expr, $js_type:ty) => {
        $js_obj
            .get(&mut $js_cx, $prop_name)?
            .downcast_or_throw::<$js_type, _>(&mut $js_cx)?
            .value(&mut $js_cx)
    }
}
```

Then you can just say `js_value!(server_options, cx, "namespace", JsString)`

packages/worker/native/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// NOTE: this interface is duplicated in the client module `packages/client/src/index.ts`
+// NOTE: this interface is duplicated in the client module `packages/client/src/index.ts` for lack of a shared library
```
Sushisource, line 1: Any reason not to just make a lower dep they can share?

bergundy: I will at some point, I just didn't want to add to this PR.

packages/worker/native/src/lib.rs
bergundy: I'll leave that for a subsequent PR, good idea, thanks.

packages/client/src/index.ts
bergundy: Yup ..

bergundy (commented):
Thanks for the input @Sushisource.

bergundy:
@Sushisource I've pushed 2 new commits, we're now testing mTLS against a Temporal Cloud deployment.
Note that I've handled an error in the bridge loop where there once was a TODO, I've reached it when starting a worker without polling and waiting for the process to exit (the worker was dropped while the loop was running).

PR #105, opened 2021-05-20
fix: Workflow overrides do not take effect before starting a Workflow

bergundy:
Improve, document and cleanup Workflow runtime

I will document that `Math.random()` and `new Date()` should not be used at the WF top level.


bergundy:
There seems to be an issue with the `WeakRef` mock being overridden when we create an isolate from snapshot in node 14, added a workaround.
Reproduced with:
```js
const ivm = require('isolated-vm');
const code = `
globalThis.WeakRef = function() { throw new Error('you should see this error message') }
`;
const snapshot = ivm.Isolate.createSnapshot([{ code }]);
const isolate = new ivm.Isolate({ snapshot });
const context = isolate.createContextSync();
context.evalSync('new WeakRef()');
```

bergundy:
@Sushisource I'm merging because I have another PR that changes the SDK directory structure and I'd rather avoid the extra work to merge everything.

PR #106, opened 2021-05-20
Remove `@interfaces` path alias and restructure internal packages

### LINKED ISSUE #78: [Bug] Path aliases easily confused with package scopes

bergundy:
I will update the docs about the alias removal after this is merged.

Closes #78


bergundy:
@Sushisource I'm merging this too, feel free to leave comments and I'll submit a new PR to address them.

PR #112, opened 2021-05-26
fix(worker): allow creating worker with zero activities

vkarpov15:
## What was changed

A couple errors happen if you don't have an `src/activities` directory. First, you get `Error: ENOENT: no such file or directory, scandir '/app/src/activities'` because `readdir()` throws. Once that's fixed, you still get a `Error: ENOENT: no such file or directory, open '/src/main.js'` if you have an empty activities map because `registerActivities()` creates the directory recursively, but `genEntrypoint()` does not.

## Why?

Both of these errors are a bit hard to grok, especially the `src/main.js` one. Either way, should be able to create a project with no activities.

## Checklist

1. Closes issue: 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/worker/src/isolate-builder.ts
+++ b/packages/worker/src/isolate-builder.ts
@@ -38,6 +38,7 @@ export class WorkflowIsolateBuilder {
     ufs.use(memfs.createFsFromVolume(vol) as any).use(realFS);
     const distDir = '/dist';
     const sourceDir = '/src';
+
     await this.registerActivities(vol, sourceDir);
     const entrypointPath = path.join(sourceDir, 'main.js');
     const workflows = await this.findWorkflows();
@@ -73,6 +74,11 @@ export class WorkflowIsolateBuilder {
     `;
     for (const wf of workflows) {
       code += `workflows[${JSON.stringify(wf)}] = require(${JSON.stringify(path.join(this.workflowsPath, wf))});\n`;
+    }
+    try {
+      vol.mkdirSync(path.dirname(target), { recursive: true });
+    } catch (err) {
+      if (err.code !== 'EEXIST') throw err;
     }
     vol.writeFileSync(target, code);
   }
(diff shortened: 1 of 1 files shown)


packages/worker/src/isolate-builder.ts
```diff
@@ -169,7 +175,12 @@ export class WorkflowIsolateBuilder {
     activitiesPath: string
   ): Promise<Map<string, Record<string, (...args: any[]) => any>>> {
     const resolvedActivities = new Map<string, Record<string, (...args: any[]) => any>>();
-    const files = await fs.readdir(activitiesPath, { encoding: 'utf8' });
+    let files = null;
```
bergundy, line 178:
```suggestion
    let files string[] | undefined = undefined;
```

packages/worker/src/isolate-builder.ts
```diff
@@ -169,7 +175,12 @@ export class WorkflowIsolateBuilder {
...
+    try {
+      files = await fs.readdir(activitiesPath, { encoding: 'utf8' });
+    } catch (err) {
+      return resolvedActivities;
```
bergundy, line 182: We should throw unless the error code is `ENOENT`.

bergundy (commented):
Thanks!

You should run `npm run format` to fix the [formatting](https://github.com/temporalio/sdk-node/pull/112/checks?check_run_id=2678472124#step:8:9) and `npx commitlint --from origin/main` to fix your [commit message](https://github.com/temporalio/sdk-node/pull/112/checks?check_run_id=2678472071#step:4:12).

vkarpov15:
@bergundy should be fixed. It would be nice to have a guide to your commit convention, there isn't really much info on subject-case other than this: https://github.com/conventional-changelog/commitlint/blob/master/docs/reference-rules.md#subject-case.

bergundy (approved):
I'll add more info about the commit format in contributing.md

PR #114, opened 2021-06-01
chore(core): Update core in prep for sticky execution

bergundy:
## What was changed

Updated the core submodule and made some of the preparations for sticky execution.

**NOTE**: We still do not support sharing a single core between multiple workers, I'll work on that in one of the next PRs.

## Why?
Keep up with upstream core changes

## Checklist

2. How was this tested:
Standard test suite

3. Any docs updates needed?
Not ATM

### DIFF
--- a/packages/worker/native/src/lib.rs
+++ b/packages/worker/native/src/lib.rs
@@ -8,7 +8,7 @@ use temporal_sdk_core::{
     init, protos::coresdk::workflow_completion::WfActivationCompletion,
     protos::coresdk::ActivityHeartbeat, protos::coresdk::ActivityTaskCompletion, tracing_init,
     ClientTlsConfig, CompleteActivityError, CompleteWfError, Core, CoreInitError, CoreInitOptions,
-    PollActivityError, PollWfError, ServerGatewayOptions, TlsConfig, Url,
+    PollActivityError, PollWfError, ServerGatewayOptions, TlsConfig, Url, WorkerConfig,
 };
 use tokio::sync::mpsc::{channel, Sender};
 
@@ -24,8 +24,17 @@ pub enum Request {
         /// Used to send the result back into JS
         callback: Root<JsFunction>,
     },
+    /// A request to register a new Worker against Core
+    RegisterWorker {
+        /// Worker configuration e.g. limits and task queue
+        config: WorkerConfig,
+        /// Used to send the result back into JS
+        callback: Root<JsFunction>,
+    },
     /// A request to poll for workflow activations
     PollWorkflowActivation {
+        /// Name of queue to poll on
+        queue_name: String,
         /// Used to send the result back into JS
         callback: Root<JsFunction>,
     },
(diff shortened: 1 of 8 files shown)


bergundy:
I'm merging this, it's a very mechanical PR, basically just adjusting the Node SDK to the recent changes to Core.

@Sushisource I'd appreciate if you took a look when you get the chance, I'll address your feedback in a future PR.

packages/worker/src/worker.ts
```diff
@@ -365,8 +365,8 @@ export class NativeWorker implements NativeWorkerLike {
   public readonly shutdown: Promisify<OmitFirstParam<typeof native.workerShutdown>>;
 
   public static async create(options: CompiledWorkerOptionsWithDefaults): Promise<NativeWorkerLike> {
-    const nativeWorker = await promisify(native.newWorker)({
-      ...options,
+    const core = await promisify(native.newCore)({
+      maxCachedWorkflows: 0,
```
Sushisource, line 369: To confirm: This will be the shared core later on, right? And that's where the max cached numbers will get set to something?

packages/worker/native/src/lib.rs
```diff
@@ -8,7 +8,7 @@ use temporal_sdk_core::{
     init, protos::coresdk::workflow_completion::WfActivationCompletion,
     protos::coresdk::ActivityHeartbeat, protos::coresdk::ActivityTaskCompletion, tracing_init,
     ClientTlsConfig, CompleteActivityError, CompleteWfError, Core, CoreInitError, CoreInitOptions,
-    PollActivityError, PollWfError, ServerGatewayOptions, TlsConfig, Url,
+    PollActivityError, PollWfError, ServerGatewayOptions, TlsConfig, Url, WorkerConfig,
```
Sushisource, line 11: I just your pipeline doesn't run clippy which would be a nice thing to do. Probably there's a handful of suggestions in here that'd be useful. You can check out the buildkite definition in Core for how we run clippy in pipelines.

packages/worker/native/src/lib.rs
```diff
@@ -402,12 +458,12 @@ macro_rules! js_value_getter {
...
 
-/// Create a new worker asynchronously.
+/// Create a new core instance asynchronously.
 /// Immediately spawns a poller thread that will block on [Request]s
-/// Worker is returned to JS using supplied callback
-fn worker_new(mut cx: FunctionContext) -> JsResult<JsUndefined> {
-    let worker_options = cx.argument::<JsObject>(0)?;
-    let server_options = worker_options
+/// Core is returned to JS using supplied callback
+fn core_new(mut cx: FunctionContext) -> JsResult<JsUndefined> {
```
Sushisource, line 464: Once you have shared core working properly, I would change this function to use `lazy_static!` or similar so that it is guaranteed there is only ever one core instance -- then this can be called `get_or_instantiate_core` or something.

packages/worker/src/worker.ts
bergundy: Yes, this will come in a later PR

packages/worker/native/src/lib.rs
bergundy: Sure, I'll open an issue.

packages/worker/native/src/lib.rs
bergundy: I don't remember what we said about multiple cores in a single process, why can't we have those?

bergundy (commented):
Thanks

PR #118, opened 2021-06-03
feat: Support injecting external dependencies into the Workflow isolate

### LINKED ISSUE #53: Figure out how to propagate logs out of the workflow
Currently we inject `console.log` into the workflow but this isn't optimal, as the workflow internals need logging capabilities as well as user workflow code.

Come up with a design for this.

### LINKED ISSUE #108: [Feature Request] Limit Workflow activation time

If a Workflow gets stuck for example by going into an infinite loop, it might cause the Worker to become unresponsive eating up precious CPU and occupying a thread in the isolate thread pool.

Use a configurable [`timeout`](https://github.com/laverdet/isolated-vm#referenceapplyreceiver-arguments-options-promise) to limit the WF activation time. 

**Additional context**

https://github.com/temporalio/proposals/pull/22

bergundy:
Implements proposal: https://github.com/temporalio/proposals/blob/acbe448d9325bb3914c91e693019c66f21646d07/node/logging-and-metrics-for-user-code.md

1. Closes issues:
Closes #53
Closes #108

2. How was this tested:
Added relevant tests

3. Any docs updates needed?
Will update the docs site once before this is released.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -8,6 +8,7 @@ module.exports = {
       ['activity', 'bench', 'client', 'docs', 'core', 'release', 'samples', 'worker', 'workflow'],
     ],
     'header-max-length': [2, 'always', 120],
+    'body-max-line-length': [1, 'always', 120],
     'footer-max-line-length': [2, 'always', 120],
   },
 };
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -8,7 +8,7 @@ module.exports = {
       ['activity', 'bench', 'client', 'docs', 'core', 'release', 'samples', 'worker', 'workflow'],
     ],
     'header-max-length': [2, 'always', 120],
-    'body-max-line-length': [1, 'always', 120],
+    'body-max-line-length': [1, 'always', 100],
     'footer-max-line-length': [2, 'always', 120],
   },
 };
--- a/packages/worker/src/dependencies.ts
+++ b/packages/worker/src/dependencies.ts
@@ -1,16 +1,35 @@
+/**
+ * Type definitions for the Worker end of the external dependencies mechanism.
+ *
+ * External dependencies are functions injected into a Workflow isolate from the main NodeJS isolate.
+ * They are an advanced feature and should be used with care.
+ *
+ * @see proposal at https://github.com/temporalio/proposals/blob/master/node/logging-and-metrics-for-user-code.md
+ *
+ * @module
+ */
+
 import ivm from 'isolated-vm';
-import { ApplyMode, Dependencies, Dependency, DependencyFunction, WorkflowInfo } from '@temporalio/workflow';
+import {
+  ApplyMode,
+  ExternalDependencies,
+  ExternalDependency,
+  ExternalDependencyFunction,
+  WorkflowInfo,
+} from '@temporalio/workflow';
 
 export { ApplyMode };
 
 /** Return T if T is not a Promise otherwise extracts the type wrapped in T */
 export type StripPromise<T> = T extends Promise<infer I> ? I : T;
 
+/** Wraps a function's parameters to isolated VM references */
 type RefParameters<T extends (...args: any) => any> = T extends (...args: infer P) => any
   ? { [I in keyof P]: ivm.Reference<P[I]> }
   : never;
 
-export type InjectedDependencyFunctionIvmVariant<F extends DependencyFunction, A extends ApplyMode, R> =
+/** Synchronous variant of injected dependency function - uses isolated-vm to inject into the isolate */
+export type InjectedDependencyFunctionIvmVariant<F extends ExternalDependencyFunction, A extends ApplyMode, R> =
   | {
       arguments: 'copy';
       fn(info: WorkflowInfo, ...args: Parameters<F>): R;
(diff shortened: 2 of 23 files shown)


commitlint.config.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
       ['activity', 'bench', 'client', 'docs', 'core', 'release', 'samples', 'worker', 'workflow'],
     ],
     'header-max-length': [2, 'always', 120],
+    'body-max-line-length': [1, 'always', 120],
```
Sushisource, line 11: Any chance you'd be willing to make this 100? Lines up with some other defaults and makes doing two-up views work nicely. Not a big deal.

packages/workflow/src/interfaces.ts
```diff
@@ -137,3 +137,94 @@ export interface Workflow {
...
+  /**
+   * Injected function is called synchronously, implementation must return a promise.
+   * Injection is done using an `isolated-vm` reference, function called with `applySyncPromise`.
+   * This is the safest sync `ApplyMode` because it can not break Workflow code determinism.
+   */
+  SYNC_PROMISE = 'applySyncPromise',
+  /**
+   * Injected function is called in the background not blocking the isolate.
+   * Implementation can be either synchronous or asynchronous.
+   * Injection is done using an `isolated-vm` reference, function called with `applyIgnored`.
```
Sushisource, line 227: Should probably be mentioned in these docs that sync calls are expected to be deterministic and fast, or alternatively link to somewhere with that explanation / that they should obey normal workflow code rules.

packages/worker/src/dependencies.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Takes a {@link DependencyFunction} and turns it into a type safe specification consisting of the function implementation type and call configuration.
+ *
+ * `InjectedDependencyFunction` consists of these attributes:
+ *
+ * - `fn` - type of the implementation function for dependency `F`
+ * - `applyMode` - defines how `fn` is called from the Workflow isolate (@see {@link ApplyMode})
+ * - `callDuringReplay` - whether or not a `fn` will be called during Workflow replay - defaults to `false`
+ * - `arguments` - configure how arguments are transferred between isolates
+ *   - only relevant to `SYNC_*` {@link ApplyMode}s
```
Sushisource, line 36: Ideally we would only enforce that `arguments` is only specified in sync call modes (and isn't needed at all if not) in the type system, but not a big deal if that really uglies things up

packages/worker/src/worker.ts
```diff
@@ -314,7 +374,7 @@ export type State =
 
 type ExtractToPromise<T> = T extends (err: any, result: infer R) => void ? Promise<R> : never;
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Sushisource, line 376: Can this lint disable be removed now?

packages/worker/src/worker.ts
```diff
@@ -768,16 +831,59 @@ export class Worker {
...
+                                this.handleExternalDependencyError(createdWF.info, impl.applyMode, error);
+                              }
+                            },
+                            impl.applyMode,
+                            getIvmTransferOptions(impl)
+                          );
+                        }
+                      }
+                    }
+                  });
```
Sushisource, line 886: Good candidate to separate into its own function

Sushisource (commented):
This looks good to me, but it's pretty complex. It'd be really helpful to have a big block comment at the top of wherever is a appropriate that just describes what's going on at a high level to make this work and why it's needed, etc. No need to be super long but just to provide some context.

commitlint.config.js
bergundy: Made it a warning (`1`) because I needed to fit an extra long URL in the body: https://github.com/temporalio/sdk-node/pull/118/commits/6f8f168205443e9e5214e68820864fe9c9b615ca

packages/worker/src/dependencies.ts
bergundy: Yes, this is only applicable in `SYNC_*` modes, I'll fix the doc.

This type is quite complex so it's hard to understand what it does from reading it.

packages/worker/src/worker.ts
bergundy: Should be, thanks.

packages/worker/src/worker.ts
bergundy: Yeah, this got longer than I expected, will extract.

packages/workflow/src/interfaces.ts
bergundy: Will extend the docs, thanks.

bergundy (commented):
Thanks!

I'll add more explanations.
It is indeed one of the more complex features we've had to build in order to workaround isolate limitations.

packages/workflow/src/interfaces.ts
bergundy: Done.
There'll be more detailed docs about this in our docs site too.

bergundy (commented):
I've added more docs, and I plan to document this on the docs site as well.

PR #121, opened 2021-06-09
fix(workflow): Report filename when reporting workflow not found

vkarpov15:
## What was changed

"Workflow not found" doesn't report the name of the workflow that wasn't found. I added that to to the error message.

## Why?

Ran into this error today and figured it would save a step for developers that run into this error in the future

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
Thanks, that's really helpful!

bergundy:
Feel free to merge once you pass the tests.

bergundy:
Apparently the tests are a bit flaky following the interceptors PR, I'll make sure to fix that.

PR #122, opened 2021-06-09
Interceptors

bergundy:
Implements proposal [#25](https://github.com/temporalio/proposals/blob/79357b9268dccccf1f8031f36e1714d4f07c9875/node/interceptors.md)

I'll add a page in the documentation site.

In an upcoming PR I'll submit the opentelemetry interceptor package I'm working on.

### DIFF
--- a/packages/worker/src/interceptors.ts
+++ b/packages/worker/src/interceptors.ts
@@ -43,7 +43,9 @@ export interface WorkerInterceptors {
    */
   activityInbound?: ActivityInboundCallsInterceptorFactory[];
   /**
-   * List of modules to search for interceptors in
+   * List of modules to search for Workflow interceptors in {@link WorkerOptions.workflowsPath}
+   * - Modules should export an `interceptors` variable of type {@link WorkflowInterceptors}
+   * - Workflow interceptors run in in the Workflow isolate
    */
   workflowModules?: string[];
 }
--- a/packages/worker/src/interceptors.ts
+++ b/packages/worker/src/interceptors.ts
@@ -1,0 +1,49 @@
+/**
+ * Definitions for Activity interceptors.
+ *
+ * (The Worker also accepts Workflow interceptors but those are passed as module names)
+ *
+ * @module
+ */
+
+import { Next, Headers } from '@temporalio/workflow';
+import { Context as ActivityContext } from '@temporalio/activity';
+
+/** Input for ActivityInboundCallsInterceptor.execute */
+export interface ActivityExecuteInput {
+  readonly args: unknown[];
+  readonly headers: Headers;
+}
+
+/**
+ * Implement any of these methods to intercept Activity inbound calls
+ */
+export interface ActivityInboundCallsInterceptor {
+  /**
+   * Called when Activity function is executed
+   *
+   * @return result of Activity function
+   */
+  execute?: (input: ActivityExecuteInput, next: Next<ActivityInboundCallsInterceptor, 'execute'>) => Promise<unknown>;
+}
+
+/**
+ * A function that takes Activity Context and returns an interceptor
+ */
+export interface ActivityInboundCallsInterceptorFactory {
+  (ctx: ActivityContext): ActivityInboundCallsInterceptor;
+}
+
+/**
+ * Structure for passing in Worker interceptors via {@link WorkerOptions}
+ */
+export interface WorkerInterceptors {
+  /**
+   * List of factory functions returning {@link ActivityInboundCallsInterceptor}s
+   */
+  activityInbound?: ActivityInboundCallsInterceptorFactory[];
+  /**
+   * List of modules to search for interceptors in
+   */
+  workflowModules?: string[];
+}
--- a/packages/worker/src/isolate-builder.ts
+++ b/packages/worker/src/isolate-builder.ts
@@ -25,7 +25,8 @@ export class WorkflowIsolateBuilder {
     public readonly nodeModulesPath: string,
     public readonly workflowsPath: string,
     public readonly activities: Map<string, Record<string, any>>,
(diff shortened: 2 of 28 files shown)


packages/workflow/src/interceptors.ts
```diff
@@ -0,0 +1,130 @@
...
+ * @param method the name of the interceptor method to compose
+ * @param next the original function to be executed at the end of the interception chain
+ */
+export function composeInterceptors<I, M extends keyof I>(interceptors: I[], method: M, next: Next<I, M>): Next<I, M> {
+  for (let i = interceptors.length - 1; i >= 0; --i) {
+    const interceptor = interceptors[i];
+    if (interceptor[method] !== undefined) {
+      const prev = next;
+      // We loose type safety here because Typescript can't deduce that interceptor[method] is of type Next<I, M>
+      next = ((input: any) => (interceptor[method] as any)(input, prev)) as any;
```
vkarpov15, line 32: Doesn't look like there's a way for an interceptor to report an error, like throwing a validation error if activity params are invalid for example. Is that something worth supporting?

vkarpov15 (approved):
This is pretty cool :+1:

packages/worker/src/interceptors.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Context as ActivityContext } from '@temporalio/activity';
+
+/** Input for ActivityInboundCallsInterceptor.execute */
+export interface ActivityExecuteInput {
+  readonly args: unknown[];
+  readonly headers: Headers;
+}
+
+/**
+ * Implement any of these methods to intercept Activity inbound calls
```
Sushisource, line 19: I don't think it's immediately clear what `inbound` means. Arguably this whole thing could just be `ActivityCallsInterceptor` since it's not  clear to me if/why there would ever be an `Outbound` half.

But, even without that, would be good to elaborate slightly here and make it clear we're talking about calls that run activities (or, not sure what else later, but interact with existing activities maybe?)

packages/worker/src/isolate-builder.ts
```diff
@@ -66,15 +67,23 @@ export class WorkflowIsolateBuilder {
     let code = dedent`
       const init = require('@temporalio/workflow/lib/init');
       const internals = require('@temporalio/workflow/lib/internals');
-      init.overrideGlobals();
       internals.state.activityDefaults = ${JSON.stringify(this.activityDefaults)};
-
+      init.overrideGlobals();
```
Sushisource, line 71: Still not loving that this code is just a string. Seems like other stuff might keep getting added here and it may be worth finding a different way.

packages/worker/src/interceptors.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export interface WorkerInterceptors {
+  /**
+   * List of factory functions returning {@link ActivityInboundCallsInterceptor}s
+   */
+  activityInbound?: ActivityInboundCallsInterceptorFactory[];
+  /**
+   * List of modules to search for interceptors in
+   */
+  workflowModules?: string[];
```
Sushisource, line 48: This is b/c interceptors are defined alongside workflows correct? Maybe worth mentioning that here

packages/workflow/src/interceptors.ts
```diff
@@ -0,0 +1,127 @@
...
+ * @param interceptors a list of interceptors
+ * @param method the name of the interceptor method to compose
+ * @param next the original function to be executed at the end of the interception chain
+ */
+export function composeInterceptors<I, M extends keyof I>(interceptors: I[], method: M, next: Next<I, M>): Next<I, M> {
+  for (let i = interceptors.length - 1; i >= 0; --i) {
+    const interceptor = interceptors[i];
+    if (interceptor[method] !== undefined) {
+      const prev = next;
+      // We loose type safety here because Typescript can't deduce that interceptor[method] is of type Next<I, M>
```
Sushisource, line 31: Can you make the `I` generic param more specific to fix this?

packages/workflow/src/interceptors.ts
```diff
@@ -0,0 +1,127 @@
...
+ * The Workflow specific interceptors are defined here.
+ *
+ * @module
+ */
+
+import { coresdk } from '@temporalio/proto';
+import { ActivityOptions } from './interfaces';
+import { AnyFunc, OmitLastParam } from './type-helpers';
+
+/** Type of the next function for a given interceptor function */
```
Sushisource, line 13: Probably worth saying what a next function is for as well here

packages/workflow/src/workflow.ts
```diff
@@ -74,12 +84,51 @@ export function validateActivityOptions(options: ActivityOptions): asserts optio
 }
 
 /**
+ * Push a scheduleActivity command into state accumulator and register completion
+ */
+function scheduleActivityNextHandler({ options, args, headers, seq, activityType }: ActivityInput): Promise<unknown> {
+  // let next: Next<WorkflowOutboundCallsInterceptor, 'scheduleActivity'> = (input) =>{
```
Sushisource, line 90: Leftover comment?

packages/workflow/src/interceptors.ts
bergundy: I can't find a worthwhile way to fix this

packages/workflow/src/interceptors.ts
bergundy: You can throw an error and fail the interception

packages/workflow/src/workflow.ts
bergundy: Thanks

bergundy (commented):
Thanks, addressed the comments

bergundy:
Looks like the test namespace we're using in the cloud is broken and failing the tests.
Force merged.

PR #124, opened 2021-06-14
Cancellation v2

### LINKED ISSUE #6: Revisit cancellation and scopes design
The design in https://github.com/temporalio/sdk-node/pull/5 was not agreed upon, we should revisit after the project restructuring.

bergundy:
- Implement proposal https://github.com/temporalio/proposals/blob/master/node/cancellation-v2.md
- Fix flaky interceptors test
- Fix loop in docusaurus watch mode

I'll update the documentation site once this is merged.

Closes #6


packages/test/src/workflows/cancel-requested-with-non-cancellable.ts
```diff
@@ -0,0 +1,20 @@
+// @@@SNIPSTART nodejs-cancel-requested-with-non-cancellable
```
Sushisource, line 1: I think these workflows which are part of the docs should have some brief high-level comments describing what they illustrate / the semantics (some already do, but the rest)

packages/test/src/workflows/cancel-timer-immediately.ts
```diff
@@ -1,18 +1,23 @@
...
 import { Empty } from '../interfaces';
 
 async function main(): Promise<void> {
-  const timer = sleep(3);
-  cancel(timer);
+  // Timers and Activities are automatically cancelled when their containing scope is cancelled.
   try {
-    await timer;
+    await CancellationScope.cancellable(async () => {
+      const promise = sleep(1); // <-- Will be cancelled because it is attached to `scope`
```
Sushisource, line 9:
```suggestion
      const promise = sleep(1); // <-- Will be cancelled because it is attached to this closure's scope
```

Maybe a bit more clear

packages/test/src/workflows/handle-external-workflow-cancellation-while-activity-running.ts
```diff
@@ -0,0 +1,20 @@
...
+import { CancelledError, CancellationScope } from '@temporalio/workflow';
+import { httpPostJSON, cleanup } from '@activities';
+
+export async function main(url: string, data: any): Promise<void> {
+  try {
+    await httpPostJSON(url, data);
+  } catch (err) {
+    if (err instanceof CancelledError) {
+      console.log('Workflow cancelled');
+      // Cleanup logic goes in a nonCancellable scope
```
Sushisource, line 12:
```suggestion
      // Cleanup logic must be in a nonCancellable scope
```

Sushisource (approved):
This looks really good to me, just a few minor doc related things

PR #134, opened 2021-06-22
Lazy workflow require

bergundy:
This will vastly improve devex.

Allows user code to use `console.log`, `Date`, `Math.random`, and injected dependency functions at the top level.

Also includes a fix for the max isolate memory config option not being passed to the isolate.

I will use this in the opentelemetry interceptor package (which should be added in the next PR).


packages/workflow/src/worker-interface.ts
```diff
@@ -0,0 +1,248 @@
...
+    appendChild(): void;
+  };
+  getElementsByTagName(): MockElement[];
+  createElement(): MockElement;
+}
+
+/**
+ * Create a mock document object with mimimal required attributes to support Webpack dynamic imports
+ */
+export function mockBrowserDocumentForWebpack(): MockDocument {
```
vkarpov15, line 96: Why is it necessary to fake a browser environment? It may cause some issues with outside modules. Most should be fine since you can't do I/O from workflows anyway, but could be a bit brittle.

packages/worker/src/isolate-builder.ts
```diff
@@ -66,32 +64,26 @@ export class WorkflowIsolateBuilder {
...
-      export { init, internals, workflows, interceptors };
+      // Bundle all Workflows and interceptor modules for lazy evaluation
+      globalThis.document = api.mockBrowserDocumentForWebpack();
+      // See https://webpack.js.org/api/module-methods/#requireensure
+      require.ensure([${bundlePaths.join(', ')}], function () {});
+      delete globalThis.document;
+
+      api.overrideGlobals();
+      api.setRequireFunc(
+        (name) => require(${JSON.stringify(this.workflowsPath)} + '${path.sep}' + name + '.js')
```
vkarpov15, line 82: Nice :+1: Does this mean for requiring activities, you can do `require('../activities/myActivity')`?

vkarpov15 (approved):
Nice :+1:

packages/worker/src/isolate-builder.ts
bergundy: No, that's for Workflow internals requiring user code.
It has nothing to do with user code imports.
Let's get rid of the activities alias soon though.

packages/workflow/src/worker-interface.ts
bergundy: We fake it just for webpack to store user code in the mock element.
If you see where this is used in the isolate builder, you'll notice that we delete the document object immediately after webpack is done using it.

PR #137, opened 2021-06-25
fix: Fix cross compilation of rust code

bergundy:
Noticed that rust is not being cross compiled anymore and the latest published version does not include the pre-built binaries.
This is a regression introduced in #128.


bergundy:
Installing from package initializer now uses pre-built binaries https://github.com/temporalio/sdk-node/pull/137/checks?check_run_id=2910291698

PR #138, opened 2021-06-25
fix(workflow): Allow passing number to ActivityOption timeouts

vkarpov15:
## What was changed

Allowed passing a number to `time.js` functions and added a check for [NaN](https://masteringjs.io/tutorials/fundamentals/nan) on timestamp parsing

## Why?

I got stuck today because I did:

```javascript
const greetWithCustomTaskQueue = Context.configure(greet, {
  taskQueue: 'tutorial446',
  startToCloseTimeout: 1000 // <-- Oops, should be string not number
});
```

and ended up with the below error

```
[TransportError: Unhandled error when calling the temporal server: Status { code: InvalidArgument, message: "BadScheduleActivityAttributes: A valid StartToClose or ScheduleToCloseTimeout is not set on command.", metadata: MetadataMap { headers: {"content-type": "application/grpc"} } }]
```

This is because [ms](https://www.npmjs.com/package/ms) converts numbers to strings, so `ms(1000) === '1s'`, and you end up with a NaN for both timestamp values.

Chalk this up as a case where using TypeScript would've saved me an hour or so :) But I think it's worth supporting this case, or at least throwing a more helpful error.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy:
I have a feeling more users are going to ask us to support passing numbers.
I personally am not a fan of using numbers (without units) for time values but I'm willing to consider.
The least we can do is check that the conversion works at initialization.
One thing we could do as allow passing a number using the `startToCloseTimeoutMs` option.

vkarpov15:
I'm open to the `startToCloseTimeoutMS` alternative. I would err on the side of allowing passing the number of milliseconds as a number, that's fairly common in JavaScript. But I'd at least throw a runtime error if `seconds` or `nanos` ends up as [`NaN`](https://masteringjs.io/tutorials/fundamentals/nan).

bergundy:
Let's support number as you've suggested.
Please find all of the configuration options in the SDK and modify the types and docstrings.

vkarpov15:
@bergundy I added `| number` to what looks like all timeouts and intervals. Let me know what you think :+1:

bergundy (approved):
LGTM, glad we ended up doing it

bergundy:
@vkarpov15 maybe just change the commit type to feat and remove the workflow scope when you squash?

vkarpov15:
@bergundy sure I did that with the auto-merge

bergundy:
The commit message that we got in main is:
`fix(workflow): Allow passing number to ActivityOption timeouts ...`
Where it should have been something like:
`feat: Allow passing number as timeout values`

That's what I meant in the comment above, I should have been more clear on that.

PR #147, opened 2021-07-01
feat(workflow): Implement queries

bergundy:


packages/test/src/test-integration.ts
```diff
@@ -103,12 +103,15 @@ if (RUN_INTEGRATION_TESTS) {
   });
 
   // Queries not yet properly implemented
```
Sushisource, line 105: Hopefully no longer true

Sushisource (approved):

PR #158, opened 2021-07-14
feat!: Use async data converter interface

bergundy:
Switched the data converter interface to be async to support non-blocking encryption and compression.

CC @rylandg @robholland

### DIFF
--- a/packages/workflow/src/converter/data-converter.ts
+++ b/packages/workflow/src/converter/data-converter.ts
@@ -13,9 +13,9 @@ import {
  * Implement this in order to customize worker data serialization or use the default data converter which supports `Uint8Array` and JSON serializables.
  */
 export interface DataConverter {
-  toPayload<T>(value: T): Payload;
+  toPayload<T>(value: T): Promise<Payload>;
 
-  fromPayload<T>(payload: Payload): T;
+  fromPayload<T>(payload: Payload): Promise<T>;
   /**
    * Implements conversion of a list of values.
    *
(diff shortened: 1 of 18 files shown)


packages/workflow/src/converter/data-converter.ts
```diff
@@ -88,16 +88,21 @@ export class CompositeDataConverter implements DataConverter {
   }
 }
 
-export function arrayFromPayloads(converter: DataConverter, content?: Payload[] | null): any[] {
+export async function arrayFromPayloads(converter: DataConverter, content?: Payload[] | null): Promise<unknown[]> {
```
vkarpov15, line 91: Is there a reason to make this function async? It doesn't use `await` and returns a promise, so all this `async` declaration does is create an unnecessary promise.

vkarpov15 (commented):
LGTM :+1: one minor comment but not a blocker.

packages/workflow/src/converter/data-converter.ts
bergundy: The reason would be to have it included in the stack trace in case of error.

```js
async function b() {
  await new Promise(r => setTimeout(r, 10));
  throw new Error('heh');
}
function a() {
  return Promise.all([b()]);
}
a();
```

```
> Uncaught Error: heh
    at b (REPL48:3:9)
    at async Promise.all (index 0)
```

But for that to happen we need to await in `a` which we don't :/
```js
async function a() {
 return await Promise.all([b()]);
}
```

```
Uncaught Error: heh
    at b (REPL48:3:9)
    at async Promise.all (index 0)
    at async a (REPL62:2:10)
```

I'll add the `await` keyword, sounds good?

PR #160, opened 2021-07-14
chore: Make bench wait for temporal during namespace creation

vitarb:
Makes bench wait on temporal server by adding a retry loop for the namespace creation.
This is better than running a separate script because it makes bench more self contained.


bergundy (approved):
LGTM

PR #161, opened 2021-07-21
Bench fixes

bergundy:
Each commit may be reviewed independently.

I'm having trouble getting jaeger to capture all of the spans, looks like some are missing in the UI, committing this because there's a lot of improvements here.

image


packages/test/src/perf/args.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export interface WorkerArgSpec extends arg.Spec {
+  '--ns': typeof String;
+  '--task-queue': typeof String;
+  '--max-cached-wfs': typeof Number;
+  '--max-concurrent-at-executions': typeof Number;
+  '--max-concurrent-wft-executions': typeof Number;
+  // NOTE: typeof this is not supported yet by Core
+  '--max-concurrent-at-polls': typeof Number;
+  // NOTE: typeof this is not supported yet by Core
```
Sushisource, line 42: typeof this?

packages/worker/src/activity.ts
```diff
@@ -53,6 +53,7 @@ export class Activity {
         }
         return { completed: { result: await this.dataConverter.toPayload(result) } };
       } catch (err) {
+        // TODO: should we only allow CancelledError to count as valid cancellation?
```
Sushisource, line 56: I think so yes. Not found would mean you timed out or something else bad happened.

PR #171, opened 2021-07-28
feat!: Add common library for WF and non WF code

### LINKED ISSUE #100: Create a shared dependency for the worker and client packages
Any reason not to just make a lower dep they can share?

_Originally posted by @Sushisource in https://github.com/temporalio/sdk-node/pull/97#discussion_r631968959_

bergundy:
Closes #100

I did this now because I need to share workflow-client between workflow and client packages for ChildWorkflow.


bergundy:
Looks like performance improved.
Results from single round of perf tests that I ran on my M1 MBP.

```shellsession
$ time node packages/test/lib/perf/all-in-one.js --iterations 30000 --workflow sync --concurrent-wf-clients 500 --max-cached-wfs 2500
```

Before change:
```
WFs complete (30000/30000), WFs/s curr 9.6 (acc 48.1)
642.31s user 103.73s system 119% cpu 10:25.78 total
```

After change:
```
WFs complete (30000/30000), WFs/s curr 9.1 (acc 66.2)
481.74s user 30.10s system 111% cpu 7:40.02 total
```

packages/worker/workflow-isolate-extension.cc
```diff
@@ -27,11 +27,16 @@ void custom_promise_hook(
...
-    Local<Context> context = promise->GetCreationContext().ToLocalChecked();
+    Local<Context> context;
+    bool valid = promise->GetCreationContext().ToLocal(&context);
+    if (!valid) return;
 #else
     Local<Context> context = promise->CreationContext();
 #endif
     Local<Function> fn = Local<Function>::Cast(context->GetEmbedderData(EMBEDDER_DATA_IDX));
+    if (!fn->IsFunction()) {
+        return;
```
Sushisource, line 38: Should this blow up in a more obvious way?

Sushisource (approved):
You know I wonder what the cost of just spawning separate node processes is v.s. this? I assume it's substantially more expensive but if it's only slightly more it might be worth saving all this complexity and possible bugs. Did you evaluate that?

bergundy:
I haven't evaluated separate node processes but I have a feeling it'll be much more expensive.
The only "valid" options I see right now are:
1. Staying with isolates for the WF runtime.
2. Running WFs in the default node isolate (in the same process as the Worker)

Going with (2) means that we lose all of the work we've put in to create a deterministic runtime for WFs and going with (1)  adds a lot of work (mostly done already) and a performance hit to the SDK which hopefully is either acceptable or can be improved to be acceptable - ~70 WF/s on my computer using up just over 100% CPU.

The first thing we did when working on this SDK is build the deterministic WF isolate environment, we've already invested a lot of energy in this path and I hope it'll pay off.

A subprocess wouldn't help either of the approaches, it'll just add overhead.
FTR Each node process takes about 30 MB of RAM just to start up and about 50ms to start on my laptop.

packages/worker/workflow-isolate-extension.cc
bergundy: This may happen when creating a promise before registering the promise hook for a context (e.g. when creating the root scope), I'll make sure to add a comment so it's clear.

PR #173, opened 2021-08-04
Port failure classes from Java SDK

### LINKED ISSUE #146: [Feature Request] Port Workflow error classes from Java SDK
- For instance for activity failures.
- Errors should have an `isRetryable` attributes.

bergundy:
Closes #146

We still need to figure out what to do for activity cancellation.

### DIFF
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -1,0 +1,428 @@
+import { temporal } from '@temporalio/proto';
+import { DataConverter, arrayFromPayloads } from './converter/data-converter';
+
+export const FAILURE_SOURCE = 'NodeSDK';
+export type ProtoFailure = temporal.api.failure.v1.IFailure;
+export type TimeoutType = temporal.api.enums.v1.TimeoutType;
+export const TimeoutType = temporal.api.enums.v1.TimeoutType;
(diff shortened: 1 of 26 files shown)


bergundy:
Depends on https://github.com/temporalio/sdk-core/pull/159

packages/common/src/failure.ts
```diff
@@ -0,0 +1,424 @@
...
+export const RetryState = coresdk.common.RetryState;
+
+/**
+ * Represents failures that can cross Workflow and Activity boundaries.
+ *
+ * Only exceptions that extend this class will be propagated to the caller.
+ *
+ * **Never extend this class or any of its derivatives.** They are to be used by the SDK code
+ * only. Throw an instance {@link ApplicationFailure} to pass application specific errors between
+ * Workflows and Activities.
```
Sushisource, line 17: Could just make this constructor private and then expose a `create` function to allow constructing them for testing purposes. That way the class can't be extended.

Sushisource (approved):

bergundy:
Note that this addition of 3 proto files increased the generated JS file from 244kb to 638kb, at some point we might want to optimize this but I'll let this go because looks like perf is still performant enough from a short test I made.

sw-yx:
good that you are watching the perf though, nice

PR #179, opened 2021-08-09
Child Workflows

bergundy:
This should not be merged until https://github.com/temporalio/sdk-core/pull/163 is merged.
We should bring back the incremented sequence for generating correlation IDs between lang and core.
That will be done in a later PR.

### DIFF
--- a/packages/workflow/src/internals.ts
+++ b/packages/workflow/src/internals.ts
@@ -6,19 +6,25 @@ import {
   composeInterceptors,
   errorToFailure,
   ensureTemporalFailure,
+  failureToError,
   optionalFailureToOptionalError,
   IllegalStateError,
   Workflow,
-  arrayFromPayloads,
+  WorkflowSignalType,
   DataConverter,
   defaultDataConverter,
-  WorkflowSignalType,
+  arrayFromPayloadsSync,
 } from '@temporalio/common';
 import { coresdk } from '@temporalio/proto/lib/coresdk';
 import { alea, RNG } from './alea';
 import { ContinueAsNew, ExternalDependencies, WorkflowInfo } from './interfaces';
 import { SignalInput, WorkflowInput, WorkflowInterceptors } from './interceptors';
-import { CancelledError, DeterminismViolationError, WorkflowCancelledError } from './errors';
+import {
+  CancelledError,
+  DeterminismViolationError,
+  WorkflowCancelledError,
+  WorkflowExecutionAlreadyStartedError,
+} from './errors';
 import { ROOT_SCOPE } from './cancellation-scope';
 
 export type ResolveFunction<T = any> = (val: T) => any;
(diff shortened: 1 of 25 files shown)


packages/workflow/src/internals.ts
```diff
@@ -107,7 +113,61 @@ export class Activator implements ActivationHandler {
...
+    } else if (activation.failed) {
+      if (
+        activation.failed.cause !==
+        coresdk.child_workflow.StartChildWorkflowExecutionFailedCause
+          .START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_EXISTS
+      ) {
+        throw new IllegalStateError('Got unknown StartChildWorkflowExecutionFailedCause');
+      }
+      if (!(activation.workflowId && activation.failed?.workflowType)) {
+        throw new TypeError('');
```
Sushisource, line 131: No detail in this error?

packages/workflow/src/internals.ts
```diff
@@ -315,9 +374,9 @@ async function handleWorkflowFailure(error: any) {
   state.completed = true;
 }
 
-async function completeQuery(queryId: string, result: unknown) {
+function completeQuery(queryId: string, result: unknown) {
   state.commands.push({
-    respondToQuery: { queryId, succeeded: { response: await state.dataConverter.toPayload(result) } },
+    respondToQuery: { queryId, succeeded: { response: state.dataConverter.toPayloadSync(result) } },
```
Sushisource, line 379: Why'd all these need to become sync? Just curious

packages/workflow/src/internals.ts
```diff
@@ -181,7 +241,7 @@ export interface ExternalCall {
   fnName: string;
   args: any[];
   /** Optional in case applyMode is ASYNC_IGNORED */
-  seq?: number;
+  seq?: string;
```
Sushisource, line 244: Will this go back to being a number when we make the ids change?

packages/workflow/src/workflow.ts
```diff
@@ -173,6 +198,79 @@ export function scheduleActivity<R>(
...
 
+async function startChildWorkflowExecutionNextHandler({
+  options,
+  args,
+  headers,
+  workflowType,
+}: StartChildWorkflowExecutionInput): Promise<[Promise<string>, Promise<unknown>]> {
+  const workflowId = options.workflowId ?? uuid4();
+
+  // TODO: validate options
```
Sushisource, line 209: TODO needs to be done?

Sushisource (approved):

packages/workflow/src/internals.ts
bergundy: Forgot to add them

packages/workflow/src/internals.ts
bergundy: I found out that I couldn't immediately cancel the workflow here:
https://github.com/temporalio/sdk-node/blob/1fea8fa89bb80de70f044644b6adc09f0072db77/packages/test/src/workflows/child-workflow-cancel.ts#L16

With async data converter instead of start and cancel reaching core, the scope is cancelled before the child schedule command is even sent.

packages/workflow/src/internals.ts
bergundy: Yes

packages/workflow/src/workflow.ts
bergundy: I realized there's no special validation required here assuming the user uses typescript.
At a later point I'll add validation for JS users (e.g. passing a number instead of a string).

bergundy:
Just realized I didn't add interceptors for child WFs, doing that before merging.

bergundy:
Luckily I've prepared everything for the interceptor, it was easy to add including a test.

PR #188, opened 2021-08-11
New `patched` API formerly known as versioning

Sushisource:
## What was changed
Added `patched` and `deprecatePatch` workflow APIs

## Why?
Support changes to live workflows

## Checklist

1. Closes https://github.com/temporalio/sdk-node/issues/68
2. How was this tested:
Added UTs and ITs

The nitty gritty logic is well tested in core

3. Any docs updates needed?
Definitely. We need to explain this one bigtime. Refer to https://github.com/temporalio/proposals/blob/master/core/versions_2.0.md

### DIFF
--- a/packages/test/src/test-integration.ts
+++ b/packages/test/src/test-integration.ts
@@ -44,6 +44,7 @@ const { EVENT_TYPE_TIMER_STARTED, EVENT_TYPE_TIMER_FIRED, EVENT_TYPE_TIMER_CANCE
   iface.temporal.api.enums.v1.EventType;
 
 const timerEventTypes = new Set([EVENT_TYPE_TIMER_STARTED, EVENT_TYPE_TIMER_FIRED, EVENT_TYPE_TIMER_CANCELED]);
+const CHANGE_MARKER_NAME = 'core_patch';
 
 export interface Context {
   worker: Worker;
@@ -346,6 +347,42 @@ if (RUN_INTEGRATION_TESTS) {
     t.is(tsToMs(timerEvents[1].timerStartedEventAttributes!.startToFireTimeout), 1);
     t.is(timerEvents[2].timerFiredEventAttributes!.timerId, '2');
     t.is(timerEvents[3].timerCanceledEventAttributes!.timerId, '1');
+  });
+
+  test('patched', async (t) => {
+    const client = new WorkflowClient();
+    const workflow = client.stub<Empty>('patched', { taskQueue: 'test' });
+    const runId = await workflow.start();
+    const res = await workflow.result();
+    t.is(res, undefined);
+    const execution = await client.service.getWorkflowExecutionHistory({
+      namespace,
+      execution: { workflowId: workflow.workflowId, runId },
+    });
+    const hasChangeEvents = execution.history!.events!.filter(
+      ({ eventType }) => eventType === iface.temporal.api.enums.v1.EventType.EVENT_TYPE_MARKER_RECORDED
+    );
+    // There will only be one marker despite there being 2 hasChange calls because they have the
+    // same ID and core will only record one marker per id.
+    t.is(hasChangeEvents.length, 1);
+    t.is(hasChangeEvents[0].markerRecordedEventAttributes!.markerName, CHANGE_MARKER_NAME);
+  });
+
+  test('deprecate-patch', async (t) => {
+    const client = new WorkflowClient();
+    const workflow = client.stub<Empty>('deprecate-patch', { taskQueue: 'test' });
+    const runId = await workflow.start();
+    const res = await workflow.result();
+    t.is(res, undefined);
+    const execution = await client.service.getWorkflowExecutionHistory({
+      namespace,
+      execution: { workflowId: workflow.workflowId, runId },
+    });
+    const hasChangeEvents = execution.history!.events!.filter(
+      ({ eventType }) => eventType === iface.temporal.api.enums.v1.EventType.EVENT_TYPE_MARKER_RECORDED
+    );
+    t.is(hasChangeEvents.length, 1);
+    t.is(hasChangeEvents[0].markerRecordedEventAttributes!.markerName, CHANGE_MARKER_NAME);
   });
 
   test('Worker default ServerOptions are generated correctly', async (t) => {
--- a/packages/test/src/test-workflows.ts
+++ b/packages/test/src/test-workflows.ts
@@ -1,10 +1,10 @@
-import anyTest, { TestInterface, ExecutionContext } from 'ava';
+import anyTest, { ExecutionContext, TestInterface } from 'ava';
 import path from 'path';
 import Long from 'long';
 import dedent from 'dedent';
 import { coresdk } from '@temporalio/proto';
 import { ApplyMode } from '@temporalio/workflow';
-import { defaultDataConverter, ApplicationFailure, errorToFailure, msToTs, RetryState } from '@temporalio/common';
+import { ApplicationFailure, defaultDataConverter, errorToFailure, msToTs, RetryState } from '@temporalio/common';
 import { Workflow } from '@temporalio/worker/lib/workflow';
 import { WorkflowIsolateBuilder } from '@temporalio/worker/lib/isolate-builder';
 import { RoundRobinIsolateContextProvider } from '@temporalio/worker/lib/isolate-context-provider';
@@ -94,10 +94,15 @@ function makeStartWorkflow(
   args?: coresdk.common.IPayload[],
   timestamp: number = Date.now()
 ): coresdk.workflow_activation.IWFActivation {
+  return makeActivation(timestamp, makeStartWorkflowJob(script, args));
+}
+
+function makeStartWorkflowJob(
+  script: string,
+  args?: coresdk.common.IPayload[]
+): coresdk.workflow_activation.IWFActivationJob {
   return {
-    runId: 'test-runId',
-    timestamp: msToTs(timestamp),
-    jobs: [{ startWorkflow: { workflowId: 'test-workflowId', workflowType: script, arguments: args } }],
+    startWorkflow: { workflowId: 'test-workflowId', workflowType: script, arguments: args },
   };
 }
 
@@ -153,6 +158,12 @@ function makeResolveActivity(
   timestamp: number = Date.now()
 ): coresdk.workflow_activation.IWFActivation {
   return makeActivation(timestamp, makeResolveActivityJob(seq, result));
+}
+
+function makeNotifyHasPatchJob(patchId: string): coresdk.workflow_activation.IWFActivationJob {
+  return {
+    notifyHasPatch: { patchId },
+  };
 }
 
 async function makeQueryWorkflow(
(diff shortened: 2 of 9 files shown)


packages/test/src/test-workflows.ts
```diff
@@ -229,6 +247,15 @@ function makeRespondToQueryCommand(
   };
 }
 
+function makeSetChangeMarker(myChangeId: string, deprecated: boolean): coresdk.workflow_commands.IWorkflowCommand {
```
mfateev, line 250: Shouldn't we rename the marker to match the function name? PatchMarker?

packages/test/src/test-workflows.ts
```diff
@@ -41,7 +41,17 @@ test.after.always((t) => {
...
+  const trimmedTitle = t.title.match(/\S+$/)![0];
+  // Trust me, you don't want to do this part with a regex.
+  const colonPos = trimmedTitle.indexOf(':');
+  let testName: string;
+  if (colonPos != -1) {
+    testName = trimmedTitle.substr(0, colonPos);
+  } else {
+    testName = trimmedTitle;
+  }
+  const testCase = trimmedTitle.substr(colonPos + 1);
```
bergundy, line 53: Github is screaming that this is unused, I'm joining the screaming party.

bergundy, line 53: Instead of all of this you could have just kept everything as is and just used test titles like: `not-replay patched`

packages/worker/src/workflow.ts
```diff
@@ -172,6 +172,20 @@ export class Workflow {
     if (!activation.jobs) {
       throw new Error('Expected workflow activation jobs to be defined');
     }
+
+    // Process notify change jobs first
+    // TODO: Only doing this b/c I need to prevent workflows that don't do anything else from
+    //   "completing" before they have a chance to run the notify change job. In general this
+    //   seems like a bit of a hole that would also apply to signals, as noted below.
+    activation.jobs.sort((a, b) => {
```
bergundy, line 180: This needs to happen in core for both changes and signals.

packages/workflow/src/workflow.ts
```diff
@@ -526,3 +526,66 @@ export function uuid4(): string {
...
+ *
+ * @param patchId An identifier that should be unique to this patch. It is OK to use multiple
+ * calls with the same ID, which means all such calls will always return the same value.
+ */
+export function deprecatePatch(patchId: string): void {
+  patchInternal(patchId, true);
+}
+
+function patchInternal(changeId: string, deprecated: boolean): boolean {
+  // Always send the command to core, it's smart enough to ignore it if it was pointless.
```
bergundy, line 573: I wouldn't always send the command to core, it's pointless.

packages/workflow/src/workflow.ts
```diff
@@ -526,3 +526,66 @@ export function uuid4(): string {
...
+    },
+  });
+
+  const inMapVal = state.knownPresentChanges.get(changeId);
+  if (inMapVal !== undefined) {
+    return inMapVal;
+  }
+  // If we don't already know about the change, that means there is no marker in history,
+  // and we should return false if we are replaying
+  const result = !state.info?.isReplaying;
```
bergundy, line 588: Better to assert here that `state.info` is present or throw an `IllegalStateError`

packages/workflow/src/workflow.ts
```diff
@@ -526,3 +526,66 @@ export function uuid4(): string {
...
+ * @param patchId An identifier that should be unique to this patch. It is OK to use multiple
+ * calls with the same ID, which means all such calls will always return the same value.
+ */
+export function deprecatePatch(patchId: string): void {
+  patchInternal(patchId, true);
+}
+
+function patchInternal(changeId: string, deprecated: boolean): boolean {
+  // Always send the command to core, it's smart enough to ignore it if it was pointless.
+  // TODO: Interceptors? Do we care?
```
bergundy, line 574: IMO we don't need interceptors here, let's change the comment to say:
```suggestion
  // Patch operation does not support interception at the moment, if it did, this would be the place to start the interception chain
```

packages/test/src/test-workflows.ts
```diff
@@ -1544,4 +1571,120 @@ test('continue-as-new-same-workflow', async (t) => {
...
+    );
+  }
+  t.deepEqual(logs, [['no change'], ['no change 2']]);
+});
+
+test('patched:replay-with-marker', async (t) => {
+  const { logs, script } = t.context;
+  {
+    const act = makeStartWorkflow(script, undefined, undefined, true);
+    act.jobs?.push(makeNotifyHasChangeJob('my-change-id'));
```
bergundy, line 1655: Probably better to use [makeActivation](https://github.com/temporalio/sdk-node/blob/13d648ef929fac3571b15fce77aae09742c9c750/packages/test/src/test-workflows.ts#L116) and add a `makeStartWorkflowJob` helper which `makeStartWorkflow` can use.
As mentioned in another comment, I'd rather have core enforce that the hasChange notification jobs come first, they have to, always, in every SDK.

packages/test/src/test-workflows.ts
```diff
@@ -1544,4 +1571,120 @@ test('continue-as-new-same-workflow', async (t) => {
...
+
+test('patched:replay-no-marker-then-not-replay', async (t) => {
+  const { logs, script } = t.context;
+  {
+    const req = await activate(t, makeStartWorkflow(script, undefined, undefined, true));
+    compareCompletion(
+      t,
+      req,
+      makeSuccess([
+        makeSetChangeMarker('my-change-id', false),
```
bergundy, line 1632: We probably shouldn't send this command out since we didn't use the change branch.

packages/test/src/test-workflows.ts
```diff
@@ -1544,4 +1571,120 @@ test('continue-as-new-same-workflow', async (t) => {
...
+
+test('patched:replay-no-marker', async (t) => {
+  const { logs, script } = t.context;
+  {
+    const req = await activate(t, makeStartWorkflow(script, undefined, undefined, true));
+    compareCompletion(
+      t,
+      req,
+      makeSuccess([
+        makeSetChangeMarker('my-change-id', false),
```
bergundy, line 1606: We probably shouldn't send this command out since we didn't use the change branch.

Same goes for the second marker.

bergundy (commented):
Thanks!
Good stuff, I had some comments about the behavior.

Minor things to help you get this merged:
- `chmod +x packages/create-project/lib/index.js` it's a generated file that has to be committed, not ideal but I work around it for now.
- Please rebase your branch on top of origin/main and use the `rebase and merge` button and not the `squash and merge` button when ready so we don't accidentally merge a commit that doesn't follow conventional commits.

packages/test/src/test-workflows.ts
Sushisource: BAHHHHHHHHHHHHHHHHHHH

packages/test/src/test-workflows.ts
Sushisource: Yep, missed this one.

packages/test/src/test-integration.ts
```diff
@@ -349,6 +349,42 @@ if (RUN_INTEGRATION_TESTS) {
...
+    const res = await workflow.result();
+    t.is(res, undefined);
+    const execution = await client.service.getWorkflowExecutionHistory({
+      namespace,
+      execution: { workflowId: workflow.workflowId, runId },
+    });
+    const hasChangeEvents = execution.history!.events!.filter(
+      ({ eventType }) => eventType === iface.temporal.api.enums.v1.EventType.EVENT_TYPE_MARKER_RECORDED
+    );
+    // There will only be one marker despite there being 2 hasChange calls because they have the
```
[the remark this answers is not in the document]
Sushisource: I dunno that I agree here. It's easy enough to rely on Core to do the right thing here and not have to think about it in lang, and there's no overhead besides a few extra bytes since this command isn't awaited, so the number of calls to core remains the same.

If this changes later, saying we just always send commands means there are less things to change lang side.

packages/test/src/test-integration.ts
```diff
@@ -349,6 +349,42 @@ if (RUN_INTEGRATION_TESTS) {
...
+    const res = await workflow.result();
+    t.is(res, undefined);
+    const execution = await client.service.getWorkflowExecutionHistory({
+      namespace,
+      execution: { workflowId: workflow.workflowId, runId },
+    });
+    const hasChangeEvents = execution.history!.events!.filter(
+      ({ eventType }) => eventType === iface.temporal.api.enums.v1.EventType.EVENT_TYPE_MARKER_RECORDED
+    );
+    t.is(hasChangeEvents.length, 1);
```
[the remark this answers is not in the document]
Sushisource: The marker name can pretty easily. The details are a bit annoying to deserialize but I'll at least check it's the right marker type.

packages/workflow/src/workflow.ts
Sushisource: See other comment

packages/test/src/test-workflows.ts
Sushisource: Oh, right, this is because that's still how it's called in the protos/core. I will rename it in core and then come back here.

packages/workflow/src/workflow.ts
bergundy: There are 2 cases I think we're doing wrong here:
1. We keep sending the same marker over and over again
2. We send the marker even if the workflow used the old code

I think we should change this to:

```ts
  if (state.info === undefined) {
    throw new IllegalStateError('Workflow info must be set when calling patch functions');
  }
  const { isReplaying } = state.info;
  const usePatch = !isReplaying || state.knownPresentChanges.has(patchId);
  if (usePatch && !state.sentPatches.has(patchId)) {
    state.commands.push({
      setPatchMarker: {
        patchId: patchId,
        deprecated: deprecated,
      },
    });
    state.sentPatches.add(patchId);
  }
  return usePatch;
```

packages/workflow/src/internals.ts
```diff
@@ -345,6 +348,11 @@ export class State {
   public require?: (filename: string) => Record<string, unknown>;
 
   public dataConverter: DataConverter = defaultDataConverter;
+
+  /**
+   * Changes we know the status of for this workflow, as in {@link patched}
+   */
+  public knownPresentChanges: Map<string, boolean> = new Map();
```
bergundy, line 355: Just use a `Set<string>` instead

bergundy (commented):
The main issue here is when we send out patch commands to core.
LMK what you think.

packages/workflow/src/internals.ts
Sushisource: It needs to be set to false sometimes

packages/workflow/src/workflow.ts
Sushisource: The thing is though it just doesn't matter. Core cleans that up, and like I mentioned there's only an incredibly small amount of overhead b/c we still have to call core the same amount of times. So at the end of the day I think this is more code for very little benefit. The way it is it's simple to test, idempotent, etc. I guess I'm just not following why we'd want to add the complexity when we don't need to?

Sushisource: EX: The code you posted doesn't work in the `replay-no-marker-then-not-replay patched` test. It's just tricky to get them all right. I've updated it to a version that just doesn't issue them more than once which I think is 90% of the benefit without the tricky logic.

It's also just easy to test -- first time it always gets sent, no need to think too hard

bergundy:
Okay, let's merge as is, I still find it a little weird that we're saying we're using the patch even though we're not in case of old history replay but it doesn't matter for correctness AFAIU.

bergundy (approved):

PR #190, opened 2021-08-16
fix(worker): Avoid throwing error when there's no activities directory

vkarpov15:
## What was changed

Added a missing `mkdir()` call

## Why?

If there's no activities, `registerActivities()` doesn't call `mkdir()`, so the `writeFileSync()` call will fail

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/worker/src/isolate-builder.ts
+++ b/packages/worker/src/isolate-builder.ts
@@ -98,6 +98,11 @@ export class WorkflowIsolateBuilder {
 
       module.exports = api;
     `;
+    try {
+      vol.mkdirSync(path.dirname(target), { recursive: true });
+    } catch (err) {
+      if (err.code !== 'EEXIST') throw err;
+    }
     vol.writeFileSync(target, code);
   }
 


bergundy (approved):
Thanks!

joebowbeer:
I'm still seeing a residual error in samples-node/progress:

> Error: ENOENT: no such file or directory, open '/src/main.js'

packages/worker/src/isolate-builder.ts
```diff
@@ -98,6 +98,11 @@ export class WorkflowIsolateBuilder {
 
       module.exports = api;
     `;
+    try {
+      vol.mkdirSync(path.dirname(target), { recursive: true });
+    } catch (err) {
+      if (err.code !== 'EEXIST') throw err;
+    }
     vol.writeFileSync(target, code);
```
joebowbeer, line 106: It will still blow up here, right?

bergundy:
@joebowbeer we still haven't released the fix which would explain why you're still seeing the error.
`genEntrypoint` is called only once and AFAICT @vkarpov15's fix addresses the issue.

joebowbeer:
@bergundy thanks for explanation. I see now: the unreleased fix is creating the dir if it doesn't exist.

PR #199, opened 2021-08-23
fix(worker): Avoid throwing error when there are no activities

joebowbeer:
## What was changed

Avoid throwing error when there are no activities.

## Why?

I'm still seeing a residual error in samples-node/progress:

Error: ENOENT: no such file or directory, open '/src/main.js'

From https://github.com/temporalio/sdk-node/blob/main/packages/worker/src/isolate-builder.ts#L45

genEntryPoint

## Checklist

2. How was this tested:

### DIFF
--- a/packages/worker/src/isolate-builder.ts
+++ b/packages/worker/src/isolate-builder.ts
@@ -100,10 +100,10 @@ export class WorkflowIsolateBuilder {
     `;
     try {
       vol.mkdirSync(path.dirname(target), { recursive: true });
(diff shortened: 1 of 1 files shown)


bergundy:
I don't think this is needed as mentioned here: https://github.com/temporalio/sdk-node/pull/190#issuecomment-903478628.
@joebowbeer I want to change the activity registration API soon so non of this will be relevant anyways.

joebowbeer:
Never mind.

The (correct) fix is not yet released.

Issue #96, opened 2021-05-13
[Feature Request] Automated load testing


bergundy:
The simple bench test stresses the system uncovering bugs that integration tests tend to miss.
Bench test should run on every PR, running against cloud will also test that passing an address and TLS options work.

bergundy:
@vkoby I just noticed we have this issue, please close once you're finished working on it.

Sushisource:
Include additional scenarios

PR #212, opened 2021-08-26
Load tests in buildkite

Sushisource:
Buildkite loadtests

Pt. 1 of #96 

Will add more scenarios after this is merged


packages/test/src/activities/index.ts
```diff
@@ -5,6 +5,15 @@ import { Connection } from '@temporalio/client';
...
 import { cancellableFetch as cancellableFetchInner } from './cancellable-fetch';
 
+// TODO: Get rid of this by providing client via activity context
+function getTestConnection(): Connection {
+  const serverURLOverride = process.env.TEMPORAL_TESTING_SERVER_URL;
+  if (serverURLOverride !== undefined && serverURLOverride.length > 0) {
+    return new Connection({ address: serverURLOverride });
+  }
+  return new Connection();
+}
```
bergundy, line 15: Not that it's even close to being critical but this would have been sufficient
```suggestion
  const address = process.env.TEMPORAL_TESTING_SERVER_URL || undefined;
  return new Connection({ address });
}
```

packages/test/src/perf/all-in-one.ts
```diff
@@ -35,6 +35,12 @@ async function main() {
   try {
     const starterArgs = argsToForward(starterArgSpec, args);
     const starter = spawn('node', [path.resolve(__dirname, 'starter.js'), ...starterArgs], { shell, stdio: 'inherit' });
+    worker.on('exit', (code) => {
+      // If worker dies unceremoniously then also make sure we can exit and not hang
+      if (code !== 0) {
+        kill(starter, 'SIGINT');
```
bergundy, line 41:
```suggestion
        await kill(starter, 'SIGINT');
```

bergundy: This was bugging me too but never got around to fixing, thanks.

packages/worker/src/worker.ts
```diff
@@ -709,7 +709,11 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
               const result = await output.activity.run(output.input);
               const status = result.failed ? 'failed' : result.completed ? 'completed' : 'cancelled';
               span.setAttributes({ status });
-              this.log.debug('Activity resolved', { activityId: output.activity.info.activityId, status });
+              this.log.debug('Activity resolved', {
+                activityId: output.activity.info.activityId,
+                status,
+                result: JSON.stringify(result),
```
bergundy, line 715: Please revert

package.json
```diff
@@ -33,12 +34,13 @@
...
   },
   "scripts": {
-    "install": "cd packages/docs && npm ci",
+    "install": "test -n $SKIP_DOCS_INSTALL || (cd packages/docs && npm ci)",
     "rebuild": "npm run clean && npm run build",
     "build": "lerna run --stream build",
     "build.watch": "lerna run --stream build.watch",
     "test": "lerna run --stream test",
     "test.watch": "lerna run --stream test.watch",
+    "ci-load": "node ./packages/test/lib/perf/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL",
```
bergundy, line 43: Why don't we call this `perf` instead of `load`?
I'd keep this consistent, AFAIC, we can rename either the test files or the script name.

package.json
```diff
@@ -33,12 +34,13 @@
     "typescript": "^4.2.3"
   },
   "scripts": {
-    "install": "cd packages/docs && npm ci",
+    "install": "test -n $SKIP_DOCS_INSTALL || (cd packages/docs && npm ci)",
```
bergundy, line 37: I'm thinking we might want to remove the docs dependencies out of this completely and only install them on demand.
Just not sure about the local dev experience.

package.json
```diff
@@ -7,6 +7,7 @@
     "@commitlint/cli": "^12.1.4",
     "@commitlint/config-conventional": "^12.1.4",
     "@opentelemetry/exporter-jaeger": "^0.20.0",
+    "@opentelemetry/api": "^0.20.0",
```
bergundy, line 10: I'm still puzzled as to why this breaks only on buildkite, but if it works, let's leave it.

bergundy: Especially puzzling is the fact that `package-lock.json` does not reflect this change.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea/
```
bergundy, line 1: Sure we can put this here but I usually would rather not put editor specifics in the repo.
Wouldn't you rather put this in a global gitignore on your computer?

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,10 @@
+steps:
+  - label: 'load-test'
+    agents:
+      queue: 'default'
+      docker: '*'
+    timeout_in_minutes: 15
+    plugins:
+      - docker-compose#v3.0.0:
```
bergundy, line 8: `3.0.0`? is this strict versioning? Can we make it `<4` a.k.a `^3.0.0`? I'm asking due to lack of familiarity with buildkite

.buildkite/docker/docker-compose.yaml
```diff
@@ -0,0 +1,51 @@
...
+      driver: none
+    ports:
+      - '7233:7233'
+      - '7234:7234'
+      - '7235:7235'
+      - '7239:7239'
+      - '6933:6933'
+      - '6934:6934'
+      - '6935:6935'
+      - '6939:6939'
```
bergundy, line 23: Not required for CI.
Instead of duplicating the entire file, why don't you download the original?
```docker-compose -f docker-compose.yml -f server.yml up```

File: .buildkite/docker/build-and-run-load.sh [hunk 10]
```diff
@@ -0,0 +1,17 @@
...
+npm --version
+
+npm set progress=false
+export SKIP_DOCS_INSTALL=true
+export TEMPORAL_TESTING_SERVER_URL="temporal:7233"
+
+npm ci
+# Try rebuilding once if it fails
+# TODO: Figure out whatever problem causes "common" package to not get compiled when it should
+npm run build || npm run build
```
bergundy, line 15: Wow, this is how you solved the build problem?
Let's keep the hack and track the issue.

.buildkite/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM rust:1.54
+
+# INSTALL NODE ============================================================================
+# Unfortunately can't use the friendly lts name here easily, because then it gets hard to add to the
+# path. We'll just have to update this as needed.
+ENV NODE_VERSION="v14.17.5"
+RUN curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.38.0/install.sh | bash
+ENV NVM_DIR=/root/.nvm
+RUN . "$NVM_DIR/nvm.sh" && nvm install ${NODE_VERSION}
+RUN . "$NVM_DIR/nvm.sh" && nvm use ${NODE_VERSION}
```
bergundy, line 10: Why is this required?

.buildkite/docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM rust:1.54
+
+# INSTALL NODE ============================================================================
+# Unfortunately can't use the friendly lts name here easily, because then it gets hard to add to the
+# path. We'll just have to update this as needed.
+ENV NODE_VERSION="v14.17.5"
```
bergundy, line 6: Could have used lts and changed the code below to:

`ENV PATH="$(dirname $(. $NVM_DIR/nvm.sh && nvm use 16 &> /dev/null && which node)):$PATH"`

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,10 @@
+steps:
+  - label: 'load-test'
```
bergundy, line 2: Wouldn't you rather have separate build and test steps?

bergundy:
I'd also add some assertion about the number WFs processed per second and perhaps check the CPU usage.
I'd run at least 10000 WFs to get more accurate numbers.

.buildkite/docker/Dockerfile
Sushisource: It's just the easiest way to get node installed in the docker image

.gitignore
Sushisource: Makes it easier for others as well. There's not a reason at all for it to be checked in, no harm in having this line in the gitignore.

.buildkite/docker/docker-compose.yaml
Sushisource: What original are you referring to here?

.buildkite/pipeline.yml
Sushisource: It appears they must be specified exactly

.buildkite/pipeline.yml
Sushisource: I kinda would, but the way this docker-compose plugin works it's just much easier to do it this way

.buildkite/docker/Dockerfile
bergundy: I was asking about `nvm use`, I don't think it does anything here.

.buildkite/docker/docker-compose.yaml
bergundy: https://github.com/temporalio/docker-compose/blob/main/docker-compose.yml

.buildkite/pipeline.yml
bergundy: It's faster now but once we add more parallel steps the build agents will slow down.

.gitignore
bergundy: I will accept this impurity :)

bergundy (approved):
This can be merged AFAIC.

Had some unresolved comments but no blockers.

PR #215, opened 2021-08-29
feat!: Revise Activity setup and registration, use single tsconfig in project skeleton

### LINKED ISSUE #109: [Bug] Activity identifier should not include the activity module name
**Describe the bug**

Currently the Node SDK uses `JSON.stringify([module, functionName])` for activity names in order to avoid conflicts.

**Expected behavior**

We should be consistent with the other SDKs and use only the function name.
The Worker should throw an error if multiple activities are registered with the same name.

NOTE: that the main reason for using only function name as the identifier is so functions can easily be refactored to a different location.

### LINKED ISSUE #168: [Feature Request] Discuss activity registration and imports
Currently we use the following to import Activity functions as stubs in a Workflow.
```ts
import { myActivity } from '@activities/something';
// NOTE: `../activities` is not supported but that can be fixed
```

This is pretty convenient since provides implicit type safety

Another approach would be to use something like `ExternalDependencies`
`workflows/example.ts`
```ts
const { myActivity } = Context.activities<ActivityInterface>();
```
`worker/index.ts`
```ts
import * as activities from '../activities';
const worker = await Worker.create<{ activities: ActivityInterface }>({ activities });
```

Here we rely on an interface definition for type safety but it allows more flexibility in registration and it seems like there's less "magic" going on.

One other option that comes to mind is to have a single `@activities` file (e.g. `src/activities/index.ts` and disallow imports from `@activities/something`, this way we'd have unique activity registration without worrying about duplicates.

### LINKED ISSUE #169: [Feature Request] Discuss project structure
From the Temporal community slack channel:

Now that we bundle Workflow code with Webpack we can have Typescript output a commonjs module instead of ES module.
This opens the door to simplifying the project structure, instead of 4 tsconfig files we can now have a single config file.
Doing this means sacrificing some type safety, for example, node types, WeakMap and WeakRef are not defined in the Workflow tsconfig file which should help newcomers avoid some easy to make mistakes.

Specifically these customizations have been made to the workflows `tsconfig.json`
```
      "lib": [
        "es5",
        "es2015.core",
        "es2015.generator",
        "es2015.iterable",
        "es2015.promise",
        "es2015.proxy",
        "es2015.reflect",
        "es2015.symbol",
        "es2015.symbol.wellknown",
        "es2016.array.include",
        "es2017.object",
        "es2017.intl",
        "es2017.sharedmemory",
        "es2017.string",
        "es2017.typedarrays"
      ],
      "typeRoots": [
        "."
      ],
```
(`Weak*` types and node types omitted)

When modifying the project to use a single TS config file, some noticeable differences are:
1.
```
import fs from 'fs';
```
In the composite project you get an error:
Cannot find module 'fs' or its corresponding type declarations
Not a compile time error in the modified project.

2.
```
new WeakMap();
```
In the composite project you get this error:
```
'WeakMap' only refers to a type, but is being used as a value here. Do you need to change your target library? Try changing the `lib` compiler option to es2015 or later.
```
No compile time error in the modified project.

When trying to use fs in the modified example the worker fails to start with:
```
2021-05-20T07:09:18.493Z [INFO]           using description file: /private/tmp/example/package.json (relative path: ./node_modules/fs)
2021-05-20T07:09:18.493Z [INFO]             no extension
2021-05-20T07:09:18.493Z [INFO]               Field 'browser' doesn't contain a valid alias configuration
2021-05-20T07:09:18.493Z [INFO]               /private/tmp/example/node_modules/fs doesn't exist
When instantiating a new WeakMap the user will get an error on Worker startup:
Error [DeterminismViolationError]: WeakMap cannot be used in workflows because v8 GC is non-deterministic
```

bergundy:
BREAKING CHANGE: `@activities` import does not work any more.
Activities registration is either done automatically with the `workDir`
option or by providing `activities` to `WorkerOptions`.

Closes #169 
Closes #109 
Closes #168


joebowbeer (commented):
See https://github.com/temporalio/proposals/pull/41#issuecomment-907820739

packages/workflow/src/workflow.ts
```diff
@@ -334,73 +330,75 @@ function signalWorkflowNextHandler({ seq, signalName, args, target }: SignalWork
...
+   * import { Context, ActivityInterface } from '@temporalio/workflow';
+   * import * as activities from '../activities';
    *
-   * const httpGetWithCustomTimeout = Context.configure(httpGet, {
+   * // Setup Activities from module exports
+   * const { httpGet, otherActivity } = Context.configureActivities<typeof activities>({
    *   type: 'remote',
-   *   scheduleToCloseTimeout: '30 minutes',
+   *   startToCloseTimeout: '30 minutes',
    * });
```
Sushisource, line 360: Would be nice to show setting different options for activities from the same import like in my question you answered in the proposal since that wasn't immediately obvious to some of us.

Sushisource (approved):
Looks much nicer

packages/workflow/src/workflow.ts
bergundy: I'll add that in another PR

packages/test/src/workflows/unblock-or-cancel.ts
```diff
@@ -1,9 +1,20 @@
...
 // @@@SNIPSTART nodejs-blocked-workflow
 import { Trigger, CancelledFailure } from '@temporalio/workflow';
 import { Blocked } from '../interfaces';
 
+let blocked = true;
 const unblocked = new Trigger<void>();
 
+const queries = {
+  isBlocked(): boolean {
+    return blocked;
```
vkarpov15, line 14: It might be neater if triggers had an `isBlocked()` method you could use instead of having to hard-code a `let blocked` variable.

bergundy: `Trigger`s can do more than just unblock the workflow but I agree we could expose the result of the trigger.
Something along the lines of:

```ts
const blocked = new Trigger<boolean>(true);
blocked.value; // true
blocked.resolve(false);
blocked.value; // false
// -- or --
blocked.reject(new Error('err'));
blocked.value // throws Error: err
```

PR #217, opened 2021-08-31
Internals interceptor

bergundy:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


packages/workflow/src/dependencies.ts
```diff
@@ -70,3 +70,14 @@ export enum ApplyMode {
...
 }
+
+/**
+ * Call information for async external dependencies
+ */
+export interface ExternalCall {
+  ifaceName: string;
+  fnName: string;
+  args: any[];
+  /** Optional in case applyMode is ASYNC_IGNORED */
```
Sushisource, line 81: Maybe expand on what one has to do with the other?

packages/workflow/src/interceptors.ts
```diff
@@ -137,19 +138,55 @@ export interface WorkflowOutboundCallsInterceptor {
...
+   *
+   * Returns a `boolean` indicating whether this job was processed or not
+   */
+  activate?(input: ActivateInput, next: Next<this, 'activate'>): Promise<boolean>;
+  /**
+   * Called after all `WFActivationJob`s have been proccesed for an activation.
+   *
+   * Can manipulate the commands generated by the Workflow
+   */
+  concludeActivation?(input: ConcludeActivationInput, next: Next<this, 'concludeActivation'>): ConcludeActivationOutput;
```
Sushisource, line 172: Should these be "activated" and "concluded"?

Sushisource (approved):

packages/workflow/src/dependencies.ts
bergundy: There's an entire page in the docs for this concept.

packages/workflow/src/interceptors.ts
bergundy: These are the names of the methods in the Workflow runtime, I'd rather keep them.

PR #218, opened 2021-09-01
OTel collector in node

Sushisource:
## What was changed
Allows us to export spans and metrics from core using the OTel collector format and then read them in node and re-export them to the user's specified exporters

## Why?
Unification of spans/metrics without needing a separate process

## Checklist

1. See https://github.com/temporalio/sdk-core/issues/85 -- is one half of that
 
2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/etc/prometheus.yml
+++ b/etc/prometheus.yml
@@ -1,0 +1,31 @@
+global:
+  scrape_interval: 15s
+  scrape_timeout: 10s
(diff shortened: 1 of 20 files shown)


Sushisource:
Closing in favor of https://github.com/temporalio/sdk-node/pull/223

PR #221, opened 2021-09-03
fix(samples): Update sample workflow.ts in create-project

joebowbeer:
## What was changed

Use Example workflow type

## Why?

It was probably imported for a reason.


bergundy:
@joebowbeer thanks for this but I don't think it's worth merging, I'm implementing https://github.com/temporalio/sdk-node/issues/181 which will change this sample completely.

PR #223, opened 2021-09-04
Export to an external OTel collector. Much easier to deal with.

Sushisource:
In contrast to https://github.com/temporalio/sdk-node/pull/218, simply just export spans/metrics to an OTel collector. Users will have to run one, but otherwise this is dramatically simpler and more maintainable.

I strongly recommend we go with this approach.

PR still draft b/c it needs some cleanup but all the pieces are there.


Sushisource:
Depends on https://github.com/temporalio/sdk-node/pull/223

packages/core-bridge/Cargo.toml
```diff
@@ -19,3 +19,8 @@ tokio = "1.4.0"
 once_cell = "1.7.2"
 # TODO: revert back to version *, can't now because of this error: prerelease package needs to be specified explicitly
 temporal-sdk-core = { version = "0.1.0-alpha.1", path = "./sdk-core" }
+temporal-sdk-core-protos = { version = "0.1.0", path = "./sdk-core/sdk-core-protos" }
+
+# Local testing w/ changes to sdk-core
+#temporal-sdk-core = { version = "0.1.0-alpha.1", path = "../../../sdk-core" }
```
bergundy, line 25: There should be a mechanism for overriding this, maybe https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#paths-overrides?
Or maybe with an env var?

packages/core-bridge/index.d.ts
```diff
@@ -65,6 +65,25 @@ export interface CoreOptions {
...
+   * OTel collector.
+   */
+  oTelCollectorUrl?: string
+  /**
+   * A string in the env filter format specified here:
+   * https://docs.rs/tracing-subscriber/0.2.20/tracing_subscriber/struct.EnvFilter.html
+   *
+   * Which determines what tracing data is collected in the Core SDK
+   */
+  tracingFilter?: string
```
bergundy, line 86: Is there a way to simplify this filter?
Are there multiple targets that need to be filtered?
It'd be nice if we could just set a log level or some other value that's nicer to use than `temporal_sdk_core=INFO`

packages/test/package.json
```diff
@@ -27,6 +27,7 @@
     "@temporalio/proto": "file:../proto",
     "@temporalio/worker": "file:../worker",
     "@temporalio/workflow": "file:../workflow",
+    "@opentelemetry/exporter-collector-grpc": "^0.24.0",
```
bergundy, line 30: The jaeger-exporter dependency should be removed from https://github.com/temporalio/sdk-node/blob/bdbd63b75f341a8d0008b7c92686a7259bf11411/package.json#L10

bergundy (commented):
I love how simple this turned out.
My main issue with this approach is that we now depend on the opentelemetry collector for getting core logs, metrics and traces.
I thought there'd be a way to let users plug in their own metrics / logging solution as we discussed in the proposal: https://github.com/temporalio/proposals/pull/23.

packages/core-bridge/index.d.ts
Sushisource: It's easy enough to add that by just transforming simple CLI options into the appropriate string. Being able to set this gives a lot of control over showing debug output for things like the underlying GRPC library if needed.

.buildkite/pipeline.yml
```diff
@@ -7,4 +7,6 @@ steps:
     plugins:
       - docker-compose#v3.8.0:
           run: load-test
-          config: .buildkite/docker/docker-compose.yaml
+          config:
+            - .buildkite/docker/docker-compose.yaml
```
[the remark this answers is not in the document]
Sushisource: Use it how?

Sushisource:
> I love how simple this turned out.
> My main issue with this approach is that we now depend on the opentelemetry collector for getting core logs, metrics and traces.
> I thought there'd be a way to let users plug in their own metrics / logging solution as we discussed in the proposal: [temporalio/proposals#23](https://github.com/temporalio/proposals/pull/23).

There's still some more to do. Forwarding logging should be pretty simple to allow users to do however they want in lang. But, metrics and traces not so much, that's why this landed here. There's just not a reasonable way to forward them, it's an enormous maintenance burden. We could expose customization points in core too, and allow users to make sure both halves point at the same thing, but I'm not sure anyone's going to prefer that over just using the collector. There's still room to iterate on this further but I think for beta expecting the collector for metrics/traces is a great starting point.

packages/core-bridge/Cargo.toml
Sushisource: The path overrides thing works, but then you've still got to toggle them in there when switching back and forth... Kinda same difference either way

bergundy:
The difference is that with overrides you can’t accidentally commit to git,
there’s no cleanup step.
I’d recommend using that and document how to achieve it.

package.json
```diff
@@ -39,7 +38,7 @@
     "build.watch": "lerna run --stream build.watch",
     "test": "lerna run --stream test",
     "test.watch": "lerna run --stream test.watch",
-    "ci-load": "node ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --iterations 10000",
+    "ci-load": "node ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --iterations 10000 --otel-url=''",
```
bergundy, line 41: Why is the otel-url empty? Do we not want to export anything?

packages/core-bridge/src/lib.rs
```diff
@@ -201,6 +201,9 @@ fn start_bridge_loop(
                         CoreInitError::TonicTransportError(err) => {
                             TRANSPORT_ERROR.from_error(cx, err)
                         }
+                        CoreInitError::TelemetryInitError(err) => {
+                            Ok(JsError::error(cx, format!("Error initializing telemetry: {:?}", err))?.upcast())
```
bergundy, line 205: We should add a typed error for this (or reuse one that we already have).

packages/test/src/load/args.ts
```diff
@@ -59,6 +59,7 @@ export const workerArgSpec: WorkerArgSpec = {
   '--max-concurrent-wft-polls': Number,
   '--log-level': String,
   '--server-address': String,
+  '--otel-url': String,
```
bergundy, line 62: This needs to be added to the type definition above.

packages/test/src/load/worker.ts
```diff
@@ -53,7 +61,7 @@ async function main() {
   console.log('Created worker');
 
   await worker.run();
-  await otel.shutdown();
+  await otel.shutdown().catch(console.error);
```
bergundy, line 64: This should be logged by the catch call below, do we want to ignore errors here?

packages/worker/src/core.ts
```diff
@@ -8,18 +8,27 @@ import {
...
 
 export interface CompiledCoreOptions extends CoreOptions {
   /** Options for communicating with the Temporal server */
   serverOptions: CompiledServerOptions;
 }
 
+function defaultTelemetryOptions(): TelemetryOptions {
+  return {
+    // Real default handled by core
+    tracingFilter: '',
```
bergundy, line 28: Might as well leave this undefined

bergundy (commented):
Had some more comments, nothing major.
We're almost there.

package.json
Sushisource: It's default curing CI because there's nowhere to look at what's exported.

packages/test/src/load/worker.ts
Sushisource: Yeah. The OTel library will error out if the collector is unavailable on shutdown which seems undesirable. I could go either way on it.

PR #224, opened 2021-09-06
Create projects from samples-node repo

lorensr:
Used Next's tool as a starter, as it has a number of nice features. I haven't tested all of them yet, but the basic command works. Had to change the tsconfig root (because code reads from package.json), which results in the weird `lib/src/...` structure. Would like to add the ability to list available examples.

image

---

It also prompts for name and sample when they're not included:

image

---

And prints what's needed when you `Ctrl-C`:

image

---
image


packages/create-project/src/helpers/is-online.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  try {
+    const httpsProxy = execSync('npm config get https-proxy').toString().trim();
+    return httpsProxy !== 'null' ? httpsProxy : undefined;
+  } catch (e) {
+    return;
+  }
+}
+
+export function getOnline(): Promise<boolean> {
```
bergundy, line 18: Document please

packages/create-project/src/helpers/should-use-yarn.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function shouldUseYarn(): boolean {
+  try {
+    const userAgent = process.env.npm_config_user_agent;
+    if (userAgent) {
+      return Boolean(userAgent && userAgent.startsWith('yarn'));
+    }
+    execSync('yarnpkg --version', { stdio: 'ignore' });
+    return true;
+  } catch (e) {
```
bergundy, line 11: Please check that this throws on unexpected error

packages/create-project/src/helpers/should-use-yarn.ts
```diff
@@ -0,0 +1,14 @@
+import { execSync } from 'child_process';
+
+export function shouldUseYarn(): boolean {
```
bergundy, line 3: Please document where this code came from

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
+  .allowUnknownOption()
+  .parse(process.argv);
+
+async function run(): Promise<void> {
+  if (typeof projectPath === 'string') {
+    projectPath = projectPath.trim();
+  }
+
+  if (!projectPath) {
+    const res = await prompts({
```
bergundy, line 56: Interesting choice going interactive, I guess some devs prefer that.

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
 
-class UsageError extends Error {
-  public readonly name: string = 'UsageError';
-}
+const program = new Commander.Command(packageJson.name)
+  .version(packageJson.version, '-v, --version')
+  .arguments('<project-directory>')
+  .usage(`${chalk.green('<project-directory>')} [options]`)
+  .action((name) => {
+    projectPath = name;
```
bergundy, line 19: Is there a way to do this without this weird global variable?
I've used [`arg`](https://www.npmjs.com/package/arg) before, it's less magic and is dependency free.

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
-  const { _: args, ...opts } = arg({
-    '--use-yarn': Boolean,
-    '--temporal-version': String,
-    '--sample': String,
-  });
-  if (args.length !== 1) {
-    throw new UsageError();
+  let example = program.example;
+
+  // `example` is true when --example is used by itself
```
bergundy, line 102: It's weird that this is even possible with `commander`.

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
+  if (typeof example !== 'string') {
+    const res = await prompts({
+      type: 'text',
+      name: 'example',
+      message: 'Which template would you like to use?',
+      initial: 'hello-world',
+      // validate: (name) => {
```
bergundy, line 109: Leftover?

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
-    throw new UsageError();
+  let example = program.example;
+
+  // `example` is true when --example is used by itself
+  if (typeof example !== 'string') {
+    const res = await prompts({
+      type: 'text',
+      name: 'example',
+      message: 'Which template would you like to use?',
+      initial: 'hello-world',
```
bergundy, line 108: Can we suggest a list of templates?

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
+
+  if (typeof example !== 'string') {
+    console.log();
+    console.log('Please specify which example:');
+    console.log(`  ${chalk.cyan(program.name())} --example ${chalk.green('[name]|[github-url]')}`);
+    console.log();
+    console.log('For example:');
+    console.log(`  ${chalk.cyan(program.name())} --example ${chalk.green('hello-world')}`);
+    console.log();
+    console.log(`Run ${chalk.cyan(`${program.name()} --help`)} to see all options.`);
```
bergundy, line 131: Should we use `console.error`? It's common practice in CLIs to log errors to stderr.

packages/create-project/src/helpers/is-folder-empty.ts
```diff
@@ -0,0 +1,55 @@
+import chalk from 'chalk';
+import fs from 'fs';
+import path from 'path';
+
+export function isFolderEmpty(root: string, name: string): boolean {
```
bergundy, line 5: Why don't we just require that the target path doesn't exist?
I'm wondering where this is coming from.

packages/create-project/src/helpers/is-folder-empty.ts
```diff
@@ -0,0 +1,55 @@
...
+    'npm-debug.log',
+    'yarn-debug.log',
+    'yarn-error.log',
+  ];
+
+  const conflicts = fs
+    .readdirSync(root)
+    .filter((file) => !validFiles.includes(file))
+    // Support IntelliJ IDEA-based editors
+    .filter((file) => !/\.iml$/.test(file));
```
bergundy, line 31: nit: could have been merged into a single filter function.

packages/create-project/src/helpers/install.ts
```diff
@@ -0,0 +1,108 @@
...
+      args.push(...npmFlags);
+    }
+    /**
+     * Spawn the installation process.
+     */
+    const child = spawn(command, args, {
+      stdio: 'inherit',
+      env: { ...process.env, ADBLOCK: '1', DISABLE_OPENCOLLECTIVE: '1' },
+    });
+    child.on('close', (code) => {
```
bergundy, line 100: AFAIR you might get an error event from child, [this](https://github.com/temporalio/sdk-node/blob/bdbd63b75f341a8d0008b7c92686a7259bf11411/packages/create-project/src/subprocess.ts#L26) helper should catch any errors.

packages/create-project/src/helpers/install.ts
```diff
@@ -0,0 +1,108 @@
...
+        args.push(...dependencies);
+      }
+    } else {
+      /**
+       * If there are no dependencies, run a variation of `{displayCommand}
+       * install`.
+       */
+      args = ['install'];
+      if (useYarn) {
+        if (!isOnline) {
```
bergundy, line 72: Why do we bother testing if the user is online?
Does this provide better DX over just letting npm or yarn detect that?

packages/create-project/src/helpers/install.ts
```diff
@@ -0,0 +1,108 @@
...
+        args = ['add', '--exact'];
+        if (!isOnline) args.push('--offline');
+        args.push('--cwd', root);
+        if (devDependencies) args.push('--dev');
+        args.push(...dependencies);
+      } else {
+        /**
+         * Call `npm install [--save|--save-dev] ...`.
+         */
+        args = ['install', '--save-exact'];
```
bergundy, line 61: `--save-exact`? what's the reason for this?

packages/create-project/src/helpers/install.ts
```diff
@@ -0,0 +1,108 @@
...
+   */
+  useYarn: boolean;
+  /**
+   * Indicate whether there is an active Internet connection.
+   */
+  isOnline: boolean;
+  /**
+   * Indicate whether the given dependencies are devDependencies.
+   */
+  devDependencies?: boolean;
```
bergundy, line 16: Maybe call this `saveAsDev` to avoid confusion with the package.json mapping field?

packages/create-project/src/helpers/git.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-empty */
+import { execSync } from 'child_process';
+import path from 'path';
+import rimraf from 'rimraf';
+
+function isInGitRepository(): boolean {
+  try {
+    execSync('git rev-parse --is-inside-work-tree', { stdio: 'ignore' });
```
bergundy, line 8: Does this return `true` if I have a `.git` directory in my home folder and I try to create a project in `~/my-app`?

packages/create-project/src/helpers/git.ts
```diff
@@ -0,0 +1,48 @@
...
+    return true;
+  } catch (_) {}
+  return false;
+}
+
+function isInMercurialRepository(): boolean {
+  try {
+    execSync('hg --cwd . root', { stdio: 'ignore' });
+    return true;
+  } catch (_) {}
```
bergundy, line 18: This could hide potential errors

packages/create-project/src/helpers/git.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable no-empty */
+import { execSync } from 'child_process';
+import path from 'path';
+import rimraf from 'rimraf';
+
+function isInGitRepository(): boolean {
+  try {
+    execSync('git rev-parse --is-inside-work-tree', { stdio: 'ignore' });
+    return true;
+  } catch (_) {}
```
bergundy, line 10: This could hide potential unexpected errors

packages/create-project/src/helpers/examples.ts
```diff
@@ -0,0 +1,74 @@
...
+    const infoResponse = await got(`https://api.github.com/repos/${username}/${name}`).catch((e) => e);
+    if (infoResponse.statusCode !== 200) {
+      return;
+    }
+    const info = JSON.parse(infoResponse.body);
+    return { username, name, branch: info['default_branch'], filePath };
+  }
+
+  // If examplePath is available, the branch name takes the entire path
+  const branch = examplePath ? `${_branch}/${file.join('/')}`.replace(new RegExp(`/${filePath}|/$`), '') : _branch;
```
bergundy, line 36: `filePath` should be escaped or it could break the regexp.

packages/create-project/src/create-project.ts
```diff
@@ -0,0 +1,188 @@
...
+      if (repoUrl.origin !== 'https://github.com') {
+        console.error(
+          `Invalid URL: ${chalk.red(
+            `"${example}"`
+          )}. Only GitHub repositories are supported. Please use a GitHub URL and try again.`
+        );
+        process.exit(1);
+      }
+
+      repoInfo = await getRepoInfo(repoUrl, examplePath);
```
bergundy, line 57: Instead of returning undefined, you can just let this throw and log in the `catch` clause.

packages/create-project/src/create-project.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  if (!(await isWriteable(path.dirname(root)))) {
+    console.error('The application path is not writable, please check folder permissions and try again.');
+    console.error('It is likely you do not have write permissions for this folder.');
+    process.exit(1);
+  }
+
+  const appName = path.basename(root);
+
+  await makeDir(root);
```
bergundy, line 98: Is this intentional that we allow overwriting an existing folder?

packages/create-project/src/create-project.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  const isOnline = !useYarn || (await getOnline());
+  const originalDirectory = process.cwd();
+
+  const displayedCommand = useYarn ? 'yarn' : 'npm';
+  console.log(`Creating a new Temporal project in ${chalk.green(root)}/.`);
+  console.log();
+
+  await makeDir(root);
```
bergundy, line 110: We just created the directory above.

packages/create-project/src/create-project.ts
```diff
@@ -0,0 +1,188 @@
...
+  process.chdir(root);
```
bergundy, line 111: I usually prefer not to rely on being in a specific directory, it doesn't work when you want to run things concurrently.

bergundy (commented):
Great work, looks like you've put a lot of thought into it.
I really like that the errors are clearly logged and provide instructions on how to recover.

packages/create-project/license.md
```diff
@@ -0,0 +1,23 @@
+Some of the code in this directory is subject to the below license:
+
+The MIT License (MIT)
+
+Copyright (c) 2021 Vercel, Inc.
```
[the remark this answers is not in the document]
lorensr: Since it's based on Next.js's code, I believe we have to include this license somewhere? Not sure what the best location is.

packages/create-project/src/helpers/install.ts
lorensr: Yes, we can provide a clearer error message—npm just hangs with an empty progress bar. Yarn is better but could be clearer.

image

packages/create-project/src/index.ts
lorensr: It feels nice to me when tools have prompts, and a lot of the modern/popular tools I've used do. It's used only as a fallback here when command-line args aren't provided.

packages/create-project/src/helpers/is-folder-empty.ts
lorensr: I guess it covers the case in which someone habitually starts out a new project by creating a dir? Like:

```
mkdir myapp
```

"okay, now maybe there's a template I can use? oh there's this tool"

```
npx @temporalio/create myapp
```

packages/create-project/src/index.ts
lorensr: It's an old version of commander. I briefly tried latest, which didn't work, and I didn't know if it would be worth upgrading.

packages/create-project/src/helpers/git.ts
lorensr: Yes, in which case we won't `git init` for them. Don't see a way to differentiate between home folder case and adding temporal inside an existing code repo.

packages/create-project/src/create-project.ts
lorensr: We use an existing empty folder but exit if the existing folder has non-negligible contents.

packages/create-project/src/index.ts
bergundy: 👍

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
+  try {
+    const res = await update;
+    if (res?.latest) {
+      const isYarn = shouldUseYarn();
+
+      console.log();
+      console.log(chalk.yellow.bold('A new version of `@temporalio/create` is available!'));
+      console.log(
+        'You can update by running: ' +
+          chalk.cyan(isYarn ? 'yarn global add @temporalio/create' : 'npm i -g @temporalio/create')
```
bergundy, line 155: Does this guarantee that the package gets upgraded?
Also, we never ask users to install this package directly, maybe suggest running with `npx @temporalio/create@latest ...`?

packages/create-project/src/index.ts
```diff
@@ -1,151 +1,181 @@
...
 
-async function writePrettyJson(path: string, obj: any) {
-  await writeFile(path, JSON.stringify(obj, null, 2) + os.EOL);
-}
+const program = new Commander.Command(packageJson.name)
+  .version(packageJson.version, '-v, --version')
+  .arguments('<project-directory>')
+  .usage(`${chalk.green('<project-directory>')} [options]`)
+  .option(
+    '--use-yarn',
```
bergundy, line 20: This option is disregarded in most of the code here, consider dropping it.

packages/create-project/src/create-project.ts
```diff
@@ -0,0 +1,189 @@
...
+  useYarn,
+  example,
+  examplePath,
+}: {
+  appPath: string;
+  useYarn: boolean;
+  example: string;
+  examplePath?: string;
+}): Promise<void> {
+  let repoInfo: RepoInfo | undefined;
```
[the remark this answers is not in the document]
lorensr: It's used in multiple `try` clauses

packages/create-project/src/index.ts
lorensr: When the option is included, `yarn install` is used, a `yarn.lock` file is created, and the suggested commands use `yarn`. I don't know anyone with a strong yarn preference, but I imagine they're out there, in which case them not having to delete the `package-lock.json` file and run `yarn` themselves would be better UX? 

image

packages/create-project/package.json
```diff
@@ -14,16 +14,36 @@
...
+    "commander": "^2.20.0",
+    "cpy": "^8.1.2",
+    "cross-spawn": "^6.0.5",
+    "fs-extra": "^10.0.0",
+    "got": "^10.7.0",
+    "prompts": "2.1.0",
+    "rimraf": "^3.0.0",
+    "tar": "^4.4.10",
+    "update-check": "1.5.4",
+    "validate-npm-package-name": "^3.0.0"
```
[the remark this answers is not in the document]
bergundy: @lorensr this is not a blocker but I think it'd be NTH

packages/create-project/src/helpers/git.ts
bergundy: As discussed f2f, let's prompt the user for creating the git directory and overwriting an existing directory.

It might also be worth adding a flag to prevent creating the git repository for non-git users.

packages/create-project/src/index.ts
bergundy: But I think `shouldUseYarn` doesn't check the flag and the `notifyUpdate` will show an npm command

packages/create-project/src/index.ts
bergundy: I think it's worth upgrading now rather than later.
If you use angle brackets in the flag definition you should only get a valid `string | undefined` and not boolean: https://www.npmjs.com/package/commander

packages/create-project/tsconfig.json
```diff
@@ -2,8 +2,10 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./lib",
-    "rootDir": "./src"
+    "rootDir": "./",
```
bergundy, line 5: Does this cause generated JS files to be placed in `lib/src/`?
You can avoid this by asking typescript not to include package.json and bypassing some checks:
https://github.com/temporalio/sdk-node/blob/main/packages/worker/src/pkg.ts

bergundy (commented):
There are still some comments that need to be addressed, let's try to handle those soon and merge.

lorensr:
> There are still some comments that need to be addressed, let's try to handle those soon and merge.

Yep, I'll have time to get back to this starting 25h from now (at first I wrote "tomorrow night" but that might be less helpful with a globally distributed team 😄)

bergundy:
Cool, that only 16 hours away :)

packages/create-project/package.json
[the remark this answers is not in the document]
lorensr: Done! I was just waiting until the end so I knew what code was left ☺️

lorensr:
Should I add the below to `.gitignore` or add `packages/create-project/package-lock.json` to git?

```
packages/*/package-lock.json
```

packages/create-project/src/index.ts
lorensr: Sounds good

bergundy:
@lorensr please fix the tests and we'll merge.
We probably want the `test-npm-init` jobs to use a specific tag in the samples repo.
That way we can update the samples before we update this repo with anything backwards-incompatible.

lorensr:
image

None of these errors show up when I build or lint locally. Lmk if you have any ideas—I'll look into it deeper tomorrow

bergundy:
@lorensr I'd run:
```
npx lerna clean -y && rm -rf package-lock.json node_modules && npm i
```

and recommit `package-lock.json`.

lorensr:
Thanks, that fixed the typescript CI errors!

Should we collect flaky tests in a GH issue?

https://github.com/temporalio/sdk-node/pull/224/checks?check_run_id=3612941219

image

Any tips for debugging Github actions other than adding logging, pushing, and waiting? I tried using act in a couple ways but ran into errors https://github.com/nektos/act

lorensr:
another flake https://github.com/temporalio/sdk-node/pull/224/checks?check_run_id=3614424338

image

lorensr:
Current issue is "No such file or directory" when trying to build inside `/tmp/registry/example`:

image

The directory should be there. And when `test-npm-init` @ 16 finishes before 14, the directory is there:

image

This line is in 14 but not 16:

```
 warn --- config file  - /tmp/registry/verdaccio.yaml
```

.github/workflows/ci.yml
```diff
@@ -131,9 +131,9 @@ jobs:
           name: packages
           path: /tmp/registry/storage
       # The 2 steps below run npm init @temporalio from a local registry with different samples
-      - run: node scripts/init-from-verdaccio.js --registry-dir /tmp/registry
+      - run: node scripts/init-from-verdaccio.js --registry-dir /tmp/registry --example hello-world
```
bergundy, line 134: I'm ok with this rename, just curios about the reason behind it.
I would go with `--sample` because the source repo is called `samples-node`.
Not critical though and I wouldn't delay this PR for it.

bergundy (approved):
Since we can't avoid committing `lib/index.js` I suggest we make a simple JS wrapper at `lib.index.js`, and rename `index.ts` to `cli.ts`.
That way we don't commit generated code at all and everything works.

`lib/index.js`
```js
const { run } = require('./cli');
run();
```

bergundy:
> Current issue is "No such file or directory" when trying to build inside `/tmp/registry/example`:

Which file or directory?

bergundy:
> Should we collect flaky tests in a GH issue?

There's already an issue for this flake, opened another one for the other flake you found.
 
> Any tips for debugging Github actions other than adding logging, pushing, and waiting? I tried using act in a couple ways but ran into errors https://github.com/nektos/act

GH actions just runs scripts, I usually just run the scripts directly on my laptop, e.g. `RUN_INTEGRATION_TESTS=1 npm test`.

.github/workflows/ci.yml
lorensr: Not intentional—it was what Next uses. Will change

lorensr:
In CI, `npm ci` is still trying to chmod `lib/index.js`, even though `create-project/package.json`'s `bin` points elsewhere:

```js
{
  "name": "@temporalio/create",
  "version": "0.5.0",
  "description": "Create a Temporal project from template",
  "main": "cli.js",
  "bin": "cli.js",
```

Any idea why? Locally, `npm ci` runs fine. 

image

bergundy:
@lorensr I bet it’s something in package lock

bergundy:
I figured out why the first `npm init` fails (the one with `-v`, it's because you're running it in the wrong directory.
I'm trying to figure out the other issue in this PR: #260

bergundy:
Also figured out that we were installing the wrong version of `@temporalio/create` because of the `-v` command that did not specify the correct registry.

bergundy:
@lorensr checkout my branch for the fix (requires cleanup).
I bet we can use `npm init @temporalio` but I didn't try it after I figured out that we were installing the wrong version.

lorensr:
@bergundy Thanks! It now downloads hello-world to `/tmp/registry/example`, but fails on `npm install` inside that dir:

image

`npm` should be configured to:

1. ask the verdaccio registry for `temporalio@0.4.2`, and if verdaccio doesn't have it,
2. verddaccio proxies it from npm's registry? 

In step 1, verdaccio should have it, because we successfully ran `'npx',    ['lerna', 'publish', 'from-package', '--yes', '--registry', 'http://localhost:4873/'],`, which includes the meta package?

lorensr:
image

Ah, we're at `0.4.3`. I guess verdaccio just isn't doing step 2 for `temporalio`. Which is fine, because we want to be testing against the current version. I think this particular issue can be fixed by changing [this line](https://github.com/temporalio/samples-node/blob/main/hello-world/package.json#L12) to `^0.4.2`, but that doesn't work for breaking changes. What update flow do we want for breaking changes? For instance, we could point sdk-node CI tests at `next` branch of `samples-node`, and then the flow is:

- Currently sdk-node temporalio@0.4.3
- Update `next` branch on `samples-node` to use (non-published) `temporalio@^0.5.0`
- Create a `new-feature` branch on `sdk-node` and set version to `0.5.0`
- Open PR

Not ideal, as now all the PRs opened w/ version `0.4.3` will fail. We could point sdk-node CI tests at a dynamic branch that matched the version number, like `0.5.0`, in which case flow is:

- Currently sdk-node temporalio@0.4.3
- Create `0.5.0` branch on `samples-node` that uses (non-published) `temporalio@0.5.0`
- Create a `new-feature` branch on `sdk-node` and set version to `0.5.0`
- Open PR

Or we could move samples-node to `sdk-node/samples`, which sounds much easier 😁

sw-yx:
the monorepo comes for us all 🙈  no objections to moving it if it simplifies

bergundy:
@lorensr I'd remove `temporalio` from the sample's package.json and run `npm install temporalio`, this will get the proper version from the local npm registry for the test.
That's what we did before this PR.

lorensr:
Merged in #273

PR #225, opened 2021-09-07
New Workflow API

### LINKED ISSUE #181: [strawman] React-like API proposal
just sketching out what a React-like API might look like. it might help solve some of our API design challenges (by starting from something already proven) and our adoption might be helped GREATLY since most JS people know React.

where we depart from React: 
- React doesn't model Sagas
- React doesn't need hook registration in advance
- React doesn't allow external invocations.

other design considerations: 
- flexible file and folder structure as much as possible.
- type safety without poor JS dev experience

## Workflows and Activities

```tsx
import { Activity, Workflow } from 'fakeTemporal'

export const activities = {
	async chargeStripe(amount) {
		res = await call('/charge', amount)
		return res // todo: handle err case
	}
	async reverseCharge(id) {
		res = await call('/reverse', id)
		return res // todo: handle err case
	}
	async function chargeStripe(amount) {
		res = await call('/charge', amount)
		return res // todo: handle err case
	}
} 


type ValidStatuses = 'EMAIL' | 'COMPLETE' | 'CANCELED'
const invoke = (activity, ...args) = {
	// convenience function; users can call Activity.create directly if they need to customize/override these
	return Activity.create({
				activity, 
				options: {
					startToCloseTimeout: 1000 // this is the default if options not specified
				}
			})(...args)
}
export class MyWorkflow extends Workflow <WFInterface> {
	status: ValidStatuses = 'EMAIL' // all non private variables are queryable
	activities = activities // a workflow registers its own activities by convenience
	signals = {
		cancelTransaction() {
			if (status === 'COMPLETE') {
				return 'Transaction already completed, cannnot be canceled"
			} else {
				status = 'CANCELED'
			}
		}
	}
	async execute(amount) {
		const transactionId = invoke(activities.chargeStripe, amount)
		sleep(3000) // wait for cancel
		if (status === 'CANCELED') {
			await invoke(activities.reverseCharge, amount)
			status = 'COMPLETE'
		} else {
			await invoke(activities.emailUser, transactionId)
		}
	}
}

// usage
import { Client } from 'fakeTemporal'

const client = new Client({ 
	register: MyWorkflow, // register both WF and associated activity
	taskQueue: 'my-workflow' // defaults to `node-workflow`
})
client.execute(amount)
console.log(client.query.status) // 'EMAIL'
client.signal.cancelTransaction() // cancel
console.log(client.query.status) // 'CANCEL'
```

## Infinite Workflows

Infinite Workflows have no `execute`, they are always live and `continueAsNew` under the hood. We might also call them `Entity`s

```tsx
import { Activity, InfiniteWorkflow } from 'fakeTemporal'

// as above
export class MyWorkflow extends InfiniteWorkflow <WFInterface> {
	status: ValidStatuses = 'WAITING'
	activities = activities // a workflow registers its own activities by convenience
	signals = {
		cancelTransaction() {
			status = 'CANCELED'
		}
		startTransaction() {
			const transactionId = invoke(activities.chargeStripe, amount)
			sleep(3000) // wait for cancel
			if (status === 'CANCELED') {
				await invoke(activities.reverseCharge, amount)
				status = 'WAITING'
			} else {
				await invoke(activities.emailUser, transactionId)
			}		
		}
	}
}

// usage
import { Client } from 'fakeTemporal'

const client = new Client({ 
	register: MyWorkflow, // register both WF and associated activity
})
client.signal.startTransaction() // started
console.log(client.query.status) // 'WAITING'
client.signal.cancelTransaction() // canceled
console.log(client.query.status) // 'CANCELED'
```

bergundy:
Closes #181 

This should not be merged until the docs have been updated.


bergundy:
@vkarpov15 I agree with your assertions and I'm not sure what's the best way to solve this, maybe export some generic helper..
The workflow type doesn't have to be explicitly declared, I added it to all of the test workflows because eslint was complaining about missing type definition for exports.

This design is pretty flexible, for example, with classes and a proper function, you get:
```ts
export class MyWorkflow {
  constructor(arg1: string, arg2: number) { /* ... */ }
  async execute(): Promise<void> {
    // ...
  }
}

export function myWorkflow(arg1: string, arg2: number): MyWorkflow {
  return new MyWorkflow(arg1, arg2);
}
```

Or with an interface (as before) and function:
```ts
export interface MyWorkflowHandlers {
  execute(): Promise<void>;
}

export function myWorkflow(arg1: string, arg2: number): MyWorkflowHandlers {
  return {
    async execute() {
      // ...
    }
  };
}
```

lorensr:
I don't have a pref on the workflow being an object vs function that returns an object. I like the rest of the API changes. Do you mean to track this in git?

image

bergundy:
> I don't have a pref on the workflow being an object vs function that returns an object. I like the rest of the API changes. Do you mean to track this in git?
> 
> image

I originally added it so you can run the script from a git URL, it's causing some pain maintaining it because the permissions change every time I run `npm run clean -- --only ts`, I think you can remove it in your PR @lorensr .

sw-yx (approved):
this is looking like such a huuuge improvement! congrats on this massive PR.

left some minor api design commentary in the docs PR https://github.com/temporalio/documentation/pull/627

PR #231, opened 2021-09-09
Ability to import logs from core

Sushisource:
## What was changed
Import exported logs from core. Depends on https://github.com/temporalio/sdk-core/pull/189

## Why?
Easy-to-enable debuggability for users

## Checklist

1. Relates to https://github.com/temporalio/sdk-core/issues/85

2. How was this tested:
Integ tests

3. Any docs updates needed?
Probably need to have a little section on various telemetry / logging options

### DIFF
--- a/packages/core-bridge/index.d.ts
+++ b/packages/core-bridge/index.d.ts
@@ -1,4 +1,5 @@
 import { TLSConfig } from '@temporalio/common';
+import { LogLevel } from '@temporalio/worker';
 
 export { TLSConfig };
 
@@ -86,7 +87,7 @@ export interface TelemetryOptions {
   tracingFilter?: string;
   /** What level, if any, logs should be forwarded from core at */
   // These strings should match the log::LevelFilter enum in rust
-  logForwardingLevel: 'OFF' | 'ERROR' | 'WARN' | 'INFO' | 'DEBUG' | 'TRACE';
+  logForwardingLevel: 'OFF' | LogLevel;
 }
 
 export interface WorkerOptions {
@@ -125,9 +126,9 @@ export interface CoreLog {
   /** Unix millis since epoch. Ideally would be bigint but Neon can't send those
    */
   // TODO: Not meaningfully usable until logger interface has a timestamp concept
-  timestamp_millis: number;
-  /** Log level */
-  level: string;
+  timestampMillis: number;
+  /** Log level - must match rust log level names */
+  level: 'TRACE' | 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';
 }
 
 export interface Worker {}
--- a/packages/core-bridge/index.d.ts
+++ b/packages/core-bridge/index.d.ts
@@ -84,6 +84,9 @@ export interface TelemetryOptions {
    * Which determines what tracing data is collected in the Core SDK
    */
   tracingFilter?: string;
+  /** What level, if any, logs should be forwarded from core at */
+  // These strings should match the log::LevelFilter enum in rust
+  logForwardingLevel: 'OFF' | 'ERROR' | 'WARN' | 'INFO' | 'DEBUG' | 'TRACE';
 }
 
 export interface WorkerOptions {
@@ -116,6 +119,17 @@ export interface WorkerOptions {
   maxCachedWorkflows: number;
 }
 
+export interface CoreLog {
+  /** Log message */
+  message: string;
+  /** Unix millis since epoch. Ideally would be bigint but Neon can't send those
+   */
+  // TODO: Not meaningfully usable until logger interface has a timestamp concept
+  timestamp_millis: number;
+  /** Log level */
+  level: string;
+}
+
 export interface Worker {}
 export interface Core {}
 
@@ -123,6 +137,7 @@ export declare type PollCallback = (err: Error, result: ArrayBuffer) => void;
 export declare type WorkerCallback = (err: Error, result: Worker) => void;
 export declare type CoreCallback = (err: Error, result: Core) => void;
 export declare type VoidCallback = (err: Error, result: void) => void;
+export declare type LogsCallback = (err: Error, result: CoreLog[]) => void;
 
 // TODO: improve type, for some reason Error is not accepted here
 export declare function registerErrors(errors: Record<string, any>): void;
@@ -130,6 +145,7 @@ export declare function newCore(coreOptions: CoreOptions, callback: CoreCallback
 export declare function newWorker(core: Core, workerOptions: WorkerOptions, callback: WorkerCallback): void;
 export declare function workerShutdown(worker: Worker, callback: VoidCallback): void;
 export declare function coreShutdown(core: Core, callback: VoidCallback): void;
+export declare function corePollLogs(core: Core, callback: LogsCallback): void;
 export declare function workerPollWorkflowActivation(worker: Worker, callback: PollCallback): void;
 export declare function workerCompleteWorkflowActivation(
   worker: Worker,
--- a/packages/core-bridge/src/lib.rs
+++ b/packages/core-bridge/src/lib.rs
@@ -252,29 +252,20 @@ fn start_bridge_loop(
                             }
                             Request::PollLogs { callback } => {
                                 let logs = core.fetch_buffered_logs();
-                                if !logs.is_empty() {
-                                    send_result(event_queue.clone(), callback, |cx| {
-                                        let logarr = cx.empty_array();
-                                        for (i, cl) in logs.into_iter().enumerate() {
-                                            let logobj = cx.empty_object();
-                                            let level = cx.string(cl.level.to_string());
-                                            logobj.set(cx, "level", level).unwrap();
-                                            // eww, time conversion. Use neon date if possible in napi-5
-                                            let ts = cx.number(cl.millis_since_epoch() as f64);
-                                            logobj.set(cx, "timestamp_nanos", ts).unwrap();
-                                            let msg = cx.string(cl.message);
-                                            logobj.set(cx, "message", msg).unwrap();
-                                            logarr.set(cx, i as u32, logobj).unwrap();
-                                        }
-                                        Ok(logarr)
-                                    });
-                                } else {
-                                    // Dispose of unused root
-                                    event_queue.send(move |mut cx| {
-                                        callback.drop(&mut cx);
-                                        Ok(())
-                                    });
-                                }
+                                send_result(event_queue.clone(), callback, |cx| {
+                                    let logarr = cx.empty_array();
+                                    for (i, cl) in logs.into_iter().enumerate() {
+                                        let logobj = cx.empty_object();
+                                        let level = cx.string(cl.level.to_string());
+                                        logobj.set(cx, "level", level).unwrap();
+                                        let ts = cx.number(cl.millis_since_epoch() as f64);
+                                        logobj.set(cx, "timestampMillis", ts).unwrap();
+                                        let msg = cx.string(cl.message);
+                                        logobj.set(cx, "message", msg).unwrap();
+                                        logarr.set(cx, i as u32, logobj).unwrap();
+                                    }
+                                    Ok(logarr)
+                                });
                             }
                             Request::ShutdownWorker {
                                 task_queue,
@@ -631,13 +622,11 @@ fn core_new(mut cx: FunctionContext) -> JsResult<JsUndefined> {
         retry_config,
     };
 
-    let log_forwarding_level = LevelFilter::from_str(&js_value_getter!(
-        cx,
-        telem_options,
-        "logForwardingLevel",
-        JsString
-    ))
-    .unwrap_or(LevelFilter::Off);
+    let log_forwarding_level_str =
+        js_value_getter!(cx, telem_options, "logForwardingLevel", JsString)
+            .replace("WARNING", "WARN");
+    let log_forwarding_level =
+        LevelFilter::from_str(&log_forwarding_level_str).unwrap_or(LevelFilter::Off);
     let telemetry_opts = TelemetryOptions {
         otel_collector_url: get_optional(&mut cx, telem_options, "oTelCollectorUrl").map(|x| {
             Url::parse(
--- a/packages/core-bridge/src/lib.rs
+++ b/packages/core-bridge/src/lib.rs
@@ -1,8 +1,10 @@
 mod errors;
 
 use errors::*;
+use log::LevelFilter;
 use neon::prelude::*;
 use prost::Message;
+use std::str::FromStr;
 use std::{fmt::Display, future::Future, sync::Arc, time::Duration};
 use temporal_sdk_core::errors::{
     CompleteActivityError, CompleteWfError, CoreInitError, PollActivityError, PollWfError,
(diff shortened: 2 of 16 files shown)


packages/test/src/test-dependencies.ts
```diff
@@ -218,35 +213,15 @@ if (RUN_INTEGRATION_TESTS) {
...
-      recordedCalls,
-      thrownErrors.map((error) => ({
-        level: 'ERROR',
-        message: 'External dependency function threw an error',
-        meta: {
-          error,
-          workflowInfo: info,
-        },
-      }))
-    );
```
Sushisource, line 249: I figured it was fine to remove this since we're already checking the exceptions get thrown. This is the only test that wants to intercept the logger this way so it felt meh to have to keep that around somehow just for it.

packages/core-bridge/index.d.ts
```diff
@@ -116,20 +119,33 @@ export interface WorkerOptions {
   maxCachedWorkflows: number;
 }
 
+export interface CoreLog {
+  /** Log message */
+  message: string;
+  /** Unix millis since epoch. Ideally would be bigint but Neon can't send those
+   */
+  // TODO: Not meaningfully usable until logger interface has a timestamp concept
+  timestamp_millis: number;
```
bergundy, line 128: Please use camel case for JS/TS

bergundy: I wonder if the linter inspects `.d.ts` files.

packages/core-bridge/src/lib.rs
```diff
@@ -243,6 +250,32 @@ fn start_bridge_loop(
...
+                                            logobj.set(cx, "timestamp_nanos", ts).unwrap();
+                                            let msg = cx.string(cl.message);
+                                            logobj.set(cx, "message", msg).unwrap();
+                                            logarr.set(cx, i as u32, logobj).unwrap();
+                                        }
+                                        Ok(logarr)
+                                    });
+                                } else {
+                                    // Dispose of unused root
+                                    event_queue.send(move |mut cx| {
```
bergundy, line 273: I think callbacks should always be called otherwise you're left with unresolved promises in JS.

packages/core-bridge/src/lib.rs
```diff
@@ -243,6 +250,32 @@ fn start_bridge_loop(
...
+                                let logs = core.fetch_buffered_logs();
+                                if !logs.is_empty() {
+                                    send_result(event_queue.clone(), callback, |cx| {
+                                        let logarr = cx.empty_array();
+                                        for (i, cl) in logs.into_iter().enumerate() {
+                                            let logobj = cx.empty_object();
+                                            let level = cx.string(cl.level.to_string());
+                                            logobj.set(cx, "level", level).unwrap();
+                                            // eww, time conversion. Use neon date if possible in napi-5
+                                            let ts = cx.number(cl.millis_since_epoch() as f64);
```
bergundy, line 263: TBH I think it might be more convenient to work with the timestamps as numbers.
Dates are a bit of a pain in JS but that's what [temporal](https://tc39.es/proposal-temporal/docs/) is for.

bergundy: Also, it should be safe to use napi-5

packages/test/src/load/worker.ts
```diff
@@ -5,6 +5,7 @@ import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
 import { Core, Worker, DefaultLogger } from '@temporalio/worker';
 import { CollectorTraceExporter } from '@opentelemetry/exporter-collector-grpc';
 import { WorkerArgSpec, workerArgSpec, getRequired } from './args';
+import { TelemetryOptions } from "@temporalio/core-bridge";
```
bergundy, line 8: This probably failed the linter / formatter.

packages/test/src/test-dependencies.ts
bergundy: If only I'd documented why this is tested...

Let's remove this for now, we can add it back later.

packages/worker/src/core.ts
```diff
@@ -8,22 +8,33 @@ import {
...
 export interface CompiledCoreOptions extends CoreOptions {
   /** Options for communicating with the Temporal server */
   serverOptions: CompiledServerOptions;
+  logger: Logger;
 }
 
 function defaultTelemetryOptions(): TelemetryOptions {
-  return {};
+  return {
+    logForwardingLevel: 'OFF'
```
bergundy, line 36: Should this be OFF by default?
I thought that we want to forward the logs to unify the log formats we see in console (at least for development).

packages/worker/src/core.ts
```diff
@@ -58,12 +69,40 @@ export class Core {
...
+    this.initLogPolling(options, native);
+
     return new this(native, compiledOptions);
   }
 
+  private static initLogPolling(options: CoreOptions, native: native.Core) {
+    if (options.logger != null) {
+      interval(3)
+        .pipe(takeUntil(this.shouldPollForLogs.pipe(filter((v) => v === false))))
+        .subscribe((_) => {
```
bergundy, line 84: This doesn't allow catching any errors, let me think about the right way to do this some more, don't have much time left today.

packages/worker/src/core.ts
```diff
@@ -58,12 +69,40 @@ export class Core {
...
+          corePollLogs(native, (_err, logs) => {
+            for (const log of logs) {
+              if (log.level === 'DEBUG') {
+                options.logger?.debug(log.message);
```
bergundy, line 88: logger is not optional in the compiled options, better use those.

packages/worker/src/core.ts
```diff
@@ -58,12 +69,40 @@ export class Core {
...
+              } else if (log.level === 'INFO') {
+                options.logger?.info(log.message);
+              } else if (log.level === 'WARN') {
+                options.logger?.warn(log.message);
+              } else if (log.level === 'ERROR') {
+                options.logger?.error(log.message);
+              } else {
```
bergundy, line 95: This shouldn't be possible.

packages/core-bridge/index.d.ts
```diff
@@ -116,20 +119,33 @@ export interface WorkerOptions {
...
+  /** Log level */
+  level: string;
```
bergundy, line 130: Use this: https://github.com/temporalio/sdk-node/blob/dfb1c3dfc7eace3e8ec5d3776b4f671f6633bc57/packages/worker/src/logger.ts#L11

packages/worker/src/core.ts
```diff
@@ -58,12 +69,40 @@ export class Core {
...
+                options.logger?.info(log.message);
+              }
+            }
+          });
+        });
+    }
+  }
+
   protected static _instance?: Promise<Core>;
+  protected static shouldPollForLogs = new BehaviorSubject<boolean>(false);
```
bergundy, line 105: Dynamic config?
I probably would attach this as a static attribute, I'd put it on the instance.

packages/worker/src/core.ts
```diff
@@ -58,12 +69,40 @@ export class Core {
...
     const native = await promisify(newCore)(compiledOptions);
+    this.initLogPolling(options, native);
+
     return new this(native, compiledOptions);
   }
 
+  private static initLogPolling(options: CoreOptions, native: native.Core) {
+    if (options.logger != null) {
+      interval(3)
+        .pipe(takeUntil(this.shouldPollForLogs.pipe(filter((v) => v === false))))
```
bergundy, line 83: This doesn't support resuming, I can show you how to add that in rxjs.

bergundy (commented):
Awesome, I see your Typescript / rxjs level === `expert` :)

packages/worker/src/core.ts
Sushisource: Good point. For me I prefer having the extra info from the spans you get with the rust logging, but probably not for most people. Changed to `INFO` and I'll just override locally.

packages/core-bridge/src/lib.rs
Sushisource: I'll just keep it as is then

packages/core-bridge/src/lib.rs
Sushisource: K, I'll just call it with the empty array then

packages/worker/src/core.ts
Sushisource: Does it need to be paused/resumed? It's on until core is gone, then on if you make a new core.

packages/worker/src/core.ts
bergundy: So after thinking about this a bit more, the main issue in this implementation apart from the lack of error handling is that nothing is waiting on `corePollLogs`, this means that we might issue more than one concurrent poll.
It'd have been better to use [`concatMap`](https://rxjs-dev.firebaseapp.com/api/operators/concatMap) on the promisified version of `corePollLogs` and make sure that we don't overpoll if lang is falling behind:

```ts
const poll = promisify(corePollLogs);
this.logPollPromise = lastValueFrom(
  of(this.shouldPollForLogs).pipe(
    map((subject) => subject.getValue()),
    takeWhile((shouldPoll) => shouldPoll),
    concatMap(() => poll(native)),
    map((logs) => { /* logging happens here */ }),
    repeat(),
  )
);
this.logPollPromise.reject(() => undefined); // Prevent unhandled rejection
```

Then each worker can subscribe to `logPollPromise` and propagate errors to `the Worker.run` promise.

This is all assuming that errors in logging are all unexpected errors (the process should be aborted).

packages/worker/src/core.ts
```diff
@@ -58,11 +72,36 @@ export class Core {
...
+  private initLogPolling(options: CompiledCoreOptions, native: native.Core) {
+    this.shouldPollForLogs.next(true);
+    if (options.logger != null) {
+      interval(3)
+        .pipe(takeUntil(this.shouldPollForLogs.pipe(filter((v) => v === false))))
+        .subscribe((_) => {
+          corePollLogs(native, (_err, logs) => {
+            for (const log of logs) {
+              if (log.level === 'DEBUG' || log.level === 'TRACE') {
+                options.logger.debug(log.message);
```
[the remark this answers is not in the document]
Sushisource: Sure, let's just add that

packages/test/src/mock-native-worker.ts
```diff
@@ -176,18 +173,6 @@ export const defaultOptions: WorkerOptions = {
   taskQueue: 'test',
 };
 
-export async function makeDefaultWorker(): Promise<Worker> {
```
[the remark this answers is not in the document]
Sushisource: Yes

packages/core-bridge/src/lib.rs
```diff
@@ -243,6 +250,23 @@ fn start_bridge_loop(
...
+                            Request::PollLogs { callback } => {
+                                let logs = core.fetch_buffered_logs();
+                                send_result(event_queue.clone(), callback, |cx| {
+                                    let logarr = cx.empty_array();
+                                    for (i, cl) in logs.into_iter().enumerate() {
+                                        let logobj = cx.empty_object();
+                                        let level = cx.string(cl.level.to_string());
+                                        logobj.set(cx, "level", level).unwrap();
+                                        let ts = cx.number(cl.millis_since_epoch() as f64);
+                                        logobj.set(cx, "timestamp_nanos", ts).unwrap();
```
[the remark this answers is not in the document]
Sushisource: Good catch thanks

packages/core-bridge/index.d.ts
```diff
@@ -116,20 +119,33 @@ export interface WorkerOptions {
   maxCachedWorkflows: number;
 }
 
+export interface CoreLog {
+  /** Log message */
+  message: string;
+  /** Unix millis since epoch. Ideally would be bigint but Neon can't send those
+   */
+  // TODO: Not meaningfully usable until logger interface has a timestamp concept
```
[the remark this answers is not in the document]
Sushisource: I'm happy to have you do it lol :)

packages/test/src/test-dependencies.ts
```diff
@@ -165,16 +164,12 @@ if (RUN_INTEGRATION_TESTS) {
   });
 
   test('Worker wraps ignored dependencies and logs when they throw', async (t) => {
```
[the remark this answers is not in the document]
Sushisource: Because that instance will be shared with other tests, no? And as a result the collected logs will have all sorts of other stuff?

packages/worker/src/core.ts
Sushisource: I added this but didn't propagate errors to worker.run because I don't know that there's anything interesting to do with them. I think simply ignoring them is fine, or alternatively just logging an error that we couldn't... log logs.

packages/test/src/test-dependencies.ts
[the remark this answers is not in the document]
bergundy: It doesn't matter because we filter specific messages that are only emitted from this test.
And if we want to avoid conflicts we can use `test.sync`.

packages/test/src/test-dependencies.ts
bergundy: If you don't mind, please add it back

packages/worker/src/core.ts
```diff
@@ -58,11 +73,46 @@ export class Core {
...
     };
     const native = await promisify(newCore)(compiledOptions);
+
     return new this(native, compiledOptions);
   }
 
+  private initLogPolling(options: CompiledCoreOptions, native: native.Core) {
+    this.shouldPollForLogs.next(true);
+
+    if (options.logger != null) {
```
bergundy, line 86: Is this even possible?
Don't we always want there to be a logger?

packages/worker/src/core.ts
```diff
@@ -34,12 +45,16 @@ function defaultTelemetryOptions(): TelemetryOptions {
 export class Core {
   /** Track the registered workers to automatically shutdown when all have been deregistered */
   protected registeredWorkers = new Set<native.Worker>();
+  protected shouldPollForLogs = new BehaviorSubject<boolean>(false);
+  protected logPollPromise = Promise.resolve();
```
bergundy, line 49: We don't want a resolved promise as the default, I'd say that the default is a promise that never resolves.
I'd initialize this in the constructor:

```ts
this.logPollPromise = this.initLogPolling(...);
```

packages/worker/src/core.ts
```diff
@@ -58,11 +73,46 @@ export class Core {
...
+              } else if (log.level === 'ERROR') {
+                options.logger.error(log.message);
+              }
+            }
+          }),
+          delay(3), // Don't go wild polling as fast as possible
+          repeat()
+        )
+      );
+      this.logPollPromise.catch(() => undefined); // Prevent unhandled rejection
```
bergundy, line 112: If we're not exposing this promise we should at least *log* when this happens (yo dawg meme)

packages/core-bridge/index.d.ts
```diff
@@ -84,6 +84,9 @@ export interface TelemetryOptions {
    * Which determines what tracing data is collected in the Core SDK
    */
   tracingFilter?: string;
+  /** What level, if any, logs should be forwarded from core at */
+  // These strings should match the log::LevelFilter enum in rust
+  logForwardingLevel: 'OFF' | 'ERROR' | 'WARN' | 'INFO' | 'DEBUG' | 'TRACE';
```
bergundy, line 89: I'd use this but we have `WARNING` as opposed to `WARN` in node, maybe worth unifying that too.
```suggestion
  logForwardingLevel: 'OFF' | LogLevel;
```

bergundy (approved):
Some minor comments.
Feel free to merge after reading them.

packages/worker/src/core.ts
Sushisource: Ah yeah I guess this isn't possible now. I do want it to be disable-able but I suppose that works just fine by setting forwarding level to off, so I'll check there instead.

PR #247, opened 2021-09-18
Remove type: 'remote' from docs

lorensr:


bergundy (approved):

bergundy:
Looks like something's up with the temporal cloud namespace.
There was some maintenance work done on it, I force merged and I'll check on the namespace.

PR #248, opened 2021-09-19
feat(activity): Properly support activity classes

### LINKED ISSUE #214: [Feature Request] Provide a method for getting a client instance on `activity.Context`

Currently there's no easy way to get the *same* client that the worker/workflows use from inside an activity. This surfaced as a problem where worker was using the right URL for the server, but an activity sending a signal was not.

In title

bergundy:
Closes #214 

FYI @lorensr

### DIFF
--- a/packages/common/src/activity-options.ts
+++ b/packages/common/src/activity-options.ts
@@ -86,3 +86,49 @@ export interface ActivityFunction<P extends any[], R> {
  * Mapping of Activity name to function
  */
 export type ActivityInterface = Record<string, ActivityFunction<any[], any>>;
+
+/**
+ * Transforms type `T` into an {@link ActivityInterface} by extracting all activity functions.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type IntoActivityInterface<T extends object> = {
+  [P in keyof T]: T[P] extends ActivityFunction<any[], any> ? T[P] : never;
+};
+
+/**
+ * Get all defined methods of an object prototype.
+ *
+ * (Taken from https://github.com/sindresorhus/auto-bind/blob/bfdc7861c61d78969912cb2a2d620cdc18de079c/index.js)
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+function getPrototypeMethods(prototype: object) {
+  const properties = new Set<string | symbol>();
+
+  for (const key of Reflect.ownKeys(prototype)) {
+    if (key === 'constructor') continue;
+    const descriptor = Reflect.getOwnPropertyDescriptor(prototype, key);
+    if (descriptor && typeof descriptor.value === 'function') {
+      properties.add(key);
+    }
+  }
+
+  return properties;
+}
+
+/**
+ * Returns a copy of provided activities `act` where all methods are bound to `act`.
+ *
+ * Use this in order to register an activity class with the Worker.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function boundActivities<T extends object>(act: T): IntoActivityInterface<T> {
+  const bindWrapper: any = {};
+  for (const k of getPrototypeMethods(act.constructor.prototype)) {
+    Object.defineProperty(bindWrapper, k, {
+      value: (act as any)[k].bind(act),
+      writable: false,
+      enumerable: true,
+    });
+  }
+  return bindWrapper;
+}
--- a/packages/test/src/workflows/activities-from-class.ts
+++ b/packages/test/src/workflows/activities-from-class.ts
@@ -1,0 +1,15 @@
+// @@@SNIPSTART nodejs-activity-from-class-workflow
+import { createActivityHandle } from '@temporalio/workflow';
+import type { ActivitiesWithDependencies } from '../activities/with-deps';
+import { Returner } from '../interfaces';
+
+const { greet } = createActivityHandle<ActivitiesWithDependencies>({
+  startToCloseTimeout: '1 minute',
+});
+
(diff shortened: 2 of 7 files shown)


packages/common/src/activity-options.ts
```diff
@@ -86,3 +86,49 @@ export interface ActivityFunction<P extends any[], R> {
...
+ * Get all defined methods of an object prototype.
+ *
+ * (Taken from https://github.com/sindresorhus/auto-bind/blob/bfdc7861c61d78969912cb2a2d620cdc18de079c/index.js)
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+function getPrototypeMethods(prototype: object) {
+  const properties = new Set<string | symbol>();
+
+  for (const key of Reflect.ownKeys(prototype)) {
+    if (key === 'constructor') continue;
```
vkarpov15, line 108: Do we also need to ignore `__proto__` here? For functions, `typeof fn.__proto__ === 'function'`.

packages/test/src/workflows/activities-from-class.ts
```diff
@@ -0,0 +1,15 @@
+// @@@SNIPSTART nodejs-activity-from-class-workflow
+import { createActivityHandle } from '@temporalio/workflow';
+import type { ActivitiesWithDependencies } from '../activities/with-deps';
+import { Returner } from '../interfaces';
+
+const { greet } = createActivityHandle<ActivitiesWithDependencies>({
```
vkarpov15, line 6: Is there a way to set `greetString` with the `createActivityHandle()` syntax?

packages/common/src/activity-options.ts
```diff
@@ -86,3 +86,49 @@ export interface ActivityFunction<P extends any[], R> {
...
+  return properties;
+}
+
+/**
+ * Returns a copy of provided activities `act` where all methods are bound to `act`.
+ *
+ * Use this in order to register an activity class with the Worker.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export function boundActivities<T extends object>(act: T): IntoActivityInterface<T> {
```
vkarpov15, line 124: The `boundActivities()` name isn't my favorite. Also, why not just have `worker.create()` call `bind()` automatically?

bergundy: I did that initially but then realized that activities may be registered from multiple sources like objects and module exports.

```ts
import * as activitiesFromModule from './activities';
const activitiesFromClassA = new ActivityClassA();
const activitiesFromClassB = new ActivityClassB();

  activities: {
    ...activitiesFromModule,
    ...boundActivities(activitiesFromClassA),
    ...boundActivities(activitiesFromClassB),
  },
});
```

bergundy: I'm open to changing the name.
@lorensr @sw-yx WDYT?

Sushisource: Agree this name is kinda confusing.

Also docstring probably makes a little more sense like "Binds all the defined activities in `act` to the instance `act`" or something

packages/test/src/workflows/activities-from-class.ts
bergundy: No, activities must be initialized in the Worker.

packages/common/src/activity-options.ts
bergundy: Not AFAICT according to the source, but we might need to do the traversal that's done here:
https://github.com/sindresorhus/auto-bind/blob/bfdc7861c61d78969912cb2a2d620cdc18de079c/index.js#L11

packages/common/src/activity-options.ts
sw-yx: i think an "activity class" is a new concept that is a bit alien to me. i hope we can write the docs for this along side the code since we're gonna have trouble explaining this one.

`bindClient`?

bergundy:
I'm putting this on hold for further discussion

sw-yx:
I'm primarily concerned about the conceptual overhead of introducing Classes (when the rest of the SDK doesnt use classes), and then introducing another API `boundActivities` to work with Classes. very easy to lead the concept count pile up and get the reputation that we are hard to learn/use.

There's a lot to recommend for being able to clearly state "Activities are just functions" so I'm keen to explore any options that let us stay there until we really have no choice.

Roey came up with two options for DI:

```ts
async greet({ redisClient }: Dependencies, name: string): Promise<string>;

async function greet() {
	const { redisClient } = activity.Context.current().getDependencies<Dependencies>();
}
```

the first is closest to "just a function", the second is lengthier but has less "magic swallowing of first argument". i spent a while trying to come up with a third option but can't come up with a better one.

is there a way we can avoid Activity classes, and do this?

```ts
import { Context } from '@temporalio/activity'; // slight change of Context behavior here
async function greet() {
	const { redisClient } = Context().deps<Dependencies>();
}
```

bergundy:
> note that I am changing the meaning of Context here - we would rearrange cancellationSignal, cancelled, heartbeat, info, and sleep elsewhere, preferably top level @temporalio/activity exports.

This is doable for anything that's a method on `Context` today, it won't work for `cancellationSignal` and `cancelled`.
I think for activities it might be better to leave `Context` and maybe get rid of the `current()` static method so it'd just be:

```
Context.cancelled
Context.dependencies<Dependencies>()
Context.sleep()
...
```

sw-yx:
@lorensr we discussed and think we can get by with just factory functions

bergundy:
Abandoning this, added a DI sample here: https://github.com/temporalio/samples-node/pull/18

PR #252, opened 2021-09-20
chore: Fix TS project references

bergundy:
I haven't confirmed this but I think our compilation is broken because we compile 2 projects (test and interceptors-opentelemetry) with shared project references.


lorensr:
Building `main`, I get:

image

After cherry picking this commit, I get:


image

bergundy:
@lorensr you should clean the generated js files first

lorensr:
Thanks again, will get in the habit. That fixed building `main`.

Sushisource (approved):

PR #253, opened 2021-09-20
fix(workflow): Mocked Date constructor now accepts parameters and exp…

bergundy:
…oses static parse and UTC methods


vkarpov15 (approved):
Nice :+1:

PR #257, opened 2021-09-21
Initial nightly long run implementation

Sushisource:
## What was changed
Add long running nightly test

## Why?
Testing good!

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Sushisource:
@bergundy I got super sucked in to RxJS land making this work. It's very "fun" but I can't honestly claim I feel at all productive with it lol... at least not yet.

Sushisource:
@bergundy My inclination is to merge this since the nightly runs are working (with the minor exception of figuring out why it "failed" when it looks like it succeeded, need to fix that) and then make any subsequent adjustments (like exposing more configuration options) to the gRPC retry stuff in a later PR. LMK

.buildkite/pipeline.yml
```diff
@@ -4,9 +4,13 @@ steps:
...
+      - 'npm run ci-load'
     plugins:
       - docker-compose#v3.8.0:
-          run: load-test
+          run: test-container
+          upload-container-logs: on-error
           config:
             - .buildkite/docker/docker-compose.yaml
-            - .buildkite/docker/docker-compose-load.yaml
+            - .buildkite/docker/docker-ci.yaml
```
bergundy, line 16: nit: Can we rename to this?
```suggestion
            - .buildkite/docker/docker-compose-ci.yaml
```

packages/client/src/grpc-retry.ts
```diff
@@ -0,0 +1,86 @@
...
+// Return backoff amount in ms
+function backOffAmount(attempt: number): number {
+  return 2 ** attempt * 20;
+}
+
+export const GrpcRetry: Interceptor = (options, nextCall) => {
+  let savedMetadata: Metadata;
+  let savedSendMessage: any;
+  let savedReceiveMessage: any;
+  let savedMessageNext: any;
```
bergundy, line 34: Are these supposed to be `any`?

packages/client/src/grpc-retry.ts
```diff
@@ -0,0 +1,86 @@
...
+  grpc.status.OUT_OF_RANGE,
+];
+
+function isRetryableError(status: StatusObject): boolean {
+  return retryableCodes.includes(status.code);
+}
+
+// Return backoff amount in ms
+function backOffAmount(attempt: number): number {
+  return 2 ** attempt * 20;
```
bergundy, line 27: Can we make this configurable?

packages/client/src/grpc-retry.ts
```diff
@@ -0,0 +1,86 @@
...
+  InterceptingCall,
+  Interceptor,
+  ListenerBuilder,
+  Metadata,
+  RequesterBuilder,
+  StatusObject,
+} from '@grpc/grpc-js';
+import * as grpc from '@grpc/grpc-js';
+
+const maxRetries = 10;
```
bergundy, line 11: I'd rather avoid hardcoding this here, it'd be nicer if this was configurable with a sane default.
We don't have to expose this knob to users, just would make it easier to test.

packages/client/src/grpc-retry.ts
```diff
@@ -0,0 +1,86 @@
...
+const retryableCodes = [
```
bergundy, line 12: It's probably far from being an issue but this could be a `Set` for "faster" lookup.

packages/test/src/load/setup.ts
```diff
@@ -7,6 +7,7 @@ async function createNamespace(connection: Connection, namespace: string, maxAtt
   for (let attempt = 1; attempt <= maxAttempts; ++attempt) {
     try {
       await connection.service.registerNamespace({ namespace, workflowExecutionRetentionPeriod: msToTs('1 day') });
+      break;
```
bergundy, line 10: Apparently this broke CI, the test we have for checking if a namespace is available is inaccurate.

packages/test/src/test-core.ts
```diff
@@ -28,7 +28,7 @@ if (RUN_INTEGRATION_TESTS) {
     {
       // Run a simple workflow
       const wf = connection.createWorkflowHandle(workflows.sleeper, { taskQueue: 'q2' });
-      await wf.start(1);
+      await wf.execute(1);
```
bergundy, line 31: 🙌

bergundy (approved):
Had a few minor comments.
Let's merge this once CI is green.

packages/client/src/grpc-retry.ts
Sushisource: Yep :(

packages/test/src/load/setup.ts
Sushisource: Well this line didn't, since it's only load and that's running OK - but the thing in `scripts` apparently doesn't work yeah

bergundy (approved):
Looking at the special handling done for `NoWorkerRegisteredError` think it might be better to postpone deregistering the worker from core (rust) until the final poll results (ShutdownError) have been returned to lang for activities and workflows.

Please don't forget to remove the extra prints before merging.

packages/worker/src/worker.ts
```diff
@@ -1106,8 +1106,17 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
...
             error: (error) => {
-              this.log.error('Worker failed', { error });
-              this.state = 'FAILED';
+              if (
+                (this.state == 'STOPPING' || this.state == 'DRAINING') &&
+                error instanceof errors.NoWorkerRegisteredError
+              ) {
+                // It's possible for the worker to be completely removed if we poll after completing
+                // everything, in which case it will be not registered.
+                this.state = 'STOPPED';
```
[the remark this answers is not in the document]
Sushisource: Won't that not matter since it's already the last value in the combined observable of all the pollers/stuff?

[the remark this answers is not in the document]
Sushisource: Basically both pollers either already have or are immediately about to encounter the not registered error.

In any case, I agree this would be better handled in core but I didn't want to have to do two merge rounds to wrap up this PR. I will do that in core and then clean this up.

Sushisource:
Merged this even though I'm not 100% sure the "mystery exit code 1" problem is solved on the nightly run. The other stuff is good / worth getting in anyway. I will continue to work on figuring out the weird exit code, just takes ages to repro.

PR #264, opened 2021-09-23
feat(workflow)!: Use vm instead of isolated-vm

bergundy:
BREAKING CHANGE: External dependencies API has been simplified.
The `applyMode` and `arguments` attributes no longer accepted.

Removes isolated-vm and node-gyp dependencies


bergundy:
Note that workflows run in a single thread at the moment, in the future, we can use worker threads in order to increase parallelism.

Sushisource:
I love how much code this made go away.

sw-yx:
+454 -1314 love to see it

bergundy:
In order to land this we'll need to add worker threads because creating contexts blocks the main thread. The operation is so slow that the integration tests time out.

rememberlenny:
Following here

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -0,0 +1,91 @@
...
+function respondOk(requestId: BigInt): void {
+  return respond({ requestId, result: { type: 'ok' } });
+}
+
+let workflowCreator: VMWorkflowCreator | undefined;
+const workflowByRunId = new Map<string, VMWorkflow>();
+
+parentPort.on('message', async ({ requestId, input }: WorkerThreadRequest) => {
+  try {
+    switch (input.type) {
```
cretz, line 32: I forget, will TypeScript complain here if all types don't have case clauses? Otherwise, might you need a `default` to catch it at runtime?

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -0,0 +1,138 @@
...
+        completion.reject(err);
+        return;
+      }
+
+      completion.resolve(result.output);
+    });
+  }
+
+  async send(input: WorkerThreadInput): Promise<WorkerThreadOutput> {
+    const requestId = this.requestIdx++;
```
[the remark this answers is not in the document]
bergundy: It's fine, it's just used as the map key

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -0,0 +1,138 @@
...
+    const promise = new Promise<WorkerThreadOutput>((resolve, reject) => {
+      this.requestIdToCompletion.set(requestId, { resolve, reject });
+    });
+    return promise;
+  }
+
+  async destroy(): Promise<void> {
+    await this.send({ type: 'destroy' });
+    // TODO: figure this out
+    this.workerThread.unref();
```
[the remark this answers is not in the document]
bergundy: I'm still not sure what's the best way to destroy the thread.
I'll look into it before I merge this PR, have way more critical problems at the moment.

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -0,0 +1,138 @@
...
+    return new this(workerThreadClients);
+  }
+
+  constructor(protected readonly workerThreadClients: WorkerThreadClient[]) {}
+
+  async createWorkflow(options: WorkflowCreateOptions): Promise<Workflow> {
+    const workflowThreadIdx = this.workflowThreadIdx;
+    const { runId } = options.info;
+    this.runIdToThreadNum.set(runId, workflowThreadIdx);
+    const workflow = await VMWorkflowThreadProxy.create(this.workerThreadClients[workflowThreadIdx], options);
```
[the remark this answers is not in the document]
bergundy: I agree that we can probably do better than round robin but it's a simple enough default.
I've added these worker threads because creating and deleting v8 context is an expensive operation which blocks the event loop.
From my experiments with isolated vm, we have a bit more workflow cache capacity when using multiple isolates and each worker thread is run in its own isolate.

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+
+export class WorkerThreadClient {
+  requestIdx = 0n;
+  requestIdToCompletion = new Map<BigInt, Completion<WorkerThreadOutput>>();
+
+  constructor(protected readonly workerThread: Worker) {
+    workerThread.on('message', ({ requestId, result }: WorkerThreadResponse) => {
+      const completion = this.requestIdToCompletion.get(requestId);
```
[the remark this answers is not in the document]
bergundy: good point, I must have missed that

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  constructor(protected readonly workerThreadClient: WorkerThreadClient, public readonly runId: string) {}
+
+  async getAndResetSinkCalls(): Promise<SinkCall[]> {
+    const output = await this.workerThreadClient.send({
+      type: 'exteract-sink-calls',
+      runId: this.runId,
+    });
+    if (!(output?.type === 'sink-calls')) {
```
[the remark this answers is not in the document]
bergundy: I think I copied this from another place where I had multiple conditions.
I'll change.

packages/worker/src/workflow/vm.ts
```diff
@@ -0,0 +1,147 @@
...
+import { partition } from '../utils';
+import { Workflow, WorkflowCreator, WorkflowCreateOptions } from './interface';
+import { SinkCall } from '@temporalio/workflow/lib/sinks';
+
+/**
+ * Maintains a pool of v8 isolates, returns Context in a round-robin manner.
+ * Pre-compiles the bundled Workflow code from provided {@link WorkflowIsolateBuilder}.
+ */
+export class VMWorkflowCreator implements WorkflowCreator {
+  script: vm.Script | undefined;
```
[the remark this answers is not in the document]
bergundy: Sure

packages/worker/src/workflow/vm.ts
```diff
@@ -0,0 +1,147 @@
...
+    let batchIndex = 0;
+
+    // Loop and invoke each batch and wait for microtasks to complete.
+    // This is done outside of the isolate because when we used isolated-vm we couldn't wait for microtasks from inside the isolate, not relevant anymore.
+    for (const jobs of [patches, signals, rest, queries]) {
+      if (jobs.length === 0) {
+        continue;
+      }
+      const arr = coresdk.workflow_activation.WFActivation.encodeDelimited({ ...activation, jobs }).finish();
+      const { numBlockedConditions } = await this.workflowModule.activate(arr, batchIndex++);
```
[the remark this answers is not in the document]
bergundy: That's okay, I feel like starting from index 0 is "right" here.

packages/worker/src/workflow/vm.ts
```diff
@@ -0,0 +1,147 @@
...
+        await this.tryUnblockConditions();
+      }
+      // Wait for microtasks to be processed
+      await new Promise((resolve) => process.nextTick(resolve));
+    }
+    return this.workflowModule.concludeActivation();
+  }
+
+  protected async tryUnblockConditions(): Promise<void> {
+    for (;;) {
```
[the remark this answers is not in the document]
bergundy: `dispose` will not be called if the workflow is still activating in this version of our runtime.

I tried to think of a way where workflow code creates new conditions which are unblocked in an endless loop but I can't find a way to do that because the condition function is eagerly evaluated in the WF isolate.

packages/worker/src/workflow/workflow-worker-thread.ts
bergundy: It doesn't complain unless you promise to return something, I'll refactor this so TS covers all cases.

bergundy (commented):
Thanks @cretz

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -0,0 +1,138 @@
...
+  async destroy(): Promise<void> {
+    await this.send({ type: 'destroy' });
+    // TODO: figure this out
+    this.workerThread.unref();
+  }
+}
+
+export class ThreadedVMWorkflowCreator implements WorkflowCreator {
+  protected workflowThreadIdx = 0;
+  protected runIdToThreadNum = new Map<string, number>();
```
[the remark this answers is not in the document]
cretz: Looks like it has been removed

.buildkite/docker/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM rust:1.56
 # INSTALL NODE ============================================================================
 # Unfortunately can't use the friendly lts name here easily, because then it gets hard to add to the
 # path. We'll just have to update this as needed.
-ENV NODE_VERSION="v14.17.5"
```
Sushisource, line 6: Does this mean we can't support 14?

Sushisource: Yeah I'm fine either way just wanted to make sure it wasn't a requirement

packages/test/src/load/setup.ts
```diff
@@ -21,14 +21,15 @@ async function createNamespace(connection: Connection, namespace: string, maxAtt
...
   for (let attempt = 1; attempt <= maxAttempts; ++attempt) {
     try {
       await connection.service.getWorkflowExecutionHistory({
         namespace,
-        execution: { workflowId: 'fake', runId: '12345678-1234-1234-1234-1234567890ab' },
+        execution: { workflowId: 'fake', runId },
       });
     } catch (err: any) {
-      if (err.details.includes('workflow history not found') || err.details.includes('operation GetCurrentExecution')) {
+      if (err.details.includes('workflow history not found') || err.details.includes(runId)) {
```
Sushisource, line 32: Lol this stupid thing. We really need to not have to do this

packages/worker/src/worker-options.ts
```diff
@@ -198,13 +173,35 @@ export type WorkerOptionsWithDefaults = WorkerOptions &
...
+     *
+     * Threads are used to create [vm](https://nodejs.org/api/vm.html)s for the
+     *
+     * isolated Workflow environment.
+     *
+     * New Workflows are created on this pool in a round-robin fashion.
+     *
+     * This value is not exposed at the moment.
+     *
+     * @default 8
```
Sushisource, line 191: Why not default to system logical cores?

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -0,0 +1,86 @@
...
+const parentPort = parentPortOrNull;
+
+function ok(requestId: BigInt): WorkerThreadResponse {
+  return { requestId, result: { type: 'ok' } };
+}
+
+let workflowCreator: VMWorkflowCreator | undefined;
+const workflowByRunId = new Map<string, VMWorkflow>();
+
+async function handleRequest({ requestId, input }: WorkerThreadRequest): Promise<WorkerThreadResponse> {
```
Sushisource, line 26: This needs a docstring. The whole message passing convention and protocol should be given at least a cursory doc treatment

Sushisource: (Though to be clear I'm not sure specifically this function is the best place, but probably the send and recv halves should refer to each other)

packages/worker/src/workflow/vm.ts
```diff
@@ -0,0 +1,166 @@
...
+    // 3. anything left except for queries
+    // 4. queries
+    const [patches, nonPatches] = partition(activation.jobs, ({ notifyHasPatch }) => notifyHasPatch != null);
+    const [signals, nonSignals] = partition(nonPatches, ({ signalWorkflow }) => signalWorkflow != null);
+    const [queries, rest] = partition(nonSignals, ({ queryWorkflow }) => queryWorkflow != null);
+    let batchIndex = 0;
+
+    // Loop and invoke each batch and wait for microtasks to complete.
+    // This is done outside of the isolate because when we used isolated-vm we couldn't wait for microtasks from inside the isolate, not relevant anymore.
+    for (const jobs of [patches, signals, rest, queries]) {
```
Sushisource, line 135: Just thinking out loud here - what if you passed the whole job list in the message to activate rather than one at a time? Might make for some significant easy savings?

cretz (approved):

packages/worker/src/workflow/workflow-worker-thread.ts
bergundy: Yes, we need way more docstrings in this whole threaded vm code.
I didn't invest too much in that initially because I wasn't sure if we'd use this code.
Forgot to add the docs before submitting the PR.
I'll do that now.

Issue #258, opened 2021-09-21
[Feature Request] Support single-file workflowsPath


lorensr:
### Is your feature request related to a problem? Please describe.

```
await Worker.create({
    workflowsPath: path.join(__dirname, 'workflows'),
```

doesn't work for `./workflows.js`—it only works for a `./workflows/` directory.

### Describe the solution you'd like

The ability to have this file structure: https://github.com/lorensr/samples-node/blob/1d6ff776fb4d1376d62045621a7179d73b890eb1/hello-world/src/worker.ts#L18

### Additional context

This would fix https://github.com/temporalio/samples-node/issues/13

joebowbeer:
@lorensr thanks! You beat me to it.

joebowbeer:
The goal of this RFE is for the isolate to use Node-style import resolution for `workflows`.

https://www.typescriptlang.org/docs/handbook/module-resolution.html#node

Because ts-loader does this, most TypeScript programmers will expect this and trip over it if it doesn't work.

There may be some leads in the following discussion:

https://github.com/webpack/webpack/issues/13252

For example, [this comment](https://github.com/webpack/webpack/issues/13252#issuecomment-828587290)

bergundy:
I think it's easy to just look if `workflows/index.js` exists and if it doesn't try `workflows.js` here: https://github.com/temporalio/sdk-node/blob/e08ffada21ea2fe10071e1cd36683ee4573185f7/packages/worker/src/worker.ts#L279

lorensr:
Started looking into this here: https://github.com/lorensr/sdk-node/tree/single-file

with these changes: https://github.com/lorensr/sdk-node/commit/270793cb620fa81036173585ac319e5a100800b1

```
workflowsPath: `${__dirname}/workflow-file.js`
```

When I comment out these 3 lines:

https://github.com/lorensr/sdk-node/commit/270793cb620fa81036173585ac319e5a100800b1#diff-6ca636db1dc7530aefb4f6e9c60d7d81b77e2c128068ff41f5940bb47fe67fc5R115-R117

The test doesn't succeed, but it stops erroring:

![image](https://user-images.githubusercontent.com/251288/134757195-bf806136-4a84-4f9a-afae-04ff5072327f.png)

With those lines uncommented, webpack build error is:

![image](https://user-images.githubusercontent.com/251288/134757280-a6f34f69-c237-42e7-be28-43a5b366632b.png)

I think most telling is the line:

[...]

PR #267, opened 2021-09-25
feat: Support workflows.ts as default workflowsPath

joebowbeer:
## What was changed

**_[DRAFT]_** workflowsPath defaults to `workflows.ts` if `workflows` path does not exist.

## Why?

Typescript programmers will expect to see Node-style import resolution.

Relates to #258

## Checklist

2. How was this tested:

This passes npm test including integration tests.

HOWEVER, the intended behavior is not covered by any of the test cases.

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -1,4 +1,5 @@
-import os from 'os';
+import { existsSync } from 'fs';
+import * as os from 'os';
 import { resolve } from 'path';
 import { promisify } from 'util';
 import * as otel from '@opentelemetry/api';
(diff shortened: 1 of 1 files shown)


packages/worker/src/worker.ts
```diff
@@ -272,11 +273,17 @@ function includesDeps<T extends WorkerSpec>(
 
 export function addDefaults<T extends WorkerSpec>(options: WorkerSpecOptions<T>): WorkerOptionsWithDefaults<T> {
   const { workDir, ...rest } = options;
+  // If workDir is defined, workflowsPath defaults to workflows.js if workflows does not exist
+  let workflowsPath = undefined;
+  if (workDir) {
+    const workflowsDir = resolve(workDir, 'workflows');
+    workflowsPath = existsSync(workflowsDir) ? workflowsDir : resolve(workDir, 'workflows.js');
```
bergundy, line 280: This is incorrect, a user might have both `workflows/workflowA.ts` and `workflows.ts`.
We also support reading either `js` or `ts` extensions (for better source maps, default when running with `ts-node`).

If you address all of these issues along with handling interceptors loading when `workflowsPath` is a file than we can merge this.

bergundy (changes requested):
Thanks @joebowbeer for yet another contribution!

packages/worker/src/worker.ts
joebowbeer: Thanks. To be rigorously correct, though, don't these checks need to be performed at runtime? I don't think watch is smart enough to know that adding or removing a workflows file/folder will invalidate this logic.

I wish I knew how this path is loaded at runtime. For example, where is the code that loads index.js now?

bergundy: What do you mean by "at runtime"?
I watch mode, nodemon will restart the process anytime a file changes and this logic will be re-evaluated.
`index.{j,t}s` is loaded by webpack in `isolate-builder.ts`.

joebowbeer: I see. I suppose this needs to work with any custom loader, e.g., babel-loader as well as ts-loader.

lorensr:
I think we'll be removing `workDir` (#269). Is there anything we can default `workflowsPath` to without that? Eg `workflows.ts` in the same directory as worker?

joebowbeer:
The list of extensions is dependent on the loader, e.g., ts-loader

The webpack module resolver or custom loader should be able to work with just "workflows"

PR #268, opened 2021-09-27
Pre beta stabilization

### LINKED ISSUE #246: Add timestamps to logger interface
If you want I'll add that after this PR or just DIY

_Originally posted by @bergundy in https://github.com/temporalio/sdk-node/pull/231#r708861439_

### LINKED ISSUE #266: [Feature Request] continueAsNew with a taskQueue

I'm looking for a way to continue a workflow as a new workflow but on a different task queue than the workflow started on.

If there's a way to do this with the current API I just don't know what it is. It seems continue as new performs this functionality but does not currently allow specifying a task queue.

### LINKED ISSUE #270: [Feature Request] workflow.sleep args should pass through `ms`
for consistent api with our timeouts

just tracking something i proposed to roey

### LINKED ISSUE #271: [Bug] activity heartbeat options are double documented
is this a symptom that something is wrong?

https://nodejs.temporal.io/api/classes/activity.context/#heartbeat

image

bergundy:
Closes #246 
Closes #271 
Closes #270
Closes #266

### DIFF
--- a/.buildkite/docker/docker-compose.yaml
+++ b/.buildkite/docker/docker-compose.yaml
@@ -1,3 +1,4 @@
+# No ports are exposed by default because tests run concurrently on a single CI instance
 version: '3.5'
 
 services:
@@ -5,20 +6,20 @@ services:
     image: cassandra:3.11
     logging:
       driver: none
-  #    ports:
-  #      - '9042:9042'
+    # ports:
+    #   - '9042:9042'
 
   temporal:
     image: temporalio/auto-setup:1.12.1
-    ports:
-      - '7233:7233'
-      - '7234:7234'
-    #      - '7235:7235'
-    #      - '7239:7239'
-    #      - '6933:6933'
-    #      - '6934:6934'
-    #      - '6935:6935'
-    #      - '6939:6939'
+    # ports:
+    #   - '7233:7233'
+    #   - '7234:7234'
+    #   - '7235:7235'
+    #   - '7239:7239'
+    #   - '6933:6933'
+    #   - '6934:6934'
+    #   - '6935:6935'
+    #   - '6939:6939'
     environment:
       - 'CASSANDRA_SEEDS=cassandra'
       - 'DYNAMIC_CONFIG_FILE_PATH=config/dynamicconfig/development.yaml'
@@ -29,8 +30,8 @@ services:
     image: temporalio/web:1.11.0
     logging:
       driver: none
-    ports:
-      - '8088:8088'
+    # ports:
+    #   - '8088:8088'
     environment:
       - 'TEMPORAL_GRPC_ENDPOINT=temporal:7233'
     depends_on:
--- a/packages/core-bridge/index.d.ts
+++ b/packages/core-bridge/index.d.ts
@@ -128,15 +128,20 @@ export interface WorkerOptions {
   maxCachedWorkflows: number;
 }
 
-export interface CoreLog {
+/** Log level - must match rust log level names */
+export type LogLevel = 'TRACE' | 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';
+
+export interface LogEntry {
   /** Log message */
   message: string;
-  /** Unix millis since epoch. Ideally would be bigint but Neon can't send those
+  /**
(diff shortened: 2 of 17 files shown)


Sushisource:
Your commit message says warn -> warning but it's actually the other way around

.buildkite/docker/docker-compose.yaml
```diff
@@ -1,24 +1,25 @@
+# No ports are exposed by default because tests run concurrently on a single CI instance
```
Sushisource, line 1: I'm almost positive this isn't actually true

packages/core-bridge/index.d.ts
```diff
@@ -128,15 +128,20 @@ export interface WorkerOptions {
...
-export interface CoreLog {
+/** Log level - must match rust log level names */
+export type LogLevel = 'TRACE' | 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';
+
+export interface LogEntry {
   /** Log message */
   message: string;
-  /** Unix millis since epoch. Ideally would be bigint but Neon can't send those
+  /**
+   * Time since epoch [seconds, nanos].
```
Sushisource, line 138: Why not just stick with millis? Nanos is pretty pointless for log messages and the single number is much easier to handle.

packages/core-bridge/src/lib.rs
```diff
@@ -900,8 +903,31 @@ fn worker_shutdown(mut cx: FunctionContext) -> JsResult<JsUndefined> {
     }
 }
 
+/// Convert Rust SystemTime into a JS array with 2 numbers (seconds, nanos)
+fn system_time_to_js<'a, C>(cx: &mut C, time: SystemTime) -> NeonResult<Handle<'a, JsArray>>
```
Sushisource, line 907: Yeah, per my other comment feels like we could maybe just avoid this whole thing

packages/core-bridge/src/lib.rs
```diff
@@ -900,8 +903,31 @@ fn worker_shutdown(mut cx: FunctionContext) -> JsResult<JsUndefined> {
...
+    let ts = cx.empty_array();
+    ts.set(cx, 0, ts_seconds).unwrap();
+    ts.set(cx, 1, only_nanos).unwrap();
+    return Ok(ts);
+}
+
+/// Helper to get the current time in nanosecond resolution.
+fn get_time_of_day(mut cx: FunctionContext) -> JsResult<JsArray> {
+    system_time_to_js(&mut cx, SystemTime::now())
+}
```
Sushisource, line 926: Do we want to call across FFI this often for this? Seems like it could be a bit of unneeded overhead but maybe I'm overestimating the expense

.buildkite/docker/docker-compose.yaml
Sushisource: Yeah, I couldn't quickly find it explicitly stated, but this page implies build agents only ever run one thing at a time: https://buildkite.com/docs/tutorials/parallel-builds

Otherwise there would be no reason to do the "vertical scaling" they describe

bergundy: This page mentions multiple agents on the same machine which implies concurrent builds.

packages/core-bridge/index.d.ts
bergundy: I switched because we need higher time resolution for merging the node / core logs in order.

packages/core-bridge/src/lib.rs
bergundy: If we're willing to compromise on the ordering.

packages/core-bridge/src/lib.rs
bergundy: I don't think this is much more expensive that calling `process.hrtime` or `Date.now()`, those are native functions exposed to JS as well.

bergundy:
@Sushisource I ended up merging this.
LMK if you want me to fix anything and I'll do it in the next PR.

.buildkite/docker/docker-compose.yaml
Sushisource: If you configure them that way. We don't, and they don't by default and we use their amazon stack

PR #273, opened 2021-09-28
feat: Enhance `@temporalio/create` and use samples-node as its source

bergundy:
This continues @lorensr's work in an attempt to land this long outstanding PR.


lorensr:
Thanks for the help! 😊

Issue #244, opened 2021-09-15
[Bug] Auto generated changelogs and versions are not working as intended


bergundy:
### Describe the bug

lerna does not understand that the meta package (`temporalio`) version should be incremented according to its dependencies.
The last 2 releases were breaking changes but this was not reflected in the package version, bumped from `0.4.0` to `0.4.1` and then `0.4.2`. It should have been on version `0.6.0` by now.

At this point, I'm thinking it might make more sense to have a single version for all of the packages in the mono-repo.

lerna changelog seems to be broken, `feat!` commits are being classified as bug fixes and the changelog messages are duplicated across any package that was part of the commit that changed.

https://github.com/temporalio/sdk-node/commit/23da8507a90049774cc6ddd28a8ec3dfc37d345f#diff-639cb7cb3eda69b4a479a3a8e11c556d1ac91b6ba1552785fd17fb5970f67150

Using a single version for all packages, we'll be able to get rid of all of the package specific changelogs and get rid the duplication.

bergundy:
Our current process is good enough, we can add a CI step to validate the PR title but it's overkill IMHO, we end up manually editing the auto-generated CHANGELOG anyways.

PR #275, opened 2021-09-28
Pass client/version info to core & core update & set span contexts!

Sushisource:
## What was changed
Sdk name/version info now passed through into core
Spans are now attached between node/core properly!

## Why?
Server can change behavior based on client. Version tracking, etc.
Cool spans!

## Checklist

1. Closes https://github.com/temporalio/sdk-core/issues/85
2. How was this tested:
Added a test which can be run by hand so the runner can inspect the spans in Jaeger or whatever. Verified spans are now connected.

3. Any docs updates needed?

### DIFF
--- a/packages/core-bridge/scripts/build.js
+++ b/packages/core-bridge/scripts/build.js
@@ -24,6 +24,13 @@ const platformMapping = { darwin: 'apple-darwin', linux: 'unknown-linux-gnu', wi
 
 function compile(target) {
   console.log('Compiling bridge', { target });
+  // Extract version from meta package for use as version number in client
+  // version field.
+  const metaPkg = require('../../meta/package.json');
+  const cargoOutputDir = path.resolve(__dirname, '..', 'target');
+  fs.mkdirSync(cargoOutputDir, { recursive: true });
+  fs.writeFileSync(path.resolve(cargoOutputDir, 'metaversion'), metaPkg.version);
+
   const out = target ? `releases/${target}/index.node` : 'index.node';
   try {
     fs.unlinkSync(out);
--- a/packages/core-bridge/scripts/build.js
+++ b/packages/core-bridge/scripts/build.js
@@ -24,12 +24,6 @@ const platformMapping = { darwin: 'apple-darwin', linux: 'unknown-linux-gnu', wi
 
 function compile(target) {
   console.log('Compiling bridge', { target });
-  // Extract version from meta package for use as version number in client
-  // version field.
-  const metaPkg = require('../../meta/package.json');
-  const cargoOutputDir = path.resolve(__dirname, '..', 'target');
-  fs.mkdirSync(cargoOutputDir, { recursive: true });
-  fs.writeFileSync(path.resolve(cargoOutputDir, 'metaversion'), metaPkg.version);
 
   const out = target ? `releases/${target}/index.node` : 'index.node';
   try {
--- a/packages/core-bridge/src/conversions.rs
+++ b/packages/core-bridge/src/conversions.rs
@@ -1,0 +1,240 @@
+use log::LevelFilter;
+use neon::{
+    context::Context,
+    handle::Handle,
+    prelude::*,
+    types::{JsNumber, JsString},
+};
+use opentelemetry::trace::{SpanContext, SpanId, TraceFlags, TraceId, TraceState};
+use std::net::SocketAddr;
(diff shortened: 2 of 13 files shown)


packages/core-bridge/scripts/build.js
```diff
@@ -24,6 +24,13 @@ const platformMapping = { darwin: 'apple-darwin', linux: 'unknown-linux-gnu', wi
 
 function compile(target) {
   console.log('Compiling bridge', { target });
+  // Extract version from meta package for use as version number in client
+  // version field.
+  const metaPkg = require('../../meta/package.json');
+  const cargoOutputDir = path.resolve(__dirname, '..', 'target');
+  fs.mkdirSync(cargoOutputDir, { recursive: true });
+  fs.writeFileSync(path.resolve(cargoOutputDir, 'metaversion'), metaPkg.version);
```
bergundy, line 32: I'd rather avoid writing to the FS for this, instead just pass it in at runtime.
We can use the worker version for now, at some point I want to sync the versions of the different packages: #244

packages/core-bridge/src/conversions.rs
```diff
@@ -0,0 +1,240 @@
+use log::LevelFilter;
```
bergundy, line 1: Nice, it was time

packages/core-bridge/src/conversions.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+        };
+
+        let mut gateway_opts = ServerGatewayOptionsBuilder::default();
+        if let Some(tls_cfg) = tls_cfg {
+            gateway_opts.tls_cfg(tls_cfg);
+        }
+        Ok(gateway_opts
+            .client_name("temporal-sdk-node".to_string())
+            .client_version(include_str!("../target/metaversion").to_string())
```
bergundy, line 166: I'd rather this be passed in from lang as mentioned above.

packages/core-bridge/scripts/build.js
bergundy: This can cause problems later on if we build before bumping the version in our publish process.

packages/test/src/test-otel.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Test the various states of a Worker.
```
bergundy, line 2: Lies

packages/test/src/test-otel.ts
```diff
@@ -0,0 +1,57 @@
...
+    nodeModulesPath: `${__dirname}/../../../node_modules`,
+    taskQueue: 'test-otel',
+  });
+  const runPromise = worker.run();
+  // Catch the error here to avoid unhandled rejection
+  runPromise.catch((err) => {
+    console.error('Caught error while worker was running', err);
+  });
+
+  const client = new WorkflowClient(undefined, { workflowDefaults: { followRuns: false } });
```
bergundy, line 48: I'd just use the defaults here, no reason not to follow runs.

packages/test/src/test-otel.ts
```diff
@@ -0,0 +1,57 @@
...
+  const workflow = client.createWorkflowHandle(workflows.cancelFakeProgress, { taskQueue: 'test-otel' });
+  await workflow.execute();
+
+  worker.shutdown();
+  await runPromise;
```
bergundy, line 53: In newer tests I started doing this instead to avoid hanging:
```suggestion
  await Promise.all([
    workflow.execute().finally(() => worker.shutdown()),
    worker.run(),
  ])
```

packages/test/src/test-otel.ts
```diff
@@ -0,0 +1,57 @@
...
+  await otel.shutdown();
+  await new Promise((resolve) => setTimeout(resolve, 5000));
```
bergundy, line 55: This requires a comment.

packages/worker/src/server-options.ts
```diff
@@ -1,13 +1,15 @@
...
 import * as native from '@temporalio/core-bridge';
 import pkg from './pkg';
+import path from 'path';
 
 type TLSConfig = native.TLSConfig;
 
 export { TLSConfig };
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const sdkVersion = require(path.resolve(__dirname, '../package.json')).version;
```
bergundy, line 11: We already read package.json in `pkg`, we should probably just use that.
Shouldn't the SDK version also include the SDK name?

PR #278, opened 2021-10-01
Link workflow run spans together

Sushisource:
## What was changed
Links all spans in a workflow run to whatever was attached to workflow execution start headers

## Why?
Best option available at the moment. Ideally everything would be part of one trace which looks best in Jaeger UI but that is currently not feasible due to https://github.com/open-telemetry/opentelemetry-rust/issues/643

In theory this is maybe the more "right" way to do it, but Jaeger isn't good at displaying this sort of thing due to
https://github.com/jaegertracing/jaeger-ui/issues/662

## Checklist

- [ ] Update otel interceptor samples

1. Closes 

2. How was this tested:
Otel manual test

3. Any docs updates needed?


packages/interceptors-opentelemetry/src/workflow/index.ts
```diff
@@ -1,33 +1,45 @@
...
 
 export * from './interfaces';
 
-export const TRACE_HEADER = 'Otel-Trace-Context';
-
-export const tracer = otel.trace.getTracer('workflow');
-
+// TODO: Literally the existence of this function causes us to blow up
+//   trying to read Date.now() too early (not calling it, just it being defined)
+//   I think the problem is importing tracing literally calls Date.now
```
Sushisource, line 20: @bergundy What do you think we can do about initializing Date.now() earlier?

I actually prefer my lazy-initted tracer to this anyway, but the problem remains that the runtime fixer-upper thing doesn't work right b/c Date.now() can't be called at import time.

packages/worker/src/worker.ts
```diff
@@ -413,6 +417,7 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
   protected readonly numInFlightActivationsSubject = new BehaviorSubject<number>(0);
   protected readonly numInFlightActivitiesSubject = new BehaviorSubject<number>(0);
   protected readonly numRunningWorkflowInstancesSubject = new BehaviorSubject<number>(0);
+  private readonly runIdsToSpanContext = new Map<string, SpanContext>();
```
Sushisource, line 420: @bergundy Let me know if there's a better place to keep this state

package.json
```diff
@@ -45,7 +45,8 @@
     "commitlint": "commitlint --from origin/main",
     "format": "prettier --write .",
     "clean": "node ./scripts/clean.js",
-    "docs": "lerna run --stream maybe-install-deps-and-build-docs"
+    "docs": "lerna run --stream maybe-install-deps-and-build-docs",
+    "fixbuild": "pushd packages/worker && rm -r ./lib/*.d.ts || tsc --build && popd && lerna run --stream build"
```
bergundy, line 49: :(

packages/common/src/otel.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Link a span to an maybe-existing span context
+ */
+export function linkSpans(fromSpan: Span, toContext?: SpanContext): void {
+  if (toContext != null) {
+    // TODO: I have to go around typescript because otel api 😢
+    //  See https://github.com/open-telemetry/opentelemetry-js-api/issues/124
+    const links = (fromSpan as any).links;
+    if (links == null) {
```
bergundy, line 40: are they actually null or undefined? Prefer to use `===`.

packages/interceptors-opentelemetry/src/workflow/index.ts
bergundy: This is a regression, Date should be available at import time.

packages/test/src/workflows/cancel-fake-progress.ts
```diff
@@ -40,3 +45,8 @@ export const cancelFakeProgress: ActivitySignalHandler = () => {
     },
   };
 };
+
+export const interceptors = (): WorkflowInterceptors => ({
+  inbound: [new OpenTelemetryInboundInterceptor()],
+  outbound: [new OpenTelemetryOutboundInterceptor()],
+});
```
bergundy, line 52: This doesn't belong here, it has nothing to do with this WF.

packages/worker/src/worker.ts
```diff
@@ -1027,6 +1052,12 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
             throw new TypeError('Got an activity task without a "variant" attribute');
           }
           parentSpan.setAttribute('taskToken', formattedTaskToken).setAttribute('variant', variant);
+          parentSpan.setAttribute('run_id', task.start?.workflowExecution?.runId ?? 'unknown');
```
bergundy, line 1055: Let's be consistent in the attribute casing.

packages/common/src/otel.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Headers } from './interceptors';
+
+export const TRACE_HEADER = 'Otel-Trace-Context';
+
+/**
+ * If found, return a span context deserialized from the provided headers
+ */
+export async function extractSpanContextFromHeaders(headers: Headers): Promise<SpanContext | undefined> {
+  const encodedSpanContext = headers[TRACE_HEADER];
+  if (encodedSpanContext == null) {
```
bergundy, line 12:
```suggestion
  if (encodedSpanContext === undefined) {
```

packages/worker/src/worker.ts
```diff
@@ -952,7 +957,27 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
...
-          span.setAttribute('runId', runId).setAttribute('numJobs', rest.jobs.length);
+
+          let linkedContext = this.runIdsToSpanContext.get(runId);
+          span.setAttribute('run_id', runId).setAttribute('numJobs', rest.jobs.length);
+          // If there is an otel context in the headers, link our trace for the workflow to the
+          // trace in the header.
+          for (const j of rest.jobs) {
+            if (j.startWorkflow != null) {
+              const ctx = await extractSpanContextFromHeaders(j.startWorkflow.headers ?? {});
+              if (ctx != null && linkedContext == null) {
```
bergundy, line 968: Again these are `undefined`s not null.
I'll add a lint rule to prohibit using `==`.

packages/worker/src/worker.ts
bergundy, line 420: Looks like we never delete from this map, is there a way around it?

packages/common/src/otel.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (isSpanContextValid(decoded)) {
+    return decoded;
+  }
+  return undefined;
+}
+
+/**
+ * Given a span context, insert it into the provided headers if they exist
+ */
+export async function insertSpanContextIntoHeaders(context: SpanContext, headers?: Headers): Promise<void> {
```
bergundy, line 25: I'd rather not mutate and instead return a new map with the added field:
```
headers = { ...oldHeaders, [TRACE_HEADER]: ... }
```

packages/common/src/otel.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (headers === undefined) {
+    return;
+  }
+  headers[TRACE_HEADER] = await defaultDataConverter.toPayload(context);
+}
+
+/**
+ * Link a span to an maybe-existing span context
+ */
+export function linkSpans(fromSpan: Span, toContext?: SpanContext): void {
```
bergundy, line 35: I think you can set the links on span creation, will that work?

bergundy (commented):
On the one hand I like that everything connects but I don't like how otel took over the SDK and I'm not sure we want to interfere with user's spans at all.
Let's discuss this.

package.json
bergundy: I use `npm run clean -- --only ts && npm run build`

packages/common/src/otel.ts
Sushisource: My understanding is this is an accepted safe way to check for both at once.

packages/interceptors-opentelemetry/src/workflow/index.ts
Sushisource: OK what do we need to do to get it initted properly?

packages/test/src/workflows/cancel-fake-progress.ts
Sushisource: Where should it go then? I think this is something we could be more clear on or make more obvious in the API. Having to load up interceptors with the workflows makes sense when they affect them but it's not clear where they're expected to go if not

packages/worker/src/worker.ts
Sushisource: These need to match what core uses as well. I hadn't changed task token yet but it should become snake case too

packages/common/src/otel.ts
Sushisource: No, we don't know the link until after the span is created, which apparently is just generally a hole in the Opentelem spec, as both setting trace ID or setting links are only expected to happen at span creation time, which just misses quite a big section of use cases :|

I might have to change this so that the polling spans don't really get linked and instead the inner spans which are only created after we have received the info about the Run Id are, but that requires more inspecting of the results in places that ideally might be free of that.

package.json
Sushisource: Well that would've been nice to know! :stuck_out_tongue_closed_eyes:

packages/worker/src/worker.ts
Sushisource: On eviction probably

packages/common/src/otel.ts
Sushisource: https://stackoverflow.com/questions/28975896/is-there-a-way-to-check-for-both-null-and-undefined/28984306#28984306 -- Seems easier than needing to remember which or check for both in various places

packages/worker/src/worker.ts
Sushisource: Fixed this

bergundy: This is fine.

bergundy (commented):
I'm still not convinced we should mix user spans with framework spans.
In case we decide to go that direction this PR is definitely the way to go.

packages/worker/src/worker.ts
Sushisource: I wouldn't just b/c we still need it for not overly verbose checking of null/undef in protos

packages/interceptors-opentelemetry/src/workflow/index.ts
```diff
@@ -1,33 +1,32 @@
...
 import { SpanName } from './interfaces';
+import * as tracing from '@opentelemetry/sdk-trace-base';
 
 export * from './interfaces';
 
-export const TRACE_HEADER = 'Otel-Trace-Context';
+let tracer: undefined | otel.Tracer = undefined;
 
-export const tracer = otel.trace.getTracer('workflow');
+function getTracer(): otel.Tracer {
```
bergundy, line 21: What's the reason for this change?

packages/interceptors-opentelemetry/src/workflow/index.ts
```diff
@@ -41,15 +40,14 @@ export class OpenTelemetryInboundInterceptor implements WorkflowInboundCallsInte
...
+    const spanContext = await extractSpanContextFromHeaders(input.headers);
 
     if (spanContext === undefined) {
-      return await instrument(tracer, SpanName.WORKFLOW_EXECUTE, () => next(input));
+      return await instrument(getTracer(), SpanName.WORKFLOW_EXECUTE, () => next(input));
     }
-    return await instrumentFromSpanContext(tracer, spanContext, SpanName.WORKFLOW_EXECUTE, () => next(input));
+    return await instrumentFromSpanContext(getTracer(), spanContext, SpanName.WORKFLOW_EXECUTE, () => {
+      return next(input);
+    });
```
bergundy, line 50: nit
```suggestion
    return await instrumentFromSpanContext(getTracer(), spanContext, SpanName.WORKFLOW_EXECUTE, () => next(input));
```

packages/test/src/test-otel.ts
```diff
@@ -9,21 +9,27 @@ import * as opentelemetry from '@opentelemetry/sdk-node';
...
 import { WorkflowClient } from '@temporalio/client';
+import { OpenTelemetryWorkflowClientCallsInterceptor } from '@temporalio/interceptors-opentelemetry/lib/client';
+import {
+  makeWorkflowExporter,
+  OpenTelemetryActivityInboundInterceptor,
+} from '@temporalio/interceptors-opentelemetry/lib/worker';
+import { OpenTelemetryDependencies } from '@temporalio/interceptors-opentelemetry/lib/workflow';
 
 // Un-skip this test and run it by hand to inspect outputted traces
 test.skip('Otel spans connected', async (t) => {
```
bergundy, line 20: How much work would it be to run this and add assertions?
Doesn't have to be in this PR but I'd at least track it.

packages/worker/src/worker.ts
```diff
@@ -774,7 +782,7 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
               { activation, parentSpan, synthetic }
             ): Promise<{
               state: Workflow | undefined;
-              output: ContextAware<{ completion?: Uint8Array; close: boolean }>;
+              output: ContextAware<{ completion?: Uint8Array; close: boolean; runId: string }>;
```
bergundy, line 785: No need to add the runId everywhere, it's available as the GroupedObservable's key: https://rxjs-dev.firebaseapp.com/api/index/interface/GroupedObservable

packages/common/src/otel.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (isSpanContextValid(decoded)) {
+    return decoded;
+  }
+  return undefined;
+}
+
+/**
+ * Given a span context & headers, return new headers with the span context inserted
+ */
+export async function insertSpanContextIntoHeaders(context: SpanContext, headers: Headers): Promise<Headers> {
```
bergundy, line 31: The name insert  is inaccurate since there's no mutation, maybe rename to something like `headersWithSpanContext`?

bergundy (approved):
I'm approving, none of my comments are critical enough to block this PR.

packages/interceptors-opentelemetry/src/workflow/index.ts
Sushisource: Easier on users. They don't have to remember to call this register function

PR #281, opened 2021-10-05
Add nightly load "sampler" run

Sushisource:
## What was changed
* Add to the nightly run a step which exercises some different kinds of workflows for more coverage
* Allow silencing workflow logging

## Why?
More coverage

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/test/src/activities/index.ts
+++ b/packages/test/src/activities/index.ts
@@ -1,9 +1,10 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 import { Context } from '@temporalio/activity';
-import { Connection, LOCAL_DOCKER_TARGET } from '@temporalio/client';
+import { Connection, LOCAL_DOCKER_TARGET, WorkflowClient } from '@temporalio/client';
 import { fakeProgress as fakeProgressInner } from './fake-progress';
 import { cancellableFetch as cancellableFetchInner } from './cancellable-fetch';
+import { QueryDefinition } from '@temporalio/common';
 
 export { throwSpecificError } from './failure-tester';
 
@@ -91,4 +92,17 @@ export async function progressiveSleep(): Promise<void> {
   Context.current().heartbeat(2);
   await sleep(100);
   Context.current().heartbeat(3);
+}
+
+export async function queryOwnWf<R, A extends any[]>(queryDef: QueryDefinition<R, A>, ...args: A): Promise<void> {
+  const ctx = Context.current();
+  const we = ctx.info.workflowExecution;
+  const client = new WorkflowClient(getTestConnection().service, { namespace: ctx.info.workflowNamespace });
+  // TODO: Until server is released with the fix below, this fails often
+  //  https://github.com/temporalio/temporal/pull/2033
+  try {
+    await client.createWorkflowHandle(we).query(queryDef, ...args);
+  } catch (e) {
+    console.log(`Workflow ${JSON.stringify(we)} query err`, e);
+  }
 }
--- a/packages/test/src/load/starter.ts
+++ b/packages/test/src/load/starter.ts
@@ -4,23 +4,9 @@ import { bufferTime, map, mergeMap, tap, takeUntil } from 'rxjs/operators';
 import { Connection, WorkflowClient } from '@temporalio/client';
 import { StarterArgSpec, starterArgSpec, getRequired } from './args';
 import * as os from 'os';
-import { smorgasbord } from '../workflows';
 
 async function runWorkflow(client: WorkflowClient, name: string, taskQueue: string) {
-  if (name === 'smorgasbord') {
-    const wf = client.createWorkflowHandle(smorgasbord, { taskQueue });
-    const runid = await wf.start();
-    // TODO: Until server is released with the fix below, this fails often
-    //  https://github.com/temporalio/temporal/pull/2033
-    try {
-      await wf.query.step();
-    } catch (e) {
-      console.log(`wfid ${wf.workflowId} runid ${runid} query err`, e);
-    }
-    await wf.result();
-  } else {
-    await client.execute(name, { taskQueue });
-  }
+  await client.execute(name, { taskQueue });
 }
 
 class NumberOfWorkflows {
@@ -89,10 +75,9 @@ async function runWorkflows({
 
   const finalWfsPerSec = numComplete / totalTime;
   if (finalWfsPerSec < minWFPS) {
-    console.error(
-      `Insufficient overall workflows per second upon test completion: ${finalWfsPerSec} less than ${minWFPS}`
+    throw new Error(
+      `Insufficient overall workflows per second upon test completion: ${finalWfsPerSec} less than ${minWFPS} for workflow ${name}`
     );
-    throw new Error(`Load test did not pass for workflow ${name}`);
   }
 }
 
--- a/packages/test/src/workflows/smorgasbord.ts
+++ b/packages/test/src/workflows/smorgasbord.ts
@@ -2,17 +2,19 @@
  * This workflow does a little bit of everything
  */
 import {
-  Trigger,
   sleep,
   createChildWorkflowHandle,
   createActivityHandle,
   ActivityCancellationType,
   CancellationScope,
   isCancellation,
+  defineQuery,
+  setListener,
+  condition,
 } from '@temporalio/workflow';
-import { Smorgasbord } from '../interfaces';
 import * as activities from '../activities/';
 import { signalTarget } from './signal-target';
+import { activityStartedSignal, unblockSignal } from './definitions';
 
 const { fakeProgress } = createActivityHandle<typeof activities>({
   startToCloseTimeout: '5s',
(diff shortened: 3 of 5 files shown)


packages/test/src/load/starter.ts
```diff
@@ -126,22 +127,33 @@ async function main() {
...
-    taskQueue,
-    stopCondition,
-    concurrency: concurrentWFClients,
-    minWFPS,
-  });
-  if (!passed) {
-    console.error('Load test did not pass');
-    process.exit(1);
+  let workflowsToRun: string[];
+  if (workflowName === 'yummy-sampler-mode') {
```
bergundy, line 134: Why have this magic string instead of passing this in as multiple args?

packages/test/src/load/starter.ts
```diff
@@ -126,22 +127,33 @@ async function main() {
...
+    // Special workflow alias to run many different load tests sequentially.
+    // Expected wf/sec should be set low since some of these by their nature have
+    // higher latency.
+    workflowsToRun = ['cancelFakeProgress', 'childWorkflowCancel', 'childWorkflowSignals', 'cancellationScopes'];
```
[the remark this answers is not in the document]
bergundy: `cancellationScopes` isn't that interesting as a load test, it's more a test of the framework code.
`childWorkflowCancel` and `childWorkflowSignals` are better used as integration tests as they test multiple scenarios but I'd leave them for now since they provide pretty good coverage for child workflows, we can replace them later.
We should add a workflow with a query, that's missing here.

packages/test/src/load/starter.ts
```diff
@@ -126,22 +127,33 @@ async function main() {
...
+  } else {
+    workflowsToRun = [workflowName];
+  }
+
+  for (const wfName of workflowsToRun) {
+    console.log(`~~~ Starting test for ${wfName} workflows`);
+    const passed = await runWorkflows({
```
bergundy, line 145: I'd just throw from `runWorkflows` instead of adding this extra condition. the process will exit and log an error if `main()` throws.

packages/worker/src/worker.ts
```diff
@@ -220,6 +220,9 @@ export interface WorkerOptions {
    * A mapping of interceptor type to a list of factories or module paths
    */
   interceptors?: WorkerInterceptors;
+
+  /** If set true, silence all `console.log` calls from workflow code */
+  silenceWorkflowLogging?: boolean;
```
bergundy, line 225: Instead, I'd let users override the console.log implementation.
I wouldn't want to expose this to users.

packages/workflow/src/workflow.ts
```diff
@@ -662,7 +662,7 @@ export function continueAsNew<F extends Workflow>(...args: Parameters<F>): Promi
 /**
  * Generate an RFC compliant V4 uuid.
  * Uses the workflow's deterministic PRNG making it safe for use within a workflow.
- * This function is cryptograpically insecure.
+ * This function is cryptographically insecure.
```
bergundy, line 665: `h`s are overrated.

packages/test/src/load/starter.ts
Sushisource: Heh, that's already how it was but sure :P

packages/test/src/load/starter.ts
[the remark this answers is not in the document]
Sushisource: OK I will remove scopes and add a smoke test that does a little bit of everything.

This is really more like integ tests with a bunch of them in parallel. It's really more of a smoke test than a proper load test but I think it's reasonable to just have it here as this general test runner.

package.json
```diff
@@ -39,7 +39,8 @@
     "test": "lerna run --stream test",
     "test.watch": "lerna run --stream test.watch",
     "ci-load": "node ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --iterations 10000 --otel-url='' --max-cached-wfs 500",
-    "ci-nightly": "node --trace-exit ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --for-seconds 14400 --min-wfs-per-sec 5 --concurrent-wf-clients 10 --otel-url=''",
+    "ci-nightly-longrun": "node ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --for-seconds 14400 --min-wfs-per-sec 5 --concurrent-wf-clients 10 --otel-url=''",
```
bergundy, line 42: nit: I'd still prefer `--otel-url` to default to undefined in the code and omit it here.

packages/test/src/load/starter.ts
```diff
@@ -4,9 +4,23 @@ import { bufferTime, map, mergeMap, tap, takeUntil } from 'rxjs/operators';
 import { Connection, WorkflowClient } from '@temporalio/client';
 import { StarterArgSpec, starterArgSpec, getRequired } from './args';
 import * as os from 'os';
+import { smorgasbord } from '../workflows';
 
 async function runWorkflow(client: WorkflowClient, name: string, taskQueue: string) {
-  await client.execute(name, { taskQueue });
+  if (name === 'smorgasbord') {
```
bergundy, line 10: This can probably be done in an activity started by the workflow, I'd rather keep the starter generic if possible.

packages/test/src/load/starter.ts
```diff
@@ -71,14 +85,15 @@ async function runWorkflows({
...
   const { numComplete, totalTime } = await observable.toPromise();
+  process.stderr.write('\n');
+
   const finalWfsPerSec = numComplete / totalTime;
   if (finalWfsPerSec < minWFPS) {
     console.error(
       `Insufficient overall workflows per second upon test completion: ${finalWfsPerSec} less than ${minWFPS}`
     );
-    return false;
+    throw new Error(`Load test did not pass for workflow ${name}`);
```
bergundy, line 95: You could probably just put the details here and delete the log

packages/test/src/load/starter.ts
bergundy: We can put it in another script, I don't like this special case.
Leaving it up to you to decide.

packages/test/src/workflows/smorgasbord.ts
```diff
@@ -0,0 +1,70 @@
...
+      },
+    },
+    signals: {
+      activityStarted(): void {
+        unblocked.resolve();
+      },
+    },
+
+    async execute(): Promise<void> {
+      while (step < 2) {
```
bergundy, line 41:
```suggestion
      for (; step < 2; ++step) {
```
And check if step `=== 1` below

packages/worker/src/worker.ts
```diff
@@ -919,7 +925,11 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
       'console.log',
       (...args: any[]) => {
         if (workflow.info.isReplaying) return;
-        console.log(`${workflow.info.workflowType} ${workflow.info.runId} >`, ...args);
+        if (this.options.workflowConsoleLog !== undefined) {
```
bergundy, line 928: We should pass workflow info into the console log function, otherwise it's a bit useless for anything other than dropping logs.
Info also contains `isReplaying` which will let users decide if they want to see logs from replays or not.

bergundy: Also, I'm pretty sure that for production code users will use a proper logger instead of using console.log.

packages/test/src/load/starter.ts
Sushisource: done

packages/worker/src/worker.ts
Sushisource: Should I just undo this then? I'm not using the noisy workflow any longer anyway

packages/test/src/activities/index.ts
```diff
@@ -92,3 +93,16 @@ export async function progressiveSleep(): Promise<void> {
   await sleep(100);
   Context.current().heartbeat(3);
 }
+
+export async function queryOwnWf(queryDef: QueryDefinition<any>): Promise<void> {
```
bergundy, line 97: If you're adding this, why not make it:
```suggestion
export async function queryOwnWf<R, A extends any[]>(queryDef: QueryDefinition<R, A>, ...args: A): Promise<R> {
```

packages/test/src/load/starter.ts
```diff
@@ -4,6 +4,7 @@ import { bufferTime, map, mergeMap, tap, takeUntil } from 'rxjs/operators';
 import { Connection, WorkflowClient } from '@temporalio/client';
 import { StarterArgSpec, starterArgSpec, getRequired } from './args';
 import * as os from 'os';
+import { smorgasbord } from '../workflows';
```
bergundy: Unused

packages/test/src/load/starter.ts
bergundy: challenge accepted ;-)
Okay, I won't press this

packages/worker/src/worker.ts
bergundy: Let's just undo for now

PR #289, opened 2021-10-08
feat(workflow): Add condition helper function

bergundy:
First part of the redesigned Workflow API: https://github.com/temporalio/proposals/pull/44

### DIFF
--- a/packages/workflow/src/worker-interface.ts
+++ b/packages/workflow/src/worker-interface.ts
@@ -89,6 +89,11 @@ export function overrideGlobals(): void {
   Math.random = () => state.random();
 }
 
+/**
+ * Initialize the isolate runtime.
+ *
+ * Sets required internal state and instantiates the workflow and interceptors.
+ */
 export async function initRuntime(
   info: WorkflowInfo,
   interceptorModules: string[],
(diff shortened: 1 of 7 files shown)


packages/workflow/src/workflow.ts
```diff
@@ -760,3 +760,41 @@ function patchInternal(patchId: string, deprecated: boolean): boolean {
...
+ * @param timeout optional duration - {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
+ */
+export function condition(fn: () => boolean, timeout?: number | string): Promise<void> {
+  if (timeout) {
+    return CancellationScope.withTimeout(msToNumber(timeout), () => conditionInner(fn));
+  }
+  return conditionInner(fn);
+}
+
+function conditionInner(fn: () => boolean): Promise<void> {
```
cretz, line 776: Any harm in making this optionally accept a `Promise` return?

packages/workflow/src/workflow.ts
```diff
@@ -760,3 +760,50 @@ function patchInternal(patchId: string, deprecated: boolean): boolean {
...
+    const seq = state.nextSeqs.condition++;
+    if (scope.cancellable) {
+      scope.cancelRequested.catch((err) => {
+        state.blockedConditions.delete(seq);
+        reject(err);
+      });
+    }
+
+    // Eager evaluation
+    if (fn()) {
```
sw-yx, line 802: should we make sure this is `true`, not just truthy?

```ts
      if (cond.fn() === true) {
```

packages/workflow/src/worker-interface.ts
```diff
@@ -285,3 +299,27 @@ export function resolveExternalDependencies(results: ExternalDependencyResult[])
...
+ * Loop through all blocked conditions, evaluate and unblock if possible.
+ *
+ * @returns number of unblocked conditions.
+ */
+export function tryUnblockConditions(): number {
+  let numUnblocked = 0;
+  for (;;) {
+    const prevUnblocked = numUnblocked;
+    for (const [seq, cond] of state.blockedConditions.entries()) {
+      if (cond.fn()) {
```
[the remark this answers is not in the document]
bergundy: As discussed on slack, we might want to do a type check of the return value and make sure it's a boolean.
Let's open an issue and track.

PR #296, opened 2021-10-13
mind NPM readmes

sw-yx:
## What was changed
tweaked our npm readme

## Why?
one of my pet peeves is npm readmes that dont give you any help to find the real docs.


packages/common/README.md
```diff
@@ -4,4 +4,13 @@
 
 Part of the [Temporal](https://temporal.io) [Node.js SDK](https://www.npmjs.com/package/temporalio).
 
-See documentation on the [documentation site](https://docs.temporal.io/docs/node/introduction).
+You should probably not be using this package directly. See instead:
```
bergundy, line 7: Some things are exported only from this package (maybe we should re-export them where relevant though).

sw-yx: like what?

PR #317, opened 2021-10-20
feat!: Restructure code in prep for vm transition

bergundy:
- Decrease Workflow bundle size from ~7.44MB to ~2.75MB
- Remove otel module from `@temporalio/common` default export
- Rename `WorkflowIsolateBuilder` to `WorkflowCodeBundler` and remove
  unused methods
- Add `Workflow` and `WorkflowCreator` interfaces to support pluggable
  workflow environments (prepare for VM)
- Simplify external dependencies mechanism to only support `void`
  functions and remove the `isolated-vm` transfer options.


bergundy:
In a follow up PR we'll probably rename external dependencies to "sinks".

packages/worker/src/dependencies.ts
```diff
@@ -2,101 +2,47 @@
  * Type definitions for the Worker end of the external dependencies mechanism.
  *
  * External dependencies are functions injected into a Workflow isolate from the main Node.js isolate.
- * They are an advanced feature and should be used with care.
  *
- * @see proposal at https://github.com/temporalio/proposals/blob/master/node/logging-and-metrics-for-user-code.md
+ * Dependency functions may not return values to the Workflow in order to prevent breaking determinism.
```
Sushisource, line 6:
```suggestion
 * Dependency functions may not return values to the Workflow in order to help prevent breaking determinism.
```

Can't actually always prevent it

packages/worker/src/worker.ts
```diff
@@ -845,48 +812,71 @@ export class Worker<T extends WorkerSpec = DefaultWorkerSpec> {
...
+                          } catch (error) {
+                            this.log.error('External dependency function threw an error', {
+                              ifaceName,
+                              fnName,
+                              error,
+                              workflowInfo: state?.info,
+                            });
+                          }
+                        })();
+                      }
```
Sushisource, line 871: Can this bit be extracted to a function w/ a docstring?

Sushisource (approved):
Definitely nicer. Nothing blocking.

packages/worker/src/dependencies.ts
bergundy: It prevents this mechanism from interfering with the Workflow execution.

PR #321, opened 2021-10-21
clarify createWorkflowHandle isnt the same as starting workflows

swyxio:
addresses @joebowbeer concerns

## What was changed
just some docstrings

I also considered linking https://docs.temporal.io/docs/typescript/client#how-to-create-workflow-clients-and-handles in the docstring.. what do you think

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -505,7 +505,7 @@ export class WorkflowClient {
   }
 
   /**
-   * Create a {@link WorkflowHandle} for starting a new Workflow execution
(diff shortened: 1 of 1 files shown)


packages/client/src/workflow-client.ts
```diff
@@ -505,23 +505,23 @@ export class WorkflowClient {
   }
 
   /**
-   * Create a {@link WorkflowHandle} for starting a new Workflow execution
+   * Create a {@link WorkflowHandle}. Handles can start/execute, cancel/terminate, signal/query, describe or signalWithStart Workflows.
    *
    * @param name workflow type name (the exported function name in the Node.js SDK)
```
bergundy, line 510: I missed one :)

PR #341, opened 2021-10-28
chore: include index.d.ts in @temporalio/core-bridge package

fubhy:
## What was changed
Typescript currently doesn't detect the exported and re-exported types from core-bridge because the index.d.ts file is not published with the npm package. This PR adds it.


bergundy:
Sorry, I missed this PR, I was looking for an issue because I remembered that you sent me a link.
I've submitted a PR of my own for this (#347)

PR #343, opened 2021-10-29
feat!: Refactor WorkflowHandle creation

bergundy:
BREAKING CHANGE:
- `WorkflowClient.start` now returns a `WorkflowHandle`
- `WorkflowHandle` no longer has `start`, `signalWithStart` and
  `execute` methods
- `WorkflowClient.signalWithStart` was added
- To get a handle to an existing Workflow use `WorkflowClient.getHandle`
- `wf.createChildWorklowHandle` was renamed to `wf.startChild` and
  immediately starts the Workflow
- `wf.executeChild` replaces `ChildWorkflowHandle.execute`
- `wf.createExternalWorkflowHandle` was renamed to
  `wf.getExternalWorkflowHandle`

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -8,15 +8,21 @@ import {
   DataConverter,
   defaultDataConverter,
   composeInterceptors,
-  IllegalStateError,
   optionalFailureToOptionalError,
   Workflow,
   BaseWorkflowHandle,
   QueryDefinition,
   SignalDefinition,
   WorkflowResultType,
+  WithWorkflowArgs,
+  WorkflowReturnType,
 } from '@temporalio/common';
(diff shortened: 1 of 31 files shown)


packages/client/src/workflow-client.ts
```diff
@@ -106,6 +103,13 @@ export interface WorkflowHandle<T extends Workflow> extends BaseWorkflowHandle<T
   readonly client: WorkflowClient;
 }
 
+export interface WorkflowHandleWithRunId<T extends Workflow> extends WorkflowHandle<T> {
+  /**
+   * The runId of the initial run of the current Workflow
```
Sushisource, line 108: Might call this "bound" instead of "current"

Sushisource (approved):

PR #346, opened 2021-10-30
feat(worker): Add SIGUSR2 to default list of shutdown signals

### LINKED ISSUE #345: [Feature Request] Add SIGUSR2 to list of default worker shutdownSignals
This is the signal nodemon uses for graceful shutdown, it should be in the default list.

bergundy:
Closes #345

### DIFF
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -62,7 +62,7 @@ export interface WorkerOptions {
    * Automatically shut down worker on any of these signals.
    * @default
    * ```ts
-   * ['SIGINT', 'SIGTERM', 'SIGQUIT']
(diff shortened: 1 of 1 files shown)


cretz:
Not the biggest fan of taking a user signal away because of third party choice there, but since it's configurable, ok.

Were I using the library, I would not have the worker listen to any signals. I would be in charge of shutdown which is a common default with many libraries instead of them running code on signal you may be unaware of. Signal handling is the job of a binary/CLI, not a library.

cretz (approved):

bergundy:
I agree that typically signal handling is not the job of a library but since this is such a common pattern I feel like our worker should support it.
It also makes the samples work well and keeps them short.

bergundy:
I think we should document this clearly though

Issue #333, opened 2021-10-24
[Bug] Package initializer does not work for some users


sw-yx:
### Describe the bug

https://temporalio.slack.com/archives/C01DKSMU94L/p1634944174028300

### To Reproduce

i could reproduce (on both node 14 and 16) with

```
 npx @temporalio/create@latest moop
✔ Which sample would you like to use? › hello-world
Creating a new Temporal project in /Users/swyx/Temporal/testnodesdk/moop/

✔ Directory /Users/swyx/Temporal/testnodesdk/moop/ already exists. Would you like to replace it? … yes
Downloading files for sample hello-world. This might take a moment.

Installing packages. This might take a couple of minutes.

npm ERR! code ENOENT
npm ERR! syscall open
npm ERR! path /Users/swyx/package.json
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, open '/Users/swyx/package.json'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/swyx/.npm/_logs/2021-10-24T10_08_05_549Z-debug.log

Aborting installation.
  npm has failed.
```

notice that my error shows a different root than Ramazan's:

```
ramazanozer@Ramazans-MBP-39 alphademov2 % npx @temporalio/create@latest
✔ What is your project named? … my-temporal
✔ Which sample would you like to use? › hello-world
Creating a new Temporal project in /Users/ramazanozer/IdeaProjects/alphademov2/my-temporal/

Downloading files for sample hello-world. This might take a moment.

Installing packages. This might take a couple of minutes.


Aborting installation.
Unexpected error. Please report it as a bug:
Error: ENOENT: no such file or directory, open '/Users/ramazanozer/IdeaProjects/alphademov2/my-temporal/package.json'
    at Object.openSync (fs.js:497:3)
    at readFileSync (fs.js:393:35)
    at updateNodeVersion (/Users/ramazanozer/.npm/_npx/84edb3507a773a27/node_modules/@temporalio/create/lib/helpers/install.js:32:52)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async createApp (/Users/ramazanozer/.npm/_npx/84edb3507a773a27/node_modules/@temporalio/create/lib/create-project.js:141:5)
[...]

PR #349, opened 2021-11-01
Display .post-create message after install

lorensr:
## What was changed

- Display sample's `.post-create` message after install (see https://github.com/temporalio/samples-typescript/pull/43)

image

Also implemented the improvements from this issue: https://github.com/temporalio/sdk-typescript/issues/333#issuecomment-955113742

- add a check that covers when the provided repo has changed names/orgs:

```
Downloading files from repo https://github.com/lorensr/samples-node/tree/post-create/hello-world-mtls. This might take a moment.

We were unable to download and extract the provided project.
 Archive URL: https://codeload.github.com/lorensr/samples-node/tar.gz/post-create
 Archive path: samples-node-post-create/hello-world-mtls
 Sometimes this is due to the repo name changing. If that's the case, try using the new repo URL.
```

- print a list of samples and link to samples-typescript

image

## Why?

- Repetitive and sometimes incorrect post-create messages: https://temporaltechnologies.slack.com/archives/C01FT8U10GK/p1635207289037300
- Unclear error messages

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/333

2. How was this tested: 

```
lerna bootstrap
cd packages/create-project
npm run build
./cli.js myProj -s https://github.com/lorensr/samples-node/tree/post-create/hello-world-mtls
```

3. Any docs updates needed?
No

### DIFF
--- a/packages/create-project/src/create-project.ts
+++ b/packages/create-project/src/create-project.ts
@@ -127,6 +127,8 @@ export async function createApp({
       console.error('Exiting. You can re-run this command with a different project name.');
       process.exit(1);
     }
+
+    await rm(root, { recursive: true });
   }
 
   try {
--- a/packages/create-project/src/create-project.ts
+++ b/packages/create-project/src/create-project.ts
@@ -131,7 +131,7 @@ export async function createApp({
       process.exit(1);
     }
 
-    await rm(root, { recursive: true });
+    await rm(root, { recursive: true, force: true, maxRetries: 5 });
   }
 
   try {
@@ -196,7 +196,16 @@ export async function createApp({
   console.log();
   console.log(chalk.bold(appPath + '/'));
   console.log();
-  await spawn('npx', ['chalk', '-t', `"$(cat ${messageFile})"`], { stdio: 'inherit' });
-  console.log();
-  await rm(messageFile);
+
+  try {
+    await access(messageFile);
+    await spawn('npx', ['chalk', '-t', `"$(cat ${messageFile})"`], { stdio: 'inherit' });
+    console.log();
+    await rm(messageFile);
+  } catch (error) {
+    const code = getErrorCode(error);
+    if (code !== 'ENOENT') {
+      throw error;
+    }
+  }
 }
--- a/packages/create-project/src/create-project.ts
+++ b/packages/create-project/src/create-project.ts
@@ -20,6 +20,7 @@ import { isWriteable } from './helpers/is-writeable';
 import { getErrorCode } from './helpers/get-error-code';
 import { stripSnipComments } from './helpers/strip-snip-comments';
 import { spawn } from './helpers/subprocess';
+import { fetchSamples } from './helpers/fetch-samples';
 
 export class DownloadError extends Error {}
 
@@ -80,8 +81,10 @@ export async function createApp({
       console.error(
         `Could not locate a sample named ${chalk.red(`"${sample}"`)}. It could be due to the following:\n`,
         `1. Your spelling of sample ${chalk.red(`"${sample}"`)} might be incorrect.\n`,
-        `2. You might not be connected to the internet.`
+        `2. You might not be connected to the internet.\n`
       );
+      const samples = await fetchSamples();
+      console.error(`Available samples:\n\n${samples.join('\n')}\n`);
       process.exit(1);
     }
   }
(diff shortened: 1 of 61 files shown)


lorensr:
(this code shouldn't be published to npm until this is merged: https://github.com/temporalio/samples-typescript/pull/43)

bergundy:
Thanks @lorensr, I'll review in the morning

packages/create-project/src/create-project.ts
```diff
@@ -184,43 +187,16 @@ export async function createApp({
...
-    console.log();
   }
 
-  let cdpath: string;
-  if (path.join(originalDirectory, appName) === appPath) {
-    cdpath = appName;
-  } else {
-    cdpath = appPath;
-  }
+  const messageFile = path.join(root, '.post-create');
```
bergundy, line 192: We should not fail the process if this file does not exist IMHO because this tool accepts arbitrary URLs.

packages/create-project/src/create-project.ts
```diff
@@ -184,43 +187,16 @@ export async function createApp({
...
-    )
-  );
+  console.log(`${chalk.green('Success!')} Created project ${chalk.bold(appName)} at:`);
   console.log();
-  console.log(`Then, in the ${chalk.bold(cdpath + '/')} directory, using two other shells, run these commands:`);
+  console.log(chalk.bold(appPath + '/'));
   console.log();
-  console.log(`  ${chalk.cyan(`${displayedCommand} ${useYarn ? '' : 'run '}start.watch`)}`);
-  console.log(`  ${chalk.cyan(`${displayedCommand} ${useYarn ? '' : 'run '}workflow`)}`);
+  await spawn('npx', ['chalk', '-t', `"$(cat ${messageFile})"`], { stdio: 'inherit' });
```
bergundy, line 199: I'd prefer reading the file contents using node and using chalk-template directly instead of spawning a process: https://github.com/chalk/chalk-template

packages/create-project/src/helpers/samples.ts
```diff
@@ -91,13 +91,29 @@ export function hasSample(name: string): Promise<boolean> {
...
   );
+
+  const files = await readdir(root);
+  const pipelineFailed = files.length === 0;
+  if (pipelineFailed) {
+    console.error(
+      'We were unable to download and extract the provided project.\n',
+      `Archive URL: ${archiveUrl}\n`,
+      `Archive path: ${archivePath}\n`,
+      `Sometimes this is due to the repo name changing. If that's the case, try using the new repo URL.`
```
bergundy, line 113: I don't know if I'd want to bother with stating the repo change cause, we're not expecting to change the repo name anymore 😬

bergundy (commented):
Just some minor comments.

packages/create-project/src/create-project.ts
lorensr: Oops! 😄

packages/create-project/src/helpers/samples.ts
lorensr: This is only when they provide their own URL, and it's the only situation I know of in which we'd reach this code path. (the `URL/package.json` is checked earlier)

packages/create-project/src/create-project.ts
lorensr: It's not currently published to npm, but it looks like it [might be soon](https://github.com/chalk/chalk-template/pull/3#pullrequestreview-794649035). After this is merged I could open a draft PR based on the future package?

bergundy: So until then please avoid the shell expansion.
You should be able to feed the template directly as an arg here using:
```ts
await fs.readFile(messageFile, 'utf8')
```

lorensr: Done in https://github.com/temporalio/sdk-typescript/pull/349/commits/eb64788a1becec562d54d545ff3af6166f45d4e0

packages/create-project/src/create-project.ts
```diff
@@ -184,43 +192,26 @@ export async function createApp({
...
-  console.log(`Then, in the ${chalk.bold(cdpath + '/')} directory, using two other shells, run these commands:`);
-  console.log();
-  console.log(`  ${chalk.cyan(`${displayedCommand} ${useYarn ? '' : 'run '}start.watch`)}`);
-  console.log(`  ${chalk.cyan(`${displayedCommand} ${useYarn ? '' : 'run '}workflow`)}`);
+  console.log(chalk.bold(appPath + '/'));
   console.log();
+
+  try {
+    await access(messageFile);
+    const message = await readFile(messageFile);
```
bergundy, line 207: I'm definitely not blocking on this, but you might as well read the file as 'utf8' and get a string back.

packages/create-project/src/create-project.ts
```diff
@@ -184,43 +192,26 @@ export async function createApp({
...
+    await spawn('npx', ['chalk', '-t', message.toString()], { stdio: 'inherit' });
```
bergundy, line 208: Just verifying, you tested this right?

packages/create-project/tsconfig.json
```diff
@@ -4,7 +4,10 @@
     "outDir": "./lib",
     "rootDir": "./src",
     "resolveJsonModule": true,
-    "lib": ["dom"]
+    "lib": ["dom"],
+    "paths": {
+      "*": ["./src/types/*"]
```
bergundy, line 9: nit: we don't need the types here, the package would probably include types when it gets published.

bergundy (approved):
Nothing blocking, thanks!

packages/create-project/src/create-project.ts
lorensr: yep!

Issue #348, opened 2021-10-30
[Bug] Algolia broken


lorensr:
### Describe the bug

403 Error: `Index not allowed with this API key`

![image](https://user-images.githubusercontent.com/251288/139555719-863e6802-d4bd-45c0-8f5f-8be19441c247.png)

PR #350, opened 2021-11-01
fix(docs): Update Algolia index name

bergundy:


lorensr:
Fixes #348

sw-yx (approved):

bergundy:
I forced merged this because it's a trivial change and the samples tests will fail because that repo's `next` branch was updated and is no longer compatible with this branch.

PR #355, opened 2021-11-03
fix: Update core to gain infinite poll retries

Sushisource:
## What was changed
Update core

## Why?
We could end up producing a fatal error if polling was failing for a long time. Avoids that.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/core-bridge/Cargo.lock
+++ b/packages/core-bridge/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.7.4"
+version = "0.7.6"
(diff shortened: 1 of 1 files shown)


cretz:
General question, might it be worth using a non-local crate reference (e.g. with `git` and a rev or tag) instead of a submodule? Probably no huge difference, but doesn't require recursive clone and other potential submodule pitfalls.

cretz (approved):

Sushisource:
@cretz Unfortunately we can't do that because we need the proto files. Otherwise it would be nice.

PR #356, opened 2021-11-03
fix(worker): Fix Worker possible hang after graceful shutdown period expires

bergundy:


sw-yx (approved):

packages/worker/src/worker.ts
```diff
@@ -635,8 +639,8 @@ export class Worker {
                     failure: await errorToFailure(error, this.options.dataConverter),
                   },
                 }).finish();
-                // TODO: should we wait to be evicted from core?
-                state?.workflow.dispose();
+                // We do not dispose of the Workflow yet, wait to be evicted from Core.
+                // This is done to simplify the Workflow lifecycle so Core the sole driver.
```
Sushisource, line 643:
```suggestion
                // This is done to simplify the Workflow lifecycle so Core is the sole driver.
```

packages/worker/src/worker.ts
```diff
@@ -924,9 +928,17 @@ export class Worker {
         ),
         { defaultValue: undefined }
       );
-    } finally {
+      // Only shutdown the native worker if we completed without an error.
+      // Otherwise Rust / TS are in an unknown state and shutdown might hang.
+      // A new process must be created in order to instantiate a new Rust Core.
+      // TODO: force shutdown in core?
```
Sushisource, line 934: If at this point we are potentially in a state where we've diverged from what core thinks is happening (IE: It isn't shut down) - which seems like it could be the case, then yes, I would forcibly just unload core from memory to ensure things are kept consistent.

Alternatively if what you're saying (from the comment above) is process must exit at this point anyway, then I wouldn't bother. I'd just make sure we have a clear error message if someone tries to re-instantiate things that says "nope, just restart the whole process"

packages/worker/src/worker.ts
```diff
@@ -924,9 +928,17 @@ export class Worker {
         ),
         { defaultValue: undefined }
       );
-    } finally {
+      // Only shutdown the native worker if we completed without an error.
```
cretz, line 931: > Only shutdown the native worker if we completed without an error.

I may be misunderstanding the comment, but native worker complete shutdown is called regardless of error

packages/worker/src/worker.ts
```diff
@@ -924,9 +928,17 @@ export class Worker {
...
+      // Otherwise Rust / TS are in an unknown state and shutdown might hang.
+      // A new process must be created in order to instantiate a new Rust Core.
+      // TODO: force shutdown in core?
       await this.nativeWorker.completeShutdown();
-      await this.workflowCreator?.destroy();
+    } finally {
+      try {
+        await this.workflowCreator?.destroy();
```
cretz, line 938: Any concerns of, in an erroring situation, this throws and swallows the original error (assuming that's what JS does, I forget)? I can understand that an error here is very fatal, but unwinding from what may be a very fatal error in the original try, I wonder if this will hide what the real fatal error was.

bergundy: Luckily we log the error before we throw it

packages/worker/src/worker.ts
bergundy: I think for now let's just ask to restart the process, it'd be simpler.
I put this todo here to think about this.

PR #366, opened 2021-11-10
chore: Update core, changes for no more WF update errors

Sushisource:
## What was changed
Core updated, less work to do now that it handles these evictions automatically

## Why?
Correctness & simplification

## Checklist
https://github.com/temporalio/sdk-core/pull/217 must be merged & then update submodule


bergundy (approved):
Thanks!

Sushisource:
I think there might be a potential race in my changes, investigating that first before continuing here.

PR #369, opened 2021-11-10
fix(workflow): Throw if patches are used at Workflow top level

### LINKED ISSUE #362: [Bug] Patches should be applied to workflow before workflow code is required
We should support patches at the top level of user code as shown here:
```ts
let runActivity = false;
if (patched('p1')) {
  runActivity = true;
}

export async function myWorkflow() {
  if (runActivity) {
    await myActivity();
  }
}
```

bergundy:
Closes #362

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -598,13 +598,29 @@ export class Worker {
                         isReplaying: activation.isReplaying,
                       };
 
+                      const patchJobs = activation.jobs.filter(
+                        (
+                          j
+                        ): j is NonNullableObject<
+                          Pick<coresdk.workflow_activation.IWFActivationJob, 'notifyHasPatch'>
+                        > => j.notifyHasPatch != null
+                      );
+                      const patches = patchJobs.map(({ notifyHasPatch }) => {
+                        const { patchId } = notifyHasPatch;
+                        if (!patchId) {
+                          throw new TypeError('Got a patch without a patchId');
+                        }
+                        return patchId;
+                      });
+
                       const workflow = await instrument(this.tracer, span, 'workflow.create', async () => {
-                        return await workflowCreator.createWorkflow(
-                          workflowInfo,
-                          this.options.interceptors?.workflowModules ?? [],
-                          randomnessSeed.toBytes(),
-                          tsToMs(activation.timestamp)
-                        );
+                        return await workflowCreator.createWorkflow({
+                          info: workflowInfo,
+                          interceptorModules: this.options.interceptors?.workflowModules ?? [],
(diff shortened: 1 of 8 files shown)


packages/worker/src/worker.ts
```diff
@@ -598,13 +598,29 @@ export class Worker {
                         isReplaying: activation.isReplaying,
                       };
 
+                      const patchJobs = activation.jobs.filter(
+                        (
+                          j
+                        ): j is NonNullableObject<
+                          Pick<coresdk.workflow_activation.IWFActivationJob, 'notifyHasPatch'>
+                        > => j.notifyHasPatch != null
```
Sushisource, line 606: Oh my god this is some nauseating formatting (I know it's auto-formatted, but, jesus)

packages/worker/src/workflow/isolated-vm.ts
```diff
@@ -210,14 +195,13 @@ export class IsolatedVMWorkflow implements Workflow {
...
 
-    await context.evalClosure(
-      'lib.initRuntime($0, $1, $2, $3, $4)',
-      [info, interceptorModules, randomnessSeed, now, isolateExtension.derefInto()],
-      { arguments: { copy: true }, timeout: isolateExecutionTimeoutMs }
-    );
+    await context.evalClosure('lib.initRuntime($0, $1)', [options, isolateExtension.derefInto()], {
+      arguments: { copy: true },
+      timeout: isolateExecutionTimeoutMs,
+    });
```
Sushisource, line 201: That's nicer

packages/workflow/src/workflow.ts
```diff
@@ -814,13 +814,13 @@ export function deprecatePatch(patchId: string): void {
...
-  if (state.info === undefined) {
-    throw new IllegalStateError('Workflow info must be set when calling patch functions');
-  }
   // Patch operation does not support interception at the moment, if it did,
   // this would be the place to start the interception chain
 
-  const { isReplaying } = state.info;
+  const { isReplaying } = workflowInfo();
+  if (state.workflow === undefined) {
+    throw new IllegalStateError('Patches cannot be used before Workflow starts');
```
Sushisource, line 822: Lift string to constant for assertion in test?

Sushisource (approved):

packages/worker/src/worker.ts
bergundy: lol, I can declare the type somewhere else so the formatting is fixed

packages/workflow/src/workflow.ts
bergundy: It's just 1 test, I can but I wouldn't bother

PR #372, opened 2021-11-10
fix(workflow): Cancel timer created by condition

### LINKED ISSUE #363: [Bug] Timer is not cancelled when condition resolves
It is better to cancel in order to reduce load on the system.

bergundy:
Also clean up resources taken by the blocked condition.

Closes #363

### DIFF
--- a/packages/workflow/src/workflow.ts
+++ b/packages/workflow/src/workflow.ts
@@ -850,9 +850,19 @@ export function condition(fn: () => boolean, timeout: number | string): Promise<
  */
 export function condition(fn: () => boolean): Promise<void>;
 
-export function condition(fn: () => boolean, timeout?: number | string): Promise<void | boolean> {
+export async function condition(fn: () => boolean, timeout?: number | string): Promise<void | boolean> {
   if (timeout) {
-    return Promise.race([sleep(timeout).then(() => false), conditionInner(fn).then(() => true)]);
+    return await CancellationScope.cancellable(async () => {
+      try {
+        return await Promise.race([sleep(timeout).then(() => false), conditionInner(fn).then(() => true)]);
+      } finally {
+        // Use patched to avoid breaking existing histories using this `condition`
+        if (patched('__temporal-internal-condition-cancels-timer')) {
+          // Cancel either the condition or the timer
+          CancellationScope.current().cancel();
+        }
+      }
+    });
   }
   return conditionInner(fn);
 }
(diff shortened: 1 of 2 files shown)


packages/workflow/src/workflow.ts
```diff
@@ -847,9 +847,20 @@ export function condition(fn: () => boolean, timeout: number | string): Promise<
...
+export async function condition(fn: () => boolean, timeout?: number | string): Promise<void | boolean> {
   if (timeout) {
-    return Promise.race([sleep(timeout).then(() => false), conditionInner(fn).then(() => true)]);
+    return await CancellationScope.cancellable(async () => {
+      return await Promise.race([
+        sleep(timeout).then(() => false),
+        conditionInner(fn).then(() => {
+          // Use patched to avoid breaking existing histories using this `condition`
+          if (patched('__temporal-internal-condition-cancels-timer')) {
+            CancellationScope.current().cancel();
```
cretz, line 858: Forgive my lack of knowledge of the API, but since this is in a different part of the promise as a result of user code in `fn`, any concern that `CancellationScope.current()` may represent something besides the outer scope? I guess this may not be a concern because we don't allow async condition functions?

I wonder if we want to have `const cancelScope = CancellationScope.current()` as the first line in `cancellable` and change this to `cancelScope.cancel()` just to be safe.

bergundy: The scope is guaranteed to be the one we created because even if `fn` created a new scope it would be exited by the time `then` is called.
Now that I think of it, we should probably cancel the scope in case the timer resolves before the condition to clean up resources created by it.

packages/workflow/src/workflow.ts
```diff
@@ -840,18 +840,28 @@ function patchInternal(patchId: string, deprecated: boolean): boolean {
  *
  * @returns a boolean indicating whether the condition was true before the timeout expires
  */
-export function condition(timeout: number | string, fn: () => boolean): Promise<boolean>;
+export function condition(fn: () => boolean, timeout: number | string): Promise<boolean>;
```
cretz, line 843: I know this came from another PR that's already merged and I know we're in alpha, but should be noted that this change is a silent runtime failure for a JS user. I know that this is a "TypeScript library" but most TypeScript libraries consider themselves JS libraries too. No worries of course in alpha.

cretz (approved):

packages/workflow/src/workflow.ts
bergundy: Next week it won't be okay to break the API like this anymore.
We're making the final changes now.

PR #380, opened 2021-11-11
fix(workflow): Ensure signals are always processed

bergundy:
BREAKING CHANGE: This fixes a critical issue where the SDK was not processing histories in the right order.
Workflows that have been processed with older versions of the SDK may get stuck if the SDK is upgraded.

### DIFF
--- a/packages/test/src/test-signals-are-always-delivered.ts
+++ b/packages/test/src/test-signals-are-always-delivered.ts
@@ -1,0 +1,48 @@
+import test from 'ava';
+import { WorkflowClient } from '@temporalio/client';
+import { Worker, DefaultLogger, Core, InjectedSinks } from '@temporalio/worker';
+import { defaultOptions } from './mock-native-worker';
+import { RUN_INTEGRATION_TESTS } from './helpers';
+import * as workflows from './workflows';
+
+if (RUN_INTEGRATION_TESTS) {
+  test.before(async () => {
+    await Core.install({ logger: new DefaultLogger('DEBUG') });
+  });
+
+  test('Signals are always delivered', async (t) => {
+    const taskQueue = 'test-signal-delivery';
+    const conn = new WorkflowClient();
+    const wf = await conn.start(workflows.signalsAreAlwaysProcessed, { taskQueue, workflowTaskTimeout: '3s' });
+
+    const sinks: InjectedSinks<workflows.SignalProcessTestSinks> = {
+      controller: {
+        sendSignal: {
+          async fn() {
+            await wf.signal(workflows.incrementSignal);
+          },
+        },
+      },
+    };
+
+    const worker = await Worker.create({
+      ...defaultOptions,
+      taskQueue,
+      sinks,
+    });
+
+    await Promise.all([
+      worker.run(),
+      (async () => {
+        try {
+          await wf.result();
+        } finally {
+          worker.shutdown();
+        }
+      })(),
+    ]);
+
+    // Workflow completes if it got the signal
+    t.pass();
+  });
+}
--- a/packages/test/src/test-signals-are-always-delivered.ts
+++ b/packages/test/src/test-signals-are-always-delivered.ts
@@ -1,3 +1,11 @@
+/**
+ * Tests that if a signal is delivered while the Worker is processing a Workflow
+ * Task, the Worker picks up a new Workflow Task (including the signal) and
+ * the Workflow library delivers the signal to user code before it starts the
+ * Workflow execution.
+ *
+ * @module
+ */
 import test from 'ava';
 import { WorkflowClient } from '@temporalio/client';
 import { Worker, DefaultLogger, Core, InjectedSinks } from '@temporalio/worker';
@@ -13,12 +21,13 @@ if (RUN_INTEGRATION_TESTS) {
   test('Signals are always delivered', async (t) => {
     const taskQueue = 'test-signal-delivery';
     const conn = new WorkflowClient();
-    const wf = await conn.start(workflows.signalsAreAlwaysProcessed, { taskQueue, workflowTaskTimeout: '3s' });
+    const wf = await conn.start(workflows.signalsAreAlwaysProcessed, { taskQueue });
 
     const sinks: InjectedSinks<workflows.SignalProcessTestSinks> = {
       controller: {
         sendSignal: {
           async fn() {
+            // Send a signal to the Workflow which will cause the WFT to fail
             await wf.signal(workflows.incrementSignal);
           },
         },
--- a/packages/test/src/workflows/signals-are-always-processed.ts
+++ b/packages/test/src/workflows/signals-are-always-processed.ts
@@ -1,3 +1,9 @@
+/**
+ * Workflow used by test-signals-are-always-processed.ts
+ *
+ * @module
+ */
+
 import { proxySinks, Sinks, setHandler, defineSignal } from '@temporalio/workflow';
 
 export const incrementSignal = defineSignal('increment');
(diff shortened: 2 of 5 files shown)


packages/test/src/test-signals-are-always-delivered.ts
```diff
@@ -0,0 +1,48 @@
+import test from 'ava';
+import { WorkflowClient } from '@temporalio/client';
+import { Worker, DefaultLogger, Core, InjectedSinks } from '@temporalio/worker';
+import { defaultOptions } from './mock-native-worker';
+import { RUN_INTEGRATION_TESTS } from './helpers';
+import * as workflows from './workflows';
+
+if (RUN_INTEGRATION_TESTS) {
```
cretz, line 8: Pedantic - can save indenting all the code if you just did a `if (!RUN_INTEGRATION_TESTS) return;`

packages/test/src/test-signals-are-always-delivered.ts
```diff
@@ -0,0 +1,48 @@
...
+  test.before(async () => {
+    await Core.install({ logger: new DefaultLogger('DEBUG') });
+  });
+
+  test('Signals are always delivered', async (t) => {
```
cretz, line 13: Can you confirm this exact test as written fails without your fix? I am not seeing an assertion that signal is ever sent or received.

Should you at least return the counter from the workflow, or set an external variable from the sink call that proves it was run, or something similar?

packages/test/src/workflows/signals-are-always-processed.ts
```diff
@@ -0,0 +1,19 @@
...
+    sendSignal(): void;
+  };
+}
+
+const { controller } = proxySinks<SignalProcessTestSinks>();
+
+export async function signalsAreAlwaysProcessed(): Promise<void> {
+  let counter = 0;
+  setHandler(incrementSignal, () => void ++counter);
+  if (counter === 0) {
```
cretz, line 16: I do not understand the workflow here (sorry a bit unfamiliar w/ the TS SDK). How will this not always be zero if the counter is not incremented until after this?

packages/test/src/workflows/signals-are-always-processed.ts
```diff
@@ -0,0 +1,19 @@
...
+    controller.sendSignal();
```
cretz, line 17: Pardon my lack of TS SDK knowledge wrt sinks. So this asynchronously invokes the `sendSignal` call in the worker JS? And of course since this is deterministic, so it surely doesn't wait for some non-deterministic worker-side stuff right? So if it doesn't wait on anything, this simply is like a beacon to the worker-side and then this workflow exits immediately?

packages/test/src/test-signals-are-always-delivered.ts
bergundy: Return can only be used in function body though

packages/test/src/test-signals-are-always-delivered.ts
bergundy: The workflow checks the counter is set.
I wrote the test before the fix.
What happens if the signal isn't received is that the workflow will never be able to complete because it'll keep receiving signals until the history exceeds the limit.
I'll clarify and maybe make this more explicit in the test.

packages/test/src/workflows/signals-are-always-processed.ts
bergundy: The counter is incremented in the signal handler.
Signals are buffered and delivered immediately on handler registration.

packages/test/src/workflows/signals-are-always-processed.ts
bergundy: The sink function is invoked and blocks workflow task completion.
If a signal is received while workflow task is being processed the task will fail and the workflow will need to be replayed.

bergundy:
Thanks @cretz I'll clarify all of this in docstrings.

packages/test/src/test-signals-are-always-delivered.ts
cretz: It works in Node IIRC, but no prob.

packages/test/src/workflows/signals-are-always-processed.ts
cretz: Hrmm, reading the code one might expect without awaits anywhere that it would immediately run and return. In normal JS (browser, node, etc) I don't think one would expect the callback in `setHandler` to run before the next statement.

packages/test/src/workflows/signals-are-always-processed.ts
cretz: Hrmm, this can definitely be confusing for normal JS writers. A synchronous call that magically blocks for an internal asynchronous call seems off from all other JS.

Also, is it ok that these sinks may do something non-deterministic? What if it errors some of the time? Or is this like side effects in other SDKs where this is only executed once and result is stored in history?

bergundy: The synchronous call doesn't block, it's buffered in the workflow runtime.
When the worker completes the activation and the isolate finishes its work, the worker will pull out all of the buffered sink calls and run the injected sink functions.
The functions will block the worker from completing the workflow.
Sinks cannot return values back to the workflow or generate any workflow commands.

bergundy: If a sink function throws we log it and ignore the error.

packages/test/src/workflows/signals-are-always-processed.ts
bergundy: We have to deliver the signal otherwise it'd be lost.

cretz: I see. I think this may be a confusion in the API. Arguably signals should not be a handler that you don't know when it will happen, but rather something that can be awaited on at a point in the workflow where you care. One might expect to be able to `await signal.receive()`. As it is now I would probably add this helper (untested/pseudo) for myself:

```typescript
waitSignal<T>(signal: SignalDefinition<T>): Promise<T> {
  return new Promise(res => setHandler(signal, res))
}
```

The other SDKs let you choose exactly where to wait for signals. For example, take a workflow example in Go from this morning at https://community.temporal.io/t/signalling-workflows/3370/8. That user wants to receive a signal in a loop, waiting to execute the the activity until after each signal.

packages/test/src/workflows/signals-are-always-processed.ts
cretz: > The synchronous call doesn't block, it's buffered in the workflow runtime.

To confirm, from the perspective of the JS writer here, that `controller.sendSignal()` does not return until the sink call is done? Since this blocks in a single-threaded runtime, how can I do something like `Promise.race([sink.sendSignal1(), sink.sendSignal2()])`? Or is it known that sink will pause the entire runtime, regardless of whether the sink call itself may be doing something asynchronous, and therefore two sink calls cannot run at once, even if they do something asynchronous like network IO?

packages/test/src/workflows/signals-are-always-processed.ts
cretz: To confirm, maybe it's Go that's weird that lets you choose each time to receive a signal value instead of having a signal forced upon you. I see Java also operates on this signal handler mechanism.

packages/test/src/workflows/signals-are-always-processed.ts
bergundy: Let’s set up some time and I’ll brief you about how this works and why it’s done that way.
Same as for why signals are handled via callbacks

cretz (approved):

PR #387, opened 2021-11-12
feat!(client): Make workflowId required

bergundy:
This also removes the `WorkflowClientOptions.workflowDefaults`.

Reasoning:
- Workflow IDs should represent a meaningful business ID
- Workflow IDs can be used as an idempotency key when starting workflows from an external signal
- `workflowDefaults` were removed because their presence made `taskQueue` optional in `WorkflowOptions`, omitting it from both the defaults and options resulted in runtime errors where we could have caught those at compile time.\

Migration:
```ts
// Before
const client = new WorkflowClient(conn.service, { workflowDefaults: { taskQueue: 'example' } });
const handle = await client.start(myWorkflow, { args: [foo, bar] });
// After
const client = new WorkflowClient(conn.service);
const handle = await client.start(myWorkflow, {
  args: [foo, bar],
  taskQueue: 'example',
  workflowId: 'a-meaningful-business-id',
});
```

### DIFF
--- a/packages/workflow/src/interfaces.ts
+++ b/packages/workflow/src/interfaces.ts
@@ -1,5 +1,5 @@
 import { coresdk } from '@temporalio/proto/lib/coresdk';
-import { WorkflowOptions } from '@temporalio/common';
+import { CommonWorkflowOptions } from '@temporalio/common';
 
 /**
  * Workflow execution information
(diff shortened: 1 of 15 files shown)


packages/workflow/src/interfaces.ts
```diff
@@ -84,7 +84,20 @@ export const ChildWorkflowCancellationType = coresdk.child_workflow.ChildWorkflo
...
 export const ParentClosePolicy = coresdk.child_workflow.ParentClosePolicy;
 
-export interface ChildWorkflowOptions extends WorkflowOptions {
+export interface ChildWorkflowOptions extends CommonWorkflowOptions {
+  /**
+   * Workflow id to use when starting. If not specified a UUID is generated. Note that it is
+   * dangerous as in case of client side retries no deduplication will happen based on the
+   * generated id. So prefer assigning business meaningful ids if possible.
+   */
+  workflowId?: string;
```
sw-yx, line 93: still making this optional?

sw-yx (approved):

packages/workflow/src/interfaces.ts
bergundy: Just for child workflows as discussed

PR #398, opened 2021-11-22
feat(worker): Add `WorkerOptions.debugMode` to enable debugging Workflows

### LINKED ISSUE #102: [Feature Request] Add ability to debug Workflow code in isolates

Currently it is not possible to set up a debugger for Workflow code, this is an important feature which must be easy to use and well documented.

Debugging Workflows should be as seamless as possible (considering the limitations when working with multiple v8 contexts).

**Additional context**

[This](https://github.com/laverdet/isolated-vm/blob/main/inspector-example.js) should be a good starting point

bergundy:
Closes #102 

See related discussion here: https://stackoverflow.com/questions/54251797/visual-studio-node-debug-into-worker-threads-node-11


cretz:
@bergundy - To confirm, from that SO link, it sounds like vscode can now debug into the separate threads? Does that mean you only really need to set this value to true if you're using something like chrome dev tools? Have you tested w/ vscode-js-debug to see if it works across threads well?

sw-yx (approved):

bergundy:
@cretz I checked and couldn't debug in threads from vscode or Chrome.

gdw2:
For me, even when `debugMode: true`, chrome devtools are only showing me my activity source files and not any of my workflow source files.  I'm running temporalio `0.16.3` which includes this change.

PR #401, opened 2021-11-26
Added MakeOptional and Replace type helpers

JoshuaKGoldberg:
## What was changed
Adds a couple of type system helpers to clarify existing code. Also deduplicates a little bit of runtime logic in `falsyMsToTs`.

## Why?
Following up from a discussion with @sw-yx: these utilities create a little more clear end result code in the type system. Less duplication, and names that semantically imply what's happening.

## Checklist

1. How was this tested:
`npm run build` & `npm run build.watch`


bergundy (approved):
LGTM, thanks

packages/common/src/time.ts
```diff
@@ -44,6 +44,10 @@ export function msNumberToTs(millis: number): Timestamp {
   return { seconds: Long.fromNumber(seconds), nanos };
 }
 
+export function falsyMsToTs(str: string | number | undefined): Timestamp | undefined {
```
bergundy, line 47: Why do we need this falsy version? We have the optional signature below

PR #402, opened 2021-11-26
Added comments for unused query and signal generics

JoshuaKGoldberg:
## What was changed
Adds comments indicating why the seemingly unused generics are useful.

## Why?
I went through a whole investigation trying to remove them, only to have @sw-yx show me that the `Handler` type does, in fact, use the generics well.

## Checklist
1. How was this tested:
`npm run build` / `npm run build.watch`

### DIFF
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -14,7 +14,7 @@ export type Workflow = (...args: any[]) => WorkflowReturnType;
 /**
  * An interface representing a Workflow signal definition, as returned from {@link defineSignal}
  *
- * @remarks `_Args` can be used for parameter type inference in handler functions.
+ * @remarks `_Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
  */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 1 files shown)


packages/common/src/interfaces.ts
```diff
@@ -13,6 +13,8 @@ export type Workflow = (...args: any[]) => WorkflowReturnType;
 
 /**
  * An interface representing a Workflow signal definition, as returned from {@link defineSignal}
+ *
+ * @remarks `_Args` can be used for parameter type inference in handler functions.
```
bergundy, line 17: Not just handler functions, also when using `*WorkflowHandle` methods

packages/common/src/interfaces.ts
```diff
@@ -22,6 +24,8 @@ export interface SignalDefinition<_Args extends any[] = []> {
 
 /**
  * An interface representing a Workflow query definition as returned from {@link defineQuery}
+ *
+ * @remarks `_Args` can be used for parameter type inference in handler functions.
```
bergundy, line 28: Here too, `_Args` and `_Ret` both are used in `WorkflowHandle`

bergundy:
Thanks!

bergundy (approved):

PR #404, opened 2021-11-26
Expose BundleOptions and remove __namedParameters

lorensr:
- chore(docs): Prevent __namedParameters params from being generated
- chore(docs): Expose BundleOptions

Current docs:

image

Proposed:

image

### DIFF
--- a/packages/worker/src/logger.ts
+++ b/packages/worker/src/logger.ts
@@ -31,7 +31,9 @@ const severities: LogLevel[] = ['TRACE', 'DEBUG', 'INFO', 'WARN', 'ERROR'];
 /**
  * Log messages using `console.error` and basic formatting
  */
-function defaultLogFunction({ level, timestampNanos, message, meta }: LogEntry): void {
+function defaultLogFunction(entry: LogEntry): void {
+  const { level, timestampNanos, message, meta } = entry;
+
   const date = new Date(Number(timestampNanos / 1_000_000n));
   if (meta === undefined) {
     console.error(date, `[${level}]`, message);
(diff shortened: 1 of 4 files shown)


packages/worker/src/logger.ts
```diff
@@ -43,7 +45,8 @@ function defaultLogFunction({ level, timestampNanos, message, meta }: LogEntry):
 /**
  * Takes a `[seconds, nanos]` tuple as returned from getTimeOfDay and turns it into bigint.
  */
-export function timeOfDayToBigint([seconds, nanos]: [number, number]): bigint {
+export function timeOfDayToBigint(timeOfDay: [number, number]): bigint {
```
bergundy, line 48: I don't think we expose this to users, we can keep it they way it was

packages/worker/src/logger.ts
```diff
@@ -31,7 +31,9 @@ const severities: LogLevel[] = ['TRACE', 'DEBUG', 'INFO', 'WARN', 'ERROR'];
 /**
  * Log messages using `console.error` and basic formatting
  */
-function defaultLogFunction({ level, timestampNanos, message, meta }: LogEntry): void {
+function defaultLogFunction(entry: LogEntry): void {
```
bergundy, line 34: We don't expose this to users, we can keep it they way it was

bergundy (commented):
Thanks, looks like a typedoc limitation, right?

lorensr:
If we upgraded, we could alternatively use `@param` comments. 

image

It would be more verbose in cases when we already have a type like `BundleOptions`.

Trying latest (`typedoc@0.22.10`) didn't immediately work for me, and after brief Googling, I didn't dig:

image

packages/worker/src/logger.ts
lorensr: I only updated functions that show up in search:

https://typescript.temporal.io/search?q=__namedPara

image

bergundy (approved):
Thanks @lorensr, I'l try to get this to pass CI after I merge https://github.com/temporalio/sdk-typescript/pull/414

PR #405, opened 2021-11-28
Added ts-prune to detect unused code

JoshuaKGoldberg:
## What was changed
Adds [ts-prune](https://github.com/nadeesha/ts-prune), a nifty tool that detects unused code.

Copying contents & threads from https://github.com/JoshuaKGoldberg/sdk-typescript/pull/2 with the addition of a CI job.

## Why?
It found a few unused items and would be useful in CI. [[Slack thread](https://temporaltechnologies.slack.com/archives/C01FT8U10GK/p1637960205341800)]

## Checklist

1. How was this tested:
See CI job, once it's running.

Note that I can't run the GHA job as I'm an external contributor working from a fork. Someone will have to try that out for me / set me up to do it.


packages/worker/src/utils.ts
```diff
@@ -1,8 +1,5 @@
-export const sleep = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
-export const identity = <T>(x: T): T => x;
```
JoshuaKGoldberg, line 2: `identity` was unused. `sleep` was only used in a test file so I moved it there.

packages/worker/src/utils.ts
```diff
@@ -1,8 +1,5 @@
...
-
-export const GiB = 1024 ** 3;
 export const MiB = 1024 ** 2;
+export const GiB = MiB * 1024;
```
JoshuaKGoldberg, line 2: `MiB` was unused but factors into `GiB` nicely.

tsconfig.prune.json
```diff
@@ -0,0 +1,18 @@
...
+  "files": [
+    "./packages/activity/src/index.ts",
+    "./packages/client/src/index.ts",
+    "./packages/common/src/index.ts",
+    "./packages/create-project/src/index.ts",
+    "./packages/meta/src/index.ts",
+    "./packages/worker/src/index.ts",
+    "./packages/worker/src/workflow/workflow-worker-thread.ts",
+    "./packages/workflow/src/index.ts"
+  ]
```
JoshuaKGoldberg, line 17: These `files` act as the entry points for unused code detection. In other words, code things that are transitively imported by these files are counted as used.

packages/worker/src/rxutils.ts
```diff
@@ -6,20 +6,9 @@ interface StateAndOptionalOutput<T, O> {
   output?: O;
 }
 
-export type StateAndOutput<T, O> = Required<StateAndOptionalOutput<T, O>>;
+export const definitelyUnusedRemoveOnceWeKnowCiWorks = true;
 
-export function mapWithState<T, I, O>(
```
JoshuaKGoldberg, line 11: Totally unused 🤘

packages/worker/src/index.ts
```diff
@@ -11,7 +11,7 @@
 export { State, Worker, ActivityOptions, DataConverter, defaultDataConverter, RetryOptions, errors } from './worker';
 export { WorkerOptions, CompiledWorkerOptions } from './worker-options';
 export { ServerOptions, TLSConfig } from './server-options';
-export { Core } from './core';
+export { Core, CompiledCoreOptions, CoreOptions, TelemetryOptions } from './core';
```
JoshuaKGoldberg, line 14: These options interfaces are used in `Core`'s functions but not exported. Perhaps not required for unused code detection, but a nice to export.

package.json
```diff
@@ -45,6 +46,7 @@
     "ci-nightly-sampler": "node ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --otel-url='' --iterations 1000 --min-wfs-per-sec 5 --max-cached-wfs 500 --workflow yummy-sampler-mode",
     "wait-namespace": "node ./scripts/wait-on-temporal.mjs",
     "lint": "eslint packages/*/src --ext .ts --no-error-on-unmatched-pattern && prettier --end-of-line auto --check .",
+    "lint.prune": "! ts-prune -p tsconfig.prune.json --ignore \"used in module\" | grep -v \".d.ts\"",
```
JoshuaKGoldberg, line 49: Until https://github.com/nadeesha/ts-prune/issues/156 is resolved, `--skip` doesn't work. Hence the `| grep -v`. 

By default `grep` returns `0` if found and `1` or not. We actually want the opposite... so the `!` at the beginning inverts the exist code. 🧠

tsconfig.prune.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "files": [
+    "./packages/activity/src/index.ts",
+    "./packages/client/src/index.ts",
+    "./packages/common/src/index.ts",
+    "./packages/create-project/src/index.ts",
+    "./packages/meta/src/index.ts",
+    "./packages/worker/src/index.ts",
+    "./packages/worker/src/workflow/workflow-worker-thread.ts",
+    "./packages/workflow/src/index.ts"
```
bergundy, line 16: We should add `./packages/workflow/src/worker-interface.ts` too

package.json
bergundy: Will this syntax work on windows?
Might be worth adding the prune CI step to the build-lint-test CI job to verify.

packages/worker/src/index.ts
bergundy: 👍

package.json
JoshuaKGoldberg: Alas, it will not -- Windows also doesn't have `!` or `grep` 😞....

JoshuaKGoldberg: Aha, the upstream issue's fixed was released!

tsconfig.prune.json
bergundy: @JoshuaKGoldberg I want to merge this, can you please add the missing file?

JoshuaKGoldberg: Ah, sorry, I'd missed this notification! Merging now.

bergundy (approved):

PR #415, opened 2021-11-30
feat(workflow): Handle unhandled rejections in workflow code

bergundy:
This PR makes the best effort to associate unhandled rejections from workflow code to a specific runId.
It also makes the unhandled rejection behavior consistent between node 14 and 16 and propagates failure back to the user.
Previously, in node 16 the process would crash and in node 14 we would incorrectly ignore rejections leading to unexpected workflow behavior (see the correct behavior in the added tests).

### DIFF
--- a/packages/test/src/run-a-worker.ts
+++ b/packages/test/src/run-a-worker.ts
@@ -1,8 +1,12 @@
+import arg from 'arg';
 import { Worker, Core, DefaultLogger } from '@temporalio/worker';
 import * as activities from './activities';
 
 async function main() {
-  if (['1', 'y', 'yes', 't', 'true'].includes((process.env.DEBUG ?? '').toLowerCase())) {
+  const argv = arg({
+    '--debug': Boolean,
+  });
+  if (argv['--debug']) {
     await Core.install({
       logger: new DefaultLogger('DEBUG'),
       telemetryOptions: {
--- a/packages/test/src/run-a-worker.ts
+++ b/packages/test/src/run-a-worker.ts
@@ -1,7 +1,16 @@
-import { Worker } from '@temporalio/worker';
+import { Worker, Core, DefaultLogger } from '@temporalio/worker';
 import * as activities from './activities';
 
 async function main() {
+  if (['1', 'y', 'yes', 't', 'true'].includes((process.env.DEBUG ?? '').toLowerCase())) {
+    await Core.install({
+      logger: new DefaultLogger('DEBUG'),
+      telemetryOptions: {
+        logForwardingLevel: 'DEBUG',
+        tracingFilter: 'temporal_sdk_core=DEBUG',
+      },
+    });
+  }
   const worker = await Worker.create({
     activities,
     workflowsPath: require.resolve('./workflows'),
--- a/packages/test/src/workflows/unhandled-rejection.ts
+++ b/packages/test/src/workflows/unhandled-rejection.ts
@@ -1,0 +1,23 @@
+import { proxyActivities } from '@temporalio/workflow';
+import type * as activities from '../activities';
+
+const { echo } = proxyActivities<typeof activities>({
+  startToCloseTimeout: '1 minute',
+});
+
+export async function throwUnhandledRejection({ crashWorker }: { crashWorker: boolean }): Promise<void> {
+  const p1 = (async () => {
+    await echo('a');
+  })();
+
+  const p2 = (async () => {
+    if (crashWorker) {
+      throw 'not an error to crash the worker';
+    } else {
+      throw new Error('unhandled rejection');
+    }
+  })();
+
+  await p1;
+  await p2;
+}
--- a/packages/test/src/workflows/unhandled-rejection.ts
+++ b/packages/test/src/workflows/unhandled-rejection.ts
@@ -12,7 +12,9 @@ export async function throwUnhandledRejection({ crashWorker }: { crashWorker: bo
 
   const p2 = (async () => {
     if (crashWorker) {
-      throw 'not an error to crash the worker';
+      // Create a Promise associated with the worker thread context
+      const Promise = globalThis.constructor.constructor('return Promise')();
+      Promise.reject(new Error('error to crash the worker'));
     } else {
       throw new Error('unhandled rejection');
     }
--- a/packages/worker/src/workflow/threaded-vm.ts
+++ b/packages/worker/src/workflow/threaded-vm.ts
@@ -48,6 +48,8 @@ function errorNameToClass(name: string): ErrorConstructor {
 export class WorkerThreadClient {
   requestIdx = 0n;
   requestIdToCompletion = new Map<BigInt, Completion<WorkerThreadOutput>>();
+  shutDownRequested = false;
+  workerExited = false;
 
   constructor(protected workerThread: Worker) {
     workerThread.on('message', ({ requestId, result }: WorkerThreadResponse) => {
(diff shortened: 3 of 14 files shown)


bergundy:
The solution here is a bit complex and might not cover enough cases in the wild.
I'm looking into alternatives, this is the best I could come up with for now.

bergundy:
v8 has [`Object::GetCreationContext()`](https://v8.github.io/api/head/classv8_1_1Object.html#ab202f34fa042f5803e8f4de2d03ed8f8) but I don't see it exposed in node, we might need native code to access this method.

packages/test/src/workflows/unhandled-rejection.ts
```diff
@@ -0,0 +1,23 @@
...
+  const p2 = (async () => {
+    if (crashWorker) {
+      throw 'not an error to crash the worker';
+    } else {
+      throw new Error('unhandled rejection');
+    }
+  })();
+
+  await p1;
+  await p2;
```
cretz, line 22: Does this not just rethrow here? Before this PR, is this treated any different than if I just replaced this line with `throw new Error('whatever')`?

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -66,6 +68,23 @@ export class WorkerThreadClient {
...
+    workerThread.on('exit', () => {
+      this.workerExited = true;
+      if (this.shutDownRequested) {
+        return; // ignore
+      }
+      const completions = this.requestIdToCompletion.values();
+      this.requestIdToCompletion = new Map();
+      for (const completion of completions) {
+        completion.reject(
+          new UnexpectedError(
```
cretz, line 80: Sorry I am unfamiliar with how the threads work here. Is there no way we can capture an error thrown from an async workflow instead of having it terminate a worker thread? Can the workflow function not be wrapped in a try+await+catch _inside_ the thread? If concerned about top-level code throwing, can the entire `require` be wrapped in a try+await+catch?

packages/test/src/workflows/unhandled-rejection.ts
bergundy: The problem is that the second async function throws without anything catching it.
The way v8 deals with it is by providing a hook for unhandled rejections: https://v8.github.io/api/head/classv8_1_1Isolate.html#a702f0ba4e5dee8a98aeb92239d58784e.
This is exposed in node with `process.on('unhandledRejection')`.

packages/worker/src/workflow/threaded-vm.ts
bergundy: We terminate the thread if it we can't determine the context that threw the handled error to avoid accidentally completing the activation successfully.
There's no way to catch these errors unfortunately.

packages/test/src/workflows/unhandled-rejection.ts
cretz: But since that promise is `await`ed on here, it throws here right? If you removed `await p2` I think it'd be unhandled, but the promise is explicitly awaited on.

packages/test/src/run-a-worker.ts
```diff
@@ -1,7 +1,16 @@
-import { Worker } from '@temporalio/worker';
+import { Worker, Core, DefaultLogger } from '@temporalio/worker';
 import * as activities from './activities';
 
 async function main() {
+  if (['1', 'y', 'yes', 't', 'true'].includes((process.env.DEBUG ?? '').toLowerCase())) {
```
Sushisource, line 5: IMO Just checking for existence is much better than this dance. Easy enough to clear an env var if you want to set it to false.

Also just `DEBUG` is too generic I think. Something more temporal specific might be good.

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -23,6 +23,31 @@ function ok(requestId: BigInt): WorkerThreadResponse {
...
+// Best effort to catch unhandled rejections from workflow code.
+// We crash the thread if we cannot find the culprit.
+process.on('unhandledRejection', (err) => {
+  if (isErrorWithStack(err)) {
+    const match = err.stack.match(/\[as temporal-workflow-enter:(\S+)\]/);
+    if (match) {
+      const runId = match[1];
+      const workflow = workflowByRunId.get(runId);
+      if (workflow !== undefined) {
+        console.log('found workflow', runId);
```
Sushisource, line 39: Leftover console log?

packages/worker/src/workflow/vm.ts
```diff
@@ -163,7 +169,24 @@ export class VMWorkflow implements Workflow {
       // Wait for microtasks to be processed
       await new Promise((resolve) => process.nextTick(resolve));
     }
-    return this.workflowModule.concludeActivation();
+    const ret = this.workflowModule.concludeActivation();
+    // Give unhandledRejection handler a chance to process.
+    // Apparently nextTick does not get it triggered so we use setTimeout here.
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
Sushisource, line 175: Seems odd. Maybe we could set a flag in the handler, and run next tick until we see the flag set (then unset it again) ?

bergundy:
At this point my proposed solution seems both incomplete and flaky.
I'm going to bench this for now because I can't find a reliable way to handle unhandled rejections on time.
Looks like even `setTimeout` doesn't work if many workflows are run concurrently.

For now I think the best thing to do is crash the worker in node 14 so the behavior is at least consistent with node 16 and we don't run into issues where workflows are stuck and cannot be replayed.
This would be the case if errors were turned into workflow task failures instead of failing the entire workflow (as they are treated in the other SDKs).

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -23,23 +23,18 @@ function ok(requestId: BigInt): WorkerThreadResponse {
...
-        return;
-      }
+process.on('unhandledRejection', (err, promise) => {
+  // Get the runId associated with the vm context.
+  // See for reference https://github.com/patriksimek/vm2/issues/32
+  const ctor = promise.constructor.constructor;
+  const runId = ctor('return __TEMPORAL__.runId')();
+  if (runId !== undefined) {
+    const workflow = workflowByRunId.get(runId);
+    if (workflow !== undefined) {
```
Sushisource, line 35: The else here seems like a weird situation that we might want to log on or something.

packages/test/src/run-a-worker.ts
cretz: FWIW, don't use `TEMPORAL_DEBUG`. That env var is used in Java and Go SDKs to remove the workflow deadlock timer so code can be stepped slowly.

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -23,6 +23,26 @@ function ok(requestId: BigInt): WorkerThreadResponse {
 let workflowCreator: VMWorkflowCreator | undefined;
 const workflowByRunId = new Map<string, VMWorkflow>();
 
+// Best effort to catch unhandled rejections from workflow code.
+// We crash the thread if we cannot find the culprit.
+process.on('unhandledRejection', (err, promise) => {
```
cretz, line 28: To confirm, this is run from the outside not the inside, right? Is there any way to catch an unhandled rejection _inside_ the VM or add such a handler to just the VM context only?

packages/worker/src/workflow/vm.ts
bergundy: I think it might not be enough

packages/worker/src/workflow/workflow-worker-thread.ts
bergundy: I logged the runId below

cretz (approved):

PR #423, opened 2021-12-07
feat: Remove proto usage from workflow runtime

bergundy:
This is now possible because we're using vm instead of isolated-vm.

- Greatly reduce workflow bundle size - our test bundle size went down from 2.77MB to 0.73MB
- Step 1 in supporting custom data converters

### DIFF
--- a/packages/common/src/activity-options.ts
+++ b/packages/common/src/activity-options.ts
@@ -1,7 +1,16 @@
-import { coresdk } from '@temporalio/proto/lib/coresdk';
+import type { coresdk } from '@temporalio/proto/lib/coresdk';
 import { RetryOptions } from './interfaces';
+import { checkExtends } from './type-helpers';
 
-export const ActivityCancellationType = coresdk.workflow_commands.ActivityCancellationType;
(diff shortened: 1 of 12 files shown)


packages/common/src/activity-options.ts
```diff
@@ -1,7 +1,16 @@
...
+import { checkExtends } from './type-helpers';
 
-export const ActivityCancellationType = coresdk.workflow_commands.ActivityCancellationType;
+// Avoid importing the proto implementation to reduce workflow bundle size
+// Copied from coresdk.workflow_commands.ActivityCancellationType
+export enum ActivityCancellationType {
+  TRY_CANCEL = 0,
+  WAIT_CANCELLATION_COMPLETED = 1,
+  ABANDON = 2,
+}
```
Sushisource, line 11: Can we not automatically do this for these enums?

bergundy: Sadly no, at least we can verify that if the original adds more cases we cover those with the check below.

Sushisource: K. Seems fine

Sushisource (approved):

PR #427, opened 2021-12-09
fix: Use bundled Workflow interceptors

### LINKED ISSUE #390: WorkerOptions.interceptors.workflowModules might not be available if workflows are pre-bundled
### Describe the bug

At the moment workflow interceptors are passed into the bundler in order for webpack to include them (using webpack magic import comments).
They're also required to be provided in WorkerOptions to be actually loaded the workflow runtime.
This double requirement is not optimal since the worker might not have the workflows available in the file system on startup (in case it is using a workflows bundle).

To clarify, the issue is that `require.resolve` might not work in the pre-bundled worker process.

### Expected behavior

One way to get around this is to store the list of workflow interceptor modules at bundling time and load those too.
Another is to automatically load anything in the workflow bundle, this option feels cleaner since the paths only need to be specified once.

bergundy:
Closes #390

### DIFF
--- a/packages/test/src/test-workflows.ts
+++ b/packages/test/src/test-workflows.ts
@@ -84,7 +84,6 @@ async function createWorkflow(
       taskQueue: 'test',
       isReplaying: false,
     },
-    interceptorModules: [],
     randomnessSeed: Long.fromInt(1337).toBytes(),
     now: startTime,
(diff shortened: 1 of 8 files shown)


bergundy:
Force merged, skipping flaky windows build

PR #428, opened 2021-12-10
feat: Implement async Activity completion

### LINKED ISSUE #243: [Feature Request] Asynchronous Activity Completion

We have a Node activity that makes a POST request to an existing (Node) AWS Lambda function, which in turn launches a process that can last over a minute.
The Lambda is behind an API gateway so has a maximum timeout of 30 seconds. So we can't wait for the lambda to return the final response.
Right now we have the lambda immediately returning a 200 upon start of the process, and then later signaling success/failure back to the workflow when done.
However that makes us add extra signal(s) (success/fail) just to complete a simple activity. It's just a few lines of code so is not onerous, but it does add some noise to an otherwise smooth flow.
In the Go SDK we see an incrementally more attractive approach.

In the Go SDK you can "asynchronously complete an activity":
https://docs.temporal.io/docs/go/activities#asynchronous-activity-completion
We would like the same in node.
That way we can preserve the separation of concerns between workflow and activities: workflows execute activities while activities take care of any internal execution complexity, saving signals for true inter-"actor" collaboration and external events.

bergundy:
Closes #243

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -30,7 +30,22 @@ import { AsyncLocalStorage } from 'async_hooks';
 import { AbortSignal } from 'abort-controller';
 import { msToNumber } from '@temporalio/common';
 
-export { CancelledFailure } from '@temporalio/common';
+export { ActivityFunction, ActivityInterface, CancelledFailure } from '@temporalio/common';
+
+/**
+ * Throw this error from an Activity in order to make the Worker
+ * forget about this Activity.
+ *
+ * The Activity can be completed asynchronously using
+ * {@link AsyncCompletionClient}
+ */
+export class CompleteAsyncError extends Error {
+  public readonly name: string = 'CompleteAsyncError';
+
+  constructor() {
+    super();
+  }
+}
 
 /** @ignore */
 export const asyncLocalStorage = new AsyncLocalStorage<Context>();
--- a/packages/client/src/async-completion-client.ts
+++ b/packages/client/src/async-completion-client.ts
@@ -1,0 +1,250 @@
+import os from 'os';
+import { ServerErrorResponse } from '@grpc/grpc-js';
+import { Status } from '@grpc/grpc-js/build/src/constants';
+import {
+  DataConverter,
+  defaultDataConverter,
+  ensureTemporalFailure,
+  errorToFailure,
+  filterNullAndUndefined,
+} from '@temporalio/common';
+import { Connection, WorkflowService } from './connection';
+
+/**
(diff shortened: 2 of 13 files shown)


packages/activity/src/index.ts
```diff
@@ -30,7 +30,22 @@ import { AsyncLocalStorage } from 'async_hooks';
...
+ * The Activity can be completed asynchronously using
+ * {@link AsyncCompletionClient}
+ */
+export class CompleteAsyncError extends Error {
+  public readonly name: string = 'CompleteAsyncError';
+
+  constructor() {
+    super();
+  }
+}
```
Sushisource, line 48: Why not make this a return value rather than an exception? Nothing went "wrong" so it seems a return value would be better representing intent here.

I can see maybe not wanting to b/c it clutters the type signature of activities, but Typescript makes that so easy I'm not sure it's a big deal

packages/client/src/async-completion-client.ts
```diff
@@ -0,0 +1,250 @@
...
+  heartbeat(fullActivityId: FullActivityId, details?: unknown): Promise<void>;
+
+  async heartbeat(taskTokenOrFullActivityId: Uint8Array | FullActivityId, details?: unknown): Promise<void> {
+    try {
+      if (taskTokenOrFullActivityId instanceof Uint8Array) {
+        const { cancelRequested } = await this.service.recordActivityTaskHeartbeat({
+          identity: this.options.identity,
+          namespace: this.options.namespace,
+          taskToken: taskTokenOrFullActivityId,
+          details: { payloads: await this.options.dataConverter.toPayloads(details) },
```
Sushisource, line 227: This struct could be lifted up and then just modified in the one branch

packages/activity/src/index.ts
bergundy: I actually started by making this a symbol and not an exception but typescript complains if you return it and I wouldn't want the activity to have to define a special return type because we use the activity return type in the workflow.

packages/client/src/async-completion-client.ts
bergundy: sure, but I feel like this is ok, it still fits in a single line.

Sushisource: I meant the whole thing, not just details

packages/activity/src/index.ts
Sushisource: Could introduce an intermediate type which is used internally and then narrowed to the user's type by the time it is returning to the workflow? I can see that being annoying though

Sushisource (approved):

bergundy:
Thanks @Sushisource

PR #430, opened 2021-12-15
Add binary/protobuf serialization and WorkerOptions.dataConverterPath

lorensr:
## What was changed

- `WorkerOptions.dataConverter` ➡️ `WorkerOptions.dataConverterPath`: path to a module with `dataConverter` named export. This is needed in order to get the data converter into the node worker thread via postMessage (that part coming in the next PR).
- Added `ProtobufPayloadConverter` that converts Payloads of type `binary/protobuf` in the Client and for Activities (but not yet for Workflows)

Design notes: 

- If you want an argument protobuf-serialized, you must pass an instance of class that has a static `encode` method (generally this class will have been generated by [protobufjs](https://www.npmjs.com/package/protobufjs)'s `pbjs`)
- If you want your workflows/activities to be able to handle protobuf-serialized arguments, you must pass the Worker your `pbjs`-generated js file (or equivalent) with class names that match the `argument.constructor.name`s of the arguments that got serialized. (Might be nice to support `.proto` file as an alternative? At the cost of longer Worker startup time.)
- Your workflow/activity functions will receive instances of the pbjs classes and can return instances of pbjs classes.
- Message types/class names passed to a Worker must be unique
- Other SDKs, when protobuf-serializing, must include the name of the class in `payload.metadata.messageType` (Or `message-type`? Is there a protobuf key naming convention?) for the TS SDK to read (can also use for UI).

Remaining work:

- Get protobufs and custom data converters working for Workflows
- ~~`json/protobuf` encoding type~~
- Add `metadata.messageType` to other SDKs' encoding

## Why?

More efficient data storage and compatibility with the other SDKs.

## Checklist

2. How was this tested:

- Added `packages/test/src/test-custom-data-converter.ts`
- Updated `packages/test/src/test-data-converter.ts`

3. Any docs updates needed?

https://docs.temporal.io/docs/typescript/workers#additional-worker-options


packages/test/protos/protobufs.d.ts
```diff
@@ -0,0 +1,186 @@
+import * as $protobuf from "protobufjs";
```
lorensr, line 1: I don't think the two generated files will change much, but we could remove from git by changing the `build.proto` script to `pretest`. It adds 1.7s to testing on my machine.

packages/common/src/converter/data-converter.ts
```diff
@@ -182,8 +188,21 @@ export function mapToPayloadsSync<K extends string>(
   ) as Record<K, Payload>;
 }
 
-export const defaultDataConverter = new CompositeDataConverter(
-  new UndefinedPayloadConverter(),
-  new BinaryPayloadConverter(),
-  new JsonPayloadConverter()
-);
+export interface DefaultDataConverterOptions {
+  protobufClasses?: Record<string, unknown>;
```
cretz, line 192: This is a bit unfortunate that the data converter doesn't accept the type/object to (de)serialize to at (de)serialize time. This is something other SDKs provide which means the implicit default data converter works with protos. Does this mean in order to use protos in TS, I have to manually configure a data converter?

One wonders if there is value in sort of proto converter proxy. For example, if the `encoding` is `json/protobuf` or `binary/protobuf`, can `fromPayload` return a `class ProtoMessage { #raw: Uint8Array; toProto<T>(obj: T); }` which then doesn't require such preconfiguration?

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+    }
+
+    // Safe to non-null assert due to above check
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const messageClass = this.validateMessageClass(this.protobufClasses![value.constructor.name]);
+
+    return {
+      metadata: {
+        [METADATA_ENCODING_KEY]: encodingKeys.METADATA_ENCODING_PROTOBUF,
+        [METADATA_MESSAGE_TYPE_KEY]: u8(value.constructor.name),
```
cretz, line 181: Hrmm, not sure this should not be required as it should be the recipient who should choose what to deserialize into, not the serializer. This approach is also not cross-SDK compatible.

While this might seem reasonable, if someone wanted to do this, they could use `google.protobuf.Any` as their proto type which _does_ embed the fully-qualified message class in it.

My suggestion elsewhere about a generic `ProtobufMessage` result in `fromPayload` makes more sense now.

packages/common/src/converter/data-converter.ts
```diff
@@ -182,8 +188,21 @@ export function mapToPayloadsSync<K extends string>(
...
+}
+
+export class DefaultDataConverter extends CompositeDataConverter {
+  constructor({ protobufClasses }: DefaultDataConverterOptions = {}) {
+    // Match the order used in other SDKs
+    // Java: https://github.com/temporalio/sdk-java/blob/0da3149c8eb16454c30fb7bda32d28f2539f3f3c/temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java#L44
+    super(
+      new UndefinedPayloadConverter(),
+      new BinaryPayloadConverter(),
+      new ProtobufPayloadConverter(protobufClasses),
```
cretz, line 202: Note, there are actually two types of payload encoding: JSON and binary. Unfortunately it appears the Java default does not include the binary one (not sure why), and unfortunately this one does not include the JSON one.

Maybe the Go SDK at https://github.com/temporalio/sdk-go/blob/5e5645f0c550dcf717c095ae32c76a7087d2e985/converter/default_data_converter.go#L28 is better to look at here because it is more robust.

Cross SDK compatibility with these is important. I will followup on the Java SDK to see why the `binary/protobuf` is not supported by default. In the meantime, I think we need both here and with JSON being listed first.

packages/test/protos/protobufs.d.ts
cretz: IMO you should definitely leave in git and any changes should require a manual regen. This is a testing asset that should not be done on each build IMO.

It is unfortunate that this auto-generated code does not have a top-of-file comment saying it is auto-generated :-(

packages/worker/src/worker-options.ts
```diff
@@ -75,9 +75,9 @@ export interface WorkerOptions {
   shutdownSignals?: NodeJS.Signals[];
 
   /**
-   * TODO: document, figure out how to propagate this to the workflow isolate
+   * Path of a module with a `dataConverter` named export. `dataConverter` should be an instance of a class that implements {@link DataConverter}.
    */
-  dataConverter?: DataConverter;
+  dataConverterPath?: string;
```
cretz, line 80: Looks like a breaking change here, but if needed I understand.

Do we _have_ to use the data converter _inside_ the workflow VM isolate? One wonders if a data converter should have a proxy to the main node process and use the data converter there. And if we don't have to use it inside the isolate, can we leave the interface as it was?

packages/common/src/converter/data-converter.ts
lorensr: > This is a bit unfortunate that the data converter doesn't accept the type/object to (de)serialize to at (de)serialize time

When encoding, the data converter gets the object, and we could just use that (something like `(arg.constructor as unknown as Encodable).encode(args).finish()`), but we need the classes and `metadata.messageType` in order to decode.

> Does this mean in order to use protos in TS, I have to manually configure a data converter?

Yes

> One wonders if there is value in sort of proto converter proxy. For example, if the encoding is json/protobuf or binary/protobuf, can fromPayload return a class ProtoMessage { #raw: Uint8Array; toProto<T>(obj: T); } which then doesn't require such reconfiguration?

The pro would be no config needed. The con is users would have to call `toProto` a lot. As a user, I'd personally prefer to do the config once than to write the `toProto` code. Current:

```ts
import type { ProtoActivityInput } from './protos'

export async function protoActivity(args: ProtoActivityInput): Promise<ProtoActivityResult> {
  return new ProtoActivityResult({ sentence: `${args.name} is ${args.age} years old.` });
}
```

With `toProto`:

```ts
import { ProtoActivityInput } from './protos'

export async function protoActivity(message: ProtoMessage): Promise<ProtoActivityResult> {
  const args = message.toProto<ProtoActivityInput>(ProtoActivityInput)
  return new ProtoActivityResult({ sentence: `${args.name} is ${args.age} years old.` });
}
```

It's only one line here, but it's a lot more code when you consider multiple arguments, everywhere arguments or return values are deserialized.

packages/common/src/converter/payload-converter.ts
lorensr: > it should be the recipient who should choose what to deserialize into, not the serializer

Currently it's the sender who decides. Are there cases in which a user would send an instance of a certain protobuf type and want to receive a different type?

> This approach is also not cross-SDK compatible.

Correct—it would require modification of other SDKs to add the `METADATA_MESSAGE_TYPE_KEY` when encoding.

packages/common/src/converter/data-converter.ts
lorensr: Ah, thanks! Will update.

packages/worker/src/worker-options.ts
lorensr: It's breaking a feature that's not fully implemented yet (currently `defaultDataConverter` is used for the Workflow, disregarding the `dataConverter` option).

We can't send functions, including those in `ProtoActivityInput` or `ProtoMessage` instances, from the main thread to the worker thread—just [basic builtin JS types](https://nodejs.org/api/worker_threads.html#portpostmessagevalue-transferlist). If we used `ProtoMessage`, we could deserialize in main thread to an intermediate format, and then convert to `ProtoMessage` inside the worker thread. But our intermediate format wouldn't support an arbitrary custom data converter. So in order to allow custom data converters, the converting code needs to run in the worker thread.

packages/common/src/converter/payload-converter.ts
cretz: > Currently it's the sender who decides.

It's both. Proto is just a binary blob. You can have different versions, different packages, etc. Same thing with JSON. The caller sends a JSON and the recipient can do what they want with it (deserialize into an existing object, sanitize in some way, etc).

> Are there cases in which a user would send an instance of a certain protobuf type and want to receive a different type?

Probably not

> Correct—it would require modification of other SDKs to add the `METADATA_MESSAGE_TYPE_KEY` when encoding.

Unfortunately the decision was made in this repository to not allow the user to specify the type in the data converter at conversion time. That decision should not require us to go add another metadata type to the other SDKs IMO.

packages/common/src/converter/data-converter.ts
```diff
@@ -66,6 +67,11 @@ export interface DataConverter {
   fromPayloadsSync<T>(index: number, content?: Payload[] | null): T;
 }
 
+export const isValidDataConverter = (dataConverter: unknown): boolean =>
+  typeof dataConverter === 'object' &&
+  dataConverter !== null &&
+  ['toPayload', 'toPayloads', 'fromPayload', 'fromPayloads'].every((method) => method in dataConverter);
+
```
Sushisource, line 74: What's this for? Wouldn't any valid converter just implement the `DataConverter` interface?

packages/common/src/converter/data-converter.ts
cretz: I am not sure we should require other SDKs to support message type or users in this SDK to change the default converter because reification/reflection is not available. I wonder if instead you should have to be more explicit at runtime about the contract of the activity in cases where the types are ambiguous.

I would have to think a bit on how one can specify the proto classes for activity/workflow arguments at runtime leveraging the existing API without infecting payloads for everyone.

I agree that the `toProto` does look ugly that way. I was thinking about explicit data converter use, but yes for implicit conversion like for activity params, a better-defined contract is needed.

Having said this, this will continue to come up for other dynamically typed languages. We have to decide for dynamically typed languages whether the sender defines the proto type or the recipient.

packages/common/src/converter/data-converter.ts
cretz: Interfaces are not JS things and are not present at runtime.

packages/common/src/converter/data-converter.ts
```diff
@@ -182,8 +188,21 @@ export function mapToPayloadsSync<K extends string>(
...
-  new UndefinedPayloadConverter(),
-  new BinaryPayloadConverter(),
-  new JsonPayloadConverter()
-);
+export interface DefaultDataConverterOptions {
+  protobufClasses?: Record<string, unknown>;
+}
+
+export class DefaultDataConverter extends CompositeDataConverter {
+  constructor({ protobufClasses }: DefaultDataConverterOptions = {}) {
```
cretz, line 196: I will have to dig into protobufjs, but all other proto impls I am aware of maintain a global type registry that is required for resolving `google.protobuf.Any`.

packages/common/src/converter/data-converter.ts
Sushisource: Right, maybe I wasn't clear - I would expect this to narrow the type the way it's being used. As in https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

packages/worker/src/worker-options.ts
cretz: That makes sense. It's unfortunate that due to how protobufjs works, it creates classes with methods instead of simple JSON representations. If we had the requirement of "all items in and out of data converters must be trivially JSON serializable, all functions and prototype pieces do not cross the boundary" then we could just serialize on one side. But then you couldn't know whether it was originally a proto message.

I guess technically the encoding could be a key on the object (e.g. `T & { _encoding: string }`) but that has many issues as well. I already pitied TS devs who were forced to bundle workflows separately (can't have code side-by-side like our other SDKs), and now this has leaked into data converters too. But the limitations definitely make sense :+1:

packages/common/src/converter/data-converter.ts
cretz: I am starting to formulate an idea...

What if _all_ TS protos had to be (google.protobuf.Any](https://developers.google.com/protocol-buffers/docs/proto3#any)? [Here](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto)'s the raw proto which is basically what you're doing here. They have [well defined JSON semantics](https://developers.google.com/protocol-buffers/docs/proto3#json) that makes them still look good when viewed (basically just `{ "@type": "type", ...restOfProto }`). They are also very cross-language compatible.

This can all be transparent and only affects cross-language compat. So basically the data converter converts every proto to an `Any`, and only accepts `Any` input. This would force other languages to either accept `Any` if executing something from TS, or send out `Any` if communicating with TS, which gets the built-in type name. Though that's still a bit gross compared to potentially changing all of them to use a new metadata header.

Well, now I can't find a global registry for `google.protobuf.Any` in protobufjs. Now I am starting to lean back to your original suggestion of putting `protoMessageType` as a metadata value in the other SDKs. Ignore my rant, heh (but I'll leave it here).

packages/common/src/converter/data-converter.ts
cretz: Well, I can't find this, nevermind.

packages/common/src/converter/data-converter.ts
lorensr: It's only used for validating `dataConverterPath`: https://github.com/temporalio/sdk-typescript/pull/430/files#diff-04516c8c3fdd8a63a8a2620f597ad8ff29fd98b3ec501026337b9e93af73668eR252

Perhaps it would be helpful to change the name to `isImportedDataConverterValid`? Or `doesImportedDataConverterLookRight` 😄

packages/common/src/converter/data-converter.ts
bergundy: +1 to what @cretz said

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+  public toDataSync(value: unknown): Payload | undefined {
+    const isProtobufMessageInstance =
+      this.protobufClasses && value && typeof value === 'object' && this.protobufClasses[value.constructor.name];
+    if (!isProtobufMessageInstance) {
+      return undefined;
+    }
+
+    // Safe to non-null assert due to above check
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const messageClass = this.validateMessageClass(this.protobufClasses![value.constructor.name]);
```
bergundy, line 176: This should be done in the constructor

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
   }
 }
+
+/**
+ * Converts between protobufjs Message instances and serialized Protobuf Payload
+ */
+export class ProtobufPayloadConverter extends AsyncFacadePayloadConverter {
+  public encodingType = encodingTypes.METADATA_ENCODING_PROTOBUF;
+
+  constructor(protected protobufClasses?: Record<string, unknown>) {
```
bergundy, line 159: the attribute itself shouldn't be optional
```suggestion
  constructor(protected protobufClasses: Record<string, unknown> = {}) {
```

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+    super();
+
+    if (protobufClasses && typeof protobufClasses !== 'object') {
+      throw new DataConverterError('protobufClasses must be an object');
```
bergundy, line 163: Maybe `TypeError` or `ValueError` would be better here because it's not an error in the data conversion itself.

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+    }
+  }
+
+  public toDataSync(value: unknown): Payload | undefined {
```
bergundy, line 167: Are you changing the API to async in the next PR?

packages/common/src/converter/payload-converter.ts
bergundy: We should change the other SDKs to support the dynamic use case

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+  }
+
+  public fromDataSync<T>(content: Payload): T {
+    if (content.data === undefined || content.data === null) {
+      throw new ValueError('Got payload with no data');
+    }
+    if (!content.metadata || !(METADATA_MESSAGE_TYPE_KEY in content.metadata)) {
+      throw new ValueError(`Got protobuf payload without metadata.${METADATA_MESSAGE_TYPE_KEY}`);
+    }
+    if (!this.protobufClasses) {
```
bergundy, line 194: This shouldn't be possible if we add a default in the constructor.

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+
+    const messageClassName = str(content.metadata[METADATA_MESSAGE_TYPE_KEY]);
+    const messageClass = this.protobufClasses[messageClassName];
+    if (!messageClass) {
+      throw new DataConverterError(
+        `Got a \`${messageClassName}\` protobuf message but cannot find corresponding message class in protobufClasses`
+      );
+    }
+
+    return this.validateMessageClass(messageClass).decode<T>(content.data);
```
bergundy, line 208: This class should be validated during construction

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+  }
+
+  protected validateMessageClass(messageClass: unknown): ProtobufSerializable {
```
bergundy, line 211: Could be:
```suggestion
  protected validateMessageClass(messageClass: unknown): messageClass is ProtobufSerializable {
```

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+    if (typeof messageClass !== 'function') {
+      throw new DataConverterError(`protobufClasses values must be classes`);
```
bergundy, line 213: TypeError

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,75 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+    super();
+
+    if (protobufClasses && typeof protobufClasses !== 'object') {
+      throw new DataConverterError('protobufClasses must be an object');
+    }
+  }
+
+  public toDataSync(value: unknown): Payload | undefined {
+    const isProtobufMessageInstance =
+      this.protobufClasses && value && typeof value === 'object' && this.protobufClasses[value.constructor.name];
```
bergundy, line 169: Let's do an instanceof check here

bergundy (commented):
I finished most of the review.
Overall LGTM.
I'll try to get this all in reviewed by tomorrow.

packages/worker/src/workflow/bundler.ts
```diff
@@ -10,7 +10,6 @@ import { resolveNodeModulesPaths } from '../worker-options';
 
 /**
  * Builds a V8 Isolate by bundling provided Workflows using webpack.
- * Activities are replaced with stubs.
```
bergundy, line 13: that's been outdated for a while now 😅

packages/worker/src/worker.ts
```diff
@@ -219,24 +222,56 @@ export class Worker {
...
-      return new this(nativeWorker, workflowCreator, compiledOptions);
+      return new this(nativeWorker, workflowCreator, compiledOptions, dataConverter);
     } catch (err) {
       // Deregister our worker in case Worker creation (Webpack) failed
       await nativeWorker.completeShutdown();
       throw err;
     }
   }
 
+  public static async getDataConverter(options: WorkerOptions): Promise<DataConverter> {
```
bergundy, line 233: You're going to need this in the worker threads soon

packages/test/src/workflows/protobufs.ts
```diff
@@ -0,0 +1,5 @@
+import type { ProtoActivityInput } from '../../protos/protobufs';
+
+export async function protobufWorkflow(args: ProtoActivityInput): Promise<ProtoActivityInput> {
+  return args;
```
bergundy, line 4: Why not run the activity?

packages/common/src/converter/payload-converter.ts
lorensr: Yep

packages/common/src/converter/payload-converter.ts
lorensr: Do you mean going through `protobufClasses` in the constructor and collecting all the values that pass validation?

packages/worker/src/worker.ts
lorensr: Here's the worker thread part. I haven't duplicated the error handling, since `getDataConverter` runs first

image

image

packages/test/src/workflows/protobufs.ts
lorensr: Runs activity in next PR

packages/common/src/converter/payload-converter.ts
lorensr: Like this: image

packages/common/src/converter/data-converter.ts
cretz: > We went in this direction because I feel like it gives the best experience in JS / TS 

It is unfortunate that this experience though is going to affect other SDKs. It is unfortunate but I guess it's the best. I haven't spent too much time deriving an alternative.

packages/common/src/converter/payload-converter.ts
cretz: FWIW, they support this use case amongst each other. TS is the abnormal one that doesn't take the type during deserialization/decode.

lorensr:
@bergundy I went through all the comments (Changes in https://github.com/temporalio/sdk-typescript/pull/430/commits/fb404b653d7dd62db99000a051b4e0da3a9e8152), and ready for review.

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +149,130 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
...
+      typeof (messageClass as unknown as ProtobufSerializable).decode === 'function'
+    );
+  }
+
+  protected isProtobufMessageInstance(value: unknown): value is ProtobufSerializable {
+    if (!value || typeof value !== 'object') {
+      return false;
+    }
+
+    const messageClass = this.messageClasses[value.constructor.name];
```
bergundy, line 194: If you have the reverse mapping I mentioned you should be able to lookup by class and get the registered name

packages/test/src/workflows/protobufs.ts
bergundy: good enough for me

packages/test/src/test-custom-data-converter.ts
```diff
@@ -0,0 +1,143 @@
...
+      dataConverterPath: require.resolve('./data-converters/data-converter-bad-export'),
+    }),
+    {
+      message:
+        /The `dataConverter` named export at dataConverterPath .* should be an instance of a class that implements the DataConverter interface./,
+    }
+  );
+});
+
+test('Worker with proto data converter runs an activity and reports completion', async (t) => {
```
bergundy, line 121: Might as well have made this a full integration test where the workflow calls the activity

packages/test/src/test-custom-data-converter.ts
```diff
@@ -0,0 +1,143 @@
...
+    client.execute(protobufWorkflow, {
+      args: [messageInstance],
+      workflowId: uuid4(),
+      taskQueue,
+    });
+    await receivedExpectedError;
+    worker.shutdown();
+    t.pass();
+
+    // const runAndShutdown = async () => {
```
bergundy, line 75: Leftover comment?

packages/test/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "tsc --build",
     "build.watch": "tsc --build --watch",
+    "build.proto": "pbjs -t static-module -w commonjs -o protos/protobufs.js protos/test-messages.proto && pbts protos/protobufs.js -o protos/protobufs.d.ts",
```
bergundy, line 9: I'd have this as part of `build` that's the step where we compile the core and api protos.

bergundy: And let's not commit the generated files please.

bergundy: Also, will this work on windows?

packages/common/src/converter/types.ts
```diff
@@ -30,3 +31,11 @@ export const encodingKeys = {
   METADATA_ENCODING_PROTOBUF_JSON: u8(encodingTypes.METADATA_ENCODING_PROTOBUF_JSON),
   METADATA_ENCODING_PROTOBUF: u8(encodingTypes.METADATA_ENCODING_PROTOBUF),
 } as const;
+
+export const METADATA_MESSAGE_TYPE_KEY = 'messageType';
+
+export interface ProtobufSerializable {
```
bergundy, line 37: We could rename this to `ProtobufCodec` because technically this refers to the class and not the serializable instance.

packages/common/src/converter/types.ts
```diff
@@ -1,3 +1,4 @@
+import { Writer } from 'protobufjs';
```
bergundy, line 1: Let's import this as a type only

bergundy (commented):
Almost there, good job

packages/common/src/converter/types.ts
lorensr: Ack, good catch

packages/test/src/test-custom-data-converter.ts
lorensr: Gets uncommented in next PR when it works

packages/test/src/test-custom-data-converter.ts
lorensr: I was trying to skip the workflow in this test since it can't receive protobufs yet. But I guess the workflow could create a protobuf just to use when calling the activity. If that's helpful I could add/replace?

packages/test/src/test-data-converter.ts
```diff
@@ -62,11 +70,120 @@ test('JsonPayloadConverter converts to object', async (t) => {
...
+  const instance = root.foo.bar.ProtoActivityInput.create({ name: 'Proto', age: 1 });
+  const converter = new ProtobufJsonPayloadConverter(root);
+  const testInstance = await converter.fromData((await converter.toData(instance))!);
+  // tests that both are instances of the same class with the same properties
+  t.deepEqual(testInstance, instance);
+});
+
+test('ProtobufJSONPayloadConverter converts binary', async (t) => {
+  // binary should be base64-encoded:
+  // https://developers.google.com/protocol-buffers/docs/proto3#json
```
lorensr, line 143: The lib we're using seems to have good tests for the types listed in the spec: https://github.com/googleapis/proto3-json-serializer-nodejs/tree/main/typescript/test/unit

Let me know if you'd like me to:
- write tests for more types
- do the compact testing against the other SDKs (I imagine they also follow the spec and thus work, but haven't tried yet)

lorensr:
- Updated to use `proto3-json-serializer` for JSON encoding (reasoning included in `docs/protobuf-libraries.md`).
- After I write docs, will link to them from some of the error messages.

docs/protobuf-libraries.md
```diff
@@ -0,0 +1,109 @@
...
+- Use `protobufjs` with `proto3-json-serializer`
+- Have users use runtime-loaded messages (not generated classes) and `Class.create` (not `new Class()`)
+- Patch `json-module` output (which adds `nested` attributes to lowercase namespaces [which causes TS error](https://github.com/protobufjs/protobuf.js/issues/1014))
+
+```ts
+// json-module.js generated with:
+// pbjs -t json-module -w commonjs -o json-module.js *.proto
+
+// protos/root.js
+const { patchRoot } = require('@temporalio/common');
```
bergundy, line 61: This should probably have a more specific name, e.g. `patchProtobufRoot`

docs/protobuf-libraries.md
```diff
@@ -0,0 +1,109 @@
...
+  args: [foo.bar.ProtoInput.create({ name: 'Proto', age: 1 })],
+  taskQueue: 'tutorial',
+  workflowId: 'my-business-id',
+});
+
+// src/workflows.ts
+import { foo } from '../protos/root';
+
+export async function protoWorkflow(input: foo.bar.ProtoInput): Promise<foo.bar.ProtoResult> {
+  return foo.bar.ProtoResult({ sentence: `Name is ${input.name}` });
```
bergundy, line 81:
```suggestion
  return foo.bar.ProtoResult.create({ sentence: `Name is ${input.name}` });
```

docs/protobuf-libraries.md
```diff
@@ -0,0 +1,109 @@
...
+}
+```
+
+On root in `root.foo.bar.ProtoActivityInput`, TS errors: `Cannot find namespace 'root'.`
+
+## Future work
+
+If we can get changes merged into `protobufjs` (or want to fork), we can do one or both of the below:
+
+1. Change the `json-module` output to not have `nested` attributes so we don't have to patch
```
bergundy, line 104: Make `pbts` generate private (or protected) constructors so users don't accidentally try to use them and have them rely on the static `create` method.

packages/common/src/converter/data-converter.ts
```diff
@@ -182,8 +191,22 @@ export function mapToPayloadsSync<K extends string>(
...
+
+export class DefaultDataConverter extends CompositeDataConverter {
+  constructor({ root }: DefaultDataConverterOptions = {}) {
+    // Match the order used in other SDKs
+    // Go SDK: https://github.com/temporalio/sdk-go/blob/5e5645f0c550dcf717c095ae32c76a7087d2e985/converter/default_data_converter.go#L28
+    super(
+      new UndefinedPayloadConverter(),
+      new BinaryPayloadConverter(),
+      new ProtobufJsonPayloadConverter(root),
+      new ProtobufBinaryPayloadConverter(root),
```
bergundy, line 206: Once Java add the binary converter to the default converter we should change the order to binary -> json.
Let's add a todo or issue.

packages/common/src/converter/helpers.ts
```diff
@@ -0,0 +1,24 @@
+export function patchRoot<T>(root: T, seen = new Set()): T {
+  for (const key in root) {
```
bergundy, line 2: Why not just go over the `nested` attributes?

packages/common/src/converter/helpers.ts
```diff
@@ -0,0 +1,24 @@
+export function patchRoot<T>(root: T, seen = new Set()): T {
```
bergundy, line 1: It'd be better to return a new object IMO

packages/common/src/converter/payload-converter.ts
```diff
@@ -140,3 +153,157 @@ export class BinaryPayloadConverter extends AsyncFacadePayloadConverter {
     return content.data as any;
   }
 }
+
+abstract class ProtobufPayloadConverter extends AsyncFacadePayloadConverter {
+  protected readonly typesByName: Record<string, ProtobufCodec> = {};
```
bergundy, line 158: Do we even need this anymore? Why not just use root directly?

packages/test/src/data-converters/data-converter.ts
```diff
@@ -0,0 +1,6 @@
+import { DefaultDataConverter } from '@temporalio/common';
+import root, { foo } from '../../protos/root';
+
+export const messageInstance = foo.bar.ProtoActivityInput.create({ name: 'Proto', age: 1 });
```
bergundy, line 4: Should this be here?

packages/test/src/test-custom-data-converter.ts
```diff
@@ -0,0 +1,139 @@
...
+  actual: coresdk.activity_result.IActivityResult | null | undefined,
+  expected: coresdk.activity_result.IActivityResult
+) {
+  if (actual?.failed?.failure) {
+    const { stackTrace, ...rest } = actual.failed.failure;
+    actual = { failed: { failure: { stackTrace: cleanStackTrace(stackTrace), ...rest } } };
+  }
+  t.deepEqual(
+    new coresdk.activity_result.ActivityResult(actual ?? undefined).toJSON(),
+    new coresdk.activity_result.ActivityResult(expected).toJSON()
```
bergundy, line 30: I thought you're supposed to use `create`

packages/test/src/test-custom-data-converter.ts
```diff
@@ -0,0 +1,139 @@
...
+      ...defaultOptions,
+      taskQueue,
+      dataConverterPath: require.resolve('./data-converters/data-converter'),
+    });
+    const connection = new Connection();
+    const client = new WorkflowClient(connection.service, { dataConverter });
+
+    // For now, just check that the protobuf message gets to the workflow
+    worker.run();
+    client.execute(protobufWorkflow, {
```
bergundy, line 64: Let's not leave loose promises around

packages/test/src/test-data-converter.ts
bergundy: No need, the initial comment was hinting that we should verify that we support the standard protobuf JSON serialization.

packages/test/src/test-data-converter.ts
```diff
@@ -62,11 +70,120 @@ test('JsonPayloadConverter converts to object', async (t) => {
...
+  t.deepEqual(encoded, {
+    metadata: {
+      [METADATA_ENCODING_KEY]: encodingKeys.METADATA_ENCODING_PROTOBUF_JSON,
+      [METADATA_MESSAGE_TYPE_KEY]: u8('BinaryMessage'),
+    },
+    data: u8(JSON.stringify({ data: Buffer.from('abc').toString('base64') })),
+  });
+
+  const testInstance = await converter.fromData<root.BinaryMessage>(encoded!);
+  // testInstance contains a Buffer, while u8() returns a Uint8Array, so deepEqual fails:
```
bergundy, line 156: You can just use `Buffer.from`

packages/worker/src/workflow/bundler.ts
```diff
@@ -107,6 +106,7 @@ export class WorkflowCodeBundler {
       resolve: {
         modules: this.nodeModulesPaths,
         extensions: ['.ts', '.js'],
+        fallback: { assert: false }, // used by `proto3-json-serializer`
```
bergundy, line 109: This should be removed in the next PR

bergundy:
So close

packages/common/src/converter/helpers.ts
lorensr: I didn't realize capitalized namespaces were both at root and in nested. Updated.

packages/test/src/data-converters/data-converter.ts
lorensr: It's used in some of the tests. Do you think it's better to create it in the tests?

packages/common/src/converter/payload-converter.ts
lorensr: Ack, thanks 😄 Was in the `protobufClasses` mindset and just looking at properties in the console and didn't think about it having methods like `lookupType()`.

packages/test/src/test-custom-data-converter.ts
lorensr: For json-modules, yes. These are coresdk generated classes (`compareCompletion()` is copied from other test files).

packages/common/package.json
```diff
@@ -23,5 +23,8 @@
   "homepage": "https://github.com/temporalio/sdk-typescript/tree/main/packages/common",
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "protobufjs": "^6.11.2"
```
bergundy, line 28: Is this just for types?

packages/common/src/converter/payload-converter.ts
```diff
@@ -276,34 +260,34 @@ export class ProtobufJsonPayloadConverter extends ProtobufPayloadConverter {
...
     type.constructor.name === 'Type' &&
     'parent' in type &&
-    typeof (type as unknown as ProtobufCodec).name === 'string' &&
-    typeof (type as unknown as ProtobufCodec).create === 'function' &&
-    typeof (type as unknown as ProtobufCodec).encode === 'function' &&
-    typeof (type as unknown as ProtobufCodec).decode === 'function'
+    typeof (type as unknown as Type).name === 'string' &&
+    typeof (type as unknown as Type).create === 'function' &&
+    typeof (type as unknown as Type).encode === 'function' &&
+    typeof (type as unknown as Type).decode === 'function'
```
bergundy, line 275: Can we cast only once here?

bergundy (approved):
Nothing blocking, good to merge whenever you're ready

packages/common/package.json
lorensr: Yes

lorensr: Yes

packages/common/src/converter/payload-converter.ts
lorensr: Removed casting:

image

lorensr:
@bergundy I can't merge (don't have write access)

`16, windows-latest` is hanging on `npm ci`, don't see indication as to why:

image

PR #432, opened 2021-12-22
fix: Fix workflow retryable Application failure fails execution

bergundy:
Makes `ApplicationFailure.retryable` fail the workflow execution and not the task as intended, this was wrongly implemented in #429.

### DIFF
--- a/packages/workflow/src/internals.ts
+++ b/packages/workflow/src/internals.ts
@@ -12,6 +12,7 @@ import {
   Workflow,
   WorkflowQueryType,
   ApplicationFailure,
+  TemporalFailure,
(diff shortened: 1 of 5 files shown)


sw-yx (approved):

packages/workflow/src/internals.ts
```diff
@@ -12,6 +12,7 @@ import {
   Workflow,
   WorkflowQueryType,
   ApplicationFailure,
```
sw-yx, line 14: some lint error here you may want to take care of

```suggestion
```

bergundy: I'll fix in the next PR

PR #447, opened 2022-01-13
fix(workflow): Fix startChild options type

lorensr:
I think `options`'s type should have `WithWorkflowArgs`. It does in the `workflowFunc` overload of `startChild`

Reported by @sw-yx  https://temporaltechnologies.slack.com/archives/C01FT8U10GK/p1642035091026400

image


sw-yx (approved):
seeing if roey has any comments

bergundy (approved):
Thanks

PR #448, opened 2022-01-13
chore: Fix linting, remove old comment

lorensr:
Removed these comments, as they seem out of date:

```
* **Override for Workflows that accept no arguments**.
```

`WithWorkflowArgs` works with or without `args`.

https://typescript.temporal.io/api/namespaces/workflow/#startchild

image


packages/workflow/src/workflow.ts
```diff
@@ -456,8 +452,6 @@ export async function startChild<T extends Workflow>(
 /**
  * Start a child Workflow execution
  *
- * **Override for Workflows that accept no arguments**.
```
bergundy, line 459: This is actually correct, so are some of the other overrides below

lorensr: All of the ones without `options`? Would it be more accurate to say "Override for when you don't need `options`"? In which case maybe we don't need to include, since that's clear from the signature, and they appear in order after the overrides that have `options`?

bergundy: All of the options are optional so the override is only for workflows that accept no arguments (see the generic type constraint: `T extends () => Promise<any>`)

bergundy: `args` is required if the workflow takes arguments

lorensr: Is the generic type constraint needed? This compiles:

image

bergundy: What is this supposed to show? There's nothing here that references the workflow arguments.

bergundy (approved):

cretz (approved):

Issue #255, opened 2021-09-20
[Feature Request] Support history replay as exported from tctl and Temporal web


bergundy:
### Is your feature request related to a problem? Please describe.

Other SDKs have the ability to replay exported Workflow history, this is a critical feature for Workflow debuggability.

### Describe the solution you'd like

Add an easy way to replay Workflow history against a given Worker.

### Additional context

Related: https://github.com/temporalio/sdk-java/pull/729

Sushisource:
We should probably do this in core, so we don't need to do it over and over. It should fit naturally into how I already do a lot of testing.

We'll just need to expose the mock poller API or something simpler that just constructs a core instance with a provided history, and uses the mock pollers under the hood.

PR #449, opened 2022-01-15
feat: Replay history from files

Sushisource:
## What was changed
Added ability to replay history from history files

## Why?
Nice to be able to verify changes to workflow code is compatible with existing histories

## Checklist
Draft until some todos are resolved, and depends on https://github.com/temporalio/sdk-core/pull/261 (CI will fail until that is merged and I update)

1. Closes https://github.com/temporalio/sdk-typescript/issues/255

2. How was this tested:
Added tests

3. Any docs updates needed?
Yes


packages/test/src/test-replay.ts
```diff
@@ -0,0 +1,38 @@
...
+});
+
+test('cancel-fake-progress-replay', async (t) => {
+  // TODO: Auto-setup of worker for these kinds of tests
+
+  const histBin = await fs.promises.readFile(
+    path.resolve(__dirname, '../src/history_files/cancel_fake_progress_history.bin')
+  );
+  const hist = History.decode(histBin);
+  await Worker.createReplay(
```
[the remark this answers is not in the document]
bergundy: I'm trying to figure out if there's a good reason to expose `Worker.createReplay` directly.
Maybe all we want is a function to create and run which throws when there's non-determinism errors.

packages/worker/src/core.ts
```diff
@@ -81,20 +93,59 @@ export class Core {
...
   }
 
   /**
-   * Default options get overridden when Core is installed and are remembered in case Core is
-   * re-instantiated after being shut down
+   * Get or instantiate the singleton Core object
+   *
+   * If Core has not been instantiated with {@link install} or this method,
+   * a new Core instance will be installed and configured to connect to
+   * the local docker compose setup.
```
bergundy, line 128: I'd use the chance to rephrase this to:
```suggestion
   * a local server.
```

packages/worker/src/core.ts
```diff
@@ -253,3 +260,65 @@ export class Core {
...
+  protected static instantiator?: 'install' | 'instance';
+  /**
+   * Default options get overridden when Core is installed and are remembered in case Core is
+   * re-instantiated after being shut down
+   */
+  protected static defaultOptions: CoreOptions = {};
+
+  // TODO: How do we de-dupe all this static goop? Maybe have some non-static
+  //   core singleton generic type which these hold a static instance of
+  public static async install(options: CoreOptions): Promise<ReplayCore> {
```
bergundy, line 275: You can reuse the original install signature if you use this as a template param:
https://github.com/temporalio/sdk-typescript/blob/afc0d964ecf21bbe5e79b28300e46bd2273dd38e/packages/worker/src/workflow/vm.ts#L77

packages/worker/src/core.ts
```diff
@@ -253,3 +260,65 @@ export class Core {
...
+   * Create a replay core instance, for replaying static histories
+   */
+  public static async create(options: CoreOptions): Promise<ReplayCore> {
+    const compiledOptions = this.compileOptions(options);
+    const native = await promisify(newReplayCore)(compiledOptions.telemetryOptions);
+
+    return new this(native, compiledOptions);
+  }
+
+  // TODO: accept either history or binary
```
bergundy, line 314: or json, but we can leave that for another pr

packages/worker/src/index.ts
```diff
@@ -11,7 +11,7 @@
 export { State, Worker, DataConverter, defaultDataConverter, errors } from './worker';
 export { WorkerOptions, CompiledWorkerOptions } from './worker-options';
 export { ServerOptions, TLSConfig } from './server-options';
-export { Core, CoreOptions } from './core';
+export { Core, CoreOptions, ReplayCore } from './core';
```
bergundy, line 14: I'm not sure we want to export it here. Users typically shouldn't use this API

packages/worker/src/worker.ts
```diff
@@ -186,6 +198,40 @@ export class Worker {
...
+      taskQueue: options.testName + '-' + this.replayWorkerCount,
+      workflowThreadPoolSize: 1,
+      ...options,
+    };
+    const compiledOptions = compileWorkerOptions(addDefaultWorkerOptions(fixedUpOptions));
+    const replayWorker = await nativeWorkerCtor.createReplay(compiledOptions, history);
+    this.replayWorkerCount += 1;
+    const constructedWorker = await this.bundleWorker(compiledOptions, replayWorker);
+
+    // TODO: Make running lazy? Return type which users can then explicitly start running?
```
bergundy, line 219: Don't see why this is needed

packages/worker/src/worker.ts
```diff
@@ -186,6 +198,40 @@ export class Worker {
...
+  }
+
+  /**
+   * Create a replay Worker.
+   */
+  public static async createReplay(options: ReplayWorkerOptions, history: History): Promise<void> {
+    const nativeWorkerCtor: WorkerConstructor = this.nativeWorkerCtor;
+    const fixedUpOptions = {
+      taskQueue: options.testName + '-' + this.replayWorkerCount,
+      workflowThreadPoolSize: 1,
```
bergundy, line 211: We should instead start the worker in debug mode which doesn't create a thread pool at all.
`debugMode` should be omitted from ReplayWorkerOptions.

packages/worker/src/worker.ts
```diff
@@ -764,10 +811,18 @@ export class Worker {
           return { activation, parentSpan };
         });
       } catch (err) {
-        parentSpan.setStatus({ code: otel.SpanStatusCode.ERROR }).end();
+        if (!(err instanceof errors.ShutdownError)) {
```
bergundy, line 814: Looks like [instrument](https://github.com/temporalio/sdk-typescript/blob/ce23c2c96b3e9bbaa46378d5658f6698d3f6bd5f/packages/worker/src/tracing.ts#L28) already marks the span with code: ERROR

packages/worker/src/worker.ts
```diff
@@ -764,10 +811,18 @@ export class Worker {
...
+          parentSpan.setStatus({ code: otel.SpanStatusCode.ERROR }).end();
+        }
         throw err;
       }
-    });
+    }).pipe(
+      tap({
+        complete: () => {
+          this.workflowPollerStateSubject.next('SHUTDOWN');
```
bergundy, line 822: We should change the state to `FAILED` on error here and perhaps even flush the cache when that state is reached.

packages/worker/src/worker.ts
```diff
@@ -908,7 +970,7 @@ export class Worker {
    */
   async run(): Promise<void> {
     if (this.state !== 'INITIALIZED') {
-      throw new IllegalStateError('Poller was aleady started');
+      throw new IllegalStateError('Poller was already started');
```
bergundy, line 973: There are already enough "r"s in this sentence.

packages/core-bridge/Cargo.toml
```diff
@@ -19,7 +19,11 @@ prost = "0.9"
...
-temporal-sdk-core-api = { version = "0.1.0", path = "./sdk-core/core-api" }
-temporal-sdk-core-protos = { version = "0.1.0", path = "./sdk-core/sdk-core-protos" }
+#temporal-sdk-core = { version = "*", path = "./sdk-core/core" }
+#temporal-sdk-core-api = { version = "*", path = "./sdk-core/core-api" }
+#temporal-sdk-core-protos = { version = "*", path = "./sdk-core/sdk-core-protos" }
+
+temporal-sdk-core = { version = "*", path = "../../../sdk-core/core" }
+temporal-sdk-core-api = { version = "*", path = "../../../sdk-core/core-api" }
+temporal-sdk-core-protos = { version = "*", path = "../../../sdk-core/sdk-core-protos" }
+temporal-sdk-core-test-utils = { version = "*", path = "../../../sdk-core/test-utils" }
```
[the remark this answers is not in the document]
Sushisource: Fixed

packages/core-bridge/Cargo.toml
```diff
@@ -19,7 +19,11 @@ prost = "0.9"
...
-temporal-sdk-core = { version = "0.1.0-alpha.1", path = "./sdk-core/core" }
-temporal-sdk-core-api = { version = "0.1.0", path = "./sdk-core/core-api" }
-temporal-sdk-core-protos = { version = "0.1.0", path = "./sdk-core/sdk-core-protos" }
+#temporal-sdk-core = { version = "*", path = "./sdk-core/core" }
+#temporal-sdk-core-api = { version = "*", path = "./sdk-core/core-api" }
+#temporal-sdk-core-protos = { version = "*", path = "./sdk-core/sdk-core-protos" }
+
+temporal-sdk-core = { version = "*", path = "../../../sdk-core/core" }
+temporal-sdk-core-api = { version = "*", path = "../../../sdk-core/core-api" }
+temporal-sdk-core-protos = { version = "*", path = "../../../sdk-core/sdk-core-protos" }
```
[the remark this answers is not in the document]
Sushisource: Makes sense, done.

packages/test/src/mock-native-worker.ts
```diff
@@ -52,6 +52,10 @@ export class MockNativeWorker implements NativeWorkerLike {
     return new this();
   }
 
+  public static async createReplay(): Promise<NativeWorkerLike> {
```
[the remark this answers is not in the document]
Sushisource: Because otherwise it doesn't adhere to the `WorkerConstructor` interface

packages/test/src/test-replay.ts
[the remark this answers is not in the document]
Sushisource: I think that's typically what people want. We could just lazily init a replay core that is always used for all replay workers.

packages/worker/src/index.ts
Sushisource: It's needed if you want to, for some reason, control your own replay core instance and register workers against it directly. I think for now we can do the lazy-initted core and just use that unless we later discover a reason not to

Sushisource: Oh, no there is a good reason, so you can control logging options when desired.

packages/worker/src/worker.ts
Sushisource: So are you suggesting we invert this to un-set it?

packages/worker/src/worker.ts
Sushisource: Seems like just falling through to https://github.com/temporalio/sdk-typescript/blob/ce23c2c96b3e9bbaa46378d5658f6698d3f6bd5f/packages/worker/src/worker.ts#L996 is the right place to handle that? I think cache flushing subject to that probably goes in another PR.

packages/worker/src/core.ts
Sushisource: I don't think that works - not without making the entire class generic at least, because there is still only one static `_instance` field and it starts being nonassignable when you do this.

Sushisource: Actually I was able to overcome that but it still doesn't work. The static `install` on the base class will use the base class' `_instance` field, not the subclass, so we end up erroneously constructing a real core.

packages/worker/src/rxutils.ts
```diff
@@ -1,5 +1,5 @@
 import { GroupedObservable, OperatorFunction, ObservableInput, pipe, Subject } from 'rxjs';
-import { groupBy, map, scan, mergeScan } from 'rxjs/operators';
```
bergundy, line 2: scan hater

packages/worker/src/worker.ts
```diff
@@ -186,6 +198,43 @@ export class Worker {
...
+      ...options,
+    };
+    this.replayWorkerCount += 1;
+    const compiledOptions = compileWorkerOptions(addDefaultWorkerOptions(fixedUpOptions));
+    const replayWorker = await nativeWorkerCtor.createReplay(compiledOptions, history);
+    const constructedWorker = await this.bundleWorker(compiledOptions, replayWorker);
+
+    const runPromise = constructedWorker.run();
+    runPromise.catch((err) => {
+      console.error('Caught error while replay worker was running', err);
```
bergundy, line 222: Let's not log directly to console

bergundy: We probably shouldn't catch this at all since we await it anyways

packages/worker/src/worker.ts
```diff
@@ -512,16 +562,16 @@ export class Worker {
...
-            filter((state) => state === 'DRAINING'),
+            filter((state) => state === 'SHUTDOWN'),
             first(),
             map((): ContextAware<{ activation: coresdk.workflow_activation.WorkflowActivation; synthetic: true }> => {
               return {
                 parentSpan: this.tracer.startSpan('workflow.shutdown.evict'),
                 activation: new coresdk.workflow_activation.WorkflowActivation({
                   runId: group$.key,
-                  jobs: [{ removeFromCache: { reason: 'Shutting down' } }],
+                  jobs: [{ removeFromCache: { message: 'Shutting down' } }],
```
bergundy, line 574: we should probably attach the reason enum here too

packages/worker/src/worker.ts
bergundy: I don't know if this is possible, we would probably need to make shutdown error not an `Error`, maybe instead we can return an optional value and break the loop if it's undefined.

packages/test/src/test-replay.ts
```diff
@@ -13,23 +13,25 @@ export interface Context {
 const test = anyTest as TestInterface<Context>;
 
 test.before(async (t) => {
+  // We don't want AVA to whine about unhandled rejections thrown by workflows
+  process.removeAllListeners('unhandledRejection');
   const logger = new DefaultLogger('DEBUG');
   // Use forwarded logging from core
-  const replayCore = await ReplayCore.install({ logger, telemetryOptions: { logForwardingLevel: 'DEBUG' } });
+  const replayCore = await ReplayCore.install({ logger, telemetryOptions: { logForwardingLevel: 'INFO' } });
```
bergundy, line 20: Do we need to configure the tracing filter too?

packages/test/tsconfig.json
```diff
@@ -13,6 +13,6 @@
     { "path": "../worker" },
     { "path": "../workflow" }
   ],
-  "include": ["./src/**/*.ts", "./src/history_files/*.json"],
+  "include": ["./src/**/*.ts", "./history_files/*.json"],
```
bergundy, line 16: Oh now I see why you put this in `src`, just so you can import JSONs directly...
We don't have any json histories AFAICT so we could probably just remove this line.
I don't like using sources outside of the `src` directory because it means that anything that was in `src` will be outputted to `lib/src`

packages/worker/src/core.ts
```diff
@@ -15,9 +15,9 @@ import { compileServerOptions, getDefaultServerOptions, RequiredServerOptions, S
 import { DefaultLogger, LogEntry, Logger, LogTimestamp, timeOfDayToBigint } from './logger';
 import * as errors from './errors';
 import { temporal } from '@temporalio/proto';
-import History = temporal.api.history.v1.History;
 import { byteArrayToBuffer } from './utils';
 import { MakeOptional } from '@temporalio/common/src/type-helpers';
+import History = temporal.api.history.v1.History;
```
bergundy, line 20: Let's export this as the interface instead:
```ts
export type History = temporal.api.history.v1.IHistory;
```

packages/worker/src/core.ts
```diff
@@ -291,24 +284,15 @@ export class Core {
...
-  }
-
-  protected static setInstance(i: Promise<ReplayCore>): void {
-    this._replayInstance = i;
-  }
-
-  protected static clearInstance(): void {
-    delete this._replayInstance;
-  }
+  // protected static _instanceCtor = (options: native.CoreOptions, callback: native.CoreCallback): void => {
```
bergundy, line 287: letftover comment

packages/worker/src/worker.ts
```diff
@@ -229,17 +229,26 @@ export class Worker {
...
+        constructedWorker.shutdown();
+      } catch {
+        // We may have already called shutdown from the eviction branch, so ignore
+        // errors if we're already shut down.
+      }
     });
 
-    const nondeterminismPromise = firstValueFrom(
-      constructedWorker.evictions.pipe(
+    const evcitionPromise = firstValueFrom(
```
bergundy, line 242:
```suggestion
    const evictionPromise = firstValueFrom(
```

packages/worker/src/worker.ts
```diff
@@ -229,17 +229,26 @@ export class Worker {
     const constructedWorker = await this.bundleWorker(compiledOptions, replayWorker);
 
     const runPromise = constructedWorker.run();
-    const pollerShutdown = constructedWorker.workflowPollerStateSubject.pipe(
-      filter((state): state is 'SHUTDOWN' => state === 'SHUTDOWN')
-    );
+    const pollerShutdown = constructedWorker.workflowPollerStateSubject.pipe(filter((state) => state !== 'POLLING'));
     pollerShutdown.subscribe((_) => {
```
bergundy, line 233: Let's use `firstValueFrom` instead of subscribing.
If you subscribe you need to unsubscribe too which is more error prone (e.g. I missed it in the first 2 reviews).
https://rxjs.dev/api/index/class/Observable#cancel-a-subscription

bergundy (commented):
Small leftovers but we're mostly there

packages/test/src/test-replay.ts
Sushisource: Nah, no need. It's easy enough to just manually tweak when desired during test

packages/worker/src/core.ts
Sushisource: I don't think so - it should be the actual reified type, which ensures that it's actually been parsed properly and doesn't have undefined fields that can't actually be that way in a real history

bergundy (approved):

PR #451, opened 2022-01-26
Optimize rust building in CI

Sushisource:
## What was changed
* Use full LTO & no debug symbols for Rust release builds to decrease package size
* Build debug for local development, release for CI & cross-compile/packaging
* Don't lint on every platform
* Don't double-build rust in cross compile jobs

Next PR:
* Add integration with sdk-features repo

## Why?
Faster CI, smaller packages, features coverage

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/core-bridge/scripts/build.js
+++ b/packages/core-bridge/scripts/build.js
@@ -27,6 +27,7 @@ const args = arg({
   '-h': '--help',
   '--version': Boolean,
   '-v': '--version',
+  '--release': Boolean,
   '--target': [String],
   '--force': Boolean,
   '-f': '--force',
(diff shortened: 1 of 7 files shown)


packages/core-bridge/scripts/build.js
```diff
@@ -65,9 +66,10 @@ if (unsupportedTargets.length) {
   process.exit(1);
 }
 const forceBuild = args['--force'];
+const buildRelease = args['--release'] || process.env.BUILD_CORE_RELEASE !== undefined;
```
bergundy, line 69: nit: `['y', 'yes', 't', 'true', '1'].includes((process.env.BUILD_CORE_RELEASE ?? '').toLowerCase())`

bergundy (approved):

packages/core-bridge/scripts/build.js
Sushisource: Ehh, I'm not a fan TBH. If you don't want the env var set, just unset it. IMO best practice here is to have boolean-like env vars always default to false and any setting means true. That way there's no reason to ever set it explicitly to false. If you want to unset it, just unset the env var.

bergundy: I just feel like anyone setting `BUILD_CORE_RELEASE=0` would be surprised with the current behavior.

Sushisource: Why would you ever do that, is my point. Just unset it, or never set it in the first place, since the default's false. It's a very common pattern IME

Issue #457, opened 2022-02-01
[Bug] Pre-built Rust bridge does not seem to be used on Windows


bergundy:
According to the build logs here it take a lot longer to install the SDK on Windows than Ubuntu / MacOS.
https://github.com/temporalio/sdk-typescript/runs/4959704412?check_suite_focus=true

Also reported in the forums: https://community.temporal.io/t/typescript-sdk-help/3865

PR #458, opened 2022-02-02
feat(core): Make Core portable

bergundy:
- `core-bridge` can now be imported in a different OS than the one the SDK was installed in.
- Closes https://github.com/temporalio/sdk-typescript/issues/457

### DIFF
--- a/packages/core-bridge/common.js
+++ b/packages/core-bridge/common.js
@@ -1,0 +1,50 @@
+/**
+ * Shared code for scripts/build.js and index.js
+ *
+ * @module
+ */
+
+const os = require('os');
+const path = require('path');
+const fs = require('fs');
+
+// List of tested compile targets
+const targets = [
+  'x86_64-apple-darwin',
+  'aarch64-apple-darwin',
+  'x86_64-unknown-linux-gnu',
+  'aarch64-unknown-linux-gnu',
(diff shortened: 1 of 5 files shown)


packages/core-bridge/common.js
```diff
@@ -0,0 +1,50 @@
...
+  if (arch === undefined) {
+    throw new PrebuildError(`No prebuilt module for arch ${os.arch()}`);
+  }
+  const platform = platformMapping[os.platform()];
+  if (platform === undefined) {
+    throw new PrebuildError(`No prebuilt module for platform ${os.platform()}`);
+  }
+  const binary = path.resolve(__dirname, 'releases', `${arch}-${platform}`, 'index.node');
+  if (fs.existsSync(binary)) {
+    console.log('Copied prebuilt bridge module', { binary });
```
Sushisource, line 43: Where'd the copy go? Log line should change if this happened elsewhere, but I don't see where it moved to?

Sushisource: OIC it's just used in place. Log line should change though

Sushisource (approved):
LGTM

packages/core-bridge/common.js
bergundy: 👍

PR #467, opened 2022-02-04
fix!: Improve failure details, fixes #462

### LINKED ISSUE #462: [Bug] Put more info into workflow activation failures
## Problem

Currently, an Error thrown from a workflow results in this Worker output (& Web history):

```ts
throw new Error('error in workflow');
// Failing workflow activation...
// { message: "Error: error in workflow", source: "TypeScriptSDK", stack_trace: "", cause: None, failure_info: None }
```

And throwing an object results in:

```ts
throw { field: 'data' };
// Failing workflow activation...
// { message: "[object Object]", source: "TypeScriptSDK", stack_trace: "", cause: None, failure_info: None }
```

### Correct

- `failure_info`: Should be None in Error case. Use [subclass of `TemporalFailure`](https://docs.temporal.io/docs/typescript/handling-failure/#failure-classes-reference) to get this filled in.

### Change

- `message`: If thrown object isn't an Error, set `message = JSON.stringify(object)`. 
- `stack_trace`: Fill it in when Error is thrown (currently only filled in when `instanceof TemporalFailure`). When non-Error is thrown, log warning saying "we recommend throwing Errors so that you can see the stack trace".
- `cause`: When thrown error has a `cause` property, put a `Failure` here (currently only works when provided to constructor of `TemporalFailure` etc)
- `source`: Maybe be more specific, like `TypeScriptSDK > Worker > myWorkflowName`

## Related

@Sushisource We can't do anything about `throw new Error(object)`, since Node calls `toString()` on it before we get it:

image

image

lorensr:
Implements all I suggested in #462  except for more specific `source`, which I think would require passing more info to `errorToFailure`. I'd like to get to that later.

This change could be breaking: if a user is checking whether `failure.message === 'Error: orig message'`, it won't catch the new `failure.message = 'orig message'`. I think the old behavior that added the `Error: ` was unintentional—at the end of `errorToFailure`, the `instanceof Error` failed because the Error was coming from the vm, so `String(err)` was used.

### DIFF
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -1,6 +1,6 @@
 import type { temporal } from '@temporalio/proto/lib/coresdk';
 import { DataConverter, arrayFromPayloads } from './converter/data-converter';
-import { checkExtends } from './type-helpers';
+import { checkExtends, hasOwnProperties, isRecord } from './type-helpers';
 
 export const FAILURE_SOURCE = 'TypeScriptSDK';
 export type ProtoFailure = temporal.api.failure.v1.IFailure;
@@ -220,7 +220,7 @@ export async function optionalErrorToOptionalFailure(
 /**
  * Stack traces will be cutoff when on of these patterns is matched
  */
-const CUTTOFF_STACK_PATTERNS = [
+const CUTOFF_STACK_PATTERNS = [
   /** Activity execution */
   /\s+at Activity\.execute \(.*[\\/]worker[\\/](?:src|lib)[\\/]activity\.[jt]s:\d+:\d+\)/,
   /** Workflow activation */
@@ -234,7 +234,7 @@ export function cutoffStackTrace(stack?: string): string {
   const lines = (stack ?? '').split(/\r?\n/);
   const acc = Array<string>();
   lineLoop: for (const line of lines) {
-    for (const pattern of CUTTOFF_STACK_PATTERNS) {
+    for (const pattern of CUTOFF_STACK_PATTERNS) {
       if (pattern.test(line)) break lineLoop;
     }
     acc.push(line);
@@ -329,15 +329,22 @@ export async function errorToFailure(err: unknown, dataConverter: DataConverter)
     source: FAILURE_SOURCE,
   };
 
-  if (err instanceof Error) {
-    return { ...base, message: err.message ?? '', stackTrace: cutoffStackTrace(err.stack) };
+  if (isRecord(err) && hasOwnProperties(err, ['message', 'stack'])) {
+    return {
+      ...base,
+      message: String(err.message) ?? '',
+      stackTrace: cutoffStackTrace(String(err.stack)),
+      cause: await optionalErrorToOptionalFailure(err.cause, dataConverter),
+    };
   }
+
+  const stackTrace = `A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.`;
 
   if (typeof err === 'string') {
-    return { ...base, message: err };
+    return { ...base, stackTrace, message: err };
   }
 
-  return { ...base, message: String(err) };
+  return { ...base, stackTrace, message: JSON.stringify(err) };
 }
 
 /**
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -338,13 +338,13 @@ export async function errorToFailure(err: unknown, dataConverter: DataConverter)
     };
   }
 
-  const stackTrace = `A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.`;
+  const recommendation = ` [A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.]`;
 
   if (typeof err === 'string') {
-    return { ...base, stackTrace, message: err };
+    return { ...base, message: err + recommendation };
   }
 
-  return { ...base, stackTrace, message: JSON.stringify(err) };
+  return { ...base, message: JSON.stringify(err) + recommendation };
 }
 
 /**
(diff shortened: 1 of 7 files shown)


packages/common/src/failure.ts
```diff
@@ -329,15 +329,22 @@ export async function errorToFailure(err: unknown, dataConverter: DataConverter)
...
 
+  const stackTrace = `A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.`;
+
   if (typeof err === 'string') {
-    return { ...base, message: err };
+    return { ...base, stackTrace, message: err };
   }
 
-  return { ...base, message: String(err) };
+  return { ...base, stackTrace, message: JSON.stringify(err) };
```
bergundy, line 347: I guess this is better in most cases

packages/common/src/failure.ts
```diff
@@ -329,15 +329,22 @@ export async function errorToFailure(err: unknown, dataConverter: DataConverter)
...
+  if (isRecord(err) && hasOwnProperties(err, ['message', 'stack'])) {
+    return {
+      ...base,
+      message: String(err.message) ?? '',
+      stackTrace: cutoffStackTrace(String(err.stack)),
+      cause: await optionalErrorToOptionalFailure(err.cause, dataConverter),
+    };
   }
 
+  const stackTrace = `A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.`;
```
bergundy, line 341: Not sure I'd use this as the stack trace because it's not formatted like one, consider putting this in the message.

packages/common/src/failure.ts
```diff
@@ -329,15 +329,22 @@ export async function errorToFailure(err: unknown, dataConverter: DataConverter)
     source: FAILURE_SOURCE,
   };
 
-  if (err instanceof Error) {
-    return { ...base, message: err.message ?? '', stackTrace: cutoffStackTrace(err.stack) };
+  if (isRecord(err) && hasOwnProperties(err, ['message', 'stack'])) {
```
bergundy, line 332: Why not check `instanceof Error`?

packages/common/src/failure.ts
```diff
@@ -329,15 +329,22 @@ export async function errorToFailure(err: unknown, dataConverter: DataConverter)
...
+    return {
+      ...base,
+      message: String(err.message) ?? '',
+      stackTrace: cutoffStackTrace(String(err.stack)),
+      cause: await optionalErrorToOptionalFailure(err.cause, dataConverter),
```
bergundy, line 337: Until this change `cause` was only used for chaining TemporalFailures, I don't know if we chain error causes in the other SDKs and if we should.

packages/common/src/failure.ts
lorensr: `err instanceof Error` is false when the Workflow does `throw new Error('unhandled rejection');`. Could also be helpful in the rare case of people throwing error-like objects

packages/common/src/failure.ts
lorensr: I haven't found where in Go SDK workflow errors are handled, but in Java, looks like `Error`s are rethrown and other non-specific throwables are wrapped in `WorkflowServiceException` with the throwable as the `cause`:

https://github.com/temporalio/sdk-java/blob/b0199150b69bb5a4eaf6c13ba4b89c2a8746a113/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java#L299-L326

I'm not sure whether uniformity is important here? I figure we have a `cause` field, and might as well use it, as it could be helpful for users who are catching and throwing new non-`ApplicationFailure`s and want to include the original error as a cause. It will probably become a more common pattern (was added to v8 last year).

bergundy: It will become popular I suspect, now it requires node > 16.9 AFAIK.

Let's keep it, sounds like a good addition to me.

packages/common/src/failure.ts
```diff
@@ -329,15 +329,22 @@ export async function errorToFailure(err: unknown, dataConverter: DataConverter)
...
 
+  const recommendation = ` [A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.]`;
+
   if (typeof err === 'string') {
-    return { ...base, message: err };
+    return { ...base, message: err + recommendation };
   }
 
-  return { ...base, message: String(err) };
+  return { ...base, message: JSON.stringify(err) + recommendation };
```
Sushisource, line 347: :tada:

Sushisource (approved):

bergundy:
@lorensr make sure you sign into buildkite (you probably need an account) so all tests can run and we don't have to force merge.
In the near future, I'd like to have our own GH actions builders so we can drop BuildKite altogether.

lorensr:
The issue isn't whether I had an account, but that I'm working on a fork. Should I check this?

image

bergundy:
> The issue isn't whether I had an account, but that I'm working on a fork. Should I check this?
> 
> image

I think this applies to GH actions only. Afaik  buildkite works for forks, @Sushisource could probably answer that, I’d need to check but I can’t atm.

lorensr:
Screenshot is from https://buildkite.com/temporal/typescript-sdk/settings/repository 
I was just wondering if it's unchecked for a security reason.

image

packages/common/src/failure.ts
bergundy: Oh, that makes sense because the `Error` is constructed in isolate context and converted into failure outside of that. We should probably work around it and ask the workflow to convert to a failure since it has an instance of the user's data converter.

bergundy (approved):
Just one comment, not anything too critical.
Keep it in mind when working on custom DC.

packages/common/src/failure.ts
lorensr: By moving this `errorToFailure`:

https://github.com/temporalio/sdk-typescript/blob/0488e8a74fa78c44d9fcad6e2631c7aae868aae5/packages/worker/src/workflow/vm.ts#L196

to here, instead of throwing? 

https://github.com/temporalio/sdk-typescript/blob/0488e8a74fa78c44d9fcad6e2631c7aae868aae5/packages/workflow/src/internals.ts#L504

PR #468, opened 2022-02-04
feat(client): Use `runId` only in handles created with `getHandle`

### LINKED ISSUE #365: [Feature Request] Better context for grpc errors in client
Right now client errors are thrown by grpc-js asynchronously and the stack trace is meaningless.
We should catch errors from WorkflowClient methods and rethrow to provide better context.

### LINKED ISSUE #377: Improve error message: Workflow executionsRow not found

In [this scenario](https://github.com/temporalio/samples-typescript/issues/56), the error from the Worker that had the correct code was:

```
Task not found when completing: status: NotFound, message: "Workflow executionsRow not found."
```

 Say what's going wrong in the Worker at a higher level and suggest possible reasons.
 
 Migrated from #359

### LINKED ISSUE #464: [Bug] WorkflowExecutionFailedEventAttributes.retry_state not respected
### Describe the bug

Seems during workflow result, retry state is hardcoded instead of using that field

bergundy:
In addition:
- Adds safety to `terminate` and `cancel` so handles created with `start` can't accidentally affect workflows that are not part of the same execution chain
- Adds optional `firstExecutionRunId` param to `getHandle` for added safety
- Closes #464
- Closes #377
- Closes #365

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -1,4 +1,5 @@
 import os from 'os';
+import { status as grpcStatus } from '@grpc/grpc-js';
 import { temporal } from '@temporalio/proto';
 import { WorkflowClientInterceptors } from './interceptors';
 import { v4 as uuid4 } from 'uuid';
@@ -21,6 +22,7 @@ import {
   TimeoutFailure,
   TimeoutType,
   compileRetryPolicy,
+  WorkflowNotFoundError,
 } from '@temporalio/common';
 import { WorkflowOptions, compileWorkflowOptions, WorkflowSignalWithStartOptions } from './workflow-options';
 import {
@@ -38,8 +40,15 @@ import {
   StartWorkflowExecutionRequest,
   TerminateWorkflowExecutionResponse,
   RequestCancelWorkflowExecutionResponse,
+  WorkflowExecution,
 } from './types';
-import { WorkflowFailedError, WorkflowContinuedAsNewError } from './errors';
+import {
+  WorkflowFailedError,
+  WorkflowContinuedAsNewError,
+  WorkflowExecutionAlreadyStartedError,
+  ServiceError,
+  isServerErrorResponse,
+} from './errors';
 import { Connection, WorkflowService } from './connection';
 
 /**
@@ -190,7 +199,32 @@ export interface WorkflowResultOptions {
    *
    * @default true
    */
-  followRuns: boolean;
+  followRuns?: boolean;
+}
+
+export interface GetWorkflowHandleOptions extends WorkflowResultOptions {
+  /**
+   * ID of the first execution in the Workflow execution chain.
+   *
+   * When getting a handle with no `runId`, pass this option to ensure some
+   * {@link WorkflowHandle} methods (e.g. `terminate` and `cancel`) don't
+   * affect executions from another chain.
+   */
+  firstExecutionRunId?: string;
+}
+
+interface WorkflowHandleOptions extends GetWorkflowHandleOptions {
+  workflowId: string;
+  runId?: string;
+  interceptors: WorkflowClientCallsInterceptor[];
+  /**
+   * A runId to use for getting the workflow's result.
+   *
+   * - When creating a handle using `getHandle`, uses the provided runId or firstExecutionRunId
+   * - When creating a handle using `start`, uses the returned runId (first in the chain)
+   * - When creating a handle using `signalWithStart`, uses the the returned runId
+   */
+  runIdForResult?: string;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


packages/test/src/test-integration.ts
```diff
@@ -773,4 +775,120 @@ if (RUN_INTEGRATION_TESTS) {
...
+
+  test('Handle from WorkflowClient.getHandle follows only own execution chain', async (t) => {
+    const { client } = t.context;
+    const workflowId = uuid4();
+    const handleFromThrowerStart = await client.start(workflows.throwAsync, {
+      taskQueue: 'test',
+      workflowId,
+    });
+    const handleFromGet = client.getHandle(workflowId, undefined, {
+      firstExecutionRunId: handleFromThrowerStart.originalRunId,
```
sw-yx, line 845: when should we tell people to use this in docs?

packages/client/src/workflow-client.ts
```diff
@@ -190,7 +199,32 @@ export interface WorkflowResultOptions {
    *
    * @default true
    */
-  followRuns: boolean;
+  followRuns?: boolean;
+}
+
+export interface GetWorkflowHandleOptions extends WorkflowResultOptions {
+  /**
+   * ID of the first execution in the Workflow execution chain.
```
lorensr, line 207: Define or link to existing definition of chain

lorensr (approved):

packages/client/src/workflow-client.ts
```diff
@@ -275,7 +309,14 @@ export class WorkflowClient {
     // Cast is needed because it's impossible to deduce the type in this situation
     const interceptors = (this.options.interceptors.calls ?? []).map((ctor) => ctor({ workflowId }));
     const runId = await this._start(workflowTypeOrFunc, { ...options, workflowId }, interceptors);
-    const handle = this._createWorkflowHandle(workflowId, runId, interceptors, {
+    // runId is not used in handles created with `start*` calls because these
+    // handles should be allow interacting with the workflow if it continues as new.
+    const handle = this._createWorkflowHandle({
+      workflowId,
+      runId: undefined,
```
cretz, line 316: I wonder if we want to document that if you want to cancel/terminate/describe/etc a specific run ID, you must use `getHandle`.

cretz (approved):

packages/client/src/workflow-client.ts
sw-yx: how else would you do it with public api's?

cretz: Can you clarify the question?

What I mean is previously if you did a start + describe, you'd be describing the run that resulted from the start (even if it had continue-as-new'd). Now you'll be describing the latest run for the workflow ID. That seems like the right behavior, but I just wonder if we need to make it clear that you're not necessarily describing the result of start, you're describing the latest.

sw-yx: ah ok. i misread your first comment then.

packages/test/src/test-integration.ts
bergundy: When they want to make sure that they don't terminate / cancel a workflow if it's not part of a specific execution chain.

packages/client/src/workflow-client.ts
bergundy: Do we have one?

cretz:
Re https://github.com/temporalio/sdk-python/pull/6#discussion_r803792659, I think we might need to revisit this.

For handles returned by start and signal-with-start, we are allowing query, signal, cancel (for signal-with-start), and terminate to occur on a completely different workflow. We should default to using the run ID for all of those I think.

```typescript
// Server 1

const handle = await client.start(example, { args: ['do not detonate'], taskQueue: 'task-queue', workflowId: 'nuclear-workflow' });

// Meanwhile on server 2

const handle = await client.start(example, { args: ['detonate'], taskQueue: 'task-queue', workflowId: 'nuclear-workflow' });

// Back on server 1

if (await handle.query('should detonate')) {
  launch();
}
```

You just ended the world :-)

bergundy:
I'm not convinced you're right @cretz.
It really depends on the use case.
AFAIK, in a lot of cases users don't reuse the same workflow ID for different execution chains.

bergundy:
I think this change is a step in the right direction and we'll make it safer with server changes.

cretz:
After discussion, while I disagree, the team has decided that for now it's ok to silently and accidentally signal, query, cancel, or terminate incorrect workflows until the server change is in place.

packages/client/src/workflow-client.ts
lorensr: No

PR #477, opened 2022-02-11
feat!: Custom data converter

### LINKED ISSUE #130: [Feature Request] Custom data converters
At the moment there is no way of customizing the Workflow data converter.
This is a complex problem because of Workflows run in isolation and do not have access to node libraries.

While we're tacking this problem we might want to make the data converter interface asynchronous to facilitate network requests and non-blocking compression.

Solution requires a proper proposal in the proposals repo.

### LINKED ISSUE #237: [Feature Request] Protobuf payload converter support

To fully support cross language calls we need to support the json and binary protobuf payload types.

Implement a payload converter that supports reading and writing protobuf payloads and add that to the default data converter.

### Additional context

https://github.com/temporalio/sdk-java/blob/7ee9034bcfc8d56f6f5dad86ce211209f18f2697/temporal-sdk/src/main/java/io/temporal/common/converter/ProtobufPayloadConverter.java
https://github.com/temporalio/sdk-java/blob/7ee9034bcfc8d56f6f5dad86ce211209f18f2697/temporal-sdk/src/main/java/io/temporal/common/converter/ProtobufJsonPayloadConverter.java

lorensr:
## What was changed

- Implemented custom data converter feature. Design doc: https://github.com/lorensr/sdk-node/blob/custom-data-converter/docs/data-converter.md
- Separated `common` package into `common` (used everywhere but the workflow) and `workflow-common` (used everywhere) to reduce workflow bundle size.
- Un-reverted https://github.com/temporalio/sdk-typescript/pull/430

Sorry, I should have un-reverted in a subsequent PR so that it wouldn't be included in this PR diff. Here's a diff without that first un-revert commit:

https://github.com/lorensr/sdk-node/compare/cc0d41435c75dc224496f78f8fc4a2e2a14213ec...custom-data-converter

## Checklist

1. Closes #130
2. Closes #237

3. How was this tested:

Draft PR: tests aren't passing yet, and I need to write more to cover codec.

4. Any docs updates needed?

Yep


packages/workflow/package.json
```diff
@@ -2,27 +2,29 @@
...
+  "types": "lib/index.d.ts",
   "files": [
     "lib"
   ],
+  "scripts": {},
+  "dependencies": {
+    "@temporalio/workflow-common": "file:../workflow-common"
+  },
+  "devDependencies": {
+    "@temporalio/proto": "file:../proto"
```
bergundy, line 26: Is this a dev dependency now? Is it even a dependency anymore?

docs/data-converter.md
```diff
@@ -0,0 +1,78 @@
...
+
+- imports `payloadConverterPath`
+- passes `payloadConverterPath` to `VMWorkflowCreator` constructor
+
+`VMWorkflowCreator.createWorkflow`:
+_worker thread (unless in debug mode)_
+
+- passes `useCustomPayloadConverter` to `initRuntime` inside Workflow vm
+
+`worker-interface.ts#initRuntime`:
```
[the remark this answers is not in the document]
bergundy: I see you've done this with aliases.

docs/protobuf-libraries.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Options
+
+### protobufjs
+
+A and B, but not C.
+
+- Most popular lib (5M downloads/wk)
+- Fairly inactive maintainers (infrequent updates, many open PRs & issues)
+- Non-standard JSON serialization
```
cretz, line 15: What makes it non-standard? I expect every message class to have a [toJSON](https://protobufjs.github.io/protobuf.js/Message.html#toJSON) method and a `fromJSON` static method.

lorensr: Every class has a `toJSON` method, but `JSON.stringify()` doesn't match the spec:

https://developers.google.com/protocol-buffers/docs/proto3#json

Eg `stringify` throws on BigInts and doesn't base64-encode binary. protobufjs hasn't implemented a number of the types that need custom encoding: https://github.com/protobufjs/protobuf.js/issues/1304

cretz: > Eg stringify throws on BigInts and doesn't base64-encode binary.

Why use stringify? Doesn't the toJSON directly work with these? Is there an issue to follow?

> protobufjs hasn't implemented a number of the types that need custom encoding: https://github.com/protobufjs/protobuf.js/issues/1304

Makes sense. They just need better support for well known types. I wonder if we can contribute? Otherwise, I wonder if we can walk the object for WKT's and convert them ourselves.

bergundy: The last commit in their repo was in last May, I fear contribution is not a viable option.

lorensr: > Why use stringify? Doesn't the toJSON directly work with these? Is there an issue to follow?

I just meant that if `toJSON` were implemented with `JSON.stringify`, it wouldn't work. I made this issue: https://github.com/temporalio/sdk-typescript/issues/481

> I wonder if we can contribute? Otherwise, I wonder if we can walk the object for WKT's and convert them ourselves.

I'd like to try contributing. If converting ourselves, unless we fork protobufjs, we'd be doing what `proto3-json-serializer` does, and would have the same requirement on how the user generates & provides their messages.

cretz: :+1: I am not sure it's reasonable to ask people to use our custom protoc. Ideally we can t do this from the outside on protobufjs-generated classes/descriptors.

Having said that, this being the only one of our languages without a canonical, supported proto-JSON implementation, I'd support just erroring in data converters if protos are provided in either direction, but allow them to use a custom converter to pick their preferred conversion (or binary only, but we should still error on from-payload when we receive `json/protobuf`).

lorensr: > to ask people to use our custom protoc

We're still having them use the protobufjs generator (`pbjs`). It may not be the way they're currently running it. If not, I think it's straightforward to switch or to do both—our way for temporal code and their way for the rest of their code. The main difference in use is doing `Message.create()` instead of `new Message()`.

If we think running the generator our way is too big an ask for some users, we could add support for just `json/binary` for them, without removing current functionality.

> the only one of our languages without a canonical, supported proto-JSON implementation

I think the library we're using is well-maintained and from Google/canonical: https://github.com/googleapis/proto3-json-serializer-nodejs

lorensr: > Can you document why in one of your docs? Is this a limitation of proto3-json-serializer?

Yes. Limitation of this way of running `pbjs`.

packages/common/src/codec-helpers.ts
```diff
@@ -0,0 +1,137 @@
...
+  toPayloads,
+} from '@temporalio/workflow-common';
+
+export async function decodeFromPayloadsAtIndex<T>(
+  converter: LoadedDataConverter,
+  index: number,
+  payloads?: Payload[] | null
+): Promise<T> {
+  const { payloadConverter, payloadCodec } = converter;
+  return fromPayloadsAtIndex(payloadConverter, index, payloads ? await payloadCodec.decode(payloads) : payloads);
```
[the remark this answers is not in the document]
lorensr: fromPayloadsAtIndex is sync

packages/common/src/codec-helpers.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export async function encodeErrorToFailure(dataConverter: LoadedDataConverter, error: unknown): Promise<ProtoFailure> {
+  const { payloadConverter, payloadCodec } = dataConverter;
+  return await encodeFailure(payloadCodec, errorToFailure(error, payloadConverter));
+}
+
+/**
+ * Run `codec.encode()` on the {@link Payload}s in a {@link ProtoFailure}. Mutates `failure`.
+ */
+export async function encodeFailure(codec: PayloadCodec, failure: ProtoFailure): Promise<ProtoFailure> {
```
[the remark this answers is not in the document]
lorensr: Do you believe the clarity is worth the less flexible use? Eg this:

image

Becomes:

image

packages/test/src/payload-converters/payload-converter.ts
```diff
@@ -0,0 +1,6 @@
+import { DefaultPayloadConverter } from '@temporalio/common';
+import root, { foo } from '../../protos/root';
+
+export const messageInstance = foo.bar.ProtoActivityInput.create({ name: 'Proto', age: 1 });
```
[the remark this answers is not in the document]
lorensr: I don't understand what kind of doc / what to explain. This line is here for internal testing. The root & `.create` API is explained in this doc: https://github.com/lorensr/sdk-node/blob/custom-data-converter/docs/protobuf-libraries.md#current-solution

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -0,0 +1,142 @@
...
+export class WorkflowCodecRunner {
+  constructor(protected readonly codec: PayloadCodec) {}
+
+  /**
+   * Run codec.decode on the Payloads in the Activation message
+   */
+  public async decodeActivation(activation: coresdk.workflow_activation.IWorkflowActivation): Promise<void> {
+    if (activation.jobs?.length) {
+      await Promise.all(
+        activation.jobs.flatMap((job) => [
```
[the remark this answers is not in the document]
lorensr: I'm trying to await all of the promises at once / in parallel. How would I do `Promise.all(arrayOfPromises)` without flatMap?

packages/worker/src/workflow/bundler.ts
```diff
@@ -126,7 +133,23 @@ export class WorkflowCodeBundler {
...
+    if (webpackConfig.resolve) {
+      if (payloadConverterPath) {
+        webpackConfig.resolve.alias = {
+          __temporal_custom_payload_converter$: payloadConverterPath,
+        };
+      } else {
+        // Save 100KB from the bundle by not including the npm module,
+        // since we can't decode without a `root` anyway.
+        webpackConfig.resolve.alias = {
+          'proto3-json-serializer$': path.resolve(__dirname, 'empty-proto3-json-serializer.js'),
```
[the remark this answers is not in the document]
lorensr: Like this? Webpack still finds it and includes in bundle. 

image

We could use https://webpack.js.org/plugins/ignore-plugin/  instead. 

Both ignore and alias have the issue of our users trying to use `proto3-json-serializer` themselves. The ideal thing would be an inline `/* webpack-ignore */` when we import it, and then explicitly saying "include this module" here when calling `webpack()` if `payloadConverterPath`. I can't find such a thing. An alternative is creating a `__temporal_proto3-json-serializer`—this seems best to me.

packages/worker/src/workflow/bundler.ts
```diff
@@ -126,7 +133,23 @@ export class WorkflowCodeBundler {
         filename: 'main.js',
         library: '__TEMPORAL__',
       },
-    });
+    };
+    if (webpackConfig.resolve) {
+      if (payloadConverterPath) {
+        webpackConfig.resolve.alias = {
+          __temporal_custom_payload_converter$: payloadConverterPath,
```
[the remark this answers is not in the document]
lorensr: The `workflowInterceptorModules`? Alias feels like a little less of a hack to me 🤷‍♂️

packages/common/src/codec-helpers.ts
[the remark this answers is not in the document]
bergundy: As a general pattern, yes. Because it's internal to the SDK it's acceptable though.
Ideally we'd not mutate and return a new object.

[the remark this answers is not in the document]
bergundy: Come to think of it, I wouldn't mutate since user code might have reference to the proto failure: https://github.com/temporalio/sdk-typescript/blob/bcc55bd726e71a8f5a134fde6f05706cf34c8bc7/packages/common/src/failure.ts#L55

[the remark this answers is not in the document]
bergundy: https://github.com/temporalio/sdk-typescript/blob/bcc55bd726e71a8f5a134fde6f05706cf34c8bc7/packages/common/src/failure.ts#L250

packages/test/src/payload-converters/payload-converter.ts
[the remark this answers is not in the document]
bergundy: Just say that it's a message used in tests

packages/worker/src/workflow-codec-runner.ts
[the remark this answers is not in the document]
bergundy: I was thinking you could use conditions and accumulate the promises but that's not critical.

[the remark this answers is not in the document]
bergundy: I'd consider not mutating here too for extra safety

packages/worker/src/workflow/bundler.ts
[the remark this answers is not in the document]
bergundy: I think with interceptors there's not much of a choice because there might be more than one module registered.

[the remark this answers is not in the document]
bergundy: Alias does feel like less of a hack but still hackish :)

packages/worker/src/workflow/vm.ts
```diff
@@ -174,7 +188,7 @@ export class VMWorkflow implements Workflow {
     if (this.unhandledRejection) {
       return coresdk.workflow_completion.WorkflowActivationCompletion.encodeDelimited({
         runId: activation.runId,
-        failed: { failure: await errorToFailure(this.unhandledRejection, defaultDataConverter) },
+        failed: { failure: errorToFailure(this.unhandledRejection, this.payloadConverter) },
```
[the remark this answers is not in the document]
lorensr: `this.unhandledRejection` is set by `process.on('unhandledRejection'`, which is outside vm. Should I do like `__unsafe_process.on('unhandledRejection'` inside vm, or wrap `activate` and `concludeActivation` in try/catch in `worker-interface`?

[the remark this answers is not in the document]
bergundy: The handler is global for the worker thread, you can't do that.

I'm just suggesting you add a method in worker-interface to convert the failure in context.

packages/workflow/package.json
lorensr: I added as devDep because we `import type { coresdk }` in a few places in this package

lorensr:
Addressed review comments and:
- Fixed and added tests
- Stopped encoding/decoding headers (pending further discussion, will probably re-enable and add to otel in future PR)
- Restructured common packages to:
```
common
internal-workflow-common
internal-non-workflow-common
```
where the first exports things you might want to use in your own common code (shared between client/worker/workflow) like data converters, failures, and errors.

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -0,0 +1,191 @@
...
+ */
+export class WorkflowCodecRunner {
+  constructor(protected readonly codec: PayloadCodec) {}
+
+  /**
+   * Run codec.decode on the Payloads in the Activation message.
+   */
+  public async decodeActivation(
+    activation: coresdk.workflow_activation.IWorkflowActivation
+  ): Promise<coresdk.workflow_activation.IWorkflowActivation> {
```
Sushisource, line 17: This (and the similar stuff in the codec helpers file) seems maybe too verbose / change-sensitive to me.

Typically serialization concerns are handled with some sort of visitor pattern. IE: All fields of the struct are visited (unless explicitly excluded), and then de/serialized according to that field's type and possibly additional metadata. Here, we're visiting certain known fields and then handling them directly. It seems we could both shorten the code and make it more flexible in the face of change by visiting (transitively) any field with the `Payload` type, and applying the decode to it.

Of course that sort of type info is lost at runtime in TS so maybe we do what's described here https://stackoverflow.com/a/57594451/295987 and we just make a list of "here's all the payload fields we care about decoding" and that can be double-checked at compile time so if the interface changes we aren't caught by surprise.

Sushisource (commented):
Like this new design! Didn't look to much into gritty details but just some comments on high-level stuff.

packages/test/src/test-custom-payload-codec.ts
```diff
@@ -0,0 +1,181 @@
...
+        },
+      },
+    };
+
+    const dataConverter = { payloadCodec: new TestEncodeCodec() };
+    const taskQueue = 'test-workflow-encoded';
+    const worker = await Worker.create({
+      ...defaultOptions,
+      taskQueue,
+      dataConverter, // TODO { payloadCodec: defaultPayloadCodec }
```
bergundy, line 55: leftover TODO?

packages/test/src/test-custom-payload-codec.ts
```diff
@@ -0,0 +1,181 @@
...
+        workflowId: uuid4(),
+        taskQueue,
+      });
+
+      t.is(result, 'encoded');
+      worker.shutdown();
+    };
+    await Promise.all([worker.run(), runAndShutdown()]);
+    t.is(workflowLogs[0], 'encoded');
+    t.is(activityLogs[0], 'Activityencodedencoded');
```
bergundy, line 142: nit: You can probably test this without the logger if you use a single codec for both encoding and decoding.

packages/worker/src/worker.ts
```diff
@@ -307,12 +316,14 @@ export class Worker {
   protected constructor(
     protected readonly nativeWorker: NativeWorkerLike,
     /**
-     * Optional WorkflowCreator - if not provided, Worker will not poll on workflows
+     * Optional WorkflowCreator - if not provided, Worker will not poll on Workflows
      */
     protected readonly workflowCreator: WorkflowCreator | undefined,
-    public readonly options: CompiledWorkerOptions
+    public readonly options: CompiledWorkerOptions,
+    protected readonly dataConverter: LoadedDataConverter
```
bergundy, line 323: I liked that we had a "usable" data converter in the public `CompiledWorkerOptions`, would be nice to get that back

packages/workflow/package.json
bergundy: It should be a non dev dependency if we export any of those types or anything that uses those types (e.g. function signatures), to be safe, let's make it non dev.

packages/workflow/src/worker-interface.ts
```diff
@@ -148,6 +148,15 @@ export async function initRuntime({ info, randomnessSeed, now, patches }: Workfl
     }
   }
 
+  // webpack doesn't know what to bundle given a dynamic import expression, so we can't do:
+  // state.payloadConverter = (await import(payloadConverterPath)).payloadConverter;
+  // @ts-expect-error this is a webpack alias to payloadConverterPath
+  const customPayloadConverter = (await import('__temporal_custom_payload_converter')).payloadConverter;
```
bergundy, line 154: Let's do this before loading interceptors

bergundy (approved):
Didn't have many comments, most of the critical ones I communicated on slack.
I want to try to use TS type checking to ensure that we always encode and decode payloads, but it doesn't block this PR.

I'm approving, this is good to merge as soon as you go over the comments.

packages/worker/src/workflow-codec-runner.ts
lorensr: I considered visiting all fields and duck typing Payloads, but I guess there's a chance of having something else with `{ data, metadata }`, and doing specific fields is faster. I guess the speed difference is negligible and maintainability is more important.

Thanks for the link, will address in next PR.

bergundy: I suggested using the type checker to verify all fields are encoded.
I sent @lorensr some sample code that can handle that.
Can wait for the next PR.

docs/data-converter.md
```diff
@@ -23,7 +23,7 @@ function workflowInclusiveInstanceOf(instance: unknown, type: Function): boolean
 
 ## Decision
 
-Given the possibility of switching or adding other isolation methods in future, we opted to convert to/from Payloads inside the vm (`PayloadConverter`). We also added another transformer layer called `PayloadCodec` that runs outside the vm, can use node modules and Promises, and operates on Payloads. A `DataConverter` is a `PayloadConverter` and a `PayloadCodec`:
+Given the possibility of switching or adding other isolation methods in future, we opted to convert to/from Payloads inside the vm (`PayloadConverter`). We also added another transformer layer called `PayloadCodec` that runs outside the vm, can use node APIs and Promises, and operates on Payloads. A `DataConverter` is a `PayloadConverter` and a `PayloadCodec`:
```
bergundy, line 26: Instead of promises maybe say Node async APIs and give examples, like encryption and compression.

packages/common/src/converter/payload-converter.ts
```diff
@@ -40,23 +39,24 @@ export class CompositePayloadConverter implements PayloadConverter {
...
     for (const converter of this.converters) {
       try {
         const result = converter.toPayload(value);
-        return result;
+        if (result !== undefined) {
+          return result;
+        }
       } catch (e: unknown) {
-        if (e instanceof UnsupportedTypeError) {
+        if (e instanceof PayloadConverterError) {
```
bergundy, line 52: I don't think we need this anymore

packages/common/src/converter/payload-converter.ts
```diff
@@ -75,20 +75,45 @@ export class CompositePayloadConverter implements PayloadConverter {
...
 
+/**
+ * Tries to convert `value` to a {@link Payload}. Throws if conversion fails.
+ *
+ * @throws {@link PayloadConverterError}
+ */
+export function toPayload(converter: PayloadConverter, value: unknown): Payload {
+  const payload = converter.toPayload(value);
+  if (payload === undefined) {
+    throw new PayloadConverterError(`Failed to convert value: ${value}`);
```
bergundy, line 86: Can we make `PayloadConverterError` subclass of `ValueError`?

packages/test/src/mock-native-worker.ts
```diff
@@ -182,7 +177,6 @@ export const defaultOptions: WorkerOptions = {
 };
 
 export async function isolateFreeWorker(options: WorkerOptions = defaultOptions): Promise<Worker> {
```
bergundy, line 179: I think this doesn't need to be async anymore

bergundy (approved):

lorensr:
I'll update `sdk-features`

PR #484, opened 2022-02-17
fix(client): Add missing error handling and interceptors for workflow describe

andreasasprou:
## What was changed

- Added missing wrapper around `handle.describe` to handle raw grpc errors in a similar way to how the other workflow handle functions are wrapped.
- Added the ability to intercept workflow describe calls

## Why?

Instead of `const isWorkflowNotFoundError = error.code === grpcStatus.NOT_FOUND`, I'd like to handle errors like I would if I called `handle.cancel` by comparing against `WorkflowNotFoundError`.

Use case example:

```typescript
export async function isWorkflowRunning(
  handle: WorkflowHandle,
): Promise<boolean> {
  try {
    const result = await handle.describe();

    return (
      result.workflowExecutionInfo?.status ===
      temporal.api.enums.v1.WorkflowExecutionStatus
        .WORKFLOW_EXECUTION_STATUS_RUNNING
    );
  } catch (error) {
    const isWorkflowNotFoundError = error instanceof WorkflowNotFoundError; // <---- Currently not possible

    if (isWorkflowNotFoundError) {
      return false;
    }

    throw error;
  }
}
```


cretz:
Hrmm, the describe operation isn't interceptable in our other SDKs because the use case wasn't really there compared to just wrapping the client/handle in your own class. While I suppose the same could be said for signal/query/etc, describe doesn't really do any data conversion or header injection which is one of the benefits of interceptors.

@bergundy - Thoughts?

packages/client/src/interceptors.ts
```diff
@@ -94,6 +104,10 @@ export interface WorkflowClientCallsInterceptor {
    * Intercept a service call to requestCancelWorkflowExecution
    */
   cancel?: (input: WorkflowCancelInput, next: Next<this, 'cancel'>) => Promise<RequestCancelWorkflowExecutionResponse>;
+  /**
+   * Intercept a service call to describeWorkflowExecution
+   */
+  describe?: (input: WorkflowDescribeInput, next: Next<this, 'cancel'>) => Promise<DescribeWorkflowExecutionResponse>;
```
bergundy, line 110:
```suggestion
  describe?: (input: WorkflowDescribeInput, next: Next<this, 'describe'>) => Promise<DescribeWorkflowExecutionResponse>;
```

packages/client/src/interceptors.ts
```diff
@@ -57,6 +62,11 @@ export interface WorkflowCancelInput {
   readonly firstExecutionRunId?: string;
 }
 
+export interface WorkflowDescribeInput {
```
bergundy, line 65:
```suggestion
/** Input for WorkflowClientCallsInterceptor.describe */
export interface WorkflowDescribeInput {
```

bergundy (commented):
I agree with @cretz that this call doesn't **need** interception.
I don't think there's much harm in adding it though.

packages/client/src/interceptors.ts
lorensr, line 110:
```suggestion
  describe?: (input: WorkflowDescribeInput, next: Next<this, 'describe'>) => Promise<DescribeWorkflowExecutionResponse>;
```

cretz:
> I don't think there's much harm in adding it though.

@bergundy - Is this something we should support in every SDK?

lorensr:
Update from slack: use case is catching all errors that flow through the temporal client & setup alerts around them. I think the only method we're missing from the [client interceptor](https://typescript.temporal.io/api/interfaces/client.workflowclientcallsinterceptor/) besides `.describe()` is `.result()`? Would we like to add that as well?

bergundy:
> Is this something we should support in every SDK?

No harm in doing so and it's trivial so why not?

bergundy:
> I think the only method we're missing from the [client interceptor](https://typescript.temporal.io/api/interfaces/client.workflowclientcallsinterceptor/) besides .describe() is .result()? Would we like to add that as well?

The original purpose of these interceptors AFAIK is to be able to modify proto headers (not to be confused with HTTP headers), if we're adding interceptors for `describe` and `result` we're not add that much value, technically users could subclass the client and wrap the methods.

That being said, there's no harm in adding them so I'm fine with it.

cretz:
> No harm in doing so and it's trivial so why not?

Agreed, just needs to be prioritized and such. I will make sure to include in Python SDK.

lorensr:
@andreasasprou Would you please sign our CLA? I'd be happy to add the other interceptors to this PR myself and then merge!

https://cla-assistant.io/temporalio/sdk-typescript?pullRequest=484

andreasasprou:
> @andreasasprou Would you please sign our CLA? I'd be happy to add the other interceptors to this PR myself and then merge!
> 
> 
> 
> https://cla-assistant.io/temporalio/sdk-typescript?pullRequest=484

Done!

packages/client/src/workflow-client.ts
```diff
@@ -747,12 +781,10 @@ export class WorkflowClient {
...
-        return this.client.service.describeWorkflowExecution({
-          namespace,
-          execution: {
-            runId,
-            workflowId,
-          },
+        const next = this.client._describeWorkflowHandler.bind(this.client);
+        const fn = interceptors.length ? composeInterceptors(interceptors, 'describe', next) : next;
+        return await fn({
+          workflowExecution: { workflowId, runId },
```
lorensr, line 787: I took out `firstExecutionRunId` because it isn't used in the grpc call

packages/test/src/test-interceptors.ts
```diff
@@ -83,6 +96,13 @@ if (RUN_INTEGRATION_TESTS) {
               const result: string = (await next(input)) as any;
               return [...result].reverse().join('');
             },
+            async result<T extends Workflow>(
+              input: WorkflowResultInput,
+              next: Next<WorkflowClientCallsInterceptor, 'result'>
+            ) {
+              const result = await next(input);
+              return (result + '-intercepted') as WorkflowResultType<T>;
+            },
```
lorensr, line 105: Is there a way to simplify the typing?

packages/client/src/workflow-client.ts
bergundy: It will later but we can leave it out for now

bergundy (approved):
We'd mentioned wrapping describe output in a friendly interface.
Not blocking this PR though

lorensr:
Yep I branched off of this, but then couldn't submit a PR with the remote as a base

PR #487, opened 2022-02-17
fix(workflow): Fix ContinueAsNew error message and name

### LINKED ISSUE #397: [Bug] ContinueAsNew shouldn't occur as error in Otel traces
It’s quite irritating to see this error in my workflow traces

CleanShot 2021-11-22 at 10 03 32

bergundy:
Also do not set an error on the otel span when the interceptor catches `ContinueAsNew`.

Closes #397

### DIFF
--- a/packages/interceptors-opentelemetry/src/workflow/index.ts
+++ b/packages/interceptors-opentelemetry/src/workflow/index.ts
@@ -10,6 +10,7 @@ import {
   WorkflowExecuteInput,
   workflowInfo,
   ContinueAsNewInput,
+  ContinueAsNew,
   WorkflowInternalsInterceptor,
 } from '@temporalio/workflow';
 import { extractContextFromHeaders, headersWithContext } from '@temporalio/common/lib/otel';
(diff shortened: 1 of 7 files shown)


packages/interceptors-opentelemetry/src/workflow/index.ts
```diff
@@ -52,7 +53,13 @@ export class OpenTelemetryInboundInterceptor implements WorkflowInboundCallsInte
...
     const context = await extractContextFromHeaders(input.headers);
     const spanName = `${SpanName.WORKFLOW_EXECUTE}${SPAN_DELIMITER}${workflowInfo().workflowType}`;
-    return await instrument(this.tracer, spanName, () => next(input), context);
+    return await instrument(
+      this.tracer,
+      spanName,
+      () => next(input),
+      context,
+      (err) => err instanceof ContinueAsNew
+    );
```
lorensr, line 62: Thoughts on functions taking a single object param, versus list of params? I prefer object for readability at call site when it's not super simple to infer the meaning.

Simple: 

```ts
client.getHandle(tripId);
```

("Okay, if getHandle takes an ID, it's probably a workflow ID")

Prefer object:

```ts
new DefaultDataConverter(root) // "What is root?"
new DefaultDataConverter({ protobufRoot: root }) // "Okay it's something for protobufs"

instrument(
  this.tracer,
  spanName,
  () => next(input),
  context,
  (err) => err instanceof ContinueAsNew
) // "What are all these args for? What does returning true from the last callback mean?"
instrument(
  tracer: this.tracer,
  name: spanName,
  fn: () => next(input),
  context,
  acceptableErrors: (err) => err instanceof ContinueAsNew
) // I have a better idea
```

lorensr (approved):

packages/interceptors-opentelemetry/src/workflow/index.ts
bergundy: I generally agree with you.
In this case I just put this together quickly and didn't put much thought into this.

Because it's mostly internal I don't feel like it's critical.

PR #492, opened 2022-02-21
Resolve ts-loader

jameslnewell:
so its found in a yarn workspaces monorepo

## What was changed
Wrapped `ts-loader` in a call to `require.resolve()`

## Why?
So its correctly located in a `yarn workspaces` monorepo

## Checklist

1. Closes 

2. How was this tested:
I edited the published source code to make this change and re-ran the sample app I was working in.

3. Any docs updates needed?
N/A

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -112,7 +112,7 @@ export class WorkflowCodeBundler {
         rules: [
           {
             test: /\.ts$/,
-            loader: 'ts-loader',
+            loader: require.resolve('ts-loader'),
             exclude: /node_modules/,
           },
         ],


lorensr:
Thanks for the contribution! It appears as though the recommended way to get [`loader`](https://webpack.js.org/configuration/module/#useentry) to resolve correctly is by setting [`context`](https://webpack.js.org/configuration/entry-context/#context), like this to get from `worker/src/workflow/` to `worker/`:

```ts
context: path.resolve(__dirname, '../..')
```

Would that solve your issue?

jameslnewell:
Yeah that works too

lorensr (approved):

packages/worker/src/workflow/bundler.ts
```diff
@@ -117,6 +117,7 @@ export class WorkflowCodeBundler {
           },
         ],
       },
+      context: path.resolve(__dirname, '../..'),
```
bergundy, line 120: This should usually be the directory where the app is located, not our SDK.
Can we `require.resolve` the `ts-loader` instead?

bergundy: @lorensr WDYT?

jameslnewell: @bergundy that's [what I had originally](https://github.com/temporalio/sdk-typescript/pull/492/commits/22ebb5c03a281c62b636cc34e02693a1906512d5) 😆

lorensr: Ah. Yeah, I guess that's better than having user provide their root dir to put for context. Sorry James 😄

bergundy:
Thanks @jameslnewell!

Issue #421, opened 2021-12-06
[Feature Request] pnpm support


mengyazhu96:
### Is your feature request related to a problem? Please describe.
When using pnpm with the Typescript SDK, the Temporal worker's webpack building step fails. As a workaround, we had to add the `shamefully-hoist=true` option so that pnpm reverts to npm/yarn style module locations, e.g. putting everything in `./node_modules` instead of in workspace node modules, e.g. `./workspaces/*/node_modules`

### Describe the solution you'd like
Support pnpm. I'm not sure what this entails 😅 

### Additional context
From Slack thread here: https://temporalio.slack.com/archives/C01DKSMU94L/p1638764279342400

cc @bergundy @spikebrehm

spikebrehm:
There may be multiple layers of issues, but the immediate one that surfaces has to do with missing dependencies listed in your monorepo packages' `package.json`. For example, grep for `protobufjs`; you’ll note that it’s listed in `packages/worker/package.json` but is only used in the `packages/workflow` package, and not listed in its `package.json`. probably a mistake during refactoring. That's the first example that errored for us but there may be more issues like this in your monorepo.

Relatedly, this is why we moved from npm -> pnpm — it fails loudly when packages/workspaces haven’t configured dependencies correctly.

flybayer:
@mengyazhu96 I'm unable to get it working even with `shamefully-hoist=true`. Was there anything else you had to do?

```
2021-12-09T18:25:30.281Z [ERROR] ERROR in ../../node_modules/.pnpm/@temporalio+common@0.16.4/node_modules/@temporalio/common/lib/time.js 7:31-46
2021-12-09T18:25:30.281Z [ERROR] Module not found: Error: Can't resolve 'long' in '/Users/b/c/flightcontrol/node_modules/.pnpm/@temporalio+common@0.16.4/node_modules/@temporalio/common/lib'
2021-12-09T18:25:30.281Z [ERROR]  @ ../../node_modules/.pnpm/@temporalio+common@0.16.4/node_modules/@temporalio/common/lib/index.js 24:13-30
2021-12-09T18:25:30.281Z [ERROR]  @ ../../node_modules/.pnpm/@temporalio+workflow@0.16.3/node_modules/@temporalio/workflow/lib/worker-interface.js 9:17-46
2021-12-09T18:25:30.281Z [ERROR]  @ ../../../../../../src/main.js 1:0-68 4:0-19 5:0-18 26:4-7
[...]

PR #497, opened 2022-02-23
fix(workflow-bundler): Enable resolution of modules in Webpack based on Node's regular algorithm

### LINKED ISSUE #489: [Bug] `npm i --production` insufficient
@idoshamun reported in https://github.com/temporalio/documentation/pull/917 that they needed to `npm i` instead of `npm i --production` (or `npm i --only=prod` or `NODE_ENV=production npm i`), which implies that we need to move at least one devDep to deps somewhere. I haven't repro'd yet. Ido—do you have the error message, or remember which package it was in?

mjameswh:
Fix #489

## What was changed
Append the string `node_module` to Webpack's configuration `resolve.modules` (unless it is already included in provided `nodeModulesPaths`).

## Why?
If `resolve.modules` contains only absolute paths, then webpack will only look at those exact directories for requested modules (ie. it will _not_ behave in a way similar to how Node resolve dependencies). See [Webpack's documentation](https://webpack.js.org/configuration/resolve/#resolvemodules) about this.

It so happens that the `resolveNodeModulesPaths(...)` function, used by the workflow bundler to calculate the default value of `resolve.modules` when none has been provided by the user, indeed returns a single absolute path. It can also be expected that, even if a user was to explicitly provide a path using the `nodeModulesPaths`, he would probably not know that he has to include `node_modules` in it.

Hence, the workflow bundler currently only resolve modules that come from that single, determined `node_modules`, but fails to resolve modules from other, related `node_modules`. This may cause various errors when the workflow code is being bundled, including "Module not found" errors or resolving a dependency to the wrong version. This is notably visible when dependencies has been installed using `npm i --production`, but may also happen for many other reasons.

See #489 for a detailed explanation of such a case.

## Checklist
1. Closes 489

2. How was this tested:
Before applying this change, I reached to the hello-world example (see discussion in #489) and installed dependencies using `npm i --production`. Then, I built the example (`npm run build`) and ran it (`node lib/worker.js`). At this point, I would get an error about module `ms` being missing.

After applying this change and building the Temporal's TS SDK, I reached to the hello-world example, and linked `@temporalio/worker` to the one I has built. Then, I built the example and ran it. No more error.

3. Any docs updates needed?
It might no longer be useful to recommend that dependencies of workers not be installed with `npm i --production` (that is, it could be possible to revert [temporalio/documentation#930](https://github.com/temporalio/documentation/pull/930).

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -78,7 +78,7 @@ export class WorkflowCodeBundler {
 
       // Bundle all Workflows and interceptor modules for lazy evaluation
       api.overrideGlobals();
-      api.setImportFuncs({ 
+      api.setImportFuncs({
         importWorkflows: () => {
           return import(/* webpackMode: "eager" */ ${JSON.stringify(this.workflowsPath)});
         },
@@ -105,7 +105,10 @@ export class WorkflowCodeBundler {
   protected async bundle(filesystem: typeof unionfs.ufs, entry: string, distDir: string): Promise<void> {
     const compiler = webpack({
       resolve: {
-        modules: this.nodeModulesPaths,
+        modules: [
+          ...this.nodeModulesPaths,
+          ...(!this.nodeModulesPaths.includes('node_modules') ? ['node_modules'] : []),
+        ],
         extensions: ['.ts', '.js'],
       },
       module: {


lorensr (approved):
Awesome, thank you! 😃

packages/worker/src/workflow/bundler.ts
```diff
@@ -105,7 +105,10 @@ export class WorkflowCodeBundler {
   protected async bundle(filesystem: typeof unionfs.ufs, entry: string, distDir: string): Promise<void> {
     const compiler = webpack({
       resolve: {
-        modules: this.nodeModulesPaths,
+        modules: [
+          ...this.nodeModulesPaths,
+          ...(!this.nodeModulesPaths.includes('node_modules') ? ['node_modules'] : []),
```
bergundy, line 110: Is this `node_modules` relative to PWD?
What happens if you run this from a different directory?
An ideal solution would take that into account.

bergundy:
@mjameswh thanks for the contribution!
I want to look into a longer term solution where users don't need to specify `nodeModulesPaths` at all (at least in 99% of the cases).

packages/worker/src/workflow/bundler.ts
mjameswh: It is _not_ relative to PWD. You may want to look at [Webpack's doc](https://webpack.js.org/configuration/resolve/#resolvemodules) for an explanation on this.

To be honest, I feel they are mixing two different concepts here. But essentially, adding the string `node_modules` in this array is required for Webpack to resolve modules the same way as Node does it (that is, if file /a/b/c/d.js request module x, then look at /a/b/c/node_modules/x, then /a/b/node_modules/x, , then /a/node_modules/x).

mjameswh:
> I want to look into a longer term solution where users don't need to specify nodeModulesPaths at all (at least in 99% of the cases).

Users don't have to specify `nodeModulesPaths`, and I personally don't see any reason why they would want to specify it. I only kept that because it was there, but I don't feel it is useful.

However, we still have to specify it to webpack because of the synthetic entrypoint that is created by the bundler in `/src/main.js`. That file requires various modules, which webpack doesn't know where to locate. And this indeed logical, as if you were to actually create a file named `/src/main.js` on your local machine, and tried to execute it (ie. `node /src/main.js`), `node` would fail, telling you that it can't locate module `@temporalio/workflow/lib/worker-interface.js`...

An alternative would be to make that synthetic entrypoint file appears as a sibling of the worker file. For example, assuming that the bundler is invoked from a worker named "/a/b/c/myworker.js"`, then the synthetic entrypoint could be named `/a/b/c/myworker-workflows-${RANDOM}.js`. From there, webpack's will be able to naturally resolve all requirements, since they were resolvable from myworker.js. The `RANDOM` suffix is there to avoid conflicting with some existing file at that location.

bergundy:
> An alternative would be to make that synthetic entrypoint file appears as a sibling of the worker file. For example, assuming that the bundler is invoked from a worker named /a/b/c/myworker.js, then the synthetic entrypoint could be named /a/b/c/myworker-workflows-${RANDOM}.js. From there, webpack's will be able to naturally resolve all requirements, since they were resolvable from myworker.js. The RANDOM suffix is there to avoid conflicting with some existing file at that location.

That's where I was going with this, it seems like a better approach

packages/worker/src/workflow/bundler.ts
bergundy: The docs don't specify what it's relative to that's why I assumed PWD. What else would it be? The entry point?

mjameswh: To put it simply, it is relative to every source file that request a module... and to all parent directories of that source file. Well, not exactly simple when I phrase it like this, eh? 🤣 

What that means is that assuming the following configuration:

```
resolve: {
    modules: ['/Users/myself/myproject/MagicDirectory', 'node_modules']
}
```

and assuming that webpack is processing file A, located at `/Users/myself/myproject/src/submodule/a.js`, which imports module `B`, webpack will look for the following locations, in that specific order:

```
/Users/myself/myproject/MagicDirectory/B/package.json
/Users/myself/myproject/src/submodule/node_modules/B/package.json
/Users/myself/myproject/src/node_modules/B/package.json
/Users/myself/myproject/node_modules/B/package.json
/Users/myself/node_modules/B/package.json
/Users/node_modules/B/package.json
/node_modules/B/package.json
```

The search succeeds as soon as it finds an entry that exists.

Now, let's suppose that B was indeed found, at `/Users/myself/myproject/src/submodule/node_modules/B/package.json`. And suppose that the package.json file specify that the entrypoint of that module is at `lib/index.js`, and requires module `C`.

Now, webpack tries to resolve module `C` from the location of module B's index.js. That means that webpack will look for the following locations:

```
/Users/myself/myproject/MagicDirectory/C/package.json
/Users/myself/myproject/src/submodule/node_modules/B/lib/node_modules/C/package.json
/Users/myself/myproject/src/submodule/node_modules/B/node_modules/C/package.json
/Users/myself/myproject/src/submodule/node_modules/node_modules/C/package.json
/Users/myself/myproject/src/submodule/node_modules/C/package.json
/Users/myself/myproject/src/node_modules/C/package.json
/Users/myself/myproject/node_modules/C/package.json
/Users/myself/node_modules/C/package.json
/Users/node_modules/C/package.json
/node_modules/C/package.json
```

Obviously, most of these attempts are ridiculous, but a few are critical, notably:

`/Users/myself/myproject/MagicDirectory/C/package.json`: We explicitly specified the `MagicDirectory`, so there might be some important things there...
`/Users/myself/myproject/src/submodule/node_modules/B/node_modules/C/package.json`: This is where NPM will place package C (assuming it is a requirement of B) if that package has not been hoisted.
`/Users/myself/myproject/node_modules/C/package.json`: This is where NPM will place package C if it has been hoisted (NPM tends to hoist whenever it is possible)

mjameswh:
> That's where I was going with this, it seems like a better approach

I absolutely agree. I had initially dismissed this option as too intrusive, but if you are willing to go there, then I will have a look at this tomorrow.

packages/worker/src/workflow/bundler.ts
bergundy: Got it, this makes sense now and could actually explain the issues I was seeing with pnpm and hoisting (#421).
If this is the behavior then the fix is pretty good, and I'd go as far as removing the [default `nodeModulesPaths` we inject in the worker](https://github.com/temporalio/sdk-typescript/blob/7ad8b1a74b5217613d2bfdd7a716b7e2e61bf1bc/packages/worker/src/worker-options.ts#L316).

Seems like putting the entrypoint next to `workflowsPath` removes the need to specify `resolve.modules` altogether so I'd say it's a preferred approach.

bergundy:
> > That's where I was going with this, it seems like a better approach
> 
> I absolutely agree. I had initially dismissed this option as too intrusive, but if you are willing to go there, then I will have a look at this tomorrow.

If you want to pick that up it'd be great! We definitely appreciate the contribution.
Otherwise we can complete it.

mjameswh:
@bergundy I submitted a distinct PR for the alternative approach. Please see #498.

bergundy:
Closing this as we've already merged its successor #498

PR #498, opened 2022-02-24
fix(workflow-bundler): Simplify resolution of modules by Webpack in workflow bundler

### LINKED ISSUE #489: [Bug] `npm i --production` insufficient

mjameswh:
Fix #489 

Note: This is a second take at this issue, distinct from #497. It takes a different route, completely removing the `nodeModulesPaths` parameter. See #497 for the discussion that led to this new proposition.

## What was changed
Modified the location of the synthetic entrypoint file created by the workflow bundler, so that the file is a sibling of the workflows directory/file. It was previously hardcoded to be `/src/main.js`. Also, removed the `nodeModulesPaths` which is no longer required since webpack can now correctly resolve dependencies.

## Why?
As it was previously written, the workflow bundler only resolved modules that came from a single, determined `node_modules`, but fails to resolve modules from other, related `node_modules` (that is, it was not following the module resolution algorithm that node's follow and on which NPM and others relies). This may cause various errors when the workflow code is being bundled, including "Module not found" errors or resolving a dependency to the wrong version. This is notably visible when dependencies has been installed using `npm i --production`, but may also happen for many other reasons. #489 demonstrate such a case.

The root issue for this is that the workflow bundler needs to create a synthetic entrypoint file, which setup a worker that runs inside the isolated execution environment. That file, created in a virtual filesystem under the path `/src/main.js`, would need to load some modules (notably `@temporalio/workflow`, user defined workflows and interceptor modules). However, given the location of that file, webpack's module resolution would fail to locate some of these dependencies, since no `node_modules` would exist in either `/src` nor `/`. This had previously led to the addition of the `nodeModulesPaths` parameter, but that solution was both incomplete (as discussed in #497) and difficult to understand for users.

This PR modify the location of the entrypoint file so that it is a sibling of the workflows directory/file. It is expected that the workflows directory is inside a nodejs package that already have access to all applicable dependencies. As before, that entrypoint file is created in a union/memfs, so that the user is actually never exposed to the existence of that file. More over, the created entrypoint file includes a large random suffix, to avoid that this virtual file could ever shadow an actual user file.

## Checklist
1. Closes 489. It may also help toward PNPM and Yarn 2 compatibility (see #421), though further investigation would be required on these matters.

2. How was this tested:
Ran and succeeded all existing unit/integration tests (except for tests that were removed by this PR).

Also, before applying this change, I reached to the hello-world example (see discussion in #489) and installed dependencies using `npm i --production`. Then, I built the example (`npm run build`) and ran it (`node lib/worker.js`). At this point, I would get an error about module `ms` being missing.

After applying this change and building the Temporal's TS SDK, I reached to the hello-world example, and linked `@temporalio/worker` to the one I has built. Then, I built the example and ran it. No more error.

3. Any docs updates needed?
It might no longer be useful to recommend that dependencies of workers not be installed with `npm i --production` (that is, it could be possible to revert [temporalio/documentation#930](https://github.com/temporalio/documentation/pull/930).


bergundy (approved):
This looks great to me.
Thanks a lot for the contribution!

I'd like to test this throughly before releasing though.

mjameswh:
> I'd like to test this throughly before releasing though.

It makes sense. I feel confident about this change, but there's a lot of quirks to how modules resolution works. Having more people testing it will definitely help.

bergundy:
I'm merging this.
@lorensr make sure we don't release (with latest tag) before we throughly test it.

PR #507, opened 2022-03-02
docs: Link to source proto; improve Publishing

lorensr:


packages/proto/lib/index.d.ts
```diff
@@ -20,8 +22,13 @@
  * import { temporal } from '@temporalio/proto';
  * const retryPolicy: temporal.api.common.v1.IRetryPolicy = { ... };
  * ```
+ * 
+ * The source protos are in the [api repo](https://github.com/temporalio/api/), for example [`RetryPolicy` in `temporal/api/common/v1/message.proto`](https://github.com/temporalio/api/blob/eff34dc3a970cdbf3a10a7be827edf1b1528b125/temporal/api/common/v1/message.proto#L96).
```
bergundy, line 26: Some of the protos are there and some in sdk-core, we take the API protos from a subtree in Core

bergundy (commented):
Just one comment about the modified docstring

packages/proto/lib/index.d.ts
lorensr: Some of the `temporal` protos are in sdk-core?

bergundy: They're in a git subtree there, the original is in the API repo

lorensr: Updated to use the sdk-core repo version of temporalio/api.

lorensr: The sdk-core protos I link to earlier in this file

bergundy (approved):

PR #509, opened 2022-03-02
fix: Re-export possibly-shared-use things in common

lorensr:


bergundy (approved):

lorensr:
Don't think `sdk-features-tests` should still have `Error: ../harness/ts/harness.ts(2,29): error TS2305: Module '"@temporalio/common"' has no exported member 'Workflow'.` Will look into after this release

PR #511, opened 2022-03-04
fix: Use default data converter for search attributes

lorensr:
Anything else that the default data converter should be used on?


bergundy (approved):
LGTM

cretz (approved):

cretz:
Same, search attributes is all I can think of.

PR #513, opened 2022-03-05
feat(client): Add methods to set grpc metadata and deadline for service calls

bergundy:


packages/client/src/connection.ts
```diff
@@ -63,6 +64,48 @@ export function defaultConnectionOpts(): ConnectionOptionsWithDefaults {
...
+
+/**
+ * Set metadata for any service requests executed in `fn`'s scope.
+ */
+export async function withMetadata<R>(metadata: Metadata, fn: () => Promise<R>): Promise<R> {
+  return await callContextStorage.run({ metadata }, fn);
+}
+
+/**
+ * Set the {@link CallContext} for any service requests executed in `fn`'s scope.
```
lorensr, line 103:
```suggestion
 * Set the {@link CallContext} for any service requests executed in `fn`'s scope.
 *
 * @returns The return value of `fn`
```

lorensr (approved):
Nice solution! API-wise, would prefer an added options object param, but we're limited by the generated code.

cretz (approved):

PR #515, opened 2022-03-07
chore: Update otel packages

lorensr:
### before

mismatch between versions at root, test, and `@temporalio/interceptors-opentelemetry`:

Pasted Graphic

### after

updated deps to match `@temporalio/interceptors-opentelemetry` and ran `npm prune`

Pasted Graphic 1


package.json
```diff
@@ -72,12 +72,14 @@
   "files": [],
   "name": "typescript-sdk",
   "dependencies": {
+    "@opentelemetry/core": "^1.0.1",
+    "@opentelemetry/semantic-conventions": "^1.0.1",
```
bergundy, line 76: nit: Seems like we _usually_ put 3rd in `devDependencies` let's keep it consistent.
I think `arg` and `uuid` are used in scripts but even those can be put in dev.

packages/test/package.json
```diff
@@ -25,7 +25,10 @@
   "author": "Roey Berman <roey@temporal.io>",
   "license": "MIT",
   "dependencies": {
-    "@opentelemetry/exporter-collector-grpc": "^0.25.0",
+    "@opentelemetry/core": "^1.0.1",
```
bergundy, line 28: No need to duplicate the dependencies that are already specified in the root package.

bergundy (approved):
Overall LGTM, nothing major. Thanks!

packages/test/src/load/worker.ts
```diff
@@ -1,11 +1,11 @@
-import arg from 'arg';
+import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-grpc';
 import * as opentelemetry from '@opentelemetry/sdk-node';
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions';
-import { Core, Worker, DefaultLogger } from '@temporalio/worker';
-import { CollectorTraceExporter } from '@opentelemetry/exporter-collector-grpc';
```
cretz, line 5: How much are these APIs moving? Or are we just switching to a different one? (otel feels as unstable as it did two years ago)

lorensr: This was a breaking change. There were a number of them in the last 2 versions. 

https://github.com/open-telemetry/opentelemetry-js/blob/main/CHANGELOG.md#101--experimental-0270

I didn't change `packages/*` versions, so don't think should be breaking for our users.

cretz (approved):

PR #518, opened 2022-03-08
fix: Update TLSConfig import location

lorensr:


packages/core-bridge/index.d.ts
```diff
@@ -1,5 +1,5 @@
-import { TLSConfig } from '@temporalio/common';
 import { SpanContext } from '@opentelemetry/api';
+import { TLSConfig } from '@temporalio/internal-non-workflow-common';
```
bergundy, line 2: You should update `package.json` to depend directly on `internal-non-...`

bergundy (approved):
Please also regenerate package-lock.json 😬

PR #519, opened 2022-03-08
feat(workflow-bundler): support excluding modules from workflows bundle

### LINKED ISSUE #516: [Bug] bundleWorkflowCode fails on "heavyweight" dependencies
### Describe the bug

We have a workflow that depends on some interfaces/enums/basic utility functions in another module, which in turn depends on some "heavyweight" external package. That external package could actually be any package that depends on node's internal packages (such as filesystem, dns, buffer, url...), or any package that contains native code. When the workflow code gets bundled (either implicitly in `Worker.create(...)` or explicitly by calling `bundleWorkflowCode(...)`), **Webpack is sometime unable to determine that the external package is not actually required at runtime, and will therefore try to load it, causing a build failure.**

The error message looks something like the following (truncated to keep it readable; I actually get 46 errors). Note that in this case, the "external" package is `got` (an HTTP client, similar to `request`) and problematic dependencies are `process`, `util`, `url`, `dns` and others, which are all packages provided internally by node.

```
[ERROR] modules with errors 390 bytes [errors]
[ERROR]   node:process 39 bytes [built] [code generated] [1 error]
[ERROR]   node:util 39 bytes [built] [code generated] [1 error]
[ERROR]   node:url 39 bytes [built] [code generated] [1 error]
[ERROR]   + 7 modules
        ...
[ERROR] ERROR in ./node_modules/cacheable-lookup/source/index.js 10:4-18
[ERROR] Module not found: Error: Can't resolve 'dns' in '/Users/jwatkins/Development/temporal/bug-heavyweight-dependency-bundling/node_modules/cacheable-lookup/source'
[ERROR] resolve 'dns' in '/Users/jwatkins/Development/temporal/bug-heavyweight-dependency-bundling/node_modules/cacheable-lookup/source'
        ...
[ERROR]   Parsed request is a module
[ERROR]   using description file: /Users/jwatkins/Development/temporal/bug-heavyweight-dependency-bundling/node_modules/cacheable-lookup/package.json (relative path: ./source)
        ...
[ERROR]  @ ./node_modules/got/dist/source/core/options.js 10:0-47 28:25-40
[ERROR]  @ ./node_modules/got/dist/source/index.js 2:0-40 4:17-24 11:0-55 11:0-55 12:0-34 12:0-34
[ERROR]  @ ./src/api/client.ts 7:30-44
[ERROR]  @ ./src/api/index.ts 17:13-32
[ERROR]  @ ./src/temporal/myWorkflow.ts 4:16-39
[ERROR]  @ ./src/temporal/myWorkflow-entrypoint-caf2c449-bc91-46c7-9b6b-dc05043a9065.js
```

### Minimal Reproduction

See https://github.com/mjameswh/temporalio-sdk-typescript-issue-heavyweight-dependency-bundling for a workspace that demonstrate this issue.

Note that, in this specific case, the issue can easily be worked around by simply importing from `../api/models` instead of `../api/index` (in myWorkflow.ts). However, our actual use case spans multiple projects in a monorepo structure, and we have not yet found any simple, practical solution to this issue.

### Environment/Versions

- Temporal typescript-sdk version: 0.19.0-rc.1

### Solutions to be considered

- **Add [`target: 'node'`](https://webpack.js.org/configuration/target/) to bundleWorkflowCode's [webpack config](https://github.com/temporalio/sdk-typescript/blob/82351ccc69a5d3876e5b44d4860c20ec71cfca28/packages/worker/src/workflow/bundler.ts#L117).** Without this option, Webpack will simply not allow importing node's internal packages. In fact, by default, Webpack assume that you are bundling for a browser environment, which could cause some weird issues at some point in the future. Still, given that most of node's internal packages are unusable in isolated vm context, it is not clear if allowing these imports would actually help or just push issues a little bit father. It also does not resolve similar issues caused by packages that contains native code.

- **Allow users to specify a list of dependencies to be ignored.** For example, an optional property `ignoreModules` could be added to [`BundleOptions`](https://github.com/temporalio/sdk-typescript/blob/82351ccc69a5d3876e5b44d4860c20ec71cfca28/packages/worker/src/workflow/bundler.ts#L189). That list would be used to dynamically add values to Webpack's `resolve.alias` or `resolve.fallback` (ie. `{ "got": false }`). That would solve this specific issue and would be relatively easy to understand for most developers facing similar issues.

- **Allow users to apply more generic overrides to the Webpack config.** For example, an optional property `webpackConfigOverrides` could be added to `BundleOptions`. If it is there, that property would be merged with the default/calculated Webpack configuration. Alternatively, it could be an override function, which receive the default/calculated Webpack configuration, and returns the modified configuration (this is much more flexible, and easier from our point of view, more harder to work with for developers). This is for example [how Gatsby works](https://www.gatsbyjs.com/docs/how-to/custom-configuration/add-custom-webpack-config/). This approach would not only solve the current issue, but also provide a way to deal with many other very custom needs that developers may have (think of.. _Hey guys! We would like to use [`define`](https://webpack.js.org/plugins/define-plugin/) for our workflows_ 😄 ) However, this would be more complex for developers to use, and arguably, there are not that many things that can be customized in Temporal's Webpack build process without risking breaking things.

mjameswh:
## What was changed

A property `ignoreModules` has been added to `bundleWorkflowCode(...)` argument (that is, inside `BundleOptions`). This property allow user to specify a list of modules to be excluded from the workflows bundle.

Also, all node built-in modules (except for `assert`, for which a stub has already been developed) are automatically ignored. A warning message is displayed at bundling time to inform users that these modules can't be used inside workflows.

## Why?

See #516 for rationale.

## Checklist
1. Closes #516

2. How was this tested:
Unit tests have been added

3. Any docs updates needed?
Yes, to be discussed.

### DIFF
--- a/packages/test/src/test-worker-from-bundle.ts
+++ b/packages/test/src/test-worker-from-bundle.ts
@@ -11,6 +11,7 @@ import { WorkflowClient } from '@temporalio/client';
 import { bundleWorkflowCode, Worker } from '@temporalio/worker';
 import { RUN_INTEGRATION_TESTS } from './helpers';
 import { successString } from './workflows';
+import { issue516 } from './mocks/workflows-with-node-dependencies/issue-516';
 
 if (RUN_INTEGRATION_TESTS) {
   test('Worker can be created from bundle code', async (t) => {
@@ -63,6 +64,30 @@ if (RUN_INTEGRATION_TESTS) {
     } finally {
       unlink(path);
     }
+    t.pass();
+  });
+
+  test('Workflow bundle can be created from code using ignoreModules', async (t) => {
+    const taskQueue = `${t.title}-${uuid4()}`;
+    const workflowBundle = await bundleWorkflowCode({
+      workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
+      // ignoreModules: ['dns'],
+    });
+    const worker = await Worker.create({
+      taskQueue,
+      workflowBundle,
+    });
+    const client = new WorkflowClient();
+    await Promise.all([
+      worker.run(),
+      (async () => {
+        try {
+          await client.execute(issue516, { taskQueue, workflowId: uuid4() });
+        } finally {
+          worker.shutdown();
+        }
+      })(),
+    ]);
     t.pass();
   });
 }
--- a/packages/test/src/test-worker-from-bundle.ts
+++ b/packages/test/src/test-worker-from-bundle.ts
@@ -8,7 +8,7 @@ import { unlink, writeFile } from 'fs/promises';
 import os from 'os';
 import { v4 as uuid4 } from 'uuid';
 import { WorkflowClient } from '@temporalio/client';
-import { bundleWorkflowCode, Worker } from '@temporalio/worker';
+import { bundleWorkflowCode, DefaultLogger, LogEntry, Worker } from '@temporalio/worker';
 import { RUN_INTEGRATION_TESTS } from './helpers';
 import { successString } from './workflows';
 import { issue516 } from './mocks/workflows-with-node-dependencies/issue-516';
@@ -71,7 +71,7 @@ if (RUN_INTEGRATION_TESTS) {
     const taskQueue = `${t.title}-${uuid4()}`;
     const workflowBundle = await bundleWorkflowCode({
       workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
-      // ignoreModules: ['dns'],
+      ignoreModules: ['dns'],
     });
     const worker = await Worker.create({
       taskQueue,
@@ -88,6 +88,25 @@ if (RUN_INTEGRATION_TESTS) {
         }
       })(),
     ]);
+    t.pass();
+  });
+
+  test('A warning is reported when workflow depends on a node built-in module', async (t) => {
+    const logs: LogEntry[] = [];
+    const logger = new DefaultLogger('WARN', (entry: LogEntry) => {
+      logs.push(entry);
+      console.warn(entry.message);
+    });
+
+    await bundleWorkflowCode({
+      workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
+      logger,
+    });
+
+    t.true(
+      logs.some((entry) => entry.message.match(/'dns'/) && entry.message.match(/'BundleOptions.ignoreModules'/)),
+      "Bundler reported a warning message about package 'dns'"
+    );
     t.pass();
   });
 }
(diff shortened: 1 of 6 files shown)


bergundy:
@mjameswh thanks again for an awesome contribution.
Let me think about this a bit. I'll reply early next week.

bergundy:
@mjameswh do you still think this is the right approach after posting [this comment](https://github.com/temporalio/sdk-typescript/issues/516#issuecomment-1062561402)?

mjameswh:
> do you still think this is the right approach after posting 

Yes. What's included in this PR reflects my _final_ understanding and I still think it is an appropriate change.

packages/worker/src/workflow/bundler.ts
```diff
@@ -8,18 +8,74 @@ import { v4 as uuid4 } from 'uuid';
 import webpack from 'webpack';
 import { DefaultLogger, Logger } from '../logger';
 
+const nodejsBuiltinModules: string[] = [
+  // assert, // A remplacement module is injected through module-overrides
```
bergundy, line 12:
```suggestion
  // assert, // A replacement module is injected through module-overrides
```

packages/test/src/test-worker-from-bundle.ts
```diff
@@ -65,4 +66,47 @@ if (RUN_INTEGRATION_TESTS) {
...
+    await bundleWorkflowCode({
+      workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
+      logger,
+    });
+
+    t.true(
+      logs.some((entry) => entry.message.match(/'dns'/) && entry.message.match(/'BundleOptions.ignoreModules'/)),
+      "Bundler reported a warning message about package 'dns'"
+    );
+    t.pass();
```
bergundy, line 110: No need to `pass` here since this test has an assertion.

packages/worker/src/workflow/bundler.ts
```diff
@@ -206,6 +293,10 @@ export interface BundleOptions {
    * `payloadConverter` should be an instance of a class that extends {@link DataConverter}.
    */
   payloadConverterPath?: string;
+  /**
+   * List of modules to be excluded from the workflows bundle.
```
bergundy, line 297: I'd elaborate here and say:
```
Use this option when your workflow code references an import that cannot be used in isolation, e.g. a Node.js built-in module.
Modules listed here **MUST** not be used at runtime.

> NOTE: This is an advanced option that should be used with care.
```

bergundy (approved):
I had a few small comments, other than that, LGTM.
I think a lot of users will appreciate this improvement.

Other than that, I would add the `ignoreModules` option to `WorkerOptions`, perhaps under a new interface (e.g. `WorkerOptions.bundlerOptions.ignoreModules`).

bergundy:
@mjameswh thanks so much!
Do you want to address my comments or should I just fix and merge?

bergundy:
I've merged the alternative PR, thanks again @mjameswh!

PR #520, opened 2022-03-09
Get static checking of encoding Failures and Completions

lorensr:
In progress:

- Statically check Failure encoding
- Statically check Completion encoding (trading slightly less parallelization)

Todo:

- Failure and Activation decoding

Any idea why I'm getting this error?

image

Ends in:

```
Types of property 'headerFields' are incompatible.
  Type '{ [k: string]: IPayload; } | null | undefined' is not assignable to type 'EncodedPayload | null | undefined'.
    Property 'encoded' is missing in type '{ [k: string]: IPayload; }' but required in type 'EncodedPayload'.ts(2322)
  types.ts(8, 3): 'encoded' is declared here.
```

Seems like `ReplaceNested<Completion, Payload, EncodedPayload>` results in `headerFields: { [k: string]: IPayload; } | null | undefined` being replaced with `EncodedPayload | null | undefined`? I failed at reproducing here:

https://www.typescriptlang.org/play?#code/C4TwDgpgBAShYBsCGBjCA5CBnYEAmAPACoA0URA9nIqhGdcmgOoCWwAFgHxQC85UEAB64AdnixQAFADpZSAE4BzLAC4oSESADaAXQCUvbhpABYAFBQoAfnLnLaogOEQxEyg1p3rseIwisOKAAfKABRQRQEAFc8CGIyd19aTi81AG8vS0stAAUoFhEoAGsIEAoAM3IdNQ80TBx8YlydBKoktHp2-zYuAG5MqABffrNzUEgfGjRCIm4+WoxsXBmyESiAWwAjCHkyHHkCxRTRs3HoAEkRYAk+DIss8ooKdIGsqEF0qC0itX3D6qgay2OyGJFeWRAaiB23kA0G5nhJxQFBEOCgFE2ACsAEw1LqES7XOZQO5vR7PEngywfEkALzUAHI-iJFAzBmD7m9LJCoAzNiAIAy4QigA

### DIFF
--- a/packages/common/src/converter/types.ts
+++ b/packages/common/src/converter/types.ts
@@ -7,6 +7,10 @@ export interface EncodedPayload extends Payload {
   encoded: true;
 }
 
+export interface DecodedPayload extends Payload {
+  decoded: true;
+}
+
 /** An object T with any nested values of type ToReplace replaced with ReplaceWith */
 export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]) => any
   ? T
@@ -22,6 +26,9 @@ export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]
 
 /** Replace `Payload`s with `EncodedPayload`s */
 export type Encoded<T> = ReplaceNested<T, Payload, EncodedPayload>;
+
+/** Replace `Payload`s with `DecodedPayload`s */
+export type Decoded<T> = ReplaceNested<T, Payload, DecodedPayload>;
 
 /**
  * Transform an *ascii* string into a Uint8Array
--- a/packages/common/src/converter/types.ts
+++ b/packages/common/src/converter/types.ts
@@ -1,7 +1,15 @@
+import { ReplaceNested } from '@temporalio/internal-workflow-common';
 import type * as iface from '@temporalio/proto/lib/coresdk';
-import { TextEncoder, TextDecoder } from './encoding';
+import { TextDecoder, TextEncoder } from './encoding';
 
 export type Payload = iface.coresdk.common.IPayload;
+
+export interface EncodedPayload extends Payload {
+  encoded: true;
+}
+
+/** Replace `Payload`s with `EncodedPayload`s */
+export type Encoded<T> = ReplaceNested<T, Payload, EncodedPayload>;
 
 /**
  * Transform an *ascii* string into a Uint8Array
--- a/packages/common/src/converter/types.ts
+++ b/packages/common/src/converter/types.ts
@@ -3,35 +3,6 @@ import { TextDecoder, TextEncoder } from './encoding';
 
 export type Payload = coresdk.common.IPayload;
 
-export interface EncodedPayload extends Payload {
-  encoded: true;
-}
-
-export interface DecodedPayload extends Payload {
-  decoded: true;
-}
-
-/** An object T with any nested values of type ToReplace replaced with ReplaceWith */
-export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]) => any
-  ? T
-  : [keyof T] extends [never]
-  ? T
-  : T extends { [k: string]: coresdk.common.IPayload } | null
-  ? {
-      [P in keyof T]: ReplaceNested<T[P], ToReplace, ReplaceWith>;
-    }
-  : T extends ToReplace
-  ? ReplaceWith | Exclude<T, ToReplace>
-  : {
-      [P in keyof T]: ReplaceNested<T[P], ToReplace, ReplaceWith>;
-    };
-
-/** Replace `Payload`s with `EncodedPayload`s */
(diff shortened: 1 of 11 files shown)


lorensr:
Wound up using:

```ts
export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]) => any
  ? T
  : T extends { [k: string]: coresdk.common.IPayload } | null
  ? {
      [P in keyof T]: ReplaceNested<T[P], ToReplace, ReplaceWith>;
    }
  : T extends ToReplace
  ? ReplaceWith | Exclude<T, ToReplace>
  : {
      [P in keyof T]: ReplaceNested<T[P], ToReplace, ReplaceWith>;
    };
```

@Roey in the version that worked for us, the whole object matched `Record<string, any>`, so nothing got replaced.

For posterity, here's my playground. `A extends B` is `true` iff A can be assigned to B.

```ts
const _a: Record<string, any> = { foo: 2 };
type _Header = { [k: string]: coresdk.common.IPayload } | null;
let _h: _Header;
_h = null;
_h = { foo: {} };
const _p: Payload = _h; // Header extends Payload
const _record: Record<string, Payload> = { foo: {} };
const _p2: Payload = _record; // Record<string, Payload> extends Payload
const _record2: Record<string, Payload> = _h; // Header extends Record<string, Payload>
const _record3: Record<string, Payload> = { test: {} }; // { test: {} } extends Record<string, Payload>
```

Because Header extends Payload, in the old `ReplaceNested`:


```ts
export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]) => any
  ? T
  : T extends ToReplace
  ? ReplaceWith | Exclude<T, ToReplace>
  : {
      [P in keyof T]: ReplaceNested<T[P], ToReplace, ReplaceWith>;
    };
```

in `ReplaceNested<Completion, Payload, EncodedPayload>`, `T extends ToReplace` was true, so:

```ts
headerFields?: ({ [k: string]: coresdk.common.IPayload }|null);
```

was replaced with

```ts
headerFields?: (EncodedPayload|null);
```

That begs the question: what other false positives are there? Since Payload is an interface with two optional properties, `EmptyInterface extends Payload`, which is why I had to add:

```ts
cancelWorkflowExecution: command.cancelWorkflowExecution ? { encoded: true } : undefined,
```

`cancelWorkflowExecution` is not of type `Payload`, it's of type:

```ts
      /** Properties of a CancelWorkflowExecution. */
      interface ICancelWorkflowExecution {
      }
```

It would be nice if we could match against an empty interface before we get to `T extends ToReplace`, but we can't do `T extends EmptyInterface` because that's true for ~everything.

image

image

and `IConcelWorkflowExecution extends Record<string, never>` is false.

If TS has a condition like `T isA EmptyInterface`, that would work, but I can't find anything. @JoshuaKGoldberg  are there condition operators besides `extends`?

image

JoshuaKGoldberg:
Hi! Conditional types with `extends` are generally the only way to add logical branches to the type system.

If you want to add a logic branch for the case of a type being the empty object `{}`, you could use a utility type like:

```ts
type IsEmptyType<T> = [keyof T] extends [never] ? true : false;
```

Reasoning: `keyof {}` is `never`, but because `never` is the empty set in type theory, `never extends never` is false. `[never] extends [never]` is true though.

lorensr:
Thanks a lot Josh, that worked! 😄 [9ed07c1](https://github.com/temporalio/sdk-typescript/pull/520/commits/9ed07c10366bfbe2da01f493f0136d60e414808b)

packages/common/src/converter/types.ts
```diff
@@ -1,7 +1,36 @@
...
+export interface DecodedPayload extends Payload {
+  decoded: true;
+}
+
+/** An object T with any nested values of type ToReplace replaced with ReplaceWith */
+export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]) => any
+  ? T
+  : [keyof T] extends [never]
+  ? T
+  : T extends { [k: string]: coresdk.common.IPayload } | null
```
bergundy, line 19: Do we need `| null`?

If we do then I'd do this:
```ts
  : T extends { [k: string]: ToReplace } | null
```

Otherwise I'd just replace with:
```ts
  : T extends { [k: string]: ToReplace }
```

packages/common/src/converter/types.ts
```diff
@@ -1,7 +1,36 @@
...
+export interface EncodedPayload extends Payload {
+  encoded: true;
+}
+
+export interface DecodedPayload extends Payload {
+  decoded: true;
+}
+
+/** An object T with any nested values of type ToReplace replaced with ReplaceWith */
+export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]) => any
```
bergundy, line 15: Make sure these aren't exported in the index

bergundy (approved):
Nice!
I had only 2 comments

PR #522, opened 2022-03-10
fix(worker)!: Mark activity errors retryable

### LINKED ISSUE #521: Activity not found and activity arg conversion errors should be retryable.
### Describe the bug

Marked as non-retryable at https://github.com/temporalio/sdk-typescript/blob/v0.18.0/packages/worker/src/worker.ts#L474 and https://github.com/temporalio/sdk-typescript/blob/v0.18.0/packages/worker/src/worker.ts#L496 but after checking our other SDKs, this should be a retryable error.

lorensr:
Before this fix, the ApplicationFailures returned by the Worker in these
two circumstances were not retryable:

- Activity not found
- Failed to parse Activity args

Closes #521

### DIFF
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -79,7 +79,7 @@ export class ServerFailure extends TemporalFailure {
  * Any unhandled exception which doesn't extend {@link TemporalFailure} is converted to an
  * instance of this class before being returned to a caller.
  *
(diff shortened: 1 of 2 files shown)


bergundy (approved):
LGTM assuming this behavior was confirmed to be correct, CC @cretz

bergundy:
Saw the original issue, thanks @cretz

PR #523, opened 2022-03-10
fix(worker): Enhance and fix heartbeat behavior

bergundy:
- Flush last heartbeat on Activity failure
- Ensure ordering of heartbeats when custom payload codec is used
- Discard heartbeats if activity sends them faster than payload codec
  can process them

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -55,6 +55,10 @@ export const asyncLocalStorage = new AsyncLocalStorage<Context>();
  */
 export interface Info {
   taskToken: Uint8Array;
+  /**
+   * Base64 encoded `taskToken`
+   */
+  base64TaskToken: string;
   activityId: string;
   /**
    * Exposed Activity function name
--- a/packages/test/src/test-worker-heartbeats.ts
+++ b/packages/test/src/test-worker-heartbeats.ts
@@ -3,12 +3,13 @@ import { Subject, firstValueFrom } from 'rxjs';
 import { v4 as uuid4 } from 'uuid';
 import { Context } from '@temporalio/activity';
 import { isolateFreeWorker, Worker } from './mock-native-worker';
+import { coresdk } from '@temporalio/proto';
 
-async function runActivity(worker: Worker, callback?: () => void) {
+async function runActivity(worker: Worker, callback?: (completion: coresdk.ActivityTaskCompletion) => void) {
   const taskToken = Buffer.from(uuid4());
   const p = worker.run();
-  await worker.native.runActivityTask({ taskToken, start: { activityType: 'rapidHeartbeater' } });
-  callback?.();
+  const completion = await worker.native.runActivityTask({ taskToken, start: { activityType: 'rapidHeartbeater' } });
+  callback?.(completion);
   worker.shutdown();
   await p;
 }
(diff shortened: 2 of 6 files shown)


bergundy:
@cretz this fixes the issues you surfaced while working on Python.
We didn't use to have this problem when the code was originally written since data converters were sync at the time.

packages/test/src/test-worker-heartbeats.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  const heartbeatsSeen = Array<number>();
+  worker.native.activityHeartbeatCallback = (_tt, details) => {
+    heartbeatsSeen.push(details);
+  };
+  await runActivity(worker);
+  t.deepEqual(heartbeatsSeen, [1, 3]);
+});
+
+test('Worker ignores last heartbeat if activity succeeds', async (t) => {
```
cretz, line 69: Why is the last heartbeat ignored on success? I think even on success the last heartbeat should be flushed. I fear people in the UI or future uses of heartbeat details may expect their heartbeat details present. It also makes sense that simply "we guarantee the last heartbeat called will be sent to the server no matter what".

Core has its own buffer and this is a sync call, so I would think it's harmless to flush to Core. Curious though, does Core also discard unflushed heartbeat on success?

bergundy: Core will only flush pending heartbeats on failure.
Heartbeats don't matter on success, details are not persisted in history.

packages/worker/src/rxutils.ts
```diff
@@ -27,6 +27,19 @@ export function mergeMapWithState<T, I, O>(
   );
 }
 
+export function mapWithState<T, I, O>(
+  fn: (state: T, input: I) => StateAndOutput<T, O>,
+  initialState: T
+): OperatorFunction<I, O> {
```
Sushisource, line 33: :eyes: Is that you, `fold`?

packages/worker/src/worker.ts
```diff
@@ -821,21 +918,94 @@ export class Worker {
...
-      })
+      closeableGroupBy(({ base64TaskToken }) => base64TaskToken),
+      mergeMap((group$) =>
+        group$.pipe(
+          mapWithState(
+            (state: HeartbeatState, input: HeartbeatInput): HeartbeatStateAndOutput => {
+              // Ignore any input if we've marked this activity heartbeat stream as closed
+              if (state.closed) return { state, output: null };
+
+              switch (input.type) {
```
Sushisource, line 951: All this extra logic is a bit unfortunate. Now you've basically done the same amount of work Core had to do to ensure they get flushed. :cry: 

I don't know that there's any way around it with the async converters though...

Sushisource (approved):

packages/worker/src/worker.ts
```diff
@@ -821,21 +918,94 @@ export class Worker {
    * Errors from core responses are translated to cancellation requests and fed back via the activityFeedbackSubject.
    */
   protected activityHeartbeat$(): Observable<void> {
+    function process(state: HeartbeatState, heartbeat: Heartbeat): HeartbeatStateAndOutput {
+      return { state: { ...state, processing: true, pending: undefined }, output: { type: 'send', heartbeat } };
+    }
+
+    function storePending(state: HeartbeatState, heartbeat: Heartbeat): HeartbeatStateAndOutput {
+      return { state: { ...state, pending: heartbeat }, output: null };
```
lorensr, line 926: Is it possible we'd ever need to store multiple pending heartbeats?

packages/worker/src/worker.ts
```diff
@@ -821,21 +918,94 @@ export class Worker {
...
+                  }
+                case 'completion':
+                  if (!input.flushRequired) {
+                    return complete(input.callback);
+                  } else if (state.processing) {
+                    // Store the completion callback until heartbeat has been flushed
+                    return { state: { ...state, completionCallback: input.callback }, output: null };
+                  } else {
+                    if (state.pending) {
+                      // const { state, out
```
lorensr, line 979:
```suggestion
```

packages/test/src/test-worker-heartbeats.ts
cretz: I understand this now. I was somehow under the impression heartbeat details could have some value after an activity was done from a UI or history POV.

packages/worker/src/rxutils.ts
bergundy: It's very much like `scan` (uses scan internally)

packages/activity/src/index.ts
```diff
@@ -55,6 +55,10 @@ export const asyncLocalStorage = new AsyncLocalStorage<Context>();
  */
 export interface Info {
   taskToken: Uint8Array;
+  /**
+   * Base64 encoded `taskToken`
+   */
+  base64TaskToken: string;
```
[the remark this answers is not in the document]
bergundy: It's just cheaper to generate once

packages/worker/src/worker.ts
bergundy: No, the whole point is that we only flush the last one because the others become obsolete

packages/worker/src/worker.ts
bergundy: Yeah, that's unfortunate..

lorensr (approved):

PR #526, opened 2022-03-11
fix: Add reason to cancelation tasks in tests

lorensr:
Without a `cancel.reason`, the test hangs after:

image

Just confirming: even though the proto says reason is optional, it should always be there, and worker should throw & fail when it's not?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,7 +188,7 @@ jobs:
           name: packages-${{ matrix.target }}
           path: /tmp/registry/storage
         if: ${{ matrix.server != 'cloud' || env.TEMPORAL_CLIENT_CERT != '' }}
-      - run: node scripts/init-from-verdaccio.js --registry-dir /tmp/registry --sample https://github.com/temporalio/samples-typescript/tree/next/${{ matrix.sample }}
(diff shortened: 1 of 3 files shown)


lorensr:
Our test-npm-init action was pointed at `next` branch of samples-typescript, which we haven’t updated since November. We could maintain `next`, but would need a process for keeping it up-to-date. For now, I'm just pointing at `main` until we next need `next`.

bergundy (approved):
Thanks for the fix!

bergundy:
> Our test-npm-init action was pointed at `next` branch of samples-typescript, which we haven’t updated since November. We could keep `next`, but would need a process for keeping it up-to-date. For now, I'm just pointing at `main` until we next need `next`.

I just saw this, we should really point at next because otherwise we can’t make breaking changes.
You have a point that next should be updated more frequently.
It’s not high priority though.

lorensr:
I was thinking we could set it to `next` in the breaking change PR. And if
there are multiple breaking change PRs at once, you'd set the second to
`next2`.

bergundy:
My preference would be that as part of our release process, we'd upgrade all samples to use the new version and point the `next` branch at `main`.
This way we know that samples `next` is set to work with `main` of the SDK.

PR #540, opened 2022-03-15
feat(workflow-bundler): Allow ignoring modules for Webpack build

bergundy:
Picks up work started by @mjameswh in #519

### DIFF
--- a/packages/test/src/test-worker-from-bundle.ts
+++ b/packages/test/src/test-worker-from-bundle.ts
@@ -11,6 +11,7 @@ import { WorkflowClient } from '@temporalio/client';
 import { bundleWorkflowCode, Worker } from '@temporalio/worker';
 import { RUN_INTEGRATION_TESTS } from './helpers';
 import { successString } from './workflows';
+import { issue516 } from './mocks/workflows-with-node-dependencies/issue-516';
 
 if (RUN_INTEGRATION_TESTS) {
   test('Worker can be created from bundle code', async (t) => {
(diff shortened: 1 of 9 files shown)


packages/worker/src/workflow/bundler.ts
```diff
@@ -181,6 +264,22 @@ export class WorkflowCodeBundler {
...
+    this.logger.warn(
+      `Your Workflow code (or a library used by your Workflow code) is importing the following built-in Node modules:\n` +
+        Array.from(foundProblematicModules)
+          .map((module) => `  - '${module}'\n`)
+          .join('') +
+        `Workflow code don't have access to built-in Node modules (in order to help enforce determinism). If your are certain ` +
+        `these modules will not be used at runtime, then you may add their names to 'BundleOptions.ignoreModules' in order to ` +
+        `dismiss this warning. However, if your code execution actually depends on these modules, then you must change the code` +
+        `or remove the library.\n` +
+        `See https://typescript.temporal.io/api/namespaces/worker/#bundleworkflowcode for details.`
```
lorensr, line 280:
```suggestion
        `Workflow code doesn't have access to built-in Node modules (in order to help enforce determinism). If you are certain ` +
        `these modules will not be used at runtime, then you may add their names to 'WorkerOptions.bundlerOptions.ignoreModules' in order to ` +
        `dismiss this warning. However, if your code execution actually depends on these modules, then you must change the code` +
        `or remove the library.\n` +
        `See https://typescript.temporal.io/api/interfaces/worker.workeroptions/#bundleroptions for details.`
```

lorensr (approved):

packages/test/src/test-worker-from-bundle.ts
```diff
@@ -65,4 +66,49 @@ if (RUN_INTEGRATION_TESTS) {
...
+    await bundleWorkflowCode({
+      workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
+      logger,
+    });
+
+    console.log('****************************** LOGS **********************');
+    console.log(logs);
+    console.log('****************************** LOGS **********************');
+    t.true(
+      logs.some((entry) => entry.message.match(/'dns'/) && entry.message.match(/'BundleOptions.ignoreModules'/)),
```
mjameswh, line 110: The second regex is now incorrect:

```
logs.some((entry) => entry.message.match(/'dns'/) && entry.message.match(**/'WorkerOptions.bundlerOptions.ignoreModules'/**)),
```

bergundy: Oh, of course :)
Thanks.

PR #543, opened 2022-03-16
feat: Add testing framework

bergundy:
Initial work, missing features:

- No ability to fast forward time outside of a workflow
- No ability to toggle normal / time skipped mode

### DIFF
--- a/packages/test/src/test-testenvironment.ts
+++ b/packages/test/src/test-testenvironment.ts
@@ -1,0 +1,40 @@
+import anyTest, { TestInterface } from 'ava';
+import { v4 as uuid4 } from 'uuid';
+import { TestWorkflowEnvironment } from '@temporalio/testing';
+import { Worker } from '@temporalio/worker';
+import { defaultOptions } from './mock-native-worker';
+import { sleeper } from './workflows';
+
+interface Context {
+  testEnv: TestWorkflowEnvironment;
+}
+
+const test = anyTest as TestInterface<Context>;
+
+test.before(async (t) => {
+  t.context = {
+    testEnv: await TestWorkflowEnvironment.create(),
+  };
+});
+
+test.after.always(async (t) => {
+  await t.context.testEnv?.teardown();
+});
+
+test('TestEnvironment sets up test server and is able to run a Workflow with time skipping', async (t) => {
+  const worker = await Worker.create(defaultOptions);
+  const client = t.context.testEnv.workflowClient;
+  const runAndShutdown = async () => {
+    try {
+      await client.execute(sleeper, {
+        workflowId: uuid4(),
+        taskQueue: 'test',
+        args: [1_000_000],
+      });
+    } finally {
+      worker.shutdown();
+    }
+  };
+  await Promise.all([worker.run(), runAndShutdown()]);
+  t.pass();
+});
--- a/packages/testing/src/child-process.ts
+++ b/packages/testing/src/child-process.ts
@@ -9,10 +9,14 @@ export class ChildProcessError extends Error {
   }
 }
 
-export async function waitOnChild(child: ChildProcess): Promise<void> {
+export interface WaitOptions {
+  validReturnCodes: number[];
+}
+
+export async function waitOnChild(child: ChildProcess, opts?: WaitOptions): Promise<void> {
   return new Promise((resolve, reject) => {
     child.on('exit', (code, signal) => {
-      if (code === 0) {
+      if (code !== null && (opts?.validReturnCodes ?? [0]).includes(code)) {
         resolve();
       } else {
         reject(new ChildProcessError('Process failed', code, signal));
@@ -22,15 +26,18 @@ export async function waitOnChild(child: ChildProcess): Promise<void> {
   });
 }
 
-export async function kill(child: ChildProcess, signal: NodeJS.Signals = 'SIGINT'): Promise<void> {
+export async function kill(child: ChildProcess, signal: NodeJS.Signals = 'SIGINT', opts?: WaitOptions) {
   if (child.pid === undefined) {
     throw new TypeError('Expected child with pid');
   }
   process.kill(child.pid, signal);
   try {
-    await waitOnChild(child);
-  } catch (err: any) {
-    if (!(err.name === 'ChildProcessError' && err.signal === signal)) {
+    await waitOnChild(child, opts);
+  } catch (err) {
+    // Should error if the error is not a child process error or it is a child
+    // process and either the platform is Windows or the signal matches.
+    const shouldError = !(err instanceof ChildProcessError) || (process.platform !== 'win32' && err.signal !== signal);
+    if (shouldError) {
       throw err;
     }
   }
--- a/packages/testing/src/child-process.ts
+++ b/packages/testing/src/child-process.ts
@@ -26,7 +26,7 @@ export async function waitOnChild(child: ChildProcess, opts?: WaitOptions): Prom
   });
 }
 
-export async function kill(child: ChildProcess, signal: NodeJS.Signals = 'SIGINT', opts?: WaitOptions) {
+export async function kill(child: ChildProcess, signal: NodeJS.Signals = 'SIGINT', opts?: WaitOptions): Promise<void> {
   if (child.pid === undefined) {
     throw new TypeError('Expected child with pid');
   }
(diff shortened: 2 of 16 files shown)


packages/testing/src/child-process.ts
```diff
@@ -22,15 +26,18 @@ export async function waitOnChild(child: ChildProcess): Promise<void> {
...
   }
   process.kill(child.pid, signal);
   try {
-    await waitOnChild(child);
-  } catch (err: any) {
-    if (!(err.name === 'ChildProcessError' && err.signal === signal)) {
+    await waitOnChild(child, opts);
+  } catch (err) {
+    // Should error if the error is not a child process error or it is a child
+    // process and either the platform is Windows or the signal matches.
```
Sushisource, line 38: Explaining why would be good here

packages/testing/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+ */
+interface TestWorkflowEnvironmentOptions {
+  testServerSpawner?(port: number): ChildProcess;
+}
+
+function defaultTestWorkflowEnvironmentOptions(): Required<TestWorkflowEnvironmentOptions> {
+  return {
+    testServerSpawner(port: number) {
+      return spawn(path.join(__dirname, '../test-server'), [`${port}`], {
+        stdio: 'ignore',
```
Sushisource, line 23: Might make sense to direct to a log file by default or something so there's at least some info available if something went wrong

packages/test/src/test-testenvironment.ts
```diff
@@ -0,0 +1,40 @@
...
+    testEnv: await TestWorkflowEnvironment.create(),
+  };
+});
+
+test.after.always(async (t) => {
+  await t.context.testEnv?.teardown();
+});
+
+test('TestEnvironment sets up test server and is able to run a Workflow with time skipping', async (t) => {
+  const worker = await Worker.create(defaultOptions);
```
Sushisource, line 25: Would be good to have more than one test in here to make sure things work properly then. We need to require they are done serially?

bergundy: Tests would probably need to be run serially.
Right now the test server is still missing some critical features, I'll add a bunch more tests when those features are availabe.

packages/testing/src/child-process.ts
bergundy: Hmm.. I don't remember, I copied this from another file.
@cretz added this when he worked on adding windows support.

packages/testing/src/index.ts
bergundy: I might just make it inherit this proc's stdio.

packages/testing/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+ * Options for {@link TestWorkflowEnvironment.create}
+ */
+interface TestWorkflowEnvironmentOptions {
+  testServerSpawner?(port: number): ChildProcess;
+}
+
+function defaultTestWorkflowEnvironmentOptions(): Required<TestWorkflowEnvironmentOptions> {
+  return {
+    testServerSpawner(port: number) {
+      return spawn(path.join(__dirname, '../test-server'), [`${port}`], {
```
lorensr, line 23: Doesn't need `.exe` on win?

packages/testing/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    // TODO: Core is a singleton at the moment, once the bridge is refactored this will change.
+    await Core.install({ serverOptions: { address } });
+
+    return new this(child, conn);
+  }
+
+  /**
+   * Get an extablished {@link Connection} to the test server
```
lorensr, line 82:
```suggestion
   * Get an established {@link Connection} to the test server
```

packages/testing/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+   */
+  get connection(): Connection {
```
lorensr, line 84: why not `public readonly connection`?

packages/testing/src/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+        execution: { workflowId: 'fake', runId },
+      });
+    } catch (err: any) {
+      console.log(err);
+      if (err.details.includes('workflow history not found') || err.details.includes(runId)) {
+        break;
+      }
+      if (attempt === maxAttempts) {
+        throw err;
+      }
```
lorensr, line 24:
```suggestion
      if (err.details.includes('workflow history not found') || err.details.includes(runId)) {
        break;
      }
      if (attempt === maxAttempts) {
        throw err;
      }
      console.log(err);
```

lorensr: moved log to end—don't need to log in success case, or when throwing?

lorensr (approved):

packages/testing/src/child-process.ts
cretz: Likely because the `signal` field is not populated on the `exit` event in Windows

packages/testing/src/index.ts
cretz: I think that is best, and just allow verbosity configuration

packages/testing/src/index.ts
cretz: Are there tests for this? I would expect a test to catch this.

packages/testing/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  return {
+    testServerSpawner(port: number) {
+      return spawn(path.join(__dirname, '../test-server'), [`${port}`], {
+        stdio: 'ignore',
+      });
+    },
+  };
+}
+
+// TS transforms `import` statements into `require`s, this is a workaround until
```
cretz, line 31: What's wrong with require?

packages/testing/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+ */
+export class TestWorkflowEnvironment {
+  protected constructor(protected readonly serverProc: ChildProcess, protected readonly conn: Connection) {}
+
+  /**
+   * Create a new test environment
+   */
+  static async create(opts?: TestWorkflowEnvironmentOptions) {
+    // No, we're not going to compile this to ESM for one dependency
+    const getPort = (await _importDynamic('get-port')).default as typeof getPortType;
```
cretz, line 49: Why imported here instead of at the top of the file? And usually these things are so small you can just inline them in the project instead of making a dependency.

packages/testing/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+   * Get an extablished {@link Connection} to the test server
+   */
+  get connection(): Connection {
+    return this.conn;
+  }
+
+  /**
+   * Get an {@link AsyncCompletionClient} for interacting with the test server
+   */
+  get asyncCompletionClient(): AsyncCompletionClient {
```
cretz, line 91: One might expect a property to not construct a value (e.g. `env.asyncCompletionClient === env.asyncCompletionClient`). I think this should either be created once (eager or lazy) or be a function.

packages/testing/src/index.ts
bergundy: I think tests are failing on windows, I'll fix.

packages/testing/src/index.ts
bergundy: 👍

packages/testing/src/index.ts
bergundy: Doesn't work, at some point sindresorhus decided he's making all of his packages ESM and you have to `import` them.

packages/testing/src/index.ts
bergundy: import is async, and you can only have top level awaits with ESM, not commonjs.
In my experience, commonjs is still the preferred way to author node libraries.

packages/testing/src/index.ts
bergundy: 👍

bergundy: Done

packages/testing/src/index.ts
bergundy: Done

packages/testing/src/index.ts
cretz: For this small use case, just grab https://github.com/sindresorhus/get-port/blob/main/index.js and embed it (or just do what https://stackoverflow.com/a/28050404 does which is the same thing but even simpler)

bergundy: Sure, I could, and I've considered it, but `get-port` is already such a small library that it doesn't matter much.
If it wasn't for the inconvenience of ESM vs. commonjs we probably wouldn't be having this discussion.

cretz: Probably true (though un-left-pad'ing a few-line test utility is reasonable). Feel free to ignore.

Issue #483, opened 2022-02-17
[Feature Request] NestJS Transport / Integration?


StevenLangbroek:
### Is your feature request related to a problem? Please describe.

It's more of a question: Temporal's Typescript SDK seems like it would maybe be a good fit as a NestJS Module, is this something there'd be interest in exploring? I'd need to spend a little bit of time on it but I'd love to validate whether it would work with a PoC, and see what kind of API we'd end up with.

### Describe the solution you'd like

I'd just like to understand whether there's interest at this point, so I can invest some of my own time in exploring this.

### Additional context

I would imagine an API to look something like:

```ts
@Workflow()
export class TripBookingWorkflow {
  @InjectActivities({
    startToCloseTimeout: ms("1hr"),
    retryOptions: {
      maxAttempts: 1,
    },
  })
  activities: Activities<TripBookingActivities>;

  public async bookTrip(name: string): Promise<void> {
    const saga = new Saga({
      parallelCompensation: true,
    });

    try {
      const carReservationId = this.activities.reserveCar(name);
      saga.addCompensation(this.activities.cancelCar, carReservationId);

      const hotelReservationId = this.activities.bookHotel(name);
      saga.addCompensation(this.activities.cancelHotel, hotelReservationId);

      const flightReservationId = this.activities.bookFlight(name);
      saga.addCompensation(this.activities.cancelFlight, flightReservationId);
    } catch (err) {
      saga.compensate();
      throw err;
    }
  }
}
```

StevenLangbroek:
I think this could be good for Temporal because of what NestJS brings to the table;

- Application lifecycle hooks instead of homegrown solutions
- Ecosystem
- Separation of concerns
- Testability

bergundy:
I think such an integration could be valuable.

I don't know if we'll want this as part of the SDK repo. It should probably start out as an external package.

I'd be happy to review what you come up with.

StevenLangbroek:
@bergundy totally fair! would you be willing to help me iterate on the API a few rounds before starting a proof-of-concept?

bergundy:
Hi,
Sorry for the late reply, sure, feel free to ping me when you have something.

StevenLangbroek:
[...]

PR #545, opened 2022-03-18
feat(worker): support exporting a workflow bootstrapper

StevenLangbroek:
## What was changed

If the module in `workflowsPath` has a `default` export, and it's a function, the results of running that function will be returned from `importWorkflows` (and it can be async! so you can run all sorts of lifecycle hooks 'n what-not). We hope this will enable additional use-cases, such as deeper NestJS integration.

## Checklist

1. Part of #483 

2. How was this tested:

It isn't yet! I'll start with testing locally through linking, but any pointers here would be greatly appreciated.

3. Any docs updates needed?

Yes! Extensively! WIP!

- [x] Validate locally
- [ ] Write tests
- [ ] Discuss documentation

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -158,10 +158,10 @@ export class WorkflowCodeBundler {
       // Bundle all Workflows and interceptor modules for lazy evaluation
       api.overrideGlobals();
       api.setImportFuncs({
-        importWorkflows: () => {
+        importWorkflows: async () => {
           let workflows = await import(/* webpackMode: "eager" */ ${JSON.stringify(this.workflowsPath)});
-          if (workflows.default && typeof workflows.default === 'function) {
-            workflows = workflows.default();
+          if (workflows.default && typeof workflows.default === 'function') {
+            return workflows.default();
           }
           return workflows;
         },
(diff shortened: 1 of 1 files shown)


packages/worker/src/workflow/bundler.ts
```diff
@@ -159,7 +159,11 @@ export class WorkflowCodeBundler {
       api.overrideGlobals();
       api.setImportFuncs({
         importWorkflows: () => {
```
bergundy, line 161: This would need to be an async function because of the `await` statement

packages/worker/src/workflow/bundler.ts
```diff
@@ -159,7 +159,11 @@ export class WorkflowCodeBundler {
...
-          return import(/* webpackMode: "eager" */ ${JSON.stringify(this.workflowsPath)});
+          let workflows = await import(/* webpackMode: "eager" */ ${JSON.stringify(this.workflowsPath)});
+          if (workflows.default && typeof workflows.default === 'function) {
+            workflows = workflows.default();
```
bergundy, line 164: Could just `return workflows.default()` here.

StevenLangbroek:
@bergundy any thoughts on how to test this?

bergundy:
@StevenLangbroek you should add some integration tests.
Create a worker with `workflowsPath` set to a file which has a default function export and verify it works.

bergundy:
Is anyone still interested in this?
So far I haven't seen any requests for this feature and I'm afraid it'd be a backwards incompatible change in case users already export default from their workflows entry point.

PR #546, opened 2022-03-18
fix: Use JsonPayloadConverter for search attributes

lorensr:
Temporal doesn't support Null or Binary Payloads for search attributes.

### DIFF
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -14,6 +14,7 @@ import {
   Payload,
   PayloadCodec,
   RetryState,
+  searchAttributePayloadConverter,
   TerminatedFailure,
   TimeoutFailure,
   TimeoutType,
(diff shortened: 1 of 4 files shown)


packages/test/src/integration-tests.ts
```diff
@@ -574,7 +575,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       'BinaryChecksums',
     ]);
 
-    const checksums = defaultPayloadConverter.fromPayload(
+    const checksums = searchAttributePayloadConverter.fromPayload(
```
bergundy, line 578: Where is this defined?

bergundy (approved):
LGTM

packages/test/src/integration-tests.ts
```diff
@@ -612,7 +613,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     );
     t.deepEqual(await fromPayload(execution.raw.workflowExecutionInfo!.memo!.fields!.a!), 'b');
     t.deepEqual(
-      await defaultPayloadConverter.fromPayload(
+      await searchAttributePayloadConverter.fromPayload(
```
bergundy, line 616: no need to await here

PR #547, opened 2022-03-19
feat: Complete test framework

bergundy:
This should not be merged yet until a new Java test server is released.


packages/test/src/test-replay.ts
```diff
@@ -18,7 +18,10 @@ test.before(async (t) => {
   process.removeAllListeners('unhandledRejection');
   const logger = new DefaultLogger('DEBUG');
   // Use forwarded logging from core
-  const replayCore = await ReplayCore.install({ logger, telemetryOptions: { logForwardingLevel: 'INFO' } });
+  const replayCore = await ReplayCore.install({
+    logger,
+    telemetryOptions: { logForwardingLevel: 'DEBUG', tracingFilter: 'temporal_sdk_core=DEBUG' },
```
lorensr, line 23: Intentional?

packages/testing/src/index.ts
```diff
@@ -9,35 +9,62 @@ import { ChildProcess, spawn, StdioOptions } from 'child_process';
...
     testServerSpawner:
-      testServerSpawner ??
-      ((port: number) =>
-        spawn(path.join(__dirname, `../${TEST_SERVER_EXECUTABLE_NAME}`), [`${port}`], {
-          stdio: testServerStdio,
-        })),
+      typeof testServer === 'function'
+        ? testServer
+        : (port: number) =>
+            spawn(testServer?.path || DEFAULT_TEST_SERVER_PATH, [`${port}`, '--enable-time-skipping'], {
```
cretz, line 64: I wonder if test server process management and grpc connection and such should be in core instead

packages/testing/src/index.ts
```diff
@@ -118,8 +144,36 @@ export class TestWorkflowEnvironment {
...
+  async skip(durationMs: number | string): Promise<void> {
+    await this.connection.testService.skip({ duration: msToTs(durationMs) });
+  }
+
+  /**
+   * Switch the test server to normal time mode while in this block.
+   *
+   * Useful for complex test setup scenarios.
+   */
+  async withNormalTime<R>(fn: () => Promise<R>): Promise<R> {
```
lorensr, line 164: If this isn't used, can tests be run in parallel?

packages/testing/src/test-service-client.ts
```diff
@@ -0,0 +1,21 @@
+import * as grpc from '@grpc/grpc-js';
+import { Connection as BaseConnection, ConnectionOptions } from '@temporalio/client';
+import { temporal } from '../generated-protos';
+
+export type TestService = temporal.api.testservice.v1.TestService;
+export const { TestService } = temporal.api.testservice.v1;
+
+/**
+ * A Connection class that can be used interacting with both the test server's TestService and WorkflowService
```
lorensr, line 9:
```suggestion
 * A Connection class that can be used to interact with both the test server's TestService and WorkflowService
```

CONTRIBUTING.md
```diff
@@ -164,3 +164,17 @@ npx lerna publish from-git # add `--dist-tag next` for pre-release versions
 ```sh
 rm $HOME/Downloads/packages-*
 ```
+
+### Updating the Java test server proto files
+
```
lorensr, line 169:
```suggestion

These include:
- `dependencies/`
- `temporal/`
- `Makefile`
- `api-linter.yaml`
- `buf.yaml`

```

lorensr (approved):
When is `make` run?

packages/test/src/test-replay.ts
bergundy: Nope, reverting

CONTRIBUTING.md
lorensr:
```suggestion

[These files](https://github.com/temporalio/sdk-java/tree/master/temporal-test-server/src/main/proto) are taken from the `sdk-java` repo.
```

tommyk:
this is the feature that is needed so `await sleep('20 hours');` just automatically doesn't wait that long, correct? 

i haven't been able to read how that time skipping works exactly, and from a look at this change it appears that magic is built into this java testing temporal server being used, is that correct?

does this mean java has to be installed to use this feature from our typescript tests?

sorry for so many questions, really excited if this unlocks the ability to test long running workflows easier.  that is currently something i am trying to figure out how to even accomplish without making things more complicated than they should be.  appreciate all the hard work, i am loving temporal so far!

Irvenae:
Hey,

is there an ETA on this feature?

packages/core-bridge/src/lib.rs
```diff
@@ -792,6 +792,13 @@ fn worker_shutdown(mut cx: FunctionContext) -> JsResult<JsUndefined> {
     }
 }
 
+/// Drop a reference to a Client, once all references are dropped, the Client will be closed.
+fn client_close(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let client = cx.argument::<BoxedClient>(0)?;
+    drop(client);
+    Ok(cx.undefined())
```
Sushisource, line 799: I think this is purely performative - do you know this is actually doing what you want? I couldn't immediately find a test for it.

AFAICT looking at Neon docs/src dropping a JsBox does nothing - the actual lifetime is managed in JS and the `Finalize` implementation is called when it's freed.

You could do something like put the client behind an option, and _really_ drop it here, but then all other accesses of the client will have this potential error path where the client is already closed, but that's kinda necessary for this to behave how you really want

packages/test/src/test-testenvironment.ts
```diff
@@ -0,0 +1,144 @@
...
+      workflowClient
+        .execute(sleep, {
+          workflowId: uuid4(),
+          taskQueue: 'test',
+          args: [1_000_000],
+          runInNormalTime,
+        })
+        .then(() => 'fake time'),
+      // Hopefully 5 seconds will be enough in CI
+      new Promise((resolve) => setTimeout(resolve, 5000)).then(() => 'real time'),
```
Sushisource, line 77: Heh, you'd be on me about this one :wink: 

I think it is adequate to set a lower value for the sleep timer, and verify that in the skipping case it took less time than it should've, and in the real case it took at least as long as it should've.

packages/testing/src/assert-to-failure-interceptor.ts
```diff
@@ -0,0 +1,38 @@
+import assert from 'assert';
+import { ApplicationFailure, WorkflowInterceptors } from '@temporalio/workflow';
+
+/**
+ * Simple interceptor that transforms assertion errors into non retryable failures.
```
Sushisource, line 5: Worth specifying you mean `assert.AssertionError`s specifically

packages/testing/src/index.ts
```diff
@@ -9,35 +15,137 @@ import { ChildProcess, spawn, StdioOptions } from 'child_process';
...
+  /**
+   * If set to `true`, waiting for the result does not unlock time skipping
+   */
+  runInNormalTime?: boolean;
+};
+
+/**
+ * A client with the exact same API as the "normal" client with 1 exception,
+ * When this client waits on a Workflow's result, it will unlock time skipping
+ * in the test server.
```
Sushisource, line 54: "unlock time skipping" is a weird phrase from a user's perspective. I'd just say "enable" or something.

packages/testing/src/index.ts
```diff
@@ -129,8 +237,26 @@ export class TestWorkflowEnvironment {
...
+
+  /**
+   * Wait for `durationMs` in "test server time".
+   *
+   * The test server toggles between skipped time and normal time dependnding on what it needs to execute.
+   *
+   * This method is likely to resolve in less than `durationMs` of "real time".
+   *
+   * Useful for simulating events far into the future like completion of long running activities.
+   *
```
Sushisource, line 249: A motivating example would be helpful here I think.

packages/worker/src/connection.ts
```diff
@@ -10,18 +11,56 @@ import { Runtime } from './runtime';
...
+    }
+    await Runtime.instance().closeNativeClient(this.nativeClient);
+  }
 }
 
 /**
- * Extract the private native client instance from a `NativeConnection` instance
+ * Extract the private native client instance from a `NativeConnection` instance.
+ *
+ * Only meant to be used by the framework.
```
Sushisource, line 46: Is there a "hidden doc" annotation?

packages/worker/src/runtime.ts
```diff
@@ -228,7 +230,17 @@ export class Runtime {
       tls: normalizeTlsConfig(compiledServerOptions.tls),
       url: options?.tls ? `https://${compiledServerOptions.address}` : `http://${compiledServerOptions.address}`,
     };
-    return await promisify(newClient)(this.native, clientOptions);
+    const client = await promisify(newClient)(this.native, clientOptions);
+    this.registeredClients.add(client);
+    return client;
+  }
+
+  public async closeNativeClient(client: native.Client): Promise<void> {
```
Sushisource, line 238: Should have docstring like the others

packages/core-bridge/src/lib.rs
bergundy: I was meaning to get back to this, forgot to verify it in the PR.

bergundy: It ended up being way more complicated than initially anticipated but I think it's fine.
Managing these object lifecycles isn't trivial but we have to do it in order to avoid leaving the runtime around, it prevents the process from exiting.

packages/test/src/test-testenvironment.ts
bergundy: I'll change to verify it slept for less than a minute, if that's not enough in CI I don't know what is.
I won't actually sleep though.

bergundy: I changed it but it's still racy, IMHO this version is more readable.
Not sure what the best way around it is, if you're testing actual time you just need to wait and the longer you're willing to wait the less flakey the test will be.

packages/testing/src/assert-to-failure-interceptor.ts
bergundy: Sure

packages/testing/src/index.ts
bergundy: Yeah, the whole concept is confusing but it works well.
I'll try to clarify.

packages/testing/src/index.ts
bergundy: 👍

packages/worker/src/connection.ts
bergundy: No need, we never re-export this

packages/worker/src/runtime.ts
bergundy: Missed that, thanks

bergundy:
@tommyk 

Hey, sorry for the late response.

> this is the feature that is needed so `await sleep('20 hours');` just automatically doesn't wait that long, correct?

Yes, the test server automatically skips time so you don't have to wait so long to test long running workflows.


> i haven't been able to read how that time skipping works exactly, and from a look at this change it appears that magic is built into this java testing temporal server being used, is that correct?
> 
> does this mean java has to be installed to use this feature from our typescript tests?

The magic is built into the Java test server, it's behaves similarly to the real server but it skips time if only workflows are running (or when calling `sleep` from non workflow code such as in an activity).

You don't have to install Java to use this, we've pre-compiled the Java test server using graalvm to native binaries for the different targets and when you install the testing package we will automatically download the proper binary for your OS (only x86 supported at the moment but you can run the test server on M1 macs if you have rosetta installed).

cretz:
Can we get a small PR to https://github.com/temporalio/samples-typescript that shows how this might look at least on one or two workflows there? At least for me, it'd be easier to grok the testing framework from a user API POV with that compared to reading this PR sans user docs.

lorensr:
API looks like this:

https://github.com/temporalio/sdk-typescript/pull/547/files#diff-ecb09996040151c8403e643d59ebfc2e6b8f6e3ffdaf71b35569bbe3d090d949

```ts
  t.context = {
    testEnv: await TestWorkflowEnvironment.create({
      testServer: {
        stdio: 'inherit',
      },
    }),
  };

...

const { workflowClient, nativeConnection, sleep } = t.context.testEnv;
```

packages/test/src/test-testenvironment.ts
```diff
@@ -0,0 +1,149 @@
...
+import { TestWorkflowEnvironment, workflowInterceptorModules } from '@temporalio/testing';
+import { Worker } from '@temporalio/worker';
+import { WorkflowFailedError } from '@temporalio/client';
+import { assertFromWorkflow, sleep, raceActivityAndTimer } from './workflows/testenv-test-workflows';
+
+interface Context {
+  testEnv: TestWorkflowEnvironment;
+}
+
+async function withWorker<R>(worker: Worker, fn: () => Promise<R>): Promise<R> {
```
lorensr, line 12: Should we export this from `/testing`? Think it's going to be a common pattern we want people to follow, so it'll either be in our samples for people to copy-paste, or they import it.

packages/testing/src/index.ts
```diff
@@ -129,8 +237,55 @@ export class TestWorkflowEnvironment {
     // TODO: the server should return exit code 0
     await kill(this.serverProc, 'SIGINT', { validReturnCodes: [0, 130] });
   }
+
+  /**
+   * Wait for `durationMs` in "test server time".
+   *
+   * The test server toggles between skipped time and normal time dependnding on what it needs to execute.
```
lorensr, line 244:
```suggestion
   * The test server toggles between skipped time and normal time depending on what it needs to execute.
```

packages/worker/src/connection.ts
lorensr: (Only things exported from package's `index.ts` show up in API reference)

cretz:
@lorensr - I was looking for more like how do I test sending a signal, then another 10 minutes later, then maybe completing an activity an hour later.

packages/test/src/test-testenvironment.ts
bergundy: Good idea, not sure where to export this from though. Maybe this could be a method on the worker..

bergundy:
> @lorensr - I was looking for more like how do I test sending a signal, then another 10 minutes later, then maybe completing an activity an hour later.

I'll add this as a test

packages/core-bridge/src/lib.rs
```diff
@@ -135,7 +135,7 @@ struct Client {
     core_client: Arc<RawClient>,
 }
 
-type BoxedClient = JsBox<Client>;
+type BoxedClient = JsBox<Mutex<Option<Client>>>;
```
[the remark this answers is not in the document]
bergundy: This lock is rarely used so not sure if it really matters that much but it doesn't hurt, changing.

Sushisource (approved):

bergundy:
@cretz https://github.com/temporalio/sdk-typescript/pull/547/commits/cab2be2444efc318d2903f7d38019167c2617a41#diff-ecb09996040151c8403e643d59ebfc2e6b8f6e3ffdaf71b35569bbe3d090d949R130

cretz:
@bergundy - Thanks! I will still be a bit curious how this looks from a user POV once y'all start adding tests to the sample workflows. `@temporalio/workflow.sleep()` is what they call from outside a workflow but inside a test to skip time? It's a bit different than what I'm used to in the Go SDK (schedule delayed callback), so it may just be my own confusion. I can wait until I see more tests from the outside, no worries.

bergundy:
@cretz `@temporalio/workflow.sleep()` is only callable from inside a workflow.
From outside of the workflow you can call `TestEnvironment.sleep()` which will fastforward time if called outside of a workflow.
It's a bit confusing that it's called `sleep` in both places but means different things.

lorensr:
Can call it something else? 

```ts
fastForward('1m')
testSleep('1m')
waitFor('1m')
```

bergundy:
At some point I named it `skip` but I think `sleep` is fine, it’s effectively a sleep call in test server time which could be normal or skipped time depending on the status of the time skipping lock

lorensr:
Could call it fastForward 😄

testSleep

testDelay

waitFor(‘1m’)

PR #551, opened 2022-03-21
refactor(proto)!: Use protobufjs json-modules

lorensr:
## What was changed

- Moved from `pbjs -t static-module` to `pbjs -t json-module`. https://github.com/protobufjs/protobuf.js/blob/master/cli/README.md#pbjs-for-javascript
- Used different `--root`s to not conflict with each other or user protos
- Updated `docs/protobuf-libraries.md`
- Fixed `patchProtobufRoot` (it warned on enums)

## Why?

Originally started the refactor because when a user tries to run `pbjs -t json-module` for their protos, they got an `addJSON` method not found error because we had already set the default protobufjs global to static classes, which don't have `addJSON`. I later found the solution for this (SDK uses `--root`), but Roey said we had to move to json-modules for something else as well, so I kept going.


packages/proto/lib/index.js
```diff
@@ -1,2 +1,2 @@
-module.exports.coresdk = require('./coresdk').coresdk;
 module.exports = require('./temporal');
+module.exports.coresdk = require('./coresdk').coresdk;
```
bergundy, line 2: We should just unify the 2 generated files into 1.
The original reason that I split them was because the bundle side exploded when I included `temporal`.
This is no longer a restriction.

packages/proto/package.json
```diff
@@ -15,6 +15,7 @@
   "author": "Roey Berman <roey@temporal.io>",
   "license": "MIT",
   "dependencies": {
+    "@temporalio/common": "file:../common",
```
bergundy, line 18: This is a circular dependency

packages/client/src/connection.ts
```diff
@@ -1,11 +1,15 @@
...
+import root, { temporal } from '@temporalio/proto';
+import { AsyncLocalStorage } from 'async_hooks';
+import type { Root } from 'protobufjs';
 import { defaultGrpcRetryOptions, makeGrpcRetryInterceptor } from './grpc-retry';
 
 export type WorkflowService = temporal.api.workflowservice.v1.WorkflowService;
-export const { WorkflowService } = temporal.api.workflowservice.v1;
+
+// use `root.lookupService` because imported `WorkflowService` doesn't have `create` method
+const WorkflowService = (root as unknown as Root).lookupService('temporal.api.workflowservice.v1.WorkflowService');
```
bergundy, line 11: Can you explain why doesn't it have `create`? It doesn't seem to be an issue with the other protos.

packages/proto/package.json
lorensr: The dep is for: 

```
const { patchProtobufRoot } = require('@temporalio/common/lib/converter/patch-protobuf-root');
```

Shall I move the `patch-protobuf-root.ts` file to `packages/protos/src/`?

lorensr: Then `packages/common/src/converter/patch-protobuf-root.ts` would have:

```ts
export { patchProtobufRoot } from '@temporalio/proto/lib/patch-protobuf-root'
```

packages/proto/lib/index.js
lorensr: Combined in https://github.com/temporalio/sdk-typescript/commit/cf12c4c1b7d2bde493c4ef7ae315b4f2dace2c52

packages/proto/lib/coresdk.js
```diff
@@ -0,0 +1,3 @@
+module.exports = require('./root').coresdk;
+module.exports.temporal = require('./root').temporal;
+module.exports.google = require('./root').google;
```
lorensr, line 3: Re-added & edited `coresdk.js` & `.d.ts` in order to get the build working (we have code importing from `@temporalio/proto/lib/coresdk`). I could either maintain backcompat for anyone importing from `/lib/*` by also adding a similar `temporal.js,.d.ts`, or I could remove this file and update our imports.

bergundy: Let's just get rid of this, it's not worth keeping, no reason for it anymore.

packages/proto/package.json
bergundy: Sure, we might just need to re-export this function in common.

docs/protobuf-libraries.md
```diff
@@ -69,18 +69,19 @@ module.exports = patchProtobufRoot(unpatchedRoot);
...
-const worker = Worker.create({ dataConverter, ... });
+const worker = Worker.create({ dataConverter: { payloadConverterPath: require.resolve('./data-converter') }, ... });
 
 // src/client.ts
 import { foo } from '../protos/root';
 import { dataConverter } from './data-converter';
 
-const client = new WorkflowClient(connection.service, { dataConverter });
+const client = new WorkflowClient(connection.service, {
+  dataConverter: { payloadConverterPath: require.resolve('./data-converter') }
```
bergundy, line 82:
```suggestion
  dataConverter: { payloadConverterPath: require.resolve('./payload-converter') }
```

bergundy: Applies above too

packages/common/src/converter/patch-protobuf-root.ts
```diff
@@ -1,49 +1 @@
...
-type Namespace = { nested: Record<string, unknown> };
-
-function isType(value: unknown): value is Type {
-  return isRecord(value) && value.constructor.name === 'Type';
-}
-
-function isNamespace(value: unknown): value is Namespace {
-  return isRecord(value) && value.constructor.name === 'Namespace';
-}
+export { patchProtobufRoot } from '@temporalio/proto/lib/patch-protobuf-root';
```
bergundy, line 1: nit: don't need an entire file for re-exporting

packages/common/src/converter/payload-converters.ts
```diff
@@ -1,9 +1,9 @@
 import { errorMessage, UnsupportedJsonTypeError, ValueError } from '@temporalio/internal-workflow-common';
 import { PayloadConverter } from './payload-converter';
-import { encodingKeys, EncodingType, encodingTypes, METADATA_ENCODING_KEY, Payload, str, u8 } from './types';
+import { encodingKeys, encodingTypes, METADATA_ENCODING_KEY, Payload, str, u8 } from './types';
 
 export interface PayloadConverterWithEncoding extends PayloadConverter {
-  readonly encodingType: EncodingType;
+  readonly encodingType: string;
```
bergundy, line 6: Hmm.. can you remind me what this interface does?
Is it just for the composite payload converter?

packages/client/src/connection.ts
```diff
@@ -1,11 +1,12 @@
...
 import * as grpc from '@grpc/grpc-js';
+import { normalizeTlsConfig, TLSConfig } from '@temporalio/internal-non-workflow-common';
 import { temporal } from '@temporalio/proto';
-import { TLSConfig, normalizeTlsConfig } from '@temporalio/internal-non-workflow-common';
+import { AsyncLocalStorage } from 'async_hooks';
 import { defaultGrpcRetryOptions, makeGrpcRetryInterceptor } from './grpc-retry';
 
 export type WorkflowService = temporal.api.workflowservice.v1.WorkflowService;
 export const { WorkflowService } = temporal.api.workflowservice.v1;
+export { TLSConfig };
```
bergundy, line 9: Do we need to export this here?

packages/proto/lib/root.js
```diff
@@ -0,0 +1,3 @@
+const { patchProtobufRoot } = require('@temporalio/common/lib/converter/patch-protobuf-root');
```
bergundy, line 1: Can we just require this from `@temporalio/common`?

packages/proto/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./lib",
```
bergundy, line 4: This conflicts with our clean script:

```
        const { outDir } = tsconfig.compilerOptions;
        if (outDir) {
          const outPath = resolve(packagePath, outDir);
          console.log(`Removing ${outPath}`);
          rmSync(outPath, { recursive: true, force: true });
```

We shouldn't mix generated code with normal code.

scripts/clean.js
```diff
@@ -31,9 +31,12 @@ function cleanTsGeneratedFiles() {
         const tsconfig = JSON5.parse(readFileSync(filePath));
         const { outDir } = tsconfig.compilerOptions;
         if (outDir) {
-          const outPath = resolve(packagePath, outDir);
-          console.log(`Removing ${outPath}`);
-          removeSync(outPath);
+          // Skip over proto/lib, which contains source files
```
bergundy, line 34: Oh I see you addressed this here.
But this now means that the TS output cannot be cleaned with this script.

bergundy (commented):
Getting there, I have a only a few comments.

packages/common/src/converter/payload-converters.ts
lorensr: Yes

packages/client/src/connection.ts
lorensr: Moving to index

scripts/clean.js
lorensr: Correct. I could allowlist the source files, or move them from `lib/` to `js/`?

bergundy: Let's use separate directories

lorensr:
@bergundy I'm rerunning, but this run:

https://github.com/temporalio/sdk-typescript/runs/6087695330?check_suite_focus=true

had:

image

lorensr:
Huh, it happened again. Didn't expect that. Passes locally with node 16.

bergundy:
I think this might be due to slow CI and short workflow task timeout.
You can increase (or just delete) the timeout in this test, I think I accidentally copied from the test above it.

bergundy:
Oops, accidentally closed

lorensr:
@bergundy thanks, removing timeout fixed. Current issue is:

image
image

bergundy (approved):

lorensr:
Solving this issue:

image

by adding `files` field to `package.json`. Before and after:

image

Issue was NPM uses `.gitignore` to determine what not to pack when there is no `.npmignore` file.

Issue #552, opened 2022-03-21
[Bug] Windows test exists with code 1 without failures


lorensr:
Sometimes on Windows this happens, where no tests fail, but exit code is 1:

[build-and-test (16, windows-latest)](https://github.com/temporalio/sdk-typescript/runs/5630121517?check_suite_focus=true#logs)

The only thing in the output that looks bad is this error, but it's intentional, and the test that causes it succeeds.

![image](https://user-images.githubusercontent.com/251288/159302964-781d3699-a27b-4442-b7fe-4d9cd803e5be.png)

lorensr:
The next time it ran, it succeeded:

[build-and-test (16, windows-latest)](https://github.com/temporalio/sdk-typescript/runs/5630419601?check_suite_focus=true#logs)

but with 111 tests (vs the failing one had 110 tests). I sorted lines & diffed and the failing one is missing the line:

```
@temporalio/test:   √ replay » cancel-fake-progress-replay (3s)
```

![image](https://user-images.githubusercontent.com/251288/159334751-b08aa7b4-50cf-45bc-bb53-00594711d970.png)

lorensr:
passing output:

```
@temporalio/test:   ✔ replay › cancel-fake-progress-replay-nondeterministic (2.2s)
@temporalio/test: 2022-03-21T18:47:44.678Z [WARN] [temporal_sdk_core::worker] Failing workflow activation run_id="a7e68cfc-0287-48dd-b86a-866caf5c9a73" failure=Failure { failure: Some(Failure { message: "WorkflowUpdateError { source: Nondeterminism(\"No command scheduled for event HistoryEvent(id: 10, Some(ActivityTaskCancelRequested))\"), run_id: \"a7e68cfc-0287-48dd-b86a-866caf5c9a73\" }", source: "", stack_trace: "", cause: None, failure_info: Some(ApplicationFailureInfo(ApplicationFailureInfo { r#type: "", non_retryable: false, details: None })) }) }
```

failing output:

```
@temporalio/test:   × Timed out while running tests
@temporalio/test:   1 tests were pending in lib\test-replay.js
@temporalio/test:   ( ) replay » cancel-fake-progress-replay-nondeterministic
```

bergundy:
I haven't see this in a long time, I think it's fixed.
@lorensr?

lorensr:
Haven't seen either

PR #554, opened 2022-03-21
fix(worker): Fix Windows bundle-writing bug

lorensr:
Opening this PR to debug this CI issue https://github.com/temporalio/sdk-typescript/issues/552

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -237,6 +237,8 @@ export class WorkflowCodeBundler {
 
     // Cast to any because the type declarations are inaccurate
     compiler.inputFileSystem = inputFilesystem as any;
+    // Don't use ufs due to a strange bug on Windows:
+    // https://github.com/temporalio/sdk-typescript/pull/554
     compiler.outputFileSystem = outputFilesystem as any;
 
     try {
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -1,3 +1,4 @@
+import { errorCode } from '@temporalio/internal-workflow-common';
 import dedent from 'dedent';
 import * as realFS from 'fs';
 import * as memfs from 'memfs';
@@ -126,6 +127,33 @@ export class WorkflowCodeBundler {
     const entrypointPath = this.makeEntrypointPath(ufs, this.workflowsPath);
 
     this.genEntrypoint(vol, entrypointPath);
+    try {
+      console.log(memoryFs.statSync('/dist'));
+    } catch (e) {
+      if (errorCode(e) === 'ENOENT') {
+        console.log('memoryFs: No /dist');
+      } else {
+        console.log('Error stating memoryFs: ', e);
+      }
+    }
+    try {
+      console.log(realFS.statSync('/dist'));
+    } catch (e) {
+      if (errorCode(e) === 'ENOENT') {
+        console.log('realFS: No /dist');
+      } else {
+        console.log('Error stating realFS: ', e);
+      }
+    }
+    try {
+      console.log(ufs.statSync('/dist'));
+    } catch (e) {
+      if (errorCode(e) === 'ENOENT') {
+        console.log('ufs: No /dist');
+      } else {
+        console.log('Error stating ufs: ', e);
+      }
+    }
     await this.bundle(ufs, memoryFs, entrypointPath, distDir);
     return ufs.readFileSync(path.join(distDir, 'main.js'), 'utf8');
   }
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -120,12 +120,14 @@ export class WorkflowCodeBundler {
     }
 
     // Cast because the type definitions are inaccurate
-    ufs.use(memfs.createFsFromVolume(vol) as any).use({ ...realFS, readdir: readdir as any });
+    const memoryFs = memfs.createFsFromVolume(vol);
+    ufs.use(memoryFs as any).use({ ...realFS, readdir: readdir as any });
     const distDir = '/dist';
     const entrypointPath = this.makeEntrypointPath(ufs, this.workflowsPath);
 
     this.genEntrypoint(vol, entrypointPath);
-    await this.bundle(ufs, entrypointPath, distDir);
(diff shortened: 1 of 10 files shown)


bergundy (approved):

lorensr:
@Sushisource wondering if you have any ideas on this intermittent CI failure. In this test:

https://github.com/temporalio/sdk-typescript/blob/main/packages/test/src/test-replay.ts

we concurrently replay history on 3 workers. Sometimes on Windows, the second (`cancel-fake-progress-replay-nondeterministic`) times out after a minute:

```
@temporalio/test:   × Timed out while running tests
@temporalio/test:   1 tests were pending in lib\test-replay.js
@temporalio/test:   ( ) replay » cancel-fake-progress-replay-nondeterministic
```

Here's a failing example:

[build-and-test (14, windows-latest)](https://github.com/temporalio/sdk-typescript/runs/5636734388?check_suite_focus=true#logs)

You can search for the replay run id

```
a7e68cfc-0287-48dd-b86a-866caf5c9a73
```

to see the activations from test 1 and 3, but don't think anything for test 2.

Versus in the succeeding case:

[build-and-test (14, ubuntu-latest)](https://github.com/temporalio/sdk-typescript/runs/5636734327?check_suite_focus=true#logs)

We see test #2 activations (with `http` workflowType):

image

and later, the expected Nondeterminism error:

```
@temporalio/test: 2022-03-22T00:34:54.060Z [WARN] [temporal_sdk_core::worker] Failing workflow activation run_id="a7e68cfc-0287-48dd-b86a-866caf5c9a73" failure=Failure { failure: Some(Failure { message: "WorkflowUpdateError { source: Nondeterminism(\"No command scheduled for event HistoryEvent(id: 10, Some(ActivityTaskCancelRequested))\"), run_id: \"a7e68cfc-0287-48dd-b86a-866caf5c9a73\" }", source: "", stack_trace: "", cause: None, failure_info: Some(ApplicationFailureInfo(ApplicationFailureInfo { r#type: "", non_retryable: false, details: None })) }) }
```

Sushisource:
Looks like which test fails changes sometimes too. [Here](https://github.com/temporalio/sdk-typescript/runs/5636323459?check_suite_focus=true#step:11:272) with the debug logging forwarded it's the first one.

Anyway, AFAICT the problem at least in that build is that the workflow poll never happens? The only log lines to do with the test that didn't complete are:

```
2022-03-21T23:51:42.7267278Z @temporalio/test: 2022-03-21T23:51:42.116Z [INFO] [temporal_sdk_core] Registering replay worker task_queue="cancel-fake-progress-replay-0"
2022-03-21T23:51:44.6165103Z @temporalio/test: 2022-03-21T23:51:44.576Z [DEBUG] [temporal_sdk_core] poll_activity_task; task_queue="cancel-fake-progress-replay-0"
```

There should be a `poll_workflow_task` call as well (obviously). So why that's not happening, unsure. It's pretty much one of the first things that would get logged by Rust so I'd be surprised if we're getting stuck in there. TS not calling the poll function seems more likely but I have no idea why that'd be. Given it only happens on Windows I'm thinking maybe there's some weird Node bug or similar bit of weirdness going on.

bergundy:
@lorensr we should add more logs throughout the poll path on the node side and in the bridge and rerun CI until we find where the poll is dropped.

lorensr:
Doesn't poll because `workflowCreator` is undefined. My guess is `createBundle()` returns an empty string. Also:

```
Workflow bundle created { size: '0.00MB' }
```

Will add more logging

image

packages/worker/src/workflow/bundler.ts
```diff
@@ -120,12 +120,14 @@ export class WorkflowCodeBundler {
...
     // Cast because the type definitions are inaccurate
-    ufs.use(memfs.createFsFromVolume(vol) as any).use({ ...realFS, readdir: readdir as any });
+    const memoryFs = memfs.createFsFromVolume(vol);
+    ufs.use(memoryFs as any).use({ ...realFS, readdir: readdir as any });
     const distDir = '/dist';
     const entrypointPath = this.makeEntrypointPath(ufs, this.workflowsPath);
 
     this.genEntrypoint(vol, entrypointPath);
-    await this.bundle(ufs, entrypointPath, distDir);
+    await this.bundle(ufs, memoryFs, entrypointPath, distDir);
```
lorensr, line 129: @bergundy tests have passed 4 times in a row after adding this change—telling webpack to use `memoryFs` instead of `ufs` as `outputFileSystem`. That sound right? If yes, I'll remove logs and DEBUG level/tracing and merge.

bergundy: I’m wondering why that fixes the problem and why it only happens on Windows.
What happens if you list /dist? Does that directory exist in the real FS?
Anyways the fix seems fine so I’m approving

bergundy (approved):

packages/worker/src/workflow/bundler.ts
lorensr: I don't know why, it was most related thing I could think of. 

Tried stating, ENOENT in all tests:

image

[build-and-test (16, windows-latest)](https://github.com/temporalio/sdk-typescript/runs/5726434268?check_suite_focus=true#logs)

[2cb3ece](https://github.com/temporalio/sdk-typescript/pull/554/commits/2cb3ecec16c1deffac60133014fe398c0a900905)

Will add a comment linking to this PR, revert logs, and merge.

bergundy: We don't have to keep investigating, don't know how much we're getting from finding out the root cause.

One thing that could be *nice* to verify is whether webpack writes to the output to the real file system in the failing tests.

mjameswh: >What happens if you list /dist? Does that directory exist in the real FS?

FYI: One of our developper who works on Windows got this issue this morning (a kitchen sink project that had no been updated to latest releases). Yes, he ended up with a `c:\dist\main.js` file on his machine, and once that happened, `bundleWorkflowCode()` would always return the content of that (stall) file instead of the actual result of bundling. Had to delete that file from his disk to get bundling to work properly again.

bergundy: Interesting, can you confirm that with newer SDK version presence of `c:\dist\main.js` doesn't affect bundling?

mjameswh: See #619.

PR #558, opened 2022-03-25
fix: Throw UnsupportedTypeError from CompositePayloadConverter

lorensr:
Previously, `JsonPayloadConverter` throws, which prevents it from being anything but last in a `CompositePayloadConverter` order. `CompositePayloadConverter` also returned `undefined` when unable to convert. 

Now,  `JsonPayloadConverter` returns `undefined`, so it can be used in any order, and `CompositePayloadConverter` throws `UnsupportedTypeError` when unable to convert.


packages/common/src/converter/payload-converters.ts
```diff
@@ -1,11 +1,65 @@
...
+ *
+ * Converts Payloads to values based on the `Payload.metadata.encoding` field, which matches the {@link PayloadConverterWithEncoding.encodingType}
+ * of the converter that created the Payload.
+ */
+export class CompositePayloadConverter implements PayloadConverter {
+  readonly converters: PayloadConverterWithEncoding[];
+  readonly converterByEncoding: Map<string, PayloadConverterWithEncoding> = new Map();
+
+  constructor(...converters: PayloadConverterWithEncoding[]) {
+    this.converters = converters;
```
bergundy, line 20: Let’s verify that this in a non empty array

packages/internal-workflow-common/src/errors.ts
```diff
@@ -6,12 +6,8 @@ export class PayloadConverterError extends ValueError {
   public readonly name: string = 'PayloadConverterError';
 }
 
-export class UnsupportedJsonTypeError extends PayloadConverterError {
-  public readonly name: string = 'UnsupportedJsonTypeError';
-
-  constructor(message: string | undefined, public readonly cause?: Error) {
-    super(message ?? undefined);
-  }
+export class UnsupportedTypeError extends PayloadConverterError {
```
bergundy, line 9: Why not reuse TypeError?

bergundy (approved):
Some minor comments.
I like what you did with the json error

packages/internal-workflow-common/src/errors.ts
lorensr: Generally, I'd make more specific errors in case a dev would want to do `instanceof` to differentiate. In this case, I don't imagine that being common—would be useful if the dev doesn't know what type of arg they're sending (eg it's provided by the user or something).

lorensr:
- Changed `PayloadConverter` interface:

```
-  public abstract toPayload<T>(value: T): Payload | undefined;
+  public abstract toPayload<T>(value: T): Payload;
```

So that implementors don't think that returning `undefined` is valid. 

- Further, added check that custom `PayloadConverter`s toPayload returns a Payload-like object (containing either a `metadata` or `data` field).

packages/workflow/src/worker-interface.ts
```diff
@@ -128,7 +129,7 @@ export async function initRuntime({ info, randomnessSeed, now, patches }: Workfl
   const customPayloadConverter = (await import('__temporal_custom_payload_converter')).payloadConverter;
   // The `payloadConverter` export is validated in the Worker
   if (customPayloadConverter !== undefined) {
-    state.payloadConverter = customPayloadConverter;
+    state.payloadConverter = new WrappedCustomPayloadConverter(customPayloadConverter);
```
lorensr, line 132: I missed this the first go around. Eventually found it through the `toPayload throwing results in task retry` test

lorensr:
Per further discussion:
- Changed signature back to `Payload | undefined`
- Removed Wrapped class, instead:
  - throw from toPayload helper
  - searched for all instances of `.toPayload` and switched non-test instances to use the helper

packages/common/src/converter/payload-converters.ts
```diff
@@ -87,3 +141,27 @@ export class BinaryPayloadConverter implements PayloadConverterWithEncoding {
...
+
+/**
+ * The default {@link PayloadConverter} used by the SDK.
+ * Supports `Uint8Array` and JSON serializables (so if [`JSON.stringify(yourArgOrRetval)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description) works, the default payload converter will work).
+ *
+ * To also support Protobufs, create a custom payload converter with {@link DefaultPayloadConverter}:
+ *
+ * `const myConverter = new DefaultPayloadConverter({ protobufRoot })`
+ */
+export const defaultPayloadConverter = new WrappedPayloadConverter(new DefaultPayloadConverter());
```
bergundy, line 167: Let's not wrap the default, instead I'd wrap in the worker / workflow runtime / client, that way any converter we accept would be wrapped.

packages/common/src/converter/payload-converters.ts
```diff
@@ -1,19 +1,75 @@
-import { errorMessage, UnsupportedJsonTypeError, ValueError } from '@temporalio/internal-workflow-common';
```
bergundy, line 1: All of the helper methods here should accept WrappedPayloadConverter to make sure we never use an unwrapped one IMHO.
WDYT?

bergundy (commented):
Skimmed this, overall LGTM, had a couple comments.

packages/common/src/converter/payload-converters.ts
lorensr: Like this?

https://github.com/temporalio/sdk-typescript/pull/558/commits/a59fe705c366852c043ffb4dcba6338c2847187a

packages/common/src/converter/payload-converters.ts
```diff
@@ -87,3 +141,27 @@ export class BinaryPayloadConverter implements PayloadConverterWithEncoding {
...
+
+export const searchAttributePayloadConverter = new WrappedPayloadConverter(new JsonPayloadConverter());
+
+export class DefaultPayloadConverter extends CompositePayloadConverter {
+  // Match the order used in other SDKs, but exclude Protobuf converters so that the code, including
+  // `proto3-json-serializer`, doesn't take space in Workflow bundles that don't use Protobufs. To use Protobufs, use
+  // {@link DefaultPayloadConverterWithProtobufs}.
+  //
+  // Go SDK:
+  // https://github.com/temporalio/sdk-go/blob/5e5645f0c550dcf717c095ae32c76a7087d2e985/converter/default_data_converter.go#L28
```
bergundy, line 153: Do we need this link?

packages/internal-non-workflow-common/src/otel.ts
```diff
@@ -41,7 +44,7 @@ export async function extractSpanContextFromHeaders(headers: Headers): Promise<o
 export async function headersWithContext(headers: Headers): Promise<Headers> {
   const carrier = {};
   otel.propagation.inject(otel.context.active(), carrier, otel.defaultTextMapSetter);
-  return { ...headers, [TRACE_HEADER]: toPayload(defaultPayloadConverter, carrier) };
+  return { ...headers, [TRACE_HEADER]: toPayload(payloadConverter, carrier) };
```
bergundy, line 47: Eventually I think we should just make interceptor headers `Record<string, unknown>` and apply the converter in framework code.
Not 100% sure about that though.

packages/test/src/test-custom-payload-converter.ts
```diff
@@ -56,6 +58,122 @@ if (RUN_INTEGRATION_TESTS) {
...
+      }),
+      {
+        instanceOf: ValueError,
+        message:
+          'The Payload Converter method TestPayloadConverter.toPayload must return a Payload. Received `undefined` of type `undefined` when trying to convert `hello` of type `string`.',
+      }
+    );
+  });
+
+  test('fromPayload throws from client.execute', async (t) => {
```
bergundy, line 150: I'd just add a clarification that throwing is done when getting the result.
Not critical.

bergundy (approved):
Great stuff!

packages/common/src/converter/payload-converters.ts
lorensr: I like it as a reference to what the order should be. Until we have a spec 😄

PR #560, opened 2022-03-25
use example with both signals and queries for Workflow "Signals and Queries" section

vkarpov15:
## What was changed

https://typescript.temporal.io/api/namespaces/workflow/#signals-and-queries is somewhat confusing in the docs because it doesn't include how to define a query, and it defines a signal in a promise executor.

## Why?

Causes some confusion if this is the first example of defining signals and queries you see.

## Checklist

1. Closes 

2. How was this tested:

I've actually been unable to test this, I haven't been able to figure out how to build the docs locally. Hoping that this PR comes with a Netlify deploy preview.

3. Any docs updates needed?


lorensr (approved):
Thanks! Looks good. Prettier:

`npm run format`

vkarpov15:
Thanks @lorensr .

I got the updated docs running locally, here's what they look like:

image

PR #565, opened 2022-03-28
fix(worker): Fix crash when Activity is cancelled with reason NOT_FOUND

bergundy:


packages/worker/src/worker.ts
```diff
@@ -618,10 +618,11 @@ export class Worker {
                     // NOTE: activity will not be considered cancelled until it confirms cancellation
                     this.log.debug('Cancelling activity', { taskToken: base64TaskToken });
                     span.setAttribute('found', true);
-                    if (!task.cancel?.reason) {
+                    const reason = task.cancel?.reason;
+                    if (reason === undefined || reason === null) {
```
Sushisource, line 622: Is this just because it was an empty string?

Sushisource (approved):

lorensr (approved):

packages/worker/src/worker.ts
bergundy: The enum value for `NOT_FOUND` is `0`, I overlooked that.

PR #568, opened 2022-03-30
feat(worker)!: Refactor Core bridge for top level Worker API

bergundy:
Implements the Worker part of [proposal #56](https://github.com/temporalio/proposals/blob/050825aba0e2e6cde91bae81945dce082bd47622/typescript/connections.md)
- Break down `Core` into `Runtime` and `NativeConnection`
- Workers now require a `NativeConnection` instance instead of using the singleton `Core` connection
- `Runtime` may be removed in the future but remains in the SDK to enable log forwarding from Core
- By default we don't forward Core logs into node anymore

This PR uses `sdk-core/master` which unblocks local activities and upsert search attributes, the last missing pieces for a feature complete SDK (apart from stack trace query).


packages/worker/src/runtime.ts
```diff
@@ -137,51 +121,38 @@ export class Core {
    * a new Core instance will be installed and configured to connect to
    * a local server.
    */
-  public static async instance<T extends typeof Core>(this: T): Promise<InstanceType<T>> {
```
bergundy, line 140: Considering making this async in case init will become async in the future.
Another option would be to remove log forwarding altogether and moving the logger into the Worker.
Not 100% sure what to do with this.
Could be in the next PR though but should be addressed before we make a release.

packages/core-bridge/src/lib.rs
```diff
@@ -202,6 +205,7 @@ where
     Ok(())
 }
 
+/// Call [callback] with an UnexpectedError created from [err]
```
Sushisource, line 208: `[foo]` is for doc links, args can't be linked to, so these should just be surrounded with `

Sushisource (approved):
Just a few fairly minor things. LGTM

packages/test/src/test-payload-converter.ts
```diff
@@ -31,39 +31,39 @@ import { protobufWorkflow } from './workflows';
 
 test('UndefinedPayloadConverter converts from undefined only', async (t) => {
   const converter = new UndefinedPayloadConverter();
-  t.is(await converter.toPayload(null), undefined);
-  t.is(await converter.toPayload({}), undefined);
-  t.is(await converter.toPayload(1), undefined);
-  t.is(await converter.toPayload(0), undefined);
-  t.is(await converter.toPayload('abc'), undefined);
+  t.is(converter.toPayload(null), undefined);
```
lorensr, line 34: I didn't merge #558 fast enough! 😄

packages/testing/src/index.ts
```diff
@@ -67,8 +67,20 @@ export class TestWorkflowEnvironment {
    */
   public readonly workflowClient: WorkflowClient;
 
-  protected constructor(protected readonly serverProc: ChildProcess, connection: Connection) {
+  /**
+   * A {@link NativeConnection} for interacting with the test server.
+   *
+   * Use this client when creating Workers for testing.
```
lorensr, line 73:
```suggestion
   * Use this connection when creating Workers for testing.
```

packages/worker/src/connection-options.ts
```diff
@@ -6,19 +6,14 @@ type TLSConfig = native.TLSConfig;
 
 export { TLSConfig };
 
-export interface ServerOptions {
+export interface NativeConnectionOptions {
```
lorensr, line 9: Would it be helpful at all for the two connection options extend a generic interface?

```
export sharedConnectionOptions: GenericConnectionOptions = { address: ... };
```

Versus:

```
export sharedConnectionOptions = { address: ... };
```

packages/worker/src/connection.ts
```diff
@@ -0,0 +1,27 @@
+import { Client } from '@temporalio/core-bridge';
+import { NativeConnectionOptions } from './connection-options';
+import { Runtime } from './runtime';
+
+/**
+ * A Native Connection object that delegates calls to the Rust Core binary extension.
+ *
+ * A Worker must use this class to connect to the server.
+ *
+ * Do not confuse this Connection class with the `@temporalio/client`'s Connection.
```
lorensr, line 10:
```suggestion
 * Do not confuse this connection class with `@temporalio/client`'s {@link Connection}.
```

packages/worker/src/worker-options.ts
```diff
@@ -20,6 +21,20 @@ export function isPathBundleOption(bundleOpt: WorkflowBundle): bundleOpt is { pa
...
+   * A connected {@link NativeConnection} instance.
+   *
+   * If not provided, the worker will default to connect insecurely to `localhost:7233`.
+   */
+  connection?: NativeConnection;
+
+  /**
+   * The namespace this worker will connect to
+   *
+   * @default default
```
lorensr, line 34:
```suggestion
   * @default `"default"`
```

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -239,14 +265,15 @@ export class WorkflowCodecRunner {
...
+                          headers: noopEncodeMap(command.signalExternalWorkflowExecution.headers),
                         }
                       : undefined,
                     scheduleLocalActivity: command.scheduleLocalActivity
                       ? {
                           ...command.scheduleLocalActivity,
                           arguments: await encodeOptional(this.codec, command.scheduleLocalActivity.arguments),
                           // don't encode headers
-                          headerFields: noopEncodeMap(command.scheduleLocalActivity.headerFields),
+                          headers: noopEncodeMap(command.scheduleLocalActivity.headers),
```
lorensr, line 276: Yay all `headers`!

lorensr (approved):
`x` typo. Also could say what to use it for, and/or "Use `nativeConnection` when creating a Worker"

image

packages/core-bridge/src/lib.rs
```diff
@@ -71,77 +77,74 @@ enum Request {
...
     /// A request to drain logs from core so they can be emitted in node
     PollLogs {
         /// Logs are sent to this function
         callback: Root<JsFunction>,
     },
 }
 
-#[derive(Clone)]
-struct CoreHandle {
+struct ThreadRuntime {
```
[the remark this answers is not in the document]
bergundy: Sure

packages/worker/src/runtime.ts
bergundy: Anything that would require calling into the bridge loop would make this async.
Changing it later from `sync` to `async` will be painful for users.

packages/testing/src/index.ts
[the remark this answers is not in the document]
bergundy: It's because I renamed it a few times while working on this PR so there's bound to be some overlooked leftovers.

bergundy:
Merging this, sdk-features build passed [here](https://github.com/temporalio/sdk-features/runs/5781640763?check_suite_focus=true) with [these changes](https://github.com/temporalio/sdk-features/pull/32)

PR #579, opened 2022-04-08
feat(worker): Cancel activities on shutdown

bergundy:


packages/worker/src/worker.ts
```diff
@@ -506,7 +511,25 @@ export class Worker {
...
+          this.activityPollerStateSubject.pipe(
+            // Core has indicated that it will not return any more poll results, evict all cached WFs
+            filter((state) => state !== 'POLLING'),
+            first(),
+            map((): ActivityTaskWithContext => {
+              return {
+                parentSpan: this.tracer.startSpan('activity.shutdown.evict'),
+                task: new coresdk.activity_task.ActivityTask({
+                  // NOTE: taskToken is not passed here
+                  // TODO: use shutdown as the reason
```
bergundy, line 525: I still need to address this TODO

packages/worker/src/worker.ts
```diff
@@ -613,14 +637,16 @@ export class Worker {
...
                       break;
                     }
-                    // NOTE: activity will not be considered cancelled until it confirms cancellation
+                    // NOTE: activity will not be considered cancelled until it confirms cancellation (by throwing a CancelledFailure)
                     this.log.debug('Cancelling activity', { taskToken: base64TaskToken });
                     span.setAttribute('found', true);
                     const reason = task.cancel?.reason;
                     if (reason === undefined || reason === null) {
-                      throw new TypeError('Got cancel activity cancel task with no reason');
+                      // Special case of Lang side cancellation during shutdown (see above)
```
lorensr, line 645:
```suggestion
                      // Special case of Lang side cancellation during shutdown (see `activity.shutdown.evict` above)
```

lorensr (approved):

PR #585, opened 2022-04-12
feat!: Implement local activities

bergundy:
BREAKING CHANGE: ActivityOptions no longer takes a `namespace`

### DIFF
--- a/packages/internal-workflow-common/src/activity-options.ts
+++ b/packages/internal-workflow-common/src/activity-options.ts
@@ -85,7 +85,7 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
  */
 export interface LocalActivityOptions {
   /**
-   * RetryPolicy that define how activity is retried in case of failure. If this is not set, then the SDK-defined default activity retry policy will be used.
+   * RetryPolicy that defines how an activity is retried in case of failure. If this is not set, then the SDK-defined default activity retry policy will be used.
    * Note that local activities are always executed at least once, even if maximum attempts is set to 1 due to Workflow task retries.
    */
   retry?: RetryPolicy;
--- a/packages/internal-workflow-common/src/activity-options.ts
+++ b/packages/internal-workflow-common/src/activity-options.ts
@@ -13,8 +13,7 @@ export enum ActivityCancellationType {
 checkExtends<coresdk.workflow_commands.ActivityCancellationType, ActivityCancellationType>();
 
 /**
- * Options for remote activity invocation - will be processed from a task queue.
- * @see https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/activity/ActivityOptions.Builder.html
+ * Options for remote activity invocation
  */
 export interface ActivityOptions {
   /**
@@ -25,12 +24,6 @@ export interface ActivityOptions {
    * @default an incremental sequence number
    */
   activityId?: string;
-
-  /**
-   * Namespace to schedule this activity in.
-   * @default current worker namespace
-   */
-  namespace?: string;
 
   /**
    * Task queue name.
@@ -75,6 +68,80 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
    * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
    */
   scheduleToCloseTimeout?: string | number;
+
+  /**
+   * Determines what the SDK does when the Activity is cancelled.
+   * - `TRY_CANCEL` - Initiate a cancellation request and immediately report cancellation to the workflow.
+   * - `WAIT_CANCELLATION_COMPLETED` - Wait for activity cancellation completion. Note that activity must heartbeat to receive a
+   *   cancellation notification. This can block the cancellation for a long time if activity doesn't
+   *   heartbeat or chooses to ignore the cancellation request.
+   * - `ABANDON` - Do not request cancellation of the activity and immediately report cancellation to the workflow.
+   */
+  cancellationType?: coresdk.workflow_commands.ActivityCancellationType;
+}
+
+/**
+ * Options for local activity invocation
+ */
+export interface LocalActivityOptions {
+  /**
+   * Identifier to use for tracking the activity in Workflow history.
+   * The `activityId` can be accessed by the activity function.
+   * Does not need to be unique.
+   *
+   * @default an incremental sequence number
+   */
+  activityId?: string;
+
+  /**
+   * RetryPolicy that define how activity is retried in case of failure. If this is not set, then the server-defined default activity retry policy will be used. To ensure zero retries, set maximum attempts to 1.
+   */
+  retry?: RetryPolicy;
+
+  /**
+   * Maximum time the local activity is allowed to execute after the task is dispatched. This
+   * timeout is always retryable.
+   *
+   * Either this option or {@link scheduleToCloseTimeout} is required.
+   * If set, this must be <= {@link scheduleToCloseTimeout}, otherwise, it will be clamped down.
+   *
+   * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
+   */
+  startToCloseTimeout?: string | number;
+
+  /**
+   * Limits time the local activity can idle internally before being executed. That can happen if
+   * the worker is currently at max concurrent local activity executions. This timeout is always
+   * non retryable as all a retry would achieve is to put it back into the same queue. Defaults
+   * to {@link scheduleToCloseTimeout} if not specified and that is set. Must be <=
+   * {@link scheduleToCloseTimeout} when set, otherwise, it will be clamped down.
+   *
+   * @default unlimited
+   * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
+   */
+  scheduleToStartTimeout?: string | number;
+
+  /**
+   * Indicates how long the caller is willing to wait for local activity completion. Limits how
+   * long retries will be attempted. When not specified defaults to the workflow execution
+   * timeout (which may be unset).
+   *
+   * Either this option or {@link startToCloseTimeout} is required.
+   *
+   * @default unlimited
+   * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
+   */
+  scheduleToCloseTimeout?: string | number;
+
+  /**
+   * If the activity is retrying and backoff would exceed this value, lang will be told to
+   * schedule a timer and retry the activity after. Otherwise, backoff will happen internally in
+   * core.
+   *
+   * @default 1 minute
+   * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
+   **/
+  localRetryThreshold?: string | number;
 
   /**
    * Determines what the SDK does when the Activity is cancelled.
--- a/packages/internal-workflow-common/src/activity-options.ts
+++ b/packages/internal-workflow-common/src/activity-options.ts
@@ -85,16 +85,8 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
  */
 export interface LocalActivityOptions {
   /**
-   * Identifier to use for tracking the activity in Workflow history.
-   * The `activityId` can be accessed by the activity function.
-   * Does not need to be unique.
-   *
-   * @default an incremental sequence number
-   */
-  activityId?: string;
-
-  /**
-   * RetryPolicy that define how activity is retried in case of failure. If this is not set, then the server-defined default activity retry policy will be used. To ensure zero retries, set maximum attempts to 1.
+   * RetryPolicy that define how activity is retried in case of failure. If this is not set, then the SDK-defined default activity retry policy will be used.
+   * Note that local activities are always executed at least once, even if maximum attempts is set to 1 due to Workflow task retries.
    */
   retry?: RetryPolicy;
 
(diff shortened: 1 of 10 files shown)


packages/internal-workflow-common/src/activity-options.ts
```diff
@@ -87,6 +80,80 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
...
+
+  /**
+   * If the activity is retrying and backoff would exceed this value, lang will be told to
+   * schedule a timer and retry the activity after. Otherwise, backoff will happen internally in
+   * core.
+   *
+   * @default 1 minute
+   * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
+   **/
+  localRetryThreshold?: string | number;
```
lorensr, line 144: When/why would a user want to set this?

lorensr, line 144: When/why would a user want to set this?

lorensr (commented):
partial review, will finish tomorrow

packages/internal-workflow-common/src/activity-options.ts
```diff
@@ -26,12 +25,6 @@ export interface ActivityOptions {
    */
   activityId?: string;
 
-  /**
-   * Namespace to schedule this activity in.
-   * @default current worker namespace
-   */
-  namespace?: string;
```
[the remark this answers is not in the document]
bergundy: Wasn't supposed to be exposed in the first place

packages/internal-workflow-common/src/activity-options.ts
bergundy: It's an advanced feature, in general local activity execution should be very short.

packages/internal-workflow-common/src/activity-options.ts
```diff
@@ -87,6 +80,80 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
...
+ */
+export interface LocalActivityOptions {
+  /**
+   * Identifier to use for tracking the activity in Workflow history.
+   * The `activityId` can be accessed by the activity function.
+   * Does not need to be unique.
+   *
+   * @default an incremental sequence number
+   */
+  activityId?: string;
```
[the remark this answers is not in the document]
bergundy: I originally had that (months ago) but it makes it harder to see all of the options when going to definition and the docs.

packages/test/src/workflows/local-activity-testers.ts
```diff
@@ -0,0 +1,97 @@
...
+  await activityNotFound();
+}
+
+export const interceptors: wf.WorkflowInterceptorsFactory = () => {
+  return {
+    outbound: [
+      {
+        async startTimer(input, next) {
+          const { test } = wf.proxySinks();
+          await next(input);
```
lorensr, line 83: When this resolves, start timer request has been received by core or Server, or timer has gone off / completed? I think former, but that feels different from the `next`s in `scheduleActivity`, so I'll document it.

lorensr (approved):

packages/internal-workflow-common/src/activity-options.ts
lorensr: It just controls when the timer is in lang vs core? Could we add to the comment why that's impt / why I'd want to change it?

packages/internal-workflow-common/src/activity-options.ts
```diff
@@ -87,6 +80,80 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
...
+
+  /**
+   * RetryPolicy that define how activity is retried in case of failure. If this is not set, then the server-defined default activity retry policy will be used. To ensure zero retries, set maximum attempts to 1.
```
cretz, line 97: Is this server-defined or core-defined? At least in Go, I think the local activity retry options are defaults set in the SDK.

packages/internal-workflow-common/src/activity-options.ts
```diff
@@ -87,6 +80,80 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
...
+   * Limits time the local activity can idle internally before being executed. That can happen if
+   * the worker is currently at max concurrent local activity executions. This timeout is always
+   * non retryable as all a retry would achieve is to put it back into the same queue. Defaults
+   * to {@link scheduleToCloseTimeout} if not specified and that is set. Must be <=
+   * {@link scheduleToCloseTimeout} when set, otherwise, it will be clamped down.
+   *
+   * @default unlimited
+   * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
+   */
+  scheduleToStartTimeout?: string | number;
```
cretz, line 122: This doesn't exist for local activities in Java and Go (but if it has value in TS, cool)

packages/internal-workflow-common/src/activity-options.ts
```diff
@@ -87,6 +80,80 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
...
+
+  /**
+   * Determines what the SDK does when the Activity is cancelled.
+   * - `TRY_CANCEL` - Initiate a cancellation request and immediately report cancellation to the workflow.
+   * - `WAIT_CANCELLATION_COMPLETED` - Wait for activity cancellation completion. Note that activity must heartbeat to receive a
+   *   cancellation notification. This can block the cancellation for a long time if activity doesn't
+   *   heartbeat or chooses to ignore the cancellation request.
+   * - `ABANDON` - Do not request cancellation of the activity and immediately report cancellation to the workflow.
+   */
+  cancellationType?: coresdk.workflow_commands.ActivityCancellationType;
```
cretz, line 154: This doesn't exist for local activities in Java and Go (but if it has value in TS, cool)

packages/internal-workflow-common/src/activity-options.ts
cretz, line 94: This doesn't exist for local activities in Java and Go (but if it has value in TS, cool)

packages/internal-workflow-common/src/activity-options.ts
Sushisource: Yeah, server-defined here does not make sense.

packages/internal-workflow-common/src/activity-options.ts
Sushisource: It's of dubious value, but, I implemented it mostly because it lets us more easily combine LA/nonLA eventually. IMO if we just don't expose it in TS for now that's completely legit.

packages/internal-workflow-common/src/activity-options.ts
Sushisource: The difference is actually whether the backoff is local or uses a server-side timer. This docstring should make that more explicitly clear. You can kinda tell from "schedule a timer" if you have a really well defined sense of what _schedule_ means in this context, but most users won't. 

Let's be more clear it means timer like a Temporal server timer.

packages/internal-workflow-common/src/activity-options.ts
Sushisource: Same reasoning as the other thing, but, arguably this is actually meaningfully useful. I think it's worth exposing.

packages/workflow/src/workflow.ts
```diff
@@ -165,7 +176,63 @@ async function scheduleActivityNextHandler({
...
+          requestCancelLocalActivity: {
+            seq,
+          },
+        });
+      });
+    }
+    state.completions.activity.set(seq, {
+      resolve,
+      reject,
+    });
```
Sushisource, line 221: This appears to be an exact dupe of the normal activity cancel handler stuff. Could be deduped.

packages/workflow/src/workflow.ts
```diff
@@ -193,6 +260,57 @@ export function scheduleActivity<R>(activityType: string, args: any[], options:
...
+  args: any[],
+  options: LocalActivityOptions
+): Promise<R> {
+  if (options === undefined) {
+    throw new TypeError('Got empty activity options');
+  }
+
+  let originalScheduleTime = undefined;
+
+  // TODO: is the inital attempt 0 or 1?
```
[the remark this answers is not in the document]
Sushisource: It should be 1

packages/workflow/src/workflow.ts
```diff
@@ -193,6 +260,57 @@ export function scheduleActivity<R>(activityType: string, args: any[], options:
...
+        seq,
+        attempt,
+        originalScheduleTime,
+      })) as Promise<R>;
+    } catch (err) {
+      if (err instanceof LocalActivityDoBackoff) {
+        await sleep(tsToMs(err.backoff.backoffDuration));
+        if (typeof err.backoff.attempt !== 'number') {
+          throw new TypeError('Invalid backoff attempt type');
+        }
```
Sushisource, line 304: Shouldn't this check come before the sleep?

packages/internal-workflow-common/src/activity-options.ts
bergundy: Hmmm.. doesn't look like it's recorded anywhere, might be worth removing this.
@Sushisource

packages/internal-workflow-common/src/activity-options.ts
bergundy: Hmmm.. probably copied mindlessly, will fix.

packages/internal-workflow-common/src/activity-options.ts
bergundy: Next time I won't copy your docstrings as is @Sushisource ;-p

packages/internal-workflow-common/src/activity-options.ts
bergundy: +1 to exposing in other SDKs

packages/test/src/test-local-activities.ts
```diff
@@ -0,0 +1,312 @@
...
+      const res = await client.execute(workflows.runOneLocalActivity, {
+        workflowId: uuid4(),
+        taskQueue,
+        args: ['message'],
+      });
+      t.is(res, 'messagemessage');
+    });
+  });
+
+  // TODO: fix Core shutdown and reenable this test
```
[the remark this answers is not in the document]
bergundy: No this is a bug in Core

packages/test/src/workflows/local-activity-testers.ts
bergundy: Timer has been completed when this resolves.
The name is of this method is misleading.

packages/workflow/src/workflow.ts
bergundy: True, originally I had `completions.localActivity` but realized that LAs are resolved the same way as non-LAs.
I'll dedupe.

packages/workflow/src/workflow.ts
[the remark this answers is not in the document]
bergundy: Thanks, I was hoping you'd see this comment @Sushisource

packages/workflow/src/workflow.ts
bergundy: Technically this should never happen. It's here because all proto fields are optional but you're right that it'd be more "correct" to sleep after validating the response.

packages/test/src/test-local-activities.ts
```diff
@@ -0,0 +1,312 @@
...
+      shutdownGraceTime: '10s',
+    });
+    const handle = await client.start(workflows.cancelALocalActivity, {
+      workflowId: uuid4(),
+      taskQueue,
+      workflowTaskTimeout: '3s',
+    });
+    const p = worker.run();
+    await firstValueFrom(subj);
+    worker.shutdown();
```
lorensr, line 288: Since firstValueFrom resolves when timer is done, and when timer is done, activity is canceled, does the cancel or the shutdown happen first? 

(I'm wanting to make sure that the activity is still running so that this test tests what it's try to test 😄)

packages/internal-workflow-common/src/activity-options.ts
bergundy: Removing.

packages/internal-workflow-common/src/activity-options.ts
bergundy: Changed the comment (not pushed yet) and mentioned that even if max attempts is 1 the LA might be executed more than once due to WFT retries.

packages/internal-workflow-common/src/activity-options.ts
bergundy: Let's keep this for uniformity.

packages/test/src/test-local-activities.ts
bergundy: We shutdown Core while just before requesting to cancel the local activity.
We expect it to shutdown once the workflow and local activity are complete.

cretz (approved):

packages/workflow/src/workflow.ts
bergundy: The shared code is a bit harder to follow I'll push it anyways

bergundy: Ended up duplicating the code, I immediately created a bug when trying to dedupe.
I think this is the right approach for now.

packages/internal-workflow-common/src/activity-options.ts
```diff
@@ -87,6 +80,71 @@ Note that the Temporal Server doesn't detect Worker process failures directly. I
   cancellationType?: coresdk.workflow_commands.ActivityCancellationType;
 }
 
+/**
+ * Options for local activity invocation
+ */
+export interface LocalActivityOptions {
+  /**
+   * RetryPolicy that define how activity is retried in case of failure. If this is not set, then the SDK-defined default activity retry policy will be used.
```
Sushisource, line 88:
```suggestion
   * RetryPolicy that defines how an activity is retried in case of failure. If this is not set, then the SDK-defined default activity retry policy will be used.
```

PR #590, opened 2022-04-14
fix: Compile linux binary linked to older glibc

bergundy:


Sushisource (approved):

lorensr:
👍 Now version is < buster 

image

image

bergundy:
Once the compile-binaries step completes I'll verify on my mac using: `docker run --rm -it --platform linux/amd64 debian:buster`

PR #591, opened 2022-04-14
Export nodejsBuiltinModules

lorensr:
For use in https://github.com/temporalio/samples-typescript/pull/133

```ts
import { nodejsBuiltinModules } from '@temporalio/worker/lib/workflow/bundler'
```


bergundy (approved):
I'm wondering if there's a better source for this list.
Make sure to also disallow these modules with `node:` prefix.

packages/worker/src/index.ts
```diff
@@ -8,13 +8,13 @@
...
+export * from './interceptors';
 export * from './logger';
+export { History, Runtime, RuntimeOptions, TelemetryOptions } from './runtime';
 export * from './sinks';
-export * from './interceptors';
-export { bundleWorkflowCode, BundleOptions } from './workflow/bundler';
-export * from './errors';
+export { DataConverter, defaultPayloadConverter, errors, State, Worker } from './worker';
+export { CompiledWorkerOptions, WorkerOptions } from './worker-options';
+export { BundleOptions, bundleWorkflowCode } from './workflow/bundler';
```
vkarpov15, line 20: Shouldn't this also export `nodejsBuiltinModules` or am I missing something?

vkarpov15:
@bergundy @lorensr it looks like [Sindre has a lib for this](https://github.com/sindresorhus/builtin-modules). Take a closer look at the source code - looks like pulling Node builtins is a one-liner with `require('module').builtInModules`

```
$ node
Welcome to Node.js v14.18.2.
Type ".help" for more information.
> const module = require('module')
undefined
> module.builtinModules
[
  '_http_agent',       '_http_client',        '_http_common',
  '_http_incoming',    '_http_outgoing',      '_http_server',
  '_stream_duplex',    '_stream_passthrough', '_stream_readable',
  '_stream_transform', '_stream_wrap',        '_stream_writable',
  '_tls_common',       '_tls_wrap',           'assert',
  'async_hooks',       'buffer',              'child_process',
  'cluster',           'console',             'constants',
  'crypto',            'dgram',               'diagnostics_channel',
  'dns',               'domain',              'events',
  'fs',                'fs/promises',         'http',
  'http2',             'https',               'inspector',
  'module',            'net',                 'os',
  'path',              'perf_hooks',          'process',
  'punycode',          'querystring',         'readline',
  'repl',              'stream',              'string_decoder',
  'sys',               'timers',              'tls',
  'trace_events',      'tty',                 'url',
  'util',              'v8',                  'vm',
  'wasi',              'worker_threads',      'zlib'
]
> 

```

Here's the [Node docs on `builtinModules`](https://nodejs.org/api/module.html#modulebuiltinmodules)

lorensr:
Thanks @vkarpov15 I'll use that. It has all we have plus subpath exports like `fs/promises`.

vkarpov15:
@lorensr it might be better to export a list of Node.js builtins that the TypeScript SDK injects a replacement for. For now, just `export const ['assert']`. That way, in https://github.com/temporalio/samples-typescript/pull/133, we could do something like `require('module').builtinModules.filter(module => !sdkInjectsReplacement.includes(module))`

packages/worker/src/index.ts
lorensr: Intended usage is:

```ts
import { nodejsBuiltinModules } from '@temporalio/worker/lib/workflow/bundler'
```

Would export from index if wanted users to import it, or to include in API reference

lorensr:
@vkarpov15  How's the current version of this PR?

vkarpov15 (approved):
Thanks @lorensr  :+1:

PR #598, opened 2022-04-19
chore(worker): Update swc-loader options

lorensr:
## What was changed

Original intent was to update `swc-loader` options to better match what we used to have (`ts-loader`). Didn't find much to do, and I think [original reported issue](https://temporalio.slack.com/archives/C01DKSMU94L/p1650267315905719?thread_ts=1649977054.551749&cid=C01DKSMU94L) was a red herring.

`swc` defaults, with my comments on whether they're defaults in `ts-loader`:

```ts
{
  "jsc": {
    "parser": {
      "syntax": "ecmascript",  // we're only using on .ts files, so safe to change to typescript
      "tsx": false, // default true, but needs to be .tsx file, which we haven't been supporting (`test: /\.ts$/,`)
      "dynamicImport": false, // true if target is 2020, which by default it isn't
      "privateMethod": false, // didn't find in tsc options https://www.typescriptlang.org/tsconfig
      "functionBind": false, // didn't find
      "exportDefaultFrom": false, // didn't find
      "exportNamespaceFrom": false, // didn't find
      "decorators": false, // not enabled by default, but we can enable just in case someone wants to use them. think doesn't hurt?
      "decoratorsBeforeExport": false, // two ways of decorating exports, leaving default
      "topLevelAwait": false, // didn't find
      "importMeta": false, // didn't find
      "preserveAllComments": false // enabled by default, but don't think it's necessary or helpful in our case? would be if people were looking at the bundle
    },
  }
}
```

## Checklist

2. How was this tested:

Just made sure it doesn't break our test bundle with `node packages/test/lib/run-a-worker.js`.

### DIFF
--- a/packages/test/src/workflows/swc.ts
+++ b/packages/test/src/workflows/swc.ts
@@ -1,0 +1,24 @@
+// eslint-disable-next-line @typescript-eslint/ban-types
+function sealed(constructor: Function) {
+  Object.seal(constructor);
+  Object.seal(constructor.prototype);
+}
+
+function decorate() {
+  console.log('decorate(): factory evaluated');
+  return function (_target: any, _propertyKey: string, _descriptor: PropertyDescriptor) {
+    console.log('decorate(): called');
+  };
+}
+
+@sealed
+class ExampleClass {
+  @decorate()
+  method() {
+    return 'hi';
+  }
+}
+
+export async function swc(): Promise<void> {
+  new ExampleClass().method();
+}
--- a/scripts/clean.mjs
+++ b/scripts/clean.mjs
@@ -54,7 +54,13 @@ function cleanProtoGeneratedFiles() {
 function cleanCompiledRustFiles() {
   const bridgeDir = resolve(packagesPath, 'core-bridge');
   console.log('Cleaning compiled rust files');
-  rmSync(resolve(bridgeDir, 'releases'), { recursive: true });
+  try {
+    rmSync(resolve(bridgeDir, 'releases'), { recursive: true });
+  } catch (e) {
+    if (e?.code !== 'ENOENT') {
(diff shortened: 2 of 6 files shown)


packages/test/src/workflows/swc.ts
```diff
@@ -0,0 +1,24 @@
+// eslint-disable-next-line @typescript-eslint/ban-types
+function sealed(constructor: Function) {
```
bergundy, line 2: nit: can use something like this and avoid using `Function`:
```ts
interface Constructor {
    new (): any;
}
```

packages/test/src/workflows/swc.ts
```diff
@@ -0,0 +1,24 @@
...
+  Object.seal(constructor);
+  Object.seal(constructor.prototype);
+}
+
+function decorate() {
+  // console.log('decorate(): factory evaluated');
+  return function (_target: any, _propertyKey: string, _descriptor: PropertyDescriptor) {
+    // console.log('decorate(): called');
```
bergundy, line 10: leftover logs

scripts/clean.mjs
```diff
@@ -54,7 +54,13 @@ function cleanProtoGeneratedFiles() {
 function cleanCompiledRustFiles() {
   const bridgeDir = resolve(packagesPath, 'core-bridge');
   console.log('Cleaning compiled rust files');
-  rmSync(resolve(bridgeDir, 'releases'), { recursive: true });
+  try {
+    rmSync(resolve(bridgeDir, 'releases'), { recursive: true });
```
bergundy, line 58: I think you can just use `force: true` here.

bergundy (approved):
LGTM, had some minor comments

PR #611, opened 2022-04-25
feat!(worker): Move shutdownSignals to RuntimeOptions

bergundy:
Makes more sense to put it on the singleton `Runtime` instead of a per-Worker

### DIFF
--- a/packages/test/src/test-local-activities.ts
+++ b/packages/test/src/test-local-activities.ts
@@ -36,8 +36,6 @@ test.beforeEach(async (t) => {
 
 async function defaultWorker(taskQueue: string) {
   return await Worker.create({
-    // Avoid creating too many signal handlers
-    shutdownSignals: [],
     taskQueue,
     workflowsPath: require.resolve('./workflows/local-activity-testers'),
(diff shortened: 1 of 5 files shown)


cretz:
General question - is there much benefit of the SDK providing nodeJS OS signal support vs asking the users to shutdown workers themselves when they listen to signals? It seems like if we offer worker shutdown, we should let the user decide when best to do so. Granted in Go we have a helper that supports signals, but I am not sure if we wrote it today whether we'd to do it again.

bergundy:
> General question - is there much benefit of the SDK providing nodeJS OS signal support vs asking the users to shutdown workers themselves when they listen to signals? It seems like if we offer worker shutdown, we should let the user decide when best to do so. Granted in Go we have a helper that supports signals, but I am not sure if we wrote it today whether we'd to do it again.

I think this makes sense in the JS ecosystem where users typically don't think about handling OS signals.
Providing it out of the box makes our samples more readable and easier to use.
It's easy enough to opt-out of.

lorensr (approved):

PR #613, opened 2022-04-26
Fix: Make sure signal handlers can be cleared

### LINKED ISSUE #612: [Bug] Heap out of memory on `setHandler(signal, undefined)`

In many of our workflows, we need to wait for a single occurrence of a given signal before execution before we continue with following steps, similar to the `workflow.GetSignalChannel(ctx, signalName).Receive(ctx, &args)` pattern in Go.

The TypeScript SDK does not yet offer an exact replacement for this idiom. It is obviously possible to register a signal handler, which sets a boolean value, and then `await condition(...that boolean variable...)`. However, this pattern results in consuming _all_ occurrences of the signal, where we would expect it to consume _only the first_ occurence of the signal.

For this reason, we constructed the function shown below. It has the particularity that, after receiving the expected signal, it clears up the signal handler, by calling `setHandler(signal, undefined)`. This is done from _inside_ the signal handler in order to ensure that only one "occurence" of the signal is consumed by the handler. It works great, except for the problem described in this ticket.

Here is the code:

```
export async function waitForSignal<T extends SignalDefinition>(
    signal: T,
    timeout?: number | string,
): Promise<ExtractSignalArgs<T> | false> {
    let signalReceived = false;
    let returnArgs: ExtractSignalArgs<T> | false = false;

    setHandler(signal, ((...inputArgs: ExtractSignalArgs<T>): void => {
        signalReceived = true;
        returnArgs = inputArgs;
        clearHandler(signal); // This one is required to ensure that we consume only one occurence of the signal
    }) as Handler<void, ExtractSignalArgs<T>, T>);

    await condition(() => signalReceived, timeout as number);
    clearHandler(signal); // This one is required in case we timed out

    return returnArgs;
}

export function clearHandler<T extends SignalDefinition>(signal: T) {
    setHandler(signal as SignalDefinition<[]>, undefined as unknown as Handler<void, [], T>);
}
```

### Describe the bug

Calling `setHandler(signal, undefined)` while there are buffered signals occurrences for that name results in heap out of memory.

The problem 

### Minimal Reproduction

```
workflow() {
    await sleep('30s');  // During this pause, send this workflow the given signal a few times
    setHandler(signal, () => {
        setHandler(signal, undefined);
    })
}
```

This results in a Node crash with the following message: `FATAL ERROR: invalid array length Allocation failed - JavaScript heap out of memory`.

### Analysis

Our analysis demonstrated that the problem is due to a subtile interaction between `setHandler` and `signalWorkflow`. After changing the handler, `setHandler` iterates over buffered signals and try to process each one. To do so, it calls `signalWorkflow`, which checks if a handler for that signal is actually defined. Now, since our handler is undefined, `signalWorkflow` instead push the signal occurence to the corresponding signal buffer. Turns out that this is the very same array over which `setHandler` is iterating at that very same time.

```
// in workflow.ts
export function setHandler<Ret, Args extends any[], T extends SignalDefinition<Args> | QueryDefinition<Ret, Args>>(
  def: T,
  handler: Handler<Ret, Args, T>
): void {
  if (def.type === 'signal') {
    state.signalHandlers.set(def.name, handler as any);
    const bufferedSignals = state.bufferedSignals.get(def.name);
    if (bufferedSignals !== undefined) {
      for (const signal of bufferedSignals) {
        state.activator.signalWorkflow(signal);
      }
      state.bufferedSignals.delete(def.name);
    }

// in internals.ts
  public signalWorkflow(activation: coresdk.workflow_activation.ISignalWorkflow): void {
    ...

    const fn = state.signalHandlers.get(signalName);
    if (fn === undefined) {
      let buffer = state.bufferedSignals.get(signalName);
      if (buffer === undefined) {
        buffer = [];
        state.bufferedSignals.set(signalName, buffer);
      }
      buffer.push(activation);  // <-- PUSH
      return;
    }

    ...
  }
```

### Proposed solutions

I propose moving the `state.bufferedSignals.delete(def.name)` line before the `for` loop. As an optimization, it would also be useful not to process buffered signals if the new handler is `undefined`.

```
    if (bufferedSignals !== undefined && handler !== undefined) {
      state.bufferedSignals.delete(def.name);
      for (const signal of bufferedSignals) {
        state.activator.signalWorkflow(signal);
      }
    }
```

mjameswh:
## What was changed

1. Fixed a heap overflow issue in `setHandler(signal, handler)` when handler is `undefined` and there are buffered signals for that signal name.

2. Changed the signature of `setHandler(...)` to indicate that `handler` can be `undefined` (in order to clear up previously configured handler for that signal name).

## Why?

Refer to #612 for discussion.

## Checklist

1. Closes #612 

2. **How was this tested?** I added appropriate integration test

3. **Any docs updates needed?** No doc that I know of that should be updated, but it could be useful to add a sample demonstrating this use case (for example based on examples in #612 or code in unit test).

### DIFF
--- a/packages/test/src/workflows/signal-handlers-clear.ts
+++ b/packages/test/src/workflows/signal-handlers-clear.ts
@@ -1,0 +1,39 @@
+/**
+ * Workflow used by integration-tests.ts: signal handlers can be cleared
+ *
+ * @module
+ */
+
+import { setHandler, sleep } from '@temporalio/workflow';
+import { unblockSignal } from './definitions';
+
+export async function signalHandlersCanBeCleared(): Promise<number> {
+  let counter = 0;
+
+  // Give time for signals to be received and buffered BEFORE the handler is configured. See issue #612 for details.
+  await sleep('20ms');
+
+  // Each handler below should catch a single occurence of the unblock signal.
+  // sleep() between each to make sure that any pending microtask has been run before the next handler is set
+  // If everything went ok, counter should equals 111 at the end.
+
+  setHandler(unblockSignal, () => {
+    counter += 1;
+    setHandler(unblockSignal, undefined);
+  });
+
+  await sleep('1ms');
+
+  setHandler(unblockSignal, () => {
+    counter += 10;
+    setHandler(unblockSignal, undefined);
+  });
+
+  await sleep('1ms');
+
+  setHandler(unblockSignal, () => {
+    counter += 100;
+  });
+
+  return counter;
+}
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -978,18 +978,4 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       await client.getHandle(workflowId).terminate();
     }
   });
-
-  test('signal handlers can be cleared', async (t) => {
-    const { client } = t.context;
-    const workflow = await client.start(workflows.signalHandlersCanBeCleared, {
-      taskQueue: 'test',
-      workflowId: uuid4(),
-    });
-
-    await workflow.signal(workflows.unblockSignal);
-    await workflow.signal(workflows.unblockSignal);
-    await workflow.signal(workflows.unblockSignal);
-
-    t.is(await workflow.result(), 111);
-  });
 }
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -978,4 +978,18 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       await client.getHandle(workflowId).terminate();
     }
   });
+
+  test('signal handlers can be cleared', async (t) => {
+    const { client } = t.context;
+    const workflow = await client.start(workflows.signalHandlersCanBeCleared, {
+      taskQueue: 'test',
+      workflowId: uuid4(),
+    });
+
+    await workflow.signal(workflows.unblockSignal);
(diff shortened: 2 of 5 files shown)


packages/test/src/workflows/signal-handlers-clear.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import { setHandler, sleep } from '@temporalio/workflow';
+import { unblockSignal } from './definitions';
+
+export async function signalHandlersCanBeCleared(): Promise<number> {
+  let counter = 0;
+
+  // Give time for signals to be received and buffered BEFORE the handler is configured. See issue #612 for details.
+  await sleep('20ms');
```
bergundy, line 14: Might be better to switch this to a unit test instead (see test-workflows.ts), I can help with that if you want.

bergundy (approved):
Overall LGTM, I'd just replace the integration test with a unit test so it's not timing sensitive.

packages/test/src/workflows/signal-handlers-clear.ts
mjameswh: Fair enough, I had also considered this option.

That doesn't look too difficult. I'll ping you if I need some help on this.

bergundy:
We have a regression in our tests where we can no longer run tests from forks:
https://github.com/temporalio/sdk-typescript/runs/6182355808?check_suite_focus=true#step:13:25

PR #615, opened 2022-04-26
Load scenarios

bergundy:


Sushisource (approved):
This LGTM but our pipeline is getting annoyingly long. We should consider moving some of the heavier tests to run after merge-to-master, but I think it's OK for now.

bergundy:
Agree, I had to limit the amount of workflows we run to make the test not painfully long

PR #616, opened 2022-04-26
Remove buildkite in favor of GHA for load/nightly

Sushisource:
Draft while testing


package.json
```diff
@@ -48,6 +48,7 @@
     "test": "lerna run --stream test",
     "test.watch": "lerna run --stream test.watch",
     "ci-build-no-rust": "lerna run --ignore @temporalio/core-bridge --stream build",
+    "ci-per-change-load": "node ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --iterations 10000 --otel-url='' --max-cached-wfs 500",
     "ci-load": "node ./packages/test/lib/load/run-all-stress-ci-scenarios.js --server-address $TEMPORAL_TESTING_SERVER_URL --iterations 10000 --otel-url='' --max-cached-wfs 500",
```
bergundy, line 52: Why not run all of the stress scenarios that we previously ran (`ci-load`)?
I'd consider running them in parallel to save some time.

.github/workflows/ci.yml
```diff
@@ -255,6 +255,10 @@ jobs:
       typescript-repo-ref: ${{github.event.pull_request.head.ref}}
 
   load-tests:
+    # Ensure that previous load tests in the same PR are cancelled if a new push is made
```
bergundy, line 258: Oh nice, very fancy, like

package.json
```diff
@@ -48,9 +48,11 @@
...
```
bergundy, line 52: Looks like I forgot to delete all of the args in this command, should be:
```suggestion
    "ci-load": "node ./packages/test/lib/load/run-all-stress-ci-scenarios.js",
```

package.json
```diff
@@ -48,9 +48,11 @@
     "test": "lerna run --stream test",
     "test.watch": "lerna run --stream test.watch",
     "ci-build-no-rust": "lerna run --ignore @temporalio/core-bridge --stream build",
+    "ci-per-change-load": "node ./packages/test/lib/load/all-in-one.js --server-address $TEMPORAL_TESTING_SERVER_URL --iterations 10000 --otel-url='' --max-cached-wfs 500",
```
bergundy, line 51: Wait, are we using this now? It's included in `ci-stress`

packages/test/src/load/starter.ts
```diff
@@ -36,8 +36,10 @@ async function runWorkflow({ client, workflowName, taskQueue, queryingOptions }:
             throw err;
           }
         }
-        if (queryingOptions.queryIntervalMs) {
+        if (queryingOptions.queryIntervalMs > 0) {
```
bergundy, line 39: Thanks

packages/test/src/load/starter.ts
```diff
@@ -172,7 +174,7 @@ async function main() {
   const concurrentWFClients = args['--concurrent-wf-clients'] || 100;
   const minWFPS = args['--min-wfs-per-sec'] || 35;
   const queryName = args['--do-query'];
-  const queryIntervalMs = args['--query-interval-ms'];
+  const queryIntervalMs = args['--query-interval-ms'] || 1000;
```
bergundy, line 177: Can't set to 0 now `0 || 1000 == 1000`.

bergundy, line 177: Might as well make the default undefined and not query at an interval..

package.json
Sushisource: That just changed recently though, no? It looks like they take quite a while to run. We could parallelize them and keep it as part of the PR pipeline if there are a small enough number.

It also looks like the self-hosted worker isn't really autoscaling. So we'd need to figure that out before changing things to work that way. I'm thinking we can follow up with another PR

package.json
Sushisource: That's fine, the idea is to have something short we can run per-pipeline and the nightly runs run stress, which maybe also includes this

.github/workflows/nightly.yml
```diff
@@ -29,16 +30,89 @@ jobs:
...
+
+      - name: Wait for Temporal Server
+        run: npm run wait-namespace
+      # end:docker-compose
+
+      - run: npm run ci-nightly-longrun
+        env:
+          TEMPORAL_TESTING_SERVER_URL: 'localhost:7233'
+
+  sampler:
```
lorensr, line 83: Purpose? Maybe link @ top of yml files to a cross-sdk testing doc, like `sdk-features/docs/testing.md`

lorensr (approved):

PR #617, opened 2022-04-26
Add GHA workflow

Sushisource:
Need to merge this to main before it can be tested. Thanks GHA.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,44 @@
+name: Nightly Tests
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron: '00 19 * * *'
+
+jobs:
+  stress-test:
+    runs-on: self-hosted
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+      # Don't build during install phase since we're going to explicitly build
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,44 @@
+name: Nightly Tests
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '00 19 * * *'
+
+jobs:
+  stress-test:
+    runs-on: self-hosted
```
bergundy, line 9: This isn't a good name for the runner.
How do I distinguish this from lets say a windows or mac runner?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          node-version: ${{ matrix.node }}
+      # Don't build during install phase since we're going to explicitly build
+      - run: npm ci --ignore-scripts
+      - run: npm run build
+        env:
+          BUILD_CORE_RELEASE: true
+
+      # docker-compose
+      - name: Checkout docker-compose repo
```
bergundy, line 27: Maybe just use temporalite, the default compose setup is heavy and slow?

bergundy (approved):
Had some comments but those can be address in the next PR when you implement the actual tests

.github/workflows/nightly.yml
Sushisource: I can't control that name. You have to use that label, and you can use additional ones, but they're all linux.

bergundy: Right, you'd need to use an array of labels.
For the M1 Mac, we have: [self-hosted, MacOS, arm64]

PR #619, opened 2022-04-28
fix(bundler): replace dedent by ts-dedent

### LINKED ISSUE #450: [Bug] workflows wont launch on Windows if project folder starts with 'n'

Renaming a project folder on Windows causes `npm run dev` to succeed or fail in the temporal worker launcher depending on whether the first letter of the directory name is an `n`. This issue prevents the nextjs-ecommerce-oneclick sample from running on Windows, but if you rename the folder to (for example) ecommerce-nextjs-oneclick the sample runs correctly under Windows.

### Describe the bug

When a project directory starts with an n, (like nextjs-ecommerce-oneclick in the samples-typescript repo), the code in [start:worker] on Windows incorrectly constructs the absolute path to workflows.js leaving out the slash between the project directory (nextjs-ecommerce-oneclick) and the parent of the project directory (samples-typescript) because it's treating the \ windows directory separator character and the n at the start of the project directory name as an offending \n to be stripped out (both the \ and the n are stripped from the constructed path).

### Minimal Reproduction

You can see this issue for example on a fresh clone of the samples-typescript repo when trying to run the nextjs-ecommerce-oneclick sample on Win11 running under Windows rather than under WSL. Renaming the nextjs-ecommerce-oneclick folder to ecommerce-nextjs-oneclick works around the issue and allows `npm run dev` to run without issue.

The error message is of the form:
```
[start:worker] 2022-01-24T00:52:35.620Z [ERROR] ERROR in ../../../../src/main.js
[start:worker] 2022-01-24T00:52:35.620Z [ERROR] Module not found: Error: Can't resolve 'C:\Path1\Path2\samples-typescriptextjs-ecommerce-oneclick\temporal\lib\workflows.js' in '/src'
```

If you look at the mangled path in the error message, you'll see something has incorrectly stripped the two characters `\n` from the desired `samples-typescript\nextjs-ecommerce-oneclick` leaving the malformed `samplestypescriptextjs-ecommerce-oneclick`.

### Environment/Versions

I'm seeing this issue under Windows 11 but I suspect it impacts all versions of Windows. WSL does not exhibit this issue because it does not use `\` as a path separator.

mjameswh:
## What was changed

- In package `worker` (that is, for the purpose of the workflow builder), replaced `dedent` by `ts-dedent`.
- Also, when returning the bundled workflow code, read `/dist/main.js` from memfs (instead of unionfs)

## Why?

- `dedent` uses a simple string search to replace all occurrences of `\n` by a line return character. This occurs even though the backslash itself is properly escaped. This caused incorrect handling of some paths on Windows (see issue #450).
- `unionfs` uses incoherent ordering when accessing its delegates file systems. The way `unionfs` is configured in bundler, when reading using `readFileSync()`, an existing file named `/dist/main.js` or `c:\dist\main.js` would get returned in priority over the file generated by the bundler, which is stored on the memfs (see https://github.com/temporalio/sdk-typescript/pull/554#discussion_r860134207).

## Checklist

1. Closes #450 and fix remaining issue in #554.

2. How was this tested: CHANGES IN THIS PR WAS NOT TESTED FOR EFFECTIVENESS. I asserted that existing tests runs properly on my Mac, but I am not in a position at present to execute these changes on a Windows machine.


packages/worker/package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
```
bergundy, line 6: Huh, I didn't realize this file is committed.
It can probably be deleted and added to `.gitignore`:
```
packages/*/package-lock.json
```

bergundy (approved):
Thanks!

packages/worker/package-lock.json
mjameswh: Done!

bergundy (commented):
Looking at this, we do need `docs/package-lock.json`, I'll fix

PR #621, opened 2022-04-29
fix(workflow): Fix case where activity or timer would try to be cancelled without being scheduled

bergundy:


bergundy:
@cretz FYI in case you copied this bug when using TS as reference

lorensr (approved):

PR #622, opened 2022-04-29
fix(workflow): Fail workflow task instead of run on workflow not found

bergundy:


lorensr (approved):
Taking this off my todo list, thanks! Whether missing Activities should fail WE or Task is still on it

bergundy:
@lorensr Not sure what you mean by activity not found failing workflow task

lorensr:
Ah, think it should be "what happens when Worker is missing an Activity, and does that match other SDKs"

bergundy:
@lorensr it was already fixed, we return a `non non retryable application failure` :)

PR #623, opened 2022-04-29
make snipsync pull from samples-typescript

### LINKED ISSUE #544: [Bug] Fix snipsync
Snip comments don't show up in API docs:

image
image

Config: https://github.com/temporalio/sdk-typescript/blob/b52999382ea72bc0bac45ed37cfaa9daf4bc6336/packages/docs/docusaurus.config.js#L130-L142

vkarpov15:
Fix #544

## What was changed

Added samples-typescript as a source for snipsync. This way our docs can reference snippets from samples-typescript.

## Why?

There are a few places in the API docs where we want to use snippets from samples-typescript. There shouldn't be any problem with recursive dependencies since snipsync doesn't run `npm install` in the samples-typescript repo.

## Checklist

1. Closes #544

2. How was this tested:

Snippets now show up on my local:

image

3. Any docs updates needed?

### DIFF
--- a/packages/docs/docusaurus.config.js
+++ b/packages/docs/docusaurus.config.js
@@ -137,8 +137,8 @@ module.exports = {
                 },
                 {
                   owner: 'temporalio',
-                  repo: 'samples-typescript'
-                }
+                  repo: 'samples-typescript',
+                },
               ],
               targets: ['docs'],
               features: {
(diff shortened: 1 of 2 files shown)


bergundy (approved):
Only thing I'm concerned with is that the docs reference the samples `main` branch but they're actually documenting sdk `main`.
We should probably only deploy the docs when we publish a new version and ideally we should have a way to display docs for all of them including `main`.

PR #626, opened 2022-05-02
fix: Do not patch global object in otel runtime if not in workflow context

### LINKED ISSUE #620: [Bug] OpenTelemetry interceptors patch global runtime outside workflow isolate too

Run an existing, large node.js project that starts Temporal workers, where some of the existing node code depends on `typeof window == "undefined"`.

### Describe the bug

The OpenTelemetry interceptors package modifies the global object to have a `window` property here: https://github.com/temporalio/sdk-typescript/blob/main/packages/interceptors-opentelemetry/src/workflow/runtime.ts 

This modification may be necessary within the workflow isolate, but, this code is also required outside the workflow isolate when just booting the worker and passing the interceptors because of this import: https://github.com/temporalio/sdk-typescript/blob/main/packages/interceptors-opentelemetry/src/worker/index.ts#L8

This violates a lot of existing code's assumptions and in the case of my project breaks a lot of things. `typeof window == "undefined"` is a generally pretty common way of checking if a piece of code is running within node or within the browser, and this global modification confuses all that code, because now in the whole node process `typeof window == "object"`. 

Is there a way that the interceptors could only modify the code inside the workflow isolate, or better yet, not modify the globals at all?

### Minimal Reproduction

Gonna skip this for now but if the issue is unclear let me know and I can put something together

### Environment/Versions

- OS and processor: macOS 12.1.3
- Temporal Version: temporalite + `@temporalio/worker@0.20.2`
- No docker, no k8s

bergundy:
Closes #620

### DIFF
--- a/packages/interceptors-opentelemetry/src/workflow/runtime.ts
+++ b/packages/interceptors-opentelemetry/src/workflow/runtime.ts
@@ -2,12 +2,17 @@
  * Sets global variables required for importing opentelemetry in isolate
  * @module
  */
+import { inWorkflowContext } from '@temporalio/workflow';
 
-// Required by opentelemetry (pretend to be a browser)
-(globalThis as any).performance = {
-  timeOrigin: Date.now(),
-  now() {
-    return Date.now() - this.timeOrigin;
-  },
-};
-(globalThis as any).window = globalThis;
+if (inWorkflowContext()) {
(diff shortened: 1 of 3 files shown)


lorensr (approved):

bergundy:
Tested in #627

PR #631, opened 2022-05-03
docs(workflow): Fix Trigger example

lorensr:
Currently uses unblock sample, which uses `condition`:

image


packages/workflow/src/trigger.ts
```diff
@@ -3,14 +3,29 @@ import { CancellationScope } from './cancellation-scope';
...
+ * Trigger is CancellationScope-aware: it is linked to the current scope on
  * construction and throws when that scope is cancelled.
  *
- * Useful for e.g. waiting for unblocking a Workflow from a signal.
+ * Useful for e.g. waiting for unblocking a Workflow from a Signal.
  *
  * @example
- * <!--SNIPSTART typescript-blocked-workflow-->
- * <!--SNIPEND-->
+ * ```ts
```
bergundy, line 12: Let's make this a sample in this repo so it's validated

bergundy (approved):
Looks like you forgot to commit the workflow file but otherwise LGTM

PR #633, opened 2022-05-04
fix(workflow): Propagate scope cancellation while waiting on ExternalWorkflowHandle cancellation

bergundy:
Wrapped this functionality in a patch to avoid breaking history compatibility.
This patch should be deprecated in a few months and eventually deleted.

### DIFF
--- a/packages/workflow/src/workflow.ts
+++ b/packages/workflow/src/workflow.ts
@@ -545,6 +545,14 @@ export function getExternalWorkflowHandle(workflowId: string, runId?: string): E
         if (state.info === undefined) {
           throw new IllegalStateError('Uninitialized workflow');
         }
+        const scope = CancellationScope.current();
+        if (scope.cancellable) {
+          scope.cancelRequested.catch(reject);
+        }
+        if (scope.consideredCancelled) {
+          return;
+        }
+
         const seq = state.nextSeqs.cancelWorkflow++;
         state.pushCommand({
           requestCancelExternalWorkflowExecution: {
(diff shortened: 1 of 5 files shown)


packages/workflow/src/workflow.ts
```diff
@@ -545,6 +545,17 @@ export function getExternalWorkflowHandle(workflowId: string, runId?: string): E
...
         }
+        // TODO: deprecate this patch after "enough" time has passed
+        if (patched('__temporal_internal_connect_external_handle_cancel_to_scope')) {
+          const scope = CancellationScope.current();
+          if (scope.cancellable) {
+            scope.cancelRequested.catch(reject);
+          }
+          if (scope.consideredCancelled) {
+            return;
+          }
```
lorensr, line 556: I'd find comments helpful—perhaps these?

```suggestion
          // reject if current scope gets cancelled
          const scope = CancellationScope.current();
          if (scope.cancellable) {
            scope.cancelRequested.catch(reject);
          }
          
          // don't send `requestCancelExternalWorkflowExecution` command if current scope is already cancelled
          if (scope.consideredCancelled) {
            return;
          }
```

lorensr (approved):

bergundy:
Thanks for the feedback, good comments @lorensr I'll address and resubmit

packages/workflow/src/workflow.ts
bergundy: I added some more informative comments, I'm not a huge fan of comments that spell out exactly what the code does, if those are required it points to bad abstractions or naming.

PR #636, opened 2022-05-05
test: Actually fail stress test when child fails

bergundy:
Not used to running `spawnSync`, luckily I caught it early


lorensr (approved):
It doesn't throw?! 😄

bergundy:
F#*king verdaccio.log timeout breaking CI again

PR #643, opened 2022-05-10
feat: Allow for multiple DataConverter.payloadCodecs

lorensr:
## What was changed

`DataConverter.payloadCodec` was changed to plural:

```ts
export interface DataConverter {
  ...
  payloadCodecs?: PayloadCodec[];
}
```

## Why?

So it's easier to provide multiple codecs (than the current solution, creating a CompositeCodec).

Wrote in a couple bugs that took a bit to track down! 

- Fixes: https://github.com/temporalio/sdk-typescript/issues/595

### DIFF
--- a/packages/common/src/converter/data-converter.ts
+++ b/packages/common/src/converter/data-converter.ts
@@ -1,4 +1,4 @@
-import { defaultPayloadCodec, PayloadCodec } from './payload-codec';
+import { PayloadCodec } from './payload-codec';
 import { defaultPayloadConverter, PayloadConverter } from './payload-converter';
 
 /**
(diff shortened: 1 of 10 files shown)


packages/common/src/converter/data-converter.ts
```diff
@@ -34,20 +34,24 @@ export interface DataConverter {
   payloadConverterPath?: string;
 
   /**
-   * A {@link PayloadCodec} instance. The default codec is a no-op.
+   * An array of {@link PayloadCodec} instances.
+   *
+   * Payloads are encoded in the order of the array and decoded in the opposite order. For example, if you have a
```
bergundy, line 39: Awesome docs!

bergundy (approved):
Noice!

PR #644, opened 2022-05-11
feat(worker): Support HTTP headers

### LINKED ISSUE #642: [Feature Request] Expose grpc static headers in NativeConnection
This is already supported by the core client, need to expose in NativeConnection and hook it up in the bridge

bergundy:
- Closes #642 (but also supports updating headers after connection has been established)
- Update neon to latest 0.8 and use napi-6 (requires Node 14+)


packages/test/src/test-native-connection-headers.ts
```diff
@@ -0,0 +1,79 @@
...
+  t.true(gotInitialHeader);
+
+  await connection.setHeaders({ update: 'true' });
+  // Create a worker so it starts polling for activities so we can check our mock server got the "update" header
+  const worker = await Worker.create({
+    connection,
+    taskQueue: 'tq',
+    activities: {
+      async noop() {
+        /* yes eslint this is meant to be empty */
```
lorensr, line 74: 😄

lorensr (approved):

PR #645, opened 2022-05-12
chore: Improve error message when misimporting in Workflow code

lorensr:
To help address users getting stuck like this (despite the warning about async_hooks that appears in Worker output): https://temporalio.slack.com/archives/C01DKSMU94L/p1652371946624169

Currently not working (see image)—wondering if there's a better condition to test?

image


packages/activity/src/index.ts
```diff
@@ -26,7 +26,13 @@
  * @module
  */
 
-import { msToNumber } from '@temporalio/internal-workflow-common';
+if ((globalThis as any).__TEMPORAL__ !== undefined) {
```
bergundy, line 29: Wondering if it's worth making this a function in `common`, not critical

packages/activity/src/index.ts
```diff
@@ -26,7 +26,13 @@
...
+  throw new IllegalStateError(
+    "You are importing from '@temporalio/activity' in your Workflow code, which doesn't work. Workflow code should only import from '@temporalio/workflow' and '@temporalio/common'."
```
bergundy, line 31: How about something like:
```suggestion
    "Importing from '@temporalio/activity' from Workflow code is not supported. Workflow code should only import from '@temporalio/workflow' and '@temporalio/common'. More information here: https://docs.temporal.io/docs/typescript/determinism/"
```

bergundy: I'm not a fan of this personal tone in error messages

lorensr:
TODO: add test of failing import/bundling

bergundy:
Closing this for now since using non workflow libs should result in failed bundle creation.
We should probably add the forbidden `@temporalio/*` libs to the list of blacklisted modules.

PR #646, opened 2022-05-12
feat!(worker): Restructure `TelemetryOptions`

bergundy:
* Also support passing HTTP headers to an OpenTelemetry Collector


lorensr (approved):
Like the new API! ☺️

PR #650, opened 2022-05-16
ci: Use top-level await in wait-on-temporal.mjs

lorensr:
Makes it a little shorter/simpler

### DIFF
--- a/scripts/wait-on-temporal.mjs
+++ b/scripts/wait-on-temporal.mjs
@@ -11,11 +11,11 @@ class Timer {
   }
 }
 
-const MAX_ATTEMPTS = 100,
-  RETRY_INTERVAL_SECS = 1;
+const maxAttempts = 100;
+const retryIntervalSecs = 1;
 
 try {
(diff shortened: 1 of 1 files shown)


scripts/wait-on-temporal.mjs
```diff
@@ -11,8 +11,11 @@ class Timer {
   }
 }
 
-async function main(maxAttempts = 100, retryIntervalSecs = 1) {
-  for (let attempt = 1; attempt <= maxAttempts; ++attempt) {
+const MAX_ATTEMPTS = 100,
+  RETRY_INTERVAL_SECS = 1;
```
bergundy, line 15: I haven't seen this syntax in a while :)

bergundy (approved):

scripts/wait-on-temporal.mjs
lorensr: all caps for finals, or multiple declarations in a single statement?

lorensr: my Java roots are showing 😄

PR #651, opened 2022-05-16
fix(worker): Fix memory leaks

### LINKED ISSUE #649: Memory leak on the temporal worker side
## Description

Some times ago we began using the temporal SDK in our microservice product, but after some times we detected that temporal-worker containers are killed by "out of memory" reason. Then we started our investigation and found the potential leak problem: a lot of 'OperatorSubscriber' objects create on temporal-worker side,  but it never clean, and it make memory leak problem. Right now it's  just a theory, but we hope that you can help us with checking it.

## Expected Behavior

We are not detecting memory leak

## Actual Behavior

We are detecting memory leak

## Steps to Reproduce the Problem

1. Run and inspect the temporal worker (e.g. `node --inspect src/worker.js`)
2. Start expect the worker process via chrome devtools
3. Execute the temporal client in infinity loop (for generate persistent loading for worker)
4. A little while later intervals via chrome devtools take memory snapshots from the worker process
5. Compare last memory snapshot with preview: there are a lot of `OperatorSubscriber` objects between the snapshots, so we can make the conclusion, that temporal typescript SDK has a memory leak
  

## Specifications

  - Version: 0.22.0
  - Platform: x86_64 GNU/Linux (distr: Ubuntu:20.04, core: 5.14)

-------------------------------------

## Help with reproduce

For comfort reproduce, we created [the separate repository](https://github.com/viavdeyenk-mts/temporal-tssdk-demo-memory-leak) for help and  describe the steps for representation the leak problem.

bergundy:
Closes #649

2 leaks were discovered and fixed from "recent" Activity refactoring.
I'll set up alerts for our nightly tests to make sure this never happens again.

### DIFF
--- a/packages/test/src/test-worker-heartbeats.ts
+++ b/packages/test/src/test-worker-heartbeats.ts
@@ -144,3 +144,29 @@ test('Activity gets cancelled if heartbeat fails', async (t) => {
   });
   t.deepEqual(heartbeatsSeen, []);
 });
+
+test('No heartbeat is emitted with rouge activity', async (t) => {
+  const subj = new Subject<void>();
+  let cx: Context | undefined = undefined;
+
+  const worker = isolateFreeWorker({
+    taskQueue: 'unused',
+    activities: {
+      async rapidHeartbeater() {
+        cx = Context.current();
+        Context.current().heartbeat(1);
+      },
+    },
+  });
+
+  const heartbeatsSeen = Array<number>();
+  worker.native.activityHeartbeatCallback = (_tt, details) => {
+    heartbeatsSeen.push(details);
+  };
+  await runActivity(worker, () => {
+    subj.next();
+    t.truthy(cx);
+    cx?.heartbeat(2);
+  });
+  t.deepEqual(heartbeatsSeen, [1]);
+});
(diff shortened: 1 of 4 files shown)


bergundy:
I'm still trying to figure out if there are other leaks, decided to leave #649 open for now to collect more info.

bergundy:
I have a good lead, will keep investigating

packages/test/src/test-worker-heartbeats.ts
```diff
@@ -144,3 +144,29 @@ test('Activity gets cancelled if heartbeat fails', async (t) => {
   });
   t.deepEqual(heartbeatsSeen, []);
 });
+
+test('No heartbeat is emitted with rouge activity', async (t) => {
```
lorensr, line 148:
```suggestion
test('No heartbeat is emitted with rogue activity', async (t) => {
```

lorensr (approved):

packages/test/src/test-worker-heartbeats.ts
bergundy: Thanks, just merged without this, I'll fix in another PR

PR #657, opened 2022-05-20
feat!: Update WorkflowInfo; Add upsertSearchAttributes

### LINKED ISSUE #314: [Feature Request] Add missing continuedAsNewExecutionRunId to WorkflowInfo interface

### LINKED ISSUE #357: [Feature Request] Expose history size to Workflow code
[As discussed on our community slack](https://temporalio.slack.com/archives/C01DKSMU94L/p1635615168067800), we'd like to expose the number of history events and the aggregated payloads size to via `WorkflowInfo` so Workflow authors can make smarter decisions for when to continue as new.

### LINKED ISSUE #445: Add parent workflow information in `WorkflowInfo`

### LINKED ISSUE #576: [Feature Request] Return string for Workflow status

We leverage `handle.describe()` to see the status of a workflow. Today, we have to build a wrapper around the response to translate the gRPC code (a number) to a meaningful status (a string).

Return a string type to represent status in the `WorkflowExecutionDescription`.

More explicitly, this:
```typescript
export interface WorkflowExecutionDescription {
    ...
    status: temporal.api.enums.v1.WorkflowExecutionStatus;
    ...
}

enum WorkflowExecutionStatus {
    WORKFLOW_EXECUTION_STATUS_UNSPECIFIED = 0,
    WORKFLOW_EXECUTION_STATUS_RUNNING = 1,
    WORKFLOW_EXECUTION_STATUS_COMPLETED = 2,
    WORKFLOW_EXECUTION_STATUS_FAILED = 3,
    WORKFLOW_EXECUTION_STATUS_CANCELED = 4,
    WORKFLOW_EXECUTION_STATUS_TERMINATED = 5,
    WORKFLOW_EXECUTION_STATUS_CONTINUED_AS_NEW = 6,
    WORKFLOW_EXECUTION_STATUS_TIMED_OUT = 7
}
```

turns into this:

```typescript
export interface WorkflowExecutionDescription {
  ...
  status: WorkflowExecutionStatus
  ...
}

export type WorkflowExecutionStatus = 'Canceled' | 'Completed' | 'Continued_As_New' | 'Failed' | ...
```

lorensr:
Working on broken tests, and will add more for new `WorkflowInfo` fields.

- Added and changed `WorkflowInfo` fields
- Add `upsertSearchAttributes` Workflow function
- Make Search Attributes always arrays
- Accept Dates for Search Attributes
- Convert ISO strings back to Dates when providing Search Attributes to user

## Fixes

- Fixes #314
- Fixes #445 
- Closes #576 
- Closes #357


packages/common/src/converter/search-attribute-payload-converter.ts
```diff
@@ -0,0 +1,34 @@
+import { ValueError } from '@temporalio/internal-workflow-common';
+import { PayloadConverter } from './payload-converter';
+import { JsonPayloadConverter } from './json-payload-converter';
+import { Payload, str } from './types';
+
+const jsonConverter = new JsonPayloadConverter();
+
+/**
+ *
+ */
```
Sushisource, line 10: Empty docstring

packages/internal-workflow-common/src/retry-policy.ts
```diff
@@ -71,3 +71,19 @@ export function compileRetryPolicy(retryPolicy: RetryPolicy): temporal.api.commo
     nonRetryableErrorTypes: retryPolicy.nonRetryableErrorTypes,
   };
 }
+
+export function uncompileRetryPolicy(
```
Sushisource, line 75: uncompile is a bit of a weird term here. If you wanna stay consistent with above, `decompile` is probably more common

Alternatively, `extract`

packages/workflow/src/workflow.ts
```diff
@@ -1133,3 +1134,52 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+ *   CustomIntField: [2, 3],
+ *   CustomKeywordField: ['durable code']
+ * });
+ * ```
+ *
+ * would result in the Workflow having these Search Attributes:
+ *
+ * ```ts
+ * {
+ *   CustomIntField: [2, 3],
```
Sushisource, line 1159: Shouldn't this be `[1,2,3]` ? Or would the user need to take, append, then replace? If so, worth showing that example too.

lorensr: No, I'm trying to make it clear here that upsert overwrites. I'll change the numbers. In the sample, I show upserting based on current value

bergundy (commented):
Looks like I was in the middle of a review a while ago and never submitted it, might be outdated by now so feel free to ignore. I'll review again.

packages/common/src/converter/search-attribute-payload-converter.ts
```diff
@@ -0,0 +1,34 @@
...
+export class SearchAttributePayloadConverter implements PayloadConverter {
+  public toPayload(value: unknown): Payload | undefined {
+    if (value instanceof Array && value.length > 0 && value[0] instanceof Date) {
+      value = (value as Date[]).map((date) => date.toISOString());
```
bergundy, line 14: Not required, this is automatically done by `JSON.stringify`

packages/internal-workflow-common/src/interfaces.ts
```diff
@@ -35,3 +35,10 @@ export interface QueryDefinition<_Ret, _Args extends any[] = []> {
 
 /** Get the "unwrapped" return type (without Promise) of the execute handler from Workflow type `W` */
 export type WorkflowResultType<W extends Workflow> = ReturnType<W> extends Promise<infer R> ? R : never;
+
+/**
+ * If another SDK creates a Search Attribute that's not an array, we wrap it in an array.
+ *
+ * Dates are serialized as ISO strings.
+ */
+export type SearchAttributeValue = string[] | number[] | boolean[] | Date[];
```
bergundy, line 44: Might as well make it single value and wrap it in array everywhere:
```suggestion
export type SearchAttributeValue = string | number | boolean | Date;
```

packages/internal-workflow-common/src/workflow-options.ts
```diff
@@ -47,18 +47,18 @@ export interface BaseWorkflowOptions {
...
-   * Specifies additional indexed information in result of list workflow. The type of value should
-   * be a primitive (e.g. string, number, boolean). For dates, use Date.toISOString()
+   * Specifies additional indexed information to attach to the Workflow Execution. More info:
+   * https://docs.temporal.io/docs/typescript/search-attributes
    *
-   * Values are always converted using {@link defaultPayloadConverter}, even when a custom data converter is provided.
+   * Values are always converted using {@link JsonPayloadConverter}, even when a custom data converter is provided.
    */
-  searchAttributes?: Record<string, string | number | boolean>;
+  searchAttributes?: Record<string, SearchAttributeValue>;
```
bergundy, line 61: If you take my suggestion above this would be:
```suggestion
  searchAttributes?: Record<string, SearchAttributeValue[]>;
```

packages/test/src/integration-tests.ts
```diff
@@ -631,7 +693,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       searchAttributePayloadConverter.fromPayload(
         execution.raw.workflowExecutionInfo!.searchAttributes!.indexedFields!.CustomIntField!
       ),
-      3
+      [3]
```
bergundy, line 696: I'd also check the not "raw" value

packages/worker/src/worker.ts
```diff
@@ -907,20 +919,75 @@ export class Worker {
...
+                          this.options.loadedDataConverter,
+                          lastCompletionResult?.payloads
+                        ),
+                        lastFailure: await decodeOptionalFailureToOptionalError(
+                          this.options.loadedDataConverter,
+                          continuedFailure
+                        ),
                         taskQueue: this.options.taskQueue,
-                        isReplaying: activation.isReplaying,
+                        more: {
```
bergundy, line 974: Why nest this?

packages/worker/src/worker.ts
```diff
@@ -1028,7 +1095,7 @@ export class Worker {
             ifaceName,
             fnName,
           });
-        } else if (dep.callDuringReplay || !info.isReplaying) {
+        } else if (dep.callDuringReplay || !info.unsafe.isReplaying) {
```
bergundy, line 1098: It's late so I might be missing something but I'm pretty sure we never update `isReplaying`

bergundy: Don't have to fix in this PR, could open an issue but this is pretty critical.

packages/workflow/src/interfaces.ts
```diff
@@ -1,39 +1,164 @@
...
+): ParentWorkflowInfo | undefined {
+  if (!parent) {
+    return undefined;
+  }
+
+  return {
+    /* eslint-disable @typescript-eslint/no-non-null-assertion */
+    workflowId: parent.workflowId!,
+    runId: parent.runId!,
+    namespace: parent.namespace!,
```
bergundy, line 160: Instead of non null assertion let's fail at runtime with an exception or our users will get undefined values if something changes.
In general we shouldn't use non null assertions in the codebase, only in tests.

packages/workflow/src/workflow.ts
```diff
@@ -1133,3 +1134,52 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+  }
+
+  const mergedSearchAttributes = { ...state.info.searchAttributes, ...searchAttributes };
+  if (!mergedSearchAttributes) {
+    throw new Error('searchAttributes must be a non-null Record<string, SearchAttributeValue>');
+  }
+
+  state.pushCommand({
+    upsertWorkflowSearchAttributesCommandAttributes: {
+      seq: state.nextSeqs.upsertSearchAttributes++,
```
bergundy, line 1179: Why is this required @Sushisource?

packages/workflow/src/internals.ts
```diff
@@ -403,6 +405,7 @@ export class State {
     signalWorkflow: 1,
     cancelWorkflow: 1,
     condition: 1,
+    upsertSearchAttributes: 1,
```
bergundy, line 408: What is this used for?

packages/workflow/src/workflow.ts
```diff
@@ -1133,3 +1134,52 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+    throw new IllegalStateError('`state.info` should be defined');
+  }
+
+  const mergedSearchAttributes = { ...state.info.searchAttributes, ...searchAttributes };
+  if (!mergedSearchAttributes) {
+    throw new Error('searchAttributes must be a non-null Record<string, SearchAttributeValue>');
+  }
+
+  state.pushCommand({
+    upsertWorkflowSearchAttributesCommandAttributes: {
```
bergundy, line 1178: @Sushisource this should be called `upsertSearchAttributes` IMHO, we definitely don't need the `CommandAttributes` suffix, this is the only place where we use this pattern.

packages/common/src/converter/search-attribute-payload-converter.ts
```diff
@@ -0,0 +1,34 @@
...
+import { JsonPayloadConverter } from './json-payload-converter';
+import { Payload, str } from './types';
+
+const jsonConverter = new JsonPayloadConverter();
+
+/**
+ *
+ */
+export class SearchAttributePayloadConverter implements PayloadConverter {
+  public toPayload(value: unknown): Payload | undefined {
```
bergundy, line 12: We should verify only valid types are passed in here (only arrays with single type)

packages/workflow/src/workflow.ts
```diff
@@ -1133,3 +1134,52 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+      seq: state.nextSeqs.upsertSearchAttributes++,
+      searchAttributes: mapToPayloads(searchAttributePayloadConverter, mergedSearchAttributes),
+    },
+  });
+
+  state.info.searchAttributes = mergedSearchAttributes;
```
bergundy, line 1184: Let's test that this is indeed updated

packages/workflow/src/workflow.ts
```diff
@@ -1133,3 +1134,52 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+
+  const mergedSearchAttributes = { ...state.info.searchAttributes, ...searchAttributes };
+  if (!mergedSearchAttributes) {
+    throw new Error('searchAttributes must be a non-null Record<string, SearchAttributeValue>');
+  }
+
+  state.pushCommand({
+    upsertWorkflowSearchAttributesCommandAttributes: {
+      seq: state.nextSeqs.upsertSearchAttributes++,
+      searchAttributes: mapToPayloads(searchAttributePayloadConverter, mergedSearchAttributes),
```
bergundy, line 1180: Why do we need to pass in all the search attributes? Just pass in the new values and merge the maps in workflow memory.

packages/workflow/src/interfaces.ts
```diff
@@ -1,39 +1,164 @@
...
+// If we get use cases / user requests for this,
+// expose startWorkflow.continueInitiator as WorkflowInfo.initiator: Initiator
+// export enum Initiator {
+//   Unspecified = 0,
+//   ContinueAsNew = 1,
+//   Retry = 2,
+//   Cron = 3,
+// }
+
+export function updateParentType(
```
bergundy, line 149: Looks like this is only used in the worker, I wouldn't put this in the workflow package.

bergundy (commented):
Overall nicely done.

Not sure we should rename `workflowType` to `type` at this point, it's a breaking change I'd rather avoid, users are relying on this. Also not sure we should have the `more` attribute, could just flatten the whole thing.

packages/workflow/src/workflow.ts
Sushisource: I suppose it isn't, really, but it's consistent and IMHO we should eventually be resolving this command so users can decide to await on it if they need to make sure the changes are persisted before proceeding with the workflow.

packages/workflow/src/workflow.ts
Sushisource: Sure, can do that.

Sushisource:
> Overall nicely done.
> 
> Not sure we should rename `workflowType` to `type` at this point, it's a breaking change I'd rather avoid, users are relying on this. Also not sure we should have the `more` attribute, could just flatten the whole thing.

FWIW I was conflicted about `more`, thought about it, and decided it was pretty nice actually. Narrows down the amount of stuff you see when you autocomplete to just the things you're more likely to care about.

packages/worker/src/worker.ts
lorensr: We set it from the start workflow activation:

image

And then update it during future activations in `vm.ts` and `worker-interface.ts`:

image

image

packages/workflow/src/workflow.ts
lorensr: You mean we should just pass the new ones to core? @Sushisource does that work atm?

packages/workflow/src/workflow.ts
lorensr: Does this suffice? `test('upsertAndReadSearchAttributes'`

lorensr:
I like `WorkflowInfo.type` since you know what type it is given the context, and `workflowInfo().workflowType` seems repetitive. `WorkflowInfo.more.namespace` is also breaking (used to be `WorkflowInfo.namespace`), but probably less used. I'm okay changing it back to `workflowType`.

I like `more` a lot for being able to differentiate in autocomplete and docs what the normal/common fields are, and reduce mental load when looking at the normal stuff. I'd also be interested in doing similar things elsewhere, like `WorkerOptions` is a lot / intimidating to take in, and takes time to search through to find the field you're looking for:

image

and many could be hidden behind `.throttling` or something.

packages/worker/src/worker.ts
bergundy: Yup which means it doesn't take affect in the main thread where the worker lives and it requires this info to run sink calls.

packages/workflow/src/workflow.ts
bergundy: That's not how search attributes work though, there's no way to await this command's completion and even if there was we'd need to change the language runtime to support that.

packages/workflow/src/workflow.ts
bergundy: https://github.com/temporalio/sdk-core/issues/332

packages/workflow/src/workflow.ts
bergundy: Yes, this would work.

packages/workflow/src/workflow.ts
bergundy: That should work, I'd just test and verify

bergundy: It's the way it's supposed to work.

packages/workflow/src/workflow.ts
```diff
@@ -1133,3 +1134,52 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
     throw new TypeError(`Invalid definition type: ${(def as any).type}`);
   }
 }
+
+/**
+ * Updates this Workflow's Search Attributes by merging the provided `searchAttributes` with the existing Search
```
bergundy, line 1139: Nice docstring

packages/test/src/workflows/return-search-attributes.ts
```diff
@@ -0,0 +1,11 @@
+import { SearchAttributeValue, workflowInfo } from '@temporalio/workflow';
+
+export async function returnSearchAttributes(): Promise<Record<string, SearchAttributeValue[]> | undefined> {
+  const sa = workflowInfo().searchAttributes!; // eslint-disable-line @typescript-eslint/no-non-null-assertion
+  const datetime = (sa.CustomDatetimeField as Array<Date>)[0];
+  return {
+    ...sa,
+    datetimeType: [Object.getPrototypeOf(datetime).constructor.name],
+    datetimeInstanceofWorks: [datetime instanceof Date],
```
bergundy, line 9: Great test

packages/workflow/src/interfaces.ts
```diff
@@ -1,41 +1,151 @@
...
+  };
+}
+
+export interface ParentWorkflowInfo {
+  workflowId: string;
+  runId: string;
+  namespace: string;
 }
 
+// If we get use cases / user requests for this,
```
bergundy, line 140: Wanna track this in an issue?

packages/common/src/converter/search-attribute-payload-converter.ts
bergundy: Why did you resolve this?

packages/common/src/converter/search-attribute-payload-converter.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Payload, str } from './types';
+
+const jsonConverter = new JsonPayloadConverter();
+
+/**
+ * Converts Search Attribute values using JsonPayloadConverter
+ */
+export class SearchAttributePayloadConverter implements PayloadConverter {
+  public toPayload(values: unknown): Payload | undefined {
+    if (!(values instanceof Array)) {
```
bergundy, line 13: I was going to ask why not use `Array.isArray` but there's no need in our case: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#instanceof_vs_isarray

packages/common/src/converter/search-attribute-payload-converter.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (firstType === 'object') {
+      for (const value of values) {
+        if (!(value instanceof Date)) {
+          throw new ValueError(
+            `SearchAttribute values must arrays of strings, numbers, booleans, or Dates. This value ${value} is a ${typeof value}.`
+          );
+        }
+      }
+    } else {
+      for (const value of values) {
```
bergundy, line 28: value could be undefined or bigint, we should should prevent that, let's have an explicit whitelist of types (string, number, boolean).

packages/worker/src/worker.ts
lorensr: https://github.com/temporalio/sdk-typescript/issues/667

packages/workflow/src/workflow.ts
Sushisource: I understand it's not how it works now, hence "should eventually". We can get rid of the seq until then, but, not exactly a big deal either way.

packages/workflow/src/interfaces.ts
```diff
@@ -1,39 +1,140 @@
...
+
+    /**
+     * Milliseconds after which the Workflow Execution is automatically terminated by Temporal Server. Set via {@link WorkflowOptions.workflowExecutionTimeout}.
+     */
+    executionTimeout?: number;
+
+    /**
+     * Time at which the Workflow Execution expires
+     */
+    executionExpirationTime?: Date;
```
cretz, line 90: Where does this value come from? Are we aware of any other SDK ever needing it?

packages/workflow/src/interfaces.ts
```diff
@@ -1,39 +1,140 @@
...
+  };
+
+  /**
+   * These fields are non-deterministic. Workflow code should not do different things based on these fields.
+   */
+  unsafe: {
+    /**
+     * Whether a Workflow is replaying history or processing new events
+     */
+    isReplaying: boolean;
```
cretz, line 130: I don't think this should be in workflow info. As a rule of thumb, I don't think workflow info should _ever_ change after the workflow is started from a user POV. I should be able to store the workflow info in a local variable and trust that it's immutable and doesn't change for the life of the workflow run. This property holds true in other SDKs I believe.

packages/workflow/src/interfaces.ts
```diff
@@ -1,39 +1,140 @@
...
-   * Whether a Workflow is replaying history or processing new events
+   * Task queue this Workflow is executing on
    */
-  isReplaying: boolean;
+  taskQueue: string;
+
+  /**
+   * Less commonly used information
+   */
+  more: {
```
cretz, line 60: Personally I think these can just move to the top level. There's no need to hide this info IMO. I can't think of a good reason to hide this stuff (a vertical scrollbar in intellisense is ok). Having said that, we shouldn't just add everything in workflow info, only valid stuff that represents information about the workflow run (_not_ the workflow task).

packages/workflow/src/interfaces.ts
```diff
@@ -1,39 +1,140 @@
...
+
+    /**
+     * Cron Schedule for this Execution. Set via {@link WorkflowOptions.cronSchedule}.
+     */
+    cronSchedule?: string;
+
+    /**
+     * Milliseconds between Cron Runs
+     */
+    cronScheduleToScheduleInterval?: number;
```
cretz, line 120: Has anyone asked for this value? Maybe we can leave it off.

packages/workflow/src/interfaces.ts
bergundy: You're right, we should move this from info, I originally missed this.

packages/workflow/src/workflow.ts
bergundy: Not sure what you mean by tested Core, you should test this against the server.

packages/client/src/workflow-client.ts
```diff
@@ -809,7 +810,7 @@ export class WorkflowClient {
           searchAttributes: mapFromPayloads(
             searchAttributePayloadConverter,
             raw.workflowExecutionInfo!.searchAttributes?.indexedFields
```
bergundy, line 812: Doesn't have to be in this PR but let's get rid of all of these non null assertions

packages/common/src/converter/json-payload-converter.ts
```diff
@@ -0,0 +1,37 @@
+import { ValueError } from '@temporalio/internal-workflow-common';
+import { PayloadConverterWithEncoding } from './payload-converters';
+import { encodingKeys, encodingTypes, METADATA_ENCODING_KEY, Payload, str, u8 } from './types';
+
+/**
+ * Converts between non-undefined values and serialized JSON Payload
```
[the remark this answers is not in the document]
bergundy: It's not just non-undefined, it's JSON serializables, but yes, it explicitly ignores `undefined`.

packages/common/src/converter/search-attribute-payload-converter.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Converts Search Attribute values using JsonPayloadConverter
+ */
+export class SearchAttributePayloadConverter implements PayloadConverter {
+  public toPayload(values: unknown): Payload | undefined {
+    if (!(values instanceof Array)) {
+      throw new ValueError(`SearchAttribute value must be an array`);
+    }
+
+    const firstValue = values[0];
```
bergundy, line 18: If you really want to be accurate with the error messages you could check that the array isn't empty or better yet, does empty array unset the value? Is unsetting supported?

bergundy (commented):
Let's leave WorkflowInfo flat like the earth, revert the change from `workflowType` to `type` and in a followup PR move `isReplaying` out of `WorkflowInfo`.

packages/workflow/src/interfaces.ts
lorensr: > not the workflow task

Looks like `taskTimeout` is the only task-specific info? I like the idea of a Workflow being able to know what settings it was started with, whether it's just used for logging or impacts logic.

bergundy:
You can add custom search attributes using the gRPC `operatorService` API.

We don't have a high level API for that yet.

There's also a need to work around the fact that the connection in the test environment isn't a concrete class (I'll fix that in a bit)

```ts
const { client } = testEnv;
(client.connection as Connection).operatorService.addSearchAttributes({
  namespace: client.options.namespace,
  searchAttributes: {
      MyAttribute: temporal.api.enums.v1.IndexedValueType.INDEXED_VALUE_TYPE_INT,
  },
});
```

siawyoung:
> You can add custom search attributes using the gRPC `operatorService` API.
> 
> We don't have a high level API for that yet.
> 
> There's also a need to work around the fact that the connection in the test environment isn't a concrete class (I'll fix that in a bit)
> 
> ```ts
> import { temporal } from '@temporalio/proto';
> const { client } = testEnv;
> (client.connection as Connection).operatorService.addSearchAttributes({
>   namespace: client.options.namespace,
>   searchAttributes: {
>       MyAttribute: temporal.api.enums.v1.IndexedValueType.INDEXED_VALUE_TYPE_INT,
>   },
> });
> ```

Holy moly, that was a quick turnaround. And it worked too!

```
  await (testEnv.connection as Connection).operatorService.addSearchAttributes({
    searchAttributes: {
      Patch: iface.temporal.api.enums.v1.IndexedValueType.INDEXED_VALUE_TYPE_KEYWORD
    }
  })
```

(Also just a weird side note but why does `KEYWORD_LIST` still exist when everything is now treated as an array? I tried using `INDEXED_VALUE_TYPE_KEYWORD_LIST` at first but it threw an error with `protoEnumName should start with INDEXED_VALUE_TYPE_ prefix`)

bergundy:
`KEYWORD_LIST` was added in server version 1.20, make sure you're using a new enough server.
We also plan to make all other search attributes single value only eventually.

PR #658, opened 2022-05-20
Hackily fix #614 by checking error names instead of instances

### LINKED ISSUE #614: [Bug] `ShutdownError: Core is shut down` error when running in jest

Integrate temporal workflows into an existing TypeScript application tested using jest

### Describe the bug

I am able to create and complete workflows in one jest test file using helpers similar to those found in `sdk-typescript`'s repo, but as soon as I start using the client / workers in multiple jest test files, I get strange `ShutdownError: Core is shut down` or `ShutdownError: Core is shut down and there are no more workflow replay tasks` errors. These errors seem to occur after the workers have successfully completed open workflows and activities, and seem to happen somewhat inconsistently.

### Minimal Reproduction

Find a repro here: https://github.com/airhorns/temporal-jest

You must run multiple test files at once to trigger the issue.

### Environment/Versions

- OS and processor: macOS 12.3.1, Intel Core i7
- Temporal Version: temporalite with `@temporalio/worker@0.21.1`
- No docker, k8s

### Additional context

jest uses some strange techniques to isolate each test suite (each *.test.js file) from each other. The tests run in the same process, but jest monkeypatches in a custom `require` function and evaluates the tests in a custom context, allowing each test file to muck with global state and not influence all the others. This is how jest supports their module mocking and a generally good practice for avoiding test flakes. That said, it can cause issues if code is not expecting to be required more than once, or somehow escapes their global state resetting approach. I am not sure it can be turned off for jest, but I do know that jest is a VERY popular test runner such that it'd be great to get support. See https://www.npmtrends.com/jest-vs-tap-vs-ava

airhorns:
## What was changed

As discussed on the support slack, jest's re-requiring of stuff means there are multiple versions of the same error class floating around. The per-process one quacked by the native runtime no longer matches in the 2nd+ jest test run, making these instanceof checks fail. Instead, we just look at their names.

## Why?

Fix using jest test suites to test temporal workflows

## Checklist

1. Closes #614

2. How was this tested:

In repro repo here: https://github.com/airhorns/temporal-jest

### DIFF
--- a/packages/worker/src/runtime.ts
+++ b/packages/worker/src/runtime.ts
@@ -210,7 +210,7 @@ export class Runtime {
       );
     } catch (error) {
       // Prevent unhandled rejection
-      if (error instanceof errors.ShutdownError) return;
+      if (error.name == 'ShutdownError') return;
       // Log using the original logger instead of buffering
       this.options.logger.warn('Error gathering forwarded logs from core', { error });
     } finally {
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -577,7 +577,7 @@ export class Worker {
           try {
             yield await pollFn();
           } catch (err) {
-            if (err instanceof errors.ShutdownError) {
+            if (err.name == 'ShutdownError') {
               break;
             }
             throw err;


airhorns:
I think the CLA assistant might be wonky, I just see a blank Version box when I go to that page:

lorensr:
For me it starts out blank, and then populates after a moment? Looks like your screenshot didn't finish uploading, so maybe it was a connection thing.

packages/worker/src/runtime.ts
```diff
@@ -210,7 +210,7 @@ export class Runtime {
       );
     } catch (error) {
       // Prevent unhandled rejection
-      if (error instanceof errors.ShutdownError) return;
+      if (error.name == 'ShutdownError') return;
```
bergundy, line 213: I don't think this is required since the exception is thrown in this file

bergundy:
I'm fixing this in #660, closing this PR, thanks @airhorns

PR #662, opened 2022-05-23
Updating Doc URLs

rachfop:
Part of this [migration](https://github.com/temporalio/documentation/pull/1136) there is a new URL.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


lorensr (changes requested):
There are other instances, eg other package readmes

lorensr (approved):
I see other updates are in #663

PR #664, opened 2022-05-23
fix!: Use firstExecutionRunId instead of originalRunId

lorensr:
`originalRunId` is a concept related to resetting workflows. None of the instances of `originalRunId` I found in the SDK seem to do with resetting, so I changed them all to `firstExecutionRunId`.


bergundy (changes requested):
This should be `originalRunId` (or some alternative name) in the case of `signalWithStart` since we don't know if in that case if it's the first execution or not.

CC @cretz for Python

bergundy (approved):
Thanks!

PR #673, opened 2022-05-30
test: Reduce smorgasbord timeout sensitivity for CI

bergundy:


packages/test/src/workflows/smorgasbord.ts
```diff
@@ -17,17 +17,10 @@ import * as activities from '../activities/';
 import { signalTarget } from './signal-target';
 import { activityStartedSignal, unblockSignal } from './definitions';
 
-const { fakeProgress } = proxyActivities<typeof activities>({
-  startToCloseTimeout: '5s',
-  scheduleToCloseTimeout: '10s',
-  heartbeatTimeout: '3s',
+const { fakeProgress, queryOwnWf } = proxyActivities<typeof activities>({
+  startToCloseTimeout: '1m',
```
Sushisource, line 21: Doesn't this still need a heartbeat timeout?

bergundy: Not anymore since we have the default on the worker side now.
Adding the timeout just increases flakiness.

Sushisource (approved):

PR #674, opened 2022-05-30
fix: Validate that RetryPolicy maximumAttempts is an integer

### LINKED ISSUE #656: [Feature Request] Validate retry policy `maxAttempts` is an integer (e.g. not `Infinity`)

vkarpov15:
## What was changed

Added a quick check that `maximumAttempts` is a number using [`Number.isInteger()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger). This check explicitly excludes numbers like 3.1415 and `Number.POSITIVE_INFINITY`.

## Why?

`maximumAttempts` should always be an integer, doesn't make much sense for it to have a decimal component. And `maximumAttempts = 0` is the preferred alternative for `maximumAttempts = Number.POSITIVE_INFINITY`

## Checklist

1. Closes #656

2. How was this tested:

Added a couple of test cases

3. Any docs updates needed?

### DIFF
--- a/packages/test/src/test-retry-policy.ts
+++ b/packages/test/src/test-retry-policy.ts
@@ -35,10 +35,17 @@ test('compileRetryPolicy validates backoffCoefficient is greater than 0', (t) =>
   });
 });
 
-test('compileRetryPolicy validates maximumAttempts is greater than 0', (t) => {
-  t.throws(() => compileRetryPolicy({ maximumAttempts: 0 }), {
+test('compileRetryPolicy validates maximumAttempts is an integer', (t) => {
+  t.throws(() => compileRetryPolicy({ maximumAttempts: 3.1415 }), {
     instanceOf: ValueError,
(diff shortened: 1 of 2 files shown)


packages/test/src/test-retry-policy.ts
```diff
@@ -35,10 +35,17 @@ test('compileRetryPolicy validates backoffCoefficient is greater than 0', (t) =>
   });
 });
 
-test('compileRetryPolicy validates maximumAttempts is greater than 0', (t) => {
```
bergundy, line 38: Why did you delete this test?

bergundy (approved):
Please restore the deleted test but otherwise LGTM.
Thanks!

packages/test/src/test-retry-policy.ts
vkarpov15: 100% accidental. Sorry about that and thanks for catching this. Fixed :+1:

PR #675, opened 2022-05-30
feat: Upgrade neon to 0.10

bergundy:


packages/core-bridge/src/lib.rs
```diff
@@ -172,13 +173,13 @@ fn tokio_runtime() -> &'static Runtime {
     })
 }
 
-/// Send a result to JS via callback using an [EventQueue]
-fn send_result<F, T>(queue: Arc<EventQueue>, callback: Root<JsFunction>, res_fn: F)
+/// Send a result to JS via callback using an [Channel]
```
Sushisource, line 176:
```suggestion
/// Send a result to JS via callback using a [Channel]
```

packages/core-bridge/src/lib.rs
```diff
@@ -189,13 +190,13 @@ where
     });
 }
 
-/// Send an error to JS via callback using an [EventQueue]
-fn send_error<E, F>(queue: Arc<EventQueue>, callback: Root<JsFunction>, error_ctor: F)
+/// Send an error to JS via callback using an [Channel]
```
Sushisource, line 193:
```suggestion
/// Send an error to JS via callback using a [Channel]
```

Sushisource (approved):

packages/core-bridge/src/lib.rs
bergundy: My search and replace didn't catch this :)

PR #676, opened 2022-05-30
fix: Add links to API and other categories

### LINKED ISSUE #648: [Feature Request] Make docs breadcrumbs clickable
Currently only the house emoji is clickable:

image

vkarpov15:
## What was changed

Made it so that breadcrumbs are live links where possible. I didn't make "Classes" in the below example a link because there's no page currently to link to. But "API" and "activity" are links:

image

The only potential downside is that this change makes "Activity" in the sidebar a link as well. So clicking on "Activity" loads a new page, instead of just opening the sidebar.

image

What do you think @lorensr @bergundy - is this change worthwhile, or should we just find a way to hide the breadcrumbs?

## Why?

Breadcrumbs not being links is bad for user experience and SEO.

## Checklist

1. Closes #648

2. How was this tested: Local docs build with `npm start`. After I commit this, I'll test these changes out in the Netlify deploy preview.

3. Any docs updates needed?

### DIFF
--- a/packages/docs/sidebars.js
+++ b/packages/docs/sidebars.js
@@ -44,8 +44,8 @@ function nestMarkdownFiles() {
             label: category,
             items: [],
             customProps: {
-              breadcrumbLink: `/api/namespaces/${parts[0]}#${category.toLowerCase()}`
-            }
+              breadcrumbLink: `/api/namespaces/${parts[0]}#${category.toLowerCase()}`,
+            },
           };
           items.push(item);
         }


packages/docs/sidebars.js
```diff
@@ -12,13 +12,21 @@ function titleCase(str) {
 function nestMarkdownFiles() {
   return markdownFiles()
     .sort((a, b) => a.split('.').length - b.split('.').length)
-    .reduce((acc, f) => {
+    .reduce((acc, f, i) => {
```
bergundy, line 15: Looks unused

vkarpov15: It is unused, I added that for some debugging purposes. I'll remove it.

bergundy:
We do but since it relies on GH secrets that CI step is skipped in forks.

vkarpov15:
@bergundy this branch isn't in a fork, it's in the main sdk-typescript repo:

image

I guess Netlify just doesn't run on non-`main` branches for this repo. Worth it to merge into master for a preview, and then revert if it doesn't work.

bergundy:
I see the docs deployed: https://github.com/temporalio/sdk-typescript/runs/6659868858?check_suite_focus=true#step:15:22

bergundy:
Saw the preview, LGTM

bergundy:
@vkarpov15 you can use anchors to make `Classes` (and other sections) clickable, e.g: https://62951537de47f013911041b3--friendly-pasteur-613935.netlify.app/api/namespaces/worker#classes

vkarpov15:
@bergundy I took a closer look and it doesn't look like it is possible to make a category link to a header in a page with docusaurus. Here's why:

1. Category sidebar item must be of type 'doc' and reference a document by id: https://github.com/facebook/docusaurus/blob/bf1513a3e357025ab1cf4cdb4c09f21316b31cfd/packages/docusaurus-plugin-content-docs/src/sidebars/types.ts#L56
2. Document ids refer to pages, adding `#Classes` to a document id causes an error

So our only options are to either 1) see if we can write a plugin to add the anchor text after the fact, or 2) see if we can modify typedoc-plugin-markdown's output format to create a separate page for `/api/namespaces/worker/classes` that we can link to. Neither option seems particularly appealing. What do you think @bergundy ?

packages/docs/src/theme/DocBreadcrumbs/index.js
```diff
@@ -0,0 +1,97 @@
+/**
```
lorensr, line 1: can you add a comment with the URL and hash the file is from?

vkarpov15: @lorensr done :+1:

packages/docs/sidebars.js
lorensr: Reminder

PR #679, opened 2022-05-31
feat!(worker): Expose missing WorkerOptions and remove unsupported ones

bergundy:
- Added `enableNonLocalActivities`, `maxConcurrentLocalActivityExecutions`, `maxTaskQueueActivitiesPerSecond`
- Removed `nonStickyToStickyPollRatio`, `maxConcurrentWorkflowTaskPolls`, `maxConcurrentWorkflowTaskPolls` since those are not supported by Core


packages/core-bridge/index.d.ts
```diff
@@ -161,16 +161,14 @@ export interface WorkerOptions {
...
+  maxConcurrentLocalActivityExecutions: number;
+
   /**
-   * `maxConcurrentWorkflowTaskPolls` * this number = the number of max pollers that will
-   * be allowed for the nonsticky queue when sticky tasks are enabled. If both defaults are used,
-   * the sticky queue will allow 4 max pollers while the nonsticky queue will allow one. The
-   * minimum for either poller is 1, so if `max_concurrent_wft_polls` is 1 and sticky queues are
-   * enabled, there will be 2 concurrent polls.
+   * If set to `false` this worker will only handle workflow tasks and local activities, it will not
+   * poll for activity tasks.
```
lorensr, line 168:
```suggestion
   * poll for activity tasks.
   * 
   * @default true
```

lorensr (approved):

packages/core-bridge/index.d.ts
bergundy: There's no default in this interface, it's in `WorkerOptionsWithDefaults`

PR #681, opened 2022-06-03
fix: Use JSON.stringify() for non-error objects and String() for other values

### LINKED ISSUE #635: [Bug] Fix message when objects are thrown
### Describe the bug

Currently, `throw {a : 'test'}` results in a failure with message `"[object Object]"`. It should be `{a: "test"} [A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.]`

vkarpov15:
Fix #635

## What was changed

A little extra logic to handle edge cases when a Workflow throws a non-error value

## Why?

`JSON.stringify()` can throw an error, so need to make sure to handle that. Plus, `JSON.stringify()` isn't necessarily the right choice for non-object values.

## Checklist

1. Closes #635

2. How was this tested:

3. Any docs updates needed?


lorensr (approved):
👍 thanks. `npm run lint`

packages/test/src/integration-tests.ts
```diff
@@ -918,6 +918,38 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+          throw new Error('No WFT failed event');
+        }
+        const failure = wftFailedEvent.workflowTaskFailedEventAttributes?.failure;
+        if (!failure) {
+          t.fail();
+          return;
+        }
+        t.is(
+          failure.message,
+          '42 [A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.]'
```
bergundy, line 945: Let's remove the period at the end of this.
In general, our error message should include periods IMHO.

bergundy (approved):

packages/common/src/failure.ts
```diff
@@ -337,13 +337,22 @@ export function errorToFailure(err: unknown, payloadConverter: PayloadConverter)
...
   }
 
-  const recommendation = ` [A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace.]`;
+  const recommendation = ` [A non-Error value was thrown from your code. We recommend throwing Error objects so that we can provide a stack trace]`;
 
   if (typeof err === 'string') {
     return { ...base, message: err + recommendation };
   }
+  if (typeof err === 'object') {
+    let message: string = '';
```
bergundy, line 346: Please fix lint and merge

vkarpov15: @bergundy done :+1:

lorensr:
Just getting CI to pass. Rerunning failed now

vkarpov15:
It looks like the same tests are failing in main, so I don't think this PR is introducing any failures

PR #683, opened 2022-06-04
fix: Use error constructor name as `applicationFailureInfo.type`

### LINKED ISSUE #637: [Bug] Fix non-TemporalFailure type
### Describe the bug

We say that we set a failure's `type` [to the exception full type name when it doesn't extend TemporalFailure](https://typescript.temporal.io/api/classes/common.applicationfailure). However, `throw new MyFailure()`, where `MyFailure extends Error`, results in `type: 'Error'`:

image

It should be `type: 'MyFailure'`

vkarpov15:
## What was changed

Use `err.constructor.name` instead of `err.name` by default, since `.name` is still 'Error' for classes that extend Error.

## Checklist

1. Closes #637

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -366,7 +366,7 @@ export function ensureTemporalFailure(err: unknown): TemporalFailure {
   if (err instanceof TemporalFailure) {
     return err;
   } else if (err instanceof Error) {
-    const name = err.constructor?.name || err.name;
+    const name = err.constructor?.name ?? err.name;
     const failure = new ApplicationFailure(err.message, name, false);
     failure.stack = err.stack;
     return failure;
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -366,7 +366,8 @@ export function ensureTemporalFailure(err: unknown): TemporalFailure {
   if (err instanceof TemporalFailure) {
     return err;
   } else if (err instanceof Error) {
-    const failure = new ApplicationFailure(err.message, err.name, false);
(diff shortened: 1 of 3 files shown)


bergundy (approved):
Nice, thanks!

packages/common/src/failure.ts
```diff
@@ -357,7 +357,8 @@ export function ensureTemporalFailure(err: unknown): TemporalFailure {
   if (err instanceof TemporalFailure) {
     return err;
   } else if (err instanceof Error) {
-    const failure = new ApplicationFailure(err.message, err.name, false);
+    const name = err.constructor?.name || err.name;
```
bergundy, line 360: Should we even use `err.name` anymore? Probably makes sense not to.

vkarpov15: @bergundy I think it can't hurt to have it as a fallback. You can overwrite `constructor`:

```
const err = new Error('test');
err.constructor = null;
err.constructor?.name; // undefined
```

bergundy: Yeah, but I wouldn't care about this too much TBH.
I'd probably change this to `err.construtor.name ?? 'Error'`

bergundy (approved):
@vkarpov15 I'll take over this PR and get it merged.

packages/common/src/failure.ts
bergundy: You know what, let's default to `err.name` as you suggested, looks like it should inherit from `Error`.

PR #689, opened 2022-06-07
fix(Worker): Throw error when importing a node built-in module from a Workflow

### LINKED ISSUE #685: [Bug] TypeScript: When workflow code imports built-in Node modules, worker logs the problem instead of loudly failing

Debug workflow failures.

### Describe the bug

When workflow code imports built-in Node modules, worker logs the problem instead of loudly failing.

In these logs from worker startup, you'll see that the problem is detected but then just logged:

```
2022-06-06T19:26:13.471Z [INFO]   modules by path ../config/environments/*.ts 13.1 KiB 5 modules
2022-06-06T19:26:13.471Z [INFO]   modules by path ../config/*.ts 5.6 KiB 3 modules
2022-06-06T19:26:13.471Z [INFO] modules by path ../shared-universal/ 130 KiB
2022-06-06T19:26:13.472Z [INFO]   modules by path ../shared-universal/report/ 62.2 KiB 6 modules
2022-06-06T19:26:13.472Z [INFO]   modules by path ../shared-universal/utils/*.ts 20.6 KiB 5 modules
2022-06-06T19:26:13.472Z [INFO]   ../shared-universal/generated/graphql.ts 46.7 KiB [built] [code generated]
2022-06-06T19:26:13.472Z [INFO] __temporal_custom_payload_converter (ignored) 15 bytes [built] [code generated]
2022-06-06T19:26:13.472Z [INFO] child_process (ignored) 15 bytes [built] [code generated]
2022-06-06T19:26:13.474Z [INFO] path (ignored) 15 bytes [built] [code generated]
2022-06-06T19:26:13.474Z [INFO] webpack 5.65.0 compiled successfully in 644 ms
2022-06-06T19:26:13.474Z [WARN] Your Workflow code (or a library used by your Workflow code) is importing the following built-in Node modules:
  - 'child_process'
  - 'path'
Workflow code doesn't have access to built-in Node modules (in order to help enforce determinism). If you are certain these modules will not be used at runtime, then you may add their names to 'WorkerOptions.bundlerOptions.ignoreModules' in order to dismiss this warning. However, if your code execution actually depends on these modules, then you must change the codeor remove the library.
See https://typescript.temporal.io/api/interfaces/worker.workeroptions/#bundleroptions for details.
2022-06-06T19:26:13.480Z [INFO] Workflow bundle created { size: '3.29MB' }
2022-06-06T19:26:13.684Z [INFO] Worker state changed { state: 'RUNNING' }
```

### Minimal Reproduction

Import `path` into workflow code and start the worker.

### Environment/Versions

- OS and processor: Mac
- Temporal Version: 0.21.0
- Are you using Docker or Kubernetes or building Temporal from source? Docker

vkarpov15:
## What was changed

Instead of logging a warning, throw an error when importing a node built-in

## Why?

If you `import path from 'path'`, you get a warning, which is easy to miss in the log output. Then you end up with a hard-to-diagnose runtime error if you use `path.join()`:

```
  2022-06-07T18:54:11.935439Z  WARN temporal_sdk_core::worker: Failing workflow activation, run_id: "c9db39d3-851f-4386-bd8e-682b679e82cb", failure: Failure { failure: Some(Failure { message: "path__WEBPACK_IMPORTED_MODULE_1___default(...).join is not a function", source: "TypeScriptSDK", stack_trace: "TypeError: path__WEBPACK_IMPORTED_MODULE_1___default(...).join is not a function\n    at example (webpack-internal:///./src/workflows.ts:16:68)", cause: None, failure_info: None }) }
```

Right now, I'm thinking that throwing an error is better DX. What do you think?

## Checklist

1. Closes #685

2. How was this tested:
Added a test

3. Any docs updates needed?

### DIFF
--- a/packages/test/src/test-worker-from-bundle.ts
+++ b/packages/test/src/test-worker-from-bundle.ts
@@ -98,23 +98,21 @@ if (RUN_INTEGRATION_TESTS) {
     t.pass();
   });
 
-  test('A warning is reported when workflow depends on a node built-in module', async (t) => {
+  test('An error is thrown when workflow depends on a node built-in module', async (t) => {
     const logs: LogEntry[] = [];
     const logger = new DefaultLogger('WARN', (entry: LogEntry) => {
       logs.push(entry);
       console.warn(entry.message);
     });
 
-    await bundleWorkflowCode({
-      workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
-      logger,
-    });
-
-    t.true(
-      logs.some(
-        (entry) => entry.message.match(/'dns'/) && entry.message.match(/'WorkerOptions.bundlerOptions.ignoreModules'/)
-      ),
-      "Bundler reported a warning message about package 'dns'"
+    await t.throwsAsync(
+      bundleWorkflowCode({
+        workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
+        logger,
+      }),
+      {
+        instanceOf: Error,
+      }
     );
   });
 }
--- a/packages/test/src/test-worker-from-bundle.ts
+++ b/packages/test/src/test-worker-from-bundle.ts
@@ -112,6 +112,7 @@ if (RUN_INTEGRATION_TESTS) {
       }),
       {
         instanceOf: Error,
+        message: /is importing the following built-in Node modules.*dns/,
       }
     );
   });
--- a/packages/test/src/test-worker-from-bundle.ts
+++ b/packages/test/src/test-worker-from-bundle.ts
@@ -112,7 +112,7 @@ if (RUN_INTEGRATION_TESTS) {
       }),
       {
         instanceOf: Error,
-        message: /is importing the following built-in Node modules.*dns/,
+        message: /is importing the following built-in Node modules.*dns/s,
       }
     );
   });
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -227,7 +227,22 @@ export class WorkflowCodeBundler {
                 )
               );
             }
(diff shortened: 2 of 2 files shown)


bergundy (approved):
Thanks, I'd maybe consider logging this long error instead of putting it in the message (depends on how it's rendered on screen) but other than than LGTM.

vkarpov15:
My 2 cents: I prefer long error messages, or error messages that link to further info to keep the message short. I don't like when there's additional console output in addition to the error. Because not everyone is looking at their console.log output. And error handling tools often just take the error message and stack. So I'm fine with the error message as is.

packages/test/src/test-worker-from-bundle.ts
```diff
@@ -98,23 +98,21 @@ if (RUN_INTEGRATION_TESTS) {
...
-        (entry) => entry.message.match(/'dns'/) && entry.message.match(/'WorkerOptions.bundlerOptions.ignoreModules'/)
-      ),
-      "Bundler reported a warning message about package 'dns'"
+    await t.throwsAsync(
+      bundleWorkflowCode({
+        workflowsPath: require.resolve('./mocks/workflows-with-node-dependencies/issue-516'),
+        logger,
+      }),
+      {
+        instanceOf: Error,
```
lorensr, line 114:
```suggestion
        instanceOf: Error,
        message: /is importing the following built-in Node modules.*dns/,
```

lorensr (approved):

packages/test/src/test-worker-from-bundle.ts
```diff
@@ -98,23 +98,22 @@ if (RUN_INTEGRATION_TESTS) {
...
+        message: /is importing the following built-in Node modules.*dns/,
```
lorensr, line 115:
```suggestion
        message: /is importing the following built-in Node modules.*dns/s,
```

PR #701, opened 2022-06-14
chore: Upgrade to latest Core

### LINKED ISSUE #697: [Bug] Move native connection fields to worker
Move these to Worker: 

```
NativeConnectionOptions.identity
NativeConnectionOptions.workerBinaryId
```

bergundy:
- Closes #697 
- Exposes `operatorService` on client `Connection`.

### DIFF
--- a/packages/core-bridge/src/lib.rs
+++ b/packages/core-bridge/src/lib.rs
@@ -3,6 +3,7 @@ mod errors;
 
 use crate::conversions::{get_optional, ObjectHandleConversionsExt};
 use errors::*;
+use futures::stream::StreamExt;
 use neon::prelude::*;
 use neon::types::buffer::TypedArray;
 use opentelemetry::trace::{FutureExt, SpanContext, TraceContextExt};
@@ -38,11 +39,12 @@ use tokio::{
     runtime::Runtime,
     sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
 };
+use tokio_stream::wrappers::UnboundedReceiverStream;
 
 type RawClient = RetryClient<ConfiguredClient<WorkflowServiceClientWithMetrics>>;
 
 /// A request from JS to bridge to core
-enum Request {
+enum RuntimeRequest {
     /// A request to shutdown the runtime, breaks from the thread loop.
     Shutdown {
         /// Used to send the result back into JS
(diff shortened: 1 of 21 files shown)


packages/worker/src/worker.ts
```diff
@@ -149,15 +163,26 @@ export class NativeWorker implements NativeWorkerLike {
...
+  public static async create(
+    connection: NativeConnection,
+    options: CompiledWorkerOptions,
+    bundle?: WorkflowBundleWithSourceMap
+  ): Promise<NativeWorkerLike> {
     const runtime = Runtime.instance();
-    const nativeWorker = await runtime.registerWorker(extractNativeClient(connection), options);
+    const nativeWorker = await runtime.registerWorker(
+      extractNativeClient(connection),
+      addBuildId(options, bundle?.code)
```
Sushisource, line 174: I don't think we should automatically attach the autogenerated build id no matter what.

Users should be able to override it if they so desire but it looks like that's not possible if they pass a bundle right now.

packages/worker/src/worker.ts
```diff
@@ -398,15 +457,19 @@ export class Worker {
     const evictionPromise = firstValueFrom(
       constructedWorker.evictionsSubject.pipe(
         // Ignore self-induced shutdowns, as they are a legit and we'll terminate normally.
-        filter((ev) => ev.evictJob != this.SELF_INDUCED_SHUTDOWN_EVICTION),
-        map((ev) => {
-          if (ev.evictJob.reason === EvictionReason.NONDETERMINISM) {
+        filter(
+          ({ evictJob }) =>
+            evictJob.message !== this.SELF_INDUCED_SHUTDOWN_EVICTION.message &&
+            evictJob.reason !== this.SELF_INDUCED_SHUTDOWN_EVICTION.reason
```
Sushisource, line 463: Why this `&&`? Seems possibly over-strict?

packages/worker/src/worker.ts
bergundy: Yeah, an oversight, thanks.

packages/worker/src/worker.ts
bergundy: What's the right logic here?

bergundy: This was something you originally implemented and I overlooked in the review, `==` doesn't work for deep equality check.

Sushisource: Oh I didn't realize it was comparing against a object before. This is fine then -- better would be a `equals` method on the fake eviction thing or a `deepEqual` call or something like that. Seeing the fields fanned out here instead of that being encapsulated just makes it a bit harder to read/fragile IMO

Sushisource (approved):

packages/core-bridge/src/lib.rs
```diff
@@ -358,17 +358,29 @@ fn start_bridge_loop(channel: Arc<Channel>, receiver: &mut UnboundedReceiver<Req
...
+                    match Arc::try_unwrap(worker) {
+                        Ok(unwrapped_worker) => {
+                            tokio::spawn(void_future_to_js(
+                                channel,
+                                callback,
+                                async move {
+                                    unwrapped_worker.finalize_shutdown().await;
+                                    // Wrap the empty result in a valid Result object
+                                    let result: Result<(), String> = Ok(());
+                                    result
```
Sushisource, line 370: Suggest isn't working, but can just be
```
Ok::<_, String>::(())
```

packages/core-bridge/src/lib.rs
```diff
@@ -157,14 +146,13 @@ impl Finalize for Client {}
...
-    runtime: Arc<RuntimeHandle>,
-    core_worker: Arc<CoreWorker>,
+struct WorkerHandle {
+    sender: UnboundedSender<WorkerRequest>,
 }
 
 /// Box it so we can use Worker from JS
-type BoxedWorker = JsBox<RefCell<Option<Worker>>>;
-impl Finalize for Worker {}
+type BoxedWorker = JsBox<RefCell<Option<WorkerHandle>>>;
```
Sushisource, line 154: I believe you shouldn't need the RefCell and option any longer

PR #703, opened 2022-06-15
feat(worker): Add `runUntil` method

bergundy:
Useful helper especially in tests.

Usage:

```ts
const worker = await Worker.create(opts);

// Wait on an async function
  client.execute(someWorkflow, wopts);
});

// -- OR --

// Wait on a promise
await worker.runUntil(client.execute(someWorkflow, wopts));
```

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -1493,6 +1493,29 @@ export class Worker {
   }
 
   /**
+   * {@link run | Run} the Worker until `fnOrPromise` completes.
+   *
+   * @returns the result of `fnOrPromise`
+   *
+   * Throws on fatal Worker errors.
+   */
+  async runUntil<R>(fnOrPromise: Promise<R> | (() => Promise<R>)): Promise<R> {
+    const runAndShutdown = async () => {
+      try {
+        if (typeof fnOrPromise === 'function') {
+          return await fnOrPromise();
+        } else {
+          return await fnOrPromise;
+        }
+      } finally {
+        this.shutdown();
+      }
+    };
+    const [_, ret] = await Promise.all([this.run(), runAndShutdown()]);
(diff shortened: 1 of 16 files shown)


Sushisource (approved):

packages/worker/src/worker.ts
```diff
@@ -1492,6 +1492,29 @@ export class Worker {
...
   }
 
+  /**
+   * {@link run | Run} the Worker until `fnOrPromise` completes.
+   *
+   * @returns the result of `fnOrPromise`
+   *
+   * Throws on fatal Worker errors.
+   */
+  async runUntil<R>(fnOrPromise: Promise<R> | (() => Promise<R>)): Promise<R> {
```
cretz, line 1502: Isn't this just `this.run().finally(typeof fnOrPromise === 'function' ? fnOrPromise : () => fnOrPromise)`?

cretz, line 1502: May be too late now, but I think we should make `shutdown()` async and not return until the entire shutdown is complete. That way, people could just `Promise.race([worker.run(), mySignalPromise]).finally(async () => await worker.shutdown())`.

bergundy: I think `runUntil` is cleaner, I clearly was missing it in all of the tests.

cretz: As just a helper, sure (though I would like to avoid a slippery slope of _too_ many helpers).

But I still wonder if making shutdown async and having a non-async worker.start() could help. E.g.

```js
worker.start()
try {
  // Do stuff
} finally {
  await worker.shutdown()
}
```

Granted I could understand that people may not want to await shutdown but rather the original run, so maybe what you have here is best.

cretz (approved):

PR #705, opened 2022-06-17
Fix error details being dropped when returning core errors

Sushisource:
## What was changed
Fixed error reporting to use debug instead of display formatting, which was hiding detail. Also cleaned up error conversion code.

## Why?
Detail good.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (approved):
Nice!

lorensr (approved):

PR #711, opened 2022-06-19
fix: Make docs /api page show all namespaces

vkarpov15:
## What was changed

Currently https://typescript.temporal.io/api/ only has testing namespace:

image

This is because we're using two instances of docusaurus-plugin-typedoc without using their [pattern for multiple instances](https://github.com/tgreyuk/typedoc-plugin-markdown/tree/master/packages/docusaurus-plugin-typedoc#multi-instance). I don't see why we don't just have one instance and pull in `testing`?

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/docs/docusaurus.config.js
+++ b/packages/docs/docusaurus.config.js
@@ -125,20 +125,6 @@ module.exports = {
         watch,
       },
     ],
-    [
-      'docusaurus-plugin-typedoc',
-      {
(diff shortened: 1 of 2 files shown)


bergundy:
This isn't the right way to do it, `@temporalio/testing` isn't a dependency of the meta package, I'll fix this.

PR #714, opened 2022-06-21
fix(activity): Set Info.isLocal correctly

lorensr:
## What was changed

- Set isLocal correctly. Was broken, reported here: https://temporalio.slack.com/archives/C01DKSMU94L/p1655837399070479
- Share workflow bundle in local activity tests to avoid an error about out of wasm memory (swc?)


packages/test/src/test-local-activities.ts
```diff
@@ -12,28 +12,36 @@ import * as workflows from './workflows/local-activity-testers';
...
+  getWorker: () => Promise<Worker>;
 }
 
 const test = anyTest as TestInterface<Context>;
 
+let workflowBundle: WorkflowBundleWithSourceMap;
+
+test.before(async () => {
+  workflowBundle = await bundleWorkflowCode({ workflowsPath: require.resolve('./workflows/local-activity-testers') });
+});
```
bergundy, line 24: Let's do this cleanly and put it in the test context.

bergundy (approved):
Nice, just one comment

packages/test/src/test-local-activities.ts
```diff
@@ -10,30 +10,40 @@ import { RUN_INTEGRATION_TESTS } from './helpers';
...
   client: WorkflowClient;
+  getWorker: () => Promise<Worker>;
 }
 
 const test = anyTest as TestInterface<Context>;
 
+test.before(async (t) => {
+  t.context = {
+    workflowBundle: await bundleWorkflowCode({ workflowsPath: require.resolve('./workflows/local-activity-testers') }),
+  } as Context;
```
lorensr, line 24: @bergundy Like this? (Without the cast, the other props are optional, which adds a lot of `?.` to the tests.)

bergundy: `t.context.workflowBundle = ...`

bergundy (approved):

PR #715, opened 2022-06-21
fix: Simplify DocBreadcrumbs so we don't need the swizzle warning comment

vkarpov15:
## What was changed

Wrote a much simplified version of breadcrumbs that is functionally equivalent to the swizzled one

## Why?

Maintaining a swizzled component with a patch will likely get messy and make it harder to upgrade docusaurus. I think a simplified version that is designed for our use case, with a minimum of docusaurus dependencies, will be easier to maintain going forward.

Plus, having our own version of breadcrumbs will make it easier for us to make future improvements without having to get the improvements merged into docusaurus. For example, I'm not fond of how the home page shows 2 breadcrumbs items instead of 1:

image

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy:
Cool.
Looks like the link for enums is broken, e.g should be: https://62b253f90cc8bc00845259d2--friendly-pasteur-613935.netlify.app/api/namespaces/common#enumerations

Let's merge once you've fix the lint error, and issue above.

vkarpov15:
@bergundy fixed :+1:

PR #716, opened 2022-06-22
docs: Document ParentClosePolicy and ChildWorkflowCancellationType

lorensr:
Spencer checked `ChildWorkflowCancellationType`

### DIFF
--- a/packages/internal-workflow-common/src/workflow-options.ts
+++ b/packages/internal-workflow-common/src/workflow-options.ts
@@ -6,10 +6,35 @@ import { checkExtends, Replace } from './type-helpers';
 
 // Avoid importing the proto implementation to reduce workflow bundle size
 // Copied from temporal.api.enums.v1.WorkflowIdReusePolicy
+/**
+ * Concept: [Workflow Id Reuse Policy](https://docs.temporal.io/concepts/what-is-a-workflow-id-reuse-policy/)
+ *
+ * Whether a Workflow can be started with a Workflow Id of a Closed Workflow.
+ *
+ * *Note: A Workflow can never be started with a Workflow Id of a Running Workflow.*
+ */
 export enum WorkflowIdReusePolicy {
+  /**
+   * No need to use this.
+   *
+   * (If a `WorkflowIdReusePolicy` is set to this, or is not set at all, the default value will be used.)
+   */
   WORKFLOW_ID_REUSE_POLICY_UNSPECIFIED = 0,
+
+  /**
+   * The Workflow can be started if the previous Workflow is in a Closed state.
+   * @default
+   */
   WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE = 1,
+
+  /**
+   * The Workflow can be started if the previous Workflow is in a Closed state that is not Completed.
+   */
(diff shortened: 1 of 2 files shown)


packages/internal-workflow-common/src/workflow-options.ts
```diff
@@ -20,10 +45,10 @@ export interface BaseWorkflowOptions {
    * Specifies server behavior if a completed workflow with the same id exists. Note that under no
    * conditions Temporal allows two workflows with the same namespace and workflow id run
    * simultaneously.
-   *   ALLOW_DUPLICATE_FAILED_ONLY is a default value. It means that workflow can start if
-   *   previous run failed or was canceled or terminated.
+   *   ALLOW_DUPLICATE_FAILED_ONLY is a default value. It means that
```
bergundy, line 48: What was the intent here?

packages/workflow/src/interfaces.ts
```diff
@@ -165,19 +165,68 @@ export interface ContinueAsNewOptions {
...
+ * Specifies:
+ * - whether cancellation requests are sent to the Child
+ * - whether and when a {@link CanceledFailure} is thrown from {@link executeChild} or
+ *   {@link ChildWorkflowHandle.result}
+ *
+ * @default {@link ChildWorkflowCancellationType.WAIT_CANCELLATION_COMPLETED}
+ */
 export enum ChildWorkflowCancellationType {
+  /**
+   * Don't send a cancellation request to the Child. Don't throw an error.
```
bergundy, line 178: Not sure this is right, I think that the error does get thrown but I'd need to verify.

I don't think we test all of these behaviors, I guess we will in the sdk-features repo.

packages/workflow/src/interfaces.ts
```diff
@@ -165,19 +165,68 @@ export interface ContinueAsNewOptions {
...
+   */
   ABANDON = 0,
+
+  /**
+   * Send a cancellation request to the Child. Immediately throw the error.
+   */
   TRY_CANCEL = 1,
+
+  /**
+   * Send a cancellation request to the Child. Throw the error if and when the Child's cancellation is completed.
```
bergundy, line 188: I'd note that the child might not cooperate and might end up in either failed / cancelled or completed states (or never complete).

packages/workflow/src/interfaces.ts
```diff
@@ -165,19 +165,68 @@ export interface ContinueAsNewOptions {
...
+/**
+ * Concept: [Parent Close Policy](https://docs.temporal.io/concepts/what-is-a-parent-close-policy/)
+ *
+ * How a Child Workflow reacts to the Parent Workflow reaching a Closed state.
+ */
 export enum ParentClosePolicy {
+  /**
+   * No need to use this.
+   *
+   * (If a `ParentClosePolicy` is set to this, or is not set at all, the default value will be used.)
```
bergundy, line 211: I'd just go with:
```suggestion
   * If a `ParentClosePolicy` is set to this, or is not set at all, the server default value will be used.
```

packages/internal-workflow-common/src/workflow-options.ts
lorensr: Oops thanks—meant to move to the enum values

bergundy (approved):

PR #638, opened 2022-05-05
chore(test): Use GITHUB_TOKEN for higher rate limit

lorensr:
## Problem

CI is rate limited by Github REST API:

```
@temporalio/testing: > node ./scripts/download-test-server.mjs
@temporalio/testing: internal/process/esm_loader.js:74
@temporalio/testing:     internalBinding('errors').triggerUncaughtException(
@temporalio/testing:                               ^
@temporalio/testing: HTTPError: Response code 403 (rate limit exceeded)
```
[build-and-test (14, macos-latest)](https://github.com/temporalio/sdk-typescript/runs/6314570053?check_suite_focus=true#logs)

Currently on the default anonymous 60 requests per hour per IP address.

## Solution

Authenticate, to get 1k requests per hour per repo:

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#requests-from-github-actions

PR #721, opened 2022-06-24
fix: Use GITHUB_TOKEN in create-project for CI

lorensr:
Continuing from https://github.com/temporalio/sdk-typescript/pull/638


bergundy (approved):
Thanks, let's get rid of all these CI flakes

PR #722, opened 2022-06-24
fix(worker): Disallow importing non-Workflow @temporalio packages

lorensr:
Following up on https://github.com/temporalio/sdk-typescript/pull/645#issuecomment-1158326034

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -17,11 +17,10 @@ export const disallowedModules = [
   '@temporalio/client',
   '@temporalio/worker',
   '@temporalio/internal-non-workflow-common',
-  '@temporalio/interceptors-opentelemetry',
-  '@temporalio/test',
+  '@temporalio/interceptors-opentelemetry/lib/client',
+  '@temporalio/interceptors-opentelemetry/lib/worker',
   '@temporalio/testing',
   '@temporalio/core-bridge',
-  '@temporalio/create',
 ];
 
 export function moduleMatches(userModule: string, modules: string[]): boolean {
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -16,6 +16,12 @@ export const disallowedModules = [
   '@temporalio/activity',
   '@temporalio/client',
   '@temporalio/worker',
+  '@temporalio/internal-non-workflow-common',
+  '@temporalio/interceptors-opentelemetry',
+  '@temporalio/test',
+  '@temporalio/testing',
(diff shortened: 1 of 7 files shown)


packages/worker/src/workflow/bundler.ts
```diff
@@ -11,6 +11,12 @@ import { DefaultLogger, Logger } from '../logger';
 
 export const allowedBuiltinModules = ['assert'];
 export const disallowedBuiltinModules = builtinModules.filter((module) => !allowedBuiltinModules.includes(module));
+export const disallowedModules = [
+  ...disallowedBuiltinModules,
+  '@temporalio/activity',
+  '@temporalio/client',
+  '@temporalio/worker',
```
bergundy, line 18: Maybe add non-workflow-common, the opentelemetry client and worker libs and most of the testing stuff?

bergundy (approved):
Approved with a comment, thanks @lorensr

packages/worker/src/workflow/bundler.ts
lorensr: 👍 I like when I remember to use `jq`

```
$ cat packages/*/package.json | jq .name
"@temporalio/activity"
"@temporalio/client"
"@temporalio/common"
"@temporalio/core-bridge"
"@temporalio/create"
"@temporalio/docs"
"@temporalio/interceptors-opentelemetry"
"@temporalio/internal-non-workflow-common"
"@temporalio/internal-workflow-common"
"temporalio"
"@temporalio/proto"
"@temporalio/test"
"@temporalio/testing"
"@temporalio/worker"
"@temporalio/workflow"
```

vkarpov15 (approved):
Nice :+1:

bergundy (approved):

bergundy:
Looks like CI found a real issue, nice to see that the fix works.

bergundy:
@lorensr we should probably see what's importing from non-workflow-common and fix it.

lorensr:
@bergundy this file imported from `otel.ts` which was located in `internal-non-workflow-common`. I moved it to `common` (not `internal-workflow-common`, since it uses `defaultDataConverter` which is in common).

bergundy:
I'd make it so otel api isn't required for workflow context

bergundy:
This would require some code duplication but I'd rather have that then the extra dependency

lorensr:
For bundle size sake? Will it be included if the user doesn't import from `@temporalio/interceptors-opentelemetry/lib/workflow`?

bergundy:
Hmm.. nevermind, it's okay, we don't export `lib/otel` from `common/index`

Issue #724, opened 2022-06-26
[Bug] Update SDK README to match 1.0.0 capabilities


yoDon:
### Describe the bug

The SDK README is out of date with the current capabilities of the SDK

### Minimal Reproduction

SDK indicates, for example, that Upsert on search attributes is not implemented.

### Environment/Versions

N/A

### Additional context

Congrats on getting the TypeScript SDK ready for release!

bergundy:
I’ve been meaning to update the readme for a while and haven’t gotten around to it.
Thanks for reporting.

PR #725, opened 2022-06-26
Update SDK README

yoDon:
Minor/trivial commit in support of fixing https://github.com/temporalio/sdk-typescript/issues/724. Readme was still reporting that the SDK lacked Upsert for search attributes. Fixed.


bergundy (approved):
The readme requires some more love before the 1.0.0 release but that’s a start, thanks

PR #735, opened 2022-06-29
fix: Add @types/long to client

lorensr:
Client package has this:

```
  historyLength: Long;
```

`Long` is defined in `@types/long`

This is the only remaining missing dep I found in https://community.temporal.io/t/error-when-starting-worker/4966/8?u=loren


bergundy (commented):
Is this imported directly by the client package code?

lorensr:
No, I want to make sure someone can go to definition of `Long` in their editor even if using pnp/pnpm, but maybe they can anyway? Maybe @mjameswh had another reason?

mjameswh:
I don't remember exactly why we had to add `@types/long`, though I know it was required for both Yarn3 and PNPM setup.

Let me try and config a small repo based on `samples-typescript/hello-world`, adapted to PNPM. That will be helpful for further work on fixing missing dependencies.

mjameswh:
> Let me try and config a small repo based on samples-typescript/hello-world, adapted to PNPM. That will be helpful for further work on fixing missing dependencies.

Here it is:
https://github.com/mjameswh/temporalio-typescriptsdk-pnpm-helloworld

> We should just add a pnpm sample and make sure to run it in the SDK integration tests

I can also prepare a yarn3 hello-world sample, thought it will have to wait next week.

mjameswh:
Note that in this case, I have had no issue about `@types/long`. Maybe this was fixed somehow during recent dependencies cleanup, or maybe that this simple demo isn't complex enough to force this requirement.

I'm trying again right now, with one of our bigger project, using a verdaccio build based on main, and removing all custom dependencies patching. This might take some time.

bergundy:
@mjameswh I think with the pnpm sample we should be able to figure out all of the missing dependencies (if any).
I wouldn't put too much time into reproducing this.

bergundy:
Looks like we need to import long in `client/src/types.ts` too.
@lorensr

bergundy (approved):
I’d just make history length a number so it’s easy to work with, not sure why it was long in the first place, just an oversight.

Approved if you want to merge for now.

lorensr:
I had automerge on. Wouldn't that technically be breaking?

bergundy:
Yeah but it’s fairly insignificant imo

PR #736, opened 2022-06-29
fix: Re-export from internal-workflow-common

### LINKED ISSUE #720: [Feature Request] `WorkflowNotFoundError` export is not consistent with `WorkflowExecutionAlreadyStartedError`
I am using version 0.23.x of the SDK (from NodeJS) and when working with common errors, the imports are always looking like the following snippet:
```
const { WorkflowExecutionAlreadyStartedError } = require('@temporalio/client');
const { WorkflowNotFoundError } = require('@temporalio/internal-workflow-common');
```
From what I've seen, there is no other place to find `WorkflowNotFoundError`, other than the `@temporalio/internal-workflow-common` module. I think it would be nice to have one place that re-exports all possible errors, so there is no need to import from "internal-workflow-common" at all.

I think that being able to import the errors from a single module would be great:
```
const {
  WorkflowExecutionAlreadyStartedError,
  WorkflowNotFoundError,
} = require('@temporalio/client'); // it doesn't necessarily have to be exported from the client module directly, that's just an example
```

### Additional context
I am using the `WorkflowNotFoundError` when checking errors due to invalid commands - like sending a signal to an already finished workflow. For example:
```
const handle = this.client.getHandle(id);
try {
	await handle.signal(CANCEL_SIGNAL_NAME, {});
} catch (err) {
	if (err instanceof WorkflowNotFoundError) {
		// handle this errors in a proper manner, so that it is not retried from the outside.
	}
        // all unknown errors are retried
	throw err;
}
```

lorensr:
Fixes #720

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -31,6 +31,7 @@ import { AbortSignal } from 'abort-controller';
 import { AsyncLocalStorage } from 'async_hooks';
 export { CancelledFailure } from '@temporalio/common';
 export { ActivityFunction, ActivityInterface } from '@temporalio/internal-workflow-common';
+export * from '@temporalio/internal-workflow-common/lib/errors';
 
 /**
  * Throw this error from an Activity in order to make the Worker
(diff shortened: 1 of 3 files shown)


bergundy (approved):
LGTM

PR #737, opened 2022-06-29
fix(bundler): avoid 'package source-map-loader not found' in PNPM

mjameswh:
## What was changed
In webpack configuration, replaced `'source-map-loader'` by `require.resolve('source-map-loader')`

## Why?
Plugins dependencies have to be resolved in the context of the `@temporalio/worker` package. This is true in general, but essential with strict package managers (including PNPM and Yarn3).

**Note: NOT TESTED**


bergundy (approved):
Thanks! That makes a lot of sense

mjameswh:
I just pushed a second (unrelated) change to this branch. Trying to fix some CI build issues that makes it error when processing PR from third party commiters (like me)...

lorensr:
Algolia thing is now fixed in main

bergundy:
Thanks @mjameswh

PR #743, opened 2022-07-06
feat: Support history from JSON

bergundy:


packages/common/src/proto-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+function fixHistory(h: Record<string, any>) {
+  return {
+    events: h.events.map(fixHistoryEvent),
+  };
+}
+
+/**
+ * Convert a proto JSON representation of History to a valid History object
+ */
+export function historyFromJSON(history: unknown): History {
```
cretz, line 62: Would `history: { events: []object }` make sense here?

packages/common/src/proto-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (typeof history !== 'object' || history == null || !Array.isArray((history as any).events)) {
+    throw new TypeError('Invalid history, expected an object with an array of events');
+  }
+  const loaded = fromProto3JSON(historyType, fixHistory(history));
+  if (loaded === null) {
+    throw new TypeError('Invalid history');
+  }
+  return loaded as any;
```
cretz, line 70: `loaded as History`?

packages/common/src/proto-utils.ts
```diff
@@ -0,0 +1,103 @@
+import proto from '@temporalio/proto';
+import { fromProto3JSON, toProto3JSON } from 'proto3-json-serializer';
+import { patchProtobufRoot } from '@temporalio/proto/lib/patch-protobuf-root';
+
+export type History = proto.temporal.api.history.v1.IHistory;
+export type Payload = proto.temporal.api.common.v1.IPayload;
+
+// Cast to any because the generated proto module types are missing the lookupType method
+const patched = patchProtobufRoot(proto) as any;
```
cretz, line 9: Safe to assume `patchProtobufRoot` can be re-executed on already-patched roots safely?

packages/common/src/proto-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+
+function fixHistoryEvent(e: Record<string, any>) {
+  const type = Object.keys(e).find((k) => k.endsWith('EventAttributes'));
+  if (!type) {
+    throw new TypeError(`Missing attributes in history event: ${JSON.stringify(e)}`);
+  }
+
+  return {
+    ...e,
+    eventType: `EVENT_TYPE_${pascalCaseToConstantCase(e.eventType)}`,
```
cretz, line 33: Nit, seems for consistency could use `...fixEnumValue(e, 'eventType', 'EVENT_TYPE')`

cretz (approved):

packages/common/src/proto-utils.ts
bergundy: No because typically users would feed the response of `JSON.parse` here

cretz: Then the user can choose to cast and/or ignore types

packages/common/src/proto-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+function fixHistory(h: Record<string, any>) {
+  return {
+    events: h.events.map(fixHistoryEvent),
+  };
+}
+
+/**
+ * Convert a proto JSON representation of History to a valid History object
```
lorensr, line 60:
```suggestion
 * Convert a proto JSON representation of History (from `tctl` or Web UI) to a valid History object
```

lorensr (approved):

packages/common/src/proto-utils.ts
bergundy: Only the return type matters here

PR #744, opened 2022-07-06
docs: Review API docs

lorensr:
## What was changed

- Adds links to docs & SDK docs on index
- Review 
- Excludes anything tagged `@internal`

## Next PR

- Add otel and testing to meta
- Finish review


packages/activity/src/index.ts
```diff
@@ -120,9 +147,14 @@ export interface Info {
 }
 
 /**
- * Activity Context manager.
+ * Activity Context, used to:
+ *
+ * - Get {@link Info} about the current Activity Execution
+ * - Send [heartbeats](https://docs.temporal.io/concepts/what-is-an-activity-heartbeat)
+ * - Get notified of Activity cancellation
+ * - Sleep
```
bergundy, line 155: I'd note that this is cancellation aware sleep

packages/internal-workflow-common/src/errors.ts
```diff
@@ -36,11 +40,11 @@ export class WorkflowExecutionAlreadyStartedError extends Error {
 }
 
 /**
- * Thrown when workflow with the given id is not known to the Temporal service.
+ * Thrown when a Workflow with the given Id is not known to Temporal Server.
  * It could be because:
- * - ID passed is incorrect
- * - Workflow execution is complete (for some calls e.g. terminate),
- * - workflow was purged from the service after reaching its retention limit.
+ * - Id passed is incorrect
```
bergundy, line 45: `ID` is clearer

packages/internal-workflow-common/src/errors.ts
```diff
@@ -67,7 +71,7 @@ interface ErrorWithCode {
   code: string;
 }
 /**
- * Get error code from an Error or return undefined
+ * * Get `error.code` (or `undefined` if not present)
```
bergundy, line 74:
```suggestion
 * Get `error.code` (or `undefined` if not present)
```

packages/workflow/src/workflow.ts
```diff
@@ -915,24 +915,21 @@ export function makeContinueAsNewFunc<F extends Workflow>(
 }
 
 /**
- * Continues current Workflow execution as new with default options.
+ * [Continues-As-New](https://docs.temporal.io/concepts/what-is-continue-as-new/) the current Workflow Execution with
+ * default options.
  *
- * Shorthand for `makeContinueAsNewFunc<F>()(...args)`.
+ * Shorthand for {@link makeContinueAsNewFunc}.
```
bergundy, line 921: I liked the extra details.

bergundy: You might want to leave as it was and add a `see also {@link make...}`

packages/workflow/src/workflow.ts
```diff
@@ -915,24 +915,21 @@ export function makeContinueAsNewFunc<F extends Workflow>(
...
  *
  * @example
  *
- * ```ts
- * import { continueAsNew } from '@temporalio/workflow';
+ *```ts
+ *import { continueAsNew } from '@temporalio/workflow';
  *
- * export function myWorkflow(n: number) {
```
bergundy, line 927: snipsync would have found this (but the example wouldn't show up in IDEs).
There's probably a better solution for testing the examples but it's not critical for this PR.

bergundy (approved):
Awesome stuff @lorensr!

All of theslinks that are dependent on content being rendered in docusaurus should change to `{@link ...}` so the docs can be used from IDEs.

packages/internal-workflow-common/src/errors.ts
lorensr: That's my pref as well, but I was trying to go by this, which says either Id or identifier:

https://github.com/temporalio/documentation/blob/master/STYLE.md#abbreviation-of-identifier

lorensr:
Omg this is the most useful example 😆

image

PR #747, opened 2022-07-07
docs: Deprecate `temporalio` meta package

### LINKED ISSUE #727: [Bug] Deprecate `temporalio` meta package
`npm install temporalio` doesn't work with package managers that do [strict dependency resolution](https://github.com/temporalio/sdk-typescript/issues/421#issuecomment-1067196052) (don't support transitive deps) like pnpm and yarn3.

We could either:
- keep `temporalio`, and warn at the top of the README and on install to not use it if you're using such a package manager
- change our setup instructions and samples `package.json`s to install the direct packages:

  ```
  npm install @temporalio/{client,worker,workflow,activity,common}
  ```

I think the latter is simpler / will result in less confusion.

lorensr:
- Updated README
- Added meta README
- Updated CONTRIBUTING
- Removed `--temporalio-version <version>` option from initializer (logic could be updated to support new deps, but option doesn't seem that useful)
- Fixes #727 (will also open PRs in docs and samples)


CONTRIBUTING.md
```diff
@@ -1,8 +1,44 @@
 # How to Contribute
 
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
bergundy, line 3: GH does this automatically, I'm fine if you want to keep this though

packages/create-project/src/create-project.ts
```diff
@@ -178,9 +176,6 @@ export async function createApp({
   console.log();
 
   await updateNodeVersion({ root });
-  if (temporalioVersion) {
-    await replaceTemporalVersion({ root, useYarn, temporalioVersion });
-  }
```
bergundy, line 183: I'd still want this to be able to specify a different SDK version

bergundy (commented):
You can export the missing packages from the meta package index file to fix the docs issue while you're at it

CONTRIBUTING.md
lorensr: Whoa this icon?!

image

I figure if I didn't know about it, many/most don't. Also nice to see the whole thing at once, vs built-in is scrolled.

packages/interceptors-opentelemetry/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@temporalio/interceptors-opentelemetry",
   "version": "1.0.0-rc.0",
   "description": "Temporal.io SDK interceptors bundle for tracing with opentelemetry",
-  "main": "lib/workflow/index.js",
-  "types": "./lib/workflow/index.d.ts",
+  "main": "lib/index.js",
+  "types": "./lib/index.d.ts",
```
lorensr, line 6: This might be breaking for people importing workflow code from `@temporalio/interceptors-opentelemetry`?

In our sample, we import from `@temporalio/interceptors-opentelemetry/lib/workflow`: https://github.com/temporalio/samples-typescript/blob/bd2e6d73dcc5de5ff41215a74745ab363ee7e52a/interceptors-opentelemetry/src/workflows.ts#L5

CONTRIBUTING.md
```diff
@@ -165,7 +201,7 @@ npx lerna publish from-git # add `--dist-tag next` for pre-release versions
 rm $HOME/Downloads/packages-*
 ```
 
-### Updating the Java test server proto files
+## Updating the Java test server proto files
```
bergundy, line 204: In hindsight as pointed out by @cretz I wish I'd just copied the files over but it's fine for now.

packages/create-project/src/helpers/install.test.ts
```diff
@@ -0,0 +1,26 @@
...
+import path from 'path';
+
+test('replaceTemporalVersion according to configured level', async (t) => {
+  console.log('debug windows');
+  console.log('import.meta.url', import.meta.url);
+  console.log('URL', new URL('../../test', import.meta.url));
+  const { pathname: testDir } = new URL('../../test', import.meta.url);
+  const packageJson = path.join(testDir, 'package.json');
+  console.log('packageJson', packageJson);
+  const packageJsonBackup = path.join(testDir, 'package.json.bak');
```
bergundy, line 14: I'd just create a temporary directory and not mess with the repo's structure.
We do this in other tests IIRC.

packages/create-project/src/index.ts
```diff
@@ -58,9 +58,9 @@ const program = new Command(packageJson.name)
 `
   )
   .option(
-    '--temporalio-version <version>',
+    '--temporal-version <version>',
```
bergundy, line 61: I'd rename to `sdk-version` it's clearer IMO

packages/interceptors-opentelemetry/src/index.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * [Documentation](https://docs.temporal.io/typescript/logging#opentelemetry-tracing)
+ *
+ * @module
+ */
+
+export * from './workflow';
+export * from './worker';
+export { OpenTelemetryWorkflowClientCallsInterceptor } from './client';
+export * from './instrumentation';
```
bergundy, line 14: These are internal helpers, don't export them.

packages/interceptors-opentelemetry/package.json
bergundy: It's fine, users should import from the individual subpackages.

bergundy (approved):
Approved with a single nit

lorensr:
Took a while to get CI to pass. Force pushed `samples-typescript/next` because I didn't want to deal with so many merge conflicts. Also ran the `npm deprecate` command. Now:

image

PR #750, opened 2022-07-08
feat: Revise SDK log attributes and levels

bergundy:
Adds 2 new default interceptors: `WorkflowInboundLogInterceptor` and `ActivityInboundLogInterceptor` and a default logger sink to provide better logging experience out of the box.

Also reduced the severity of internal SDK logs to `trace` level to reduce log noise.


bergundy:
@cretz you might also want to take a look at this

lorensr (approved):

Sushisource (commented):
If we're defaulting internal logging to trace, there should probably be a way to switch that level easily at runtime (or at least after a restart) without rebuilding the worker because otherwise it's going to make getting important info from users in the field more challenging. 

If that isn't easily possible right now, we should add a way with an env var or something.

bergundy:
If we want to expose an easy way to change the log level (which I don't think we do at the moment) I'd probably add a worker option, in the meantime if anybody wants to customize logging, they can set intercept the level in their log handler.
I'm against accessing env vars in any of our libraries, I don't think it's the library's responsibility.

bergundy:
We accept a logger interface with a minimal default implementation.
Users taking workers to production usually have their own logging stacks with their own methods of configuring logging.
If anything, users will want to reconfigure their own logger to bump up the log level to debug issues but that's out of our hands.

Adding an environment variable just for making these trace logs into debug logs is overdoing it IMHO.

PR #751, opened 2022-07-09
fix: Fix nightly and activity failing with non ApplicationFailure

bergundy:


packages/test/src/test-worker-activities.ts
```diff
@@ -253,3 +253,27 @@ test('Worker cancels activities after shutdown', async (t) => {
   t.truthy(result?.failed);
   t.true(activityCancelled);
 });
+
+test('Non ApplicationFailure TemporalFailures thrown from Activity are wrapped with ApplicationFailure', async (t) => {
```
lorensr, line 257: Should `CancelledFailure`s be wrapped in ApplicationFailures?

packages/worker/src/worker-options.ts
```diff
@@ -93,7 +93,7 @@ export interface WorkerOptions {
    * Time to wait for pending tasks to drain after shutdown was requested.
    *
    * @format {@link https://www.npmjs.com/package/ms | ms} formatted string or number of milliseconds
-   * @default 5s
+   * @default 10s
```
lorensr, line 96: I don't see this changed in the code—was the number just wrong before?

lorensr (approved):
I think wrapping more things in ApplicationFailures is potentially breaking, for those who do a specific number of `.cause`s in their error checks.

packages/test/src/test-worker-activities.ts
bergundy: Yes, unless the activity was requested to be cancelled

packages/worker/src/worker-options.ts
bergundy: 🤦‍♂️

bergundy:
@lorensr we’re not wrapping more things, we just ensure that only ApplicationFailure can be the failure returned from an activity. The server expects that

PR #757, opened 2022-07-14
fix: Add operatorService to ConnectionLike

lorensr:


bergundy:
Hmm..
I'd considered adding this initially but decided against it to reduce the implementation to a bare minimum.
`ConnectionLike` is used by the high level clients and they don't require access to the operator service.

lorensr:
I wanted to be able to do `client.connection.operatorService`, but I could create the Connection before creating the `WorkflowClient` and keep it around.

bergundy:
Let's revisit this later, we can add this addition interface constraint if we need to access the operator service APIs or when we have other implementations

PR #761, opened 2022-07-20
refactor(workflow)!: Move TaskInfo to WorkflowInfo

lorensr:
We want Tasks to be as much of an implementation detail as possible so that normally devs don't have to know/think about them. 

Added note to `workflowInfo()` about recommended usage due to the fact that fields change (and which fields change will also change).


bergundy (approved):
Perfect

PR #764, opened 2022-07-20
refactor!: Update activity and worker exports

lorensr:
BREAKING CHANGE: If you were importing any of these from `@temporalio/activity`, instead import from `@temporalio/common`: `ValueError, PayloadConverterError, IllegalStateError, WorkflowExecutionAlreadyStartedError, WorkflowNotFoundError`

### DIFF
--- a/packages/docs/docusaurus.config.js
+++ b/packages/docs/docusaurus.config.js
@@ -53,7 +53,7 @@ module.exports = {
             },
             {
               label: 'Community Slack',
-              href: 'https://join.slack.com/t/temporalio/shared_invite/zt-onhti57l-J0bl~Tr7MqSUnIc1upjRkw',
+              href: 'https://temporal.io/slack',
             },
           ],
         },
--- a/packages/docs/docusaurus.config.js
+++ b/packages/docs/docusaurus.config.js
@@ -40,7 +40,11 @@ module.exports = {
         {
           items: [
             {
-              label: 'NPM',
+              label: 'SDK Docs',
+              href: 'https://docs.temporal.io/typescript/introduction/',
+            },
+            {
+              label: 'npm',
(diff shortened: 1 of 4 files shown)


packages/docs/docusaurus.config.js
```diff
@@ -40,29 +40,25 @@ module.exports = {
...
+            {
+              label: 'npm',
               href: 'https://www.npmjs.com/package/temporalio',
             },
             {
               label: 'Support Forum',
               href: 'https://community.temporal.io/',
             },
             {
-              label: 'Public Slack',
```
bergundy, line 51: Yeah that's much better :)

packages/docs/docusaurus.config.js
```diff
@@ -40,29 +40,25 @@ module.exports = {
...
+              label: 'Community Slack',
               href: 'https://join.slack.com/t/temporalio/shared_invite/zt-onhti57l-J0bl~Tr7MqSUnIc1upjRkw',
```
bergundy, line 56: Can probably use the stable `temporal.io/slack` URL

bergundy (approved):

bergundy:
This is technically a breaking change although I'd bet that the vast majority of users never referenced these extra exports.

lorensr:
Right, thanks. I thought that but then forgot the !. Edited.

packages/worker/src/index.ts
```diff
@@ -21,8 +21,10 @@ export {
   CompiledWorkerOptions,
   WorkerOptions,
   WorkflowBundleOption,
+  WorkflowBundlePathWithSourceMap,
```
bergundy, line 24: nit: I'd update the PR description to note these additional imports

Issue #762, opened 2022-07-20
[Bug] worker client OOM when running as a container in a k8s cluster.


Vincechen611:
### What are you really trying to do?
Consider calculate default maxCachedWorkflows by v8.getHeapStatistics().total_available_size instead of os.totalmem()

### Describe the bug

Recently, We have implement our temporal worker client with sdk-typescript and run as a container in a k8s cluster.
We have set `NODE_OPTIONS=--max-old-space-size=XXX` when starting up worker, but we found the worker pod always get OOM killed by k8s when processing task.
After checking the detail, we found that the default `maxCachedWorkflows` is calculated by `os.totalmem()` if we didn't set.
When running in container, the `os.totalmem()` will get the host machine memory not container resource setting (which is in [cgroup](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/))
Could we calculate default [maxCachedWorkflows](https://github.com/temporalio/sdk-typescript/blob/a261054130708744c692a50b07d3f189a43b9934/packages/worker/src/worker-options.ts#L414) with v8.getHeapStatistics().total_available_size which might reduce the chance of OOM if user didn't assign maxCachedWorkflows.

### Environment/Versions

- OS and processor: [Linux], [container](https://github.com/nodejs/docker-node) based on `node:16.14.0`
- Temporal Version: 1.17.1 , SDK version : v1.0.0-rc.1
- Are you using Docker or Kubernetes? Yes

### Additional context
Our worker k8s manifest.
![image](https://user-images.githubusercontent.com/6711819/179886299-eb43b2c7-da47-4ca6-9d3b-272183531000.png)

![image](https://user-images.githubusercontent.com/6711819/179885599-62a9d068-87bc-4cda-8d57-2343f888ee69.png)

bergundy:
Thanks for reporting, we'll use `total_available_size` as the default instead.

PR #765, opened 2022-07-20
fix(worker) use total_available_size to calculate maxCachedWorkflows

Vincechen611:
## What was changed
Calculate default maxCachedWorkflows by `v8.getHeapStatistics()` instead of `os.totalmem()`

## Why?
When running in container, the os.totalmem() will get the host machine memory not container resource setting (which is in [cgroup](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/))

## issue
#762

### DIFF
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -173,7 +173,7 @@ export interface WorkerOptions {
    * You should be able to fit about 500 Workflows per GB of memory dependening on your Workflow bundle size.
    * For the SDK test Workflows, we managed to fit 750 Workflows per GB.
    *
-   * @default `max(v8.getHeapStatistics().total_available_size / 1GiB - 1, 1) * 200`
+   * @default `max(v8.getHeapStatistics().heap_size_limit / 1GiB - 1, 1) * 200`
    */
   maxCachedWorkflows?: number;
 
@@ -412,7 +412,7 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
     // 4294967295ms is the maximum allowed time
     isolateExecutionTimeout: debugMode ? '4294967295ms' : '5s',
     workflowThreadPoolSize: 8,
-    maxCachedWorkflows: maxCachedWorkflows ?? Math.max(v8.getHeapStatistics().total_available_size / GiB - 1, 1) * 200,
+    maxCachedWorkflows: maxCachedWorkflows ?? Math.max(v8.getHeapStatistics().heap_size_limit / GiB - 1, 1) * 200,
     enableSDKTracing: false,
     debugMode: debugMode ?? false,
     interceptors: appendDefaultInterceptors({}),
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -2,6 +2,7 @@ import { DataConverter, LoadedDataConverter } from '@temporalio/common';
 import { loadDataConverter } from '@temporalio/internal-non-workflow-common';
 import { msToNumber } from '@temporalio/internal-workflow-common';
 import os from 'os';
+import v8 from 'v8';
 import { ActivityInboundLogInterceptor } from './activity-log-interceptor';
 import { NativeConnection } from './connection';
 import { WorkerInterceptors } from './interceptors';
@@ -172,7 +173,7 @@ export interface WorkerOptions {
    * You should be able to fit about 500 Workflows per GB of memory dependening on your Workflow bundle size.
    * For the SDK test Workflows, we managed to fit 750 Workflows per GB.
    *
-   * @default `max(os.totalmem() / 1GiB - 1, 1) * 200`
+   * @default `max(v8.getHeapStatistics().total_available_size / 1GiB - 1, 1) * 200`
    */
   maxCachedWorkflows?: number;
 
@@ -411,7 +412,7 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
     // 4294967295ms is the maximum allowed time
     isolateExecutionTimeout: debugMode ? '4294967295ms' : '5s',
     workflowThreadPoolSize: 8,
-    maxCachedWorkflows: maxCachedWorkflows ?? Math.max(os.totalmem() / GiB - 1, 1) * 200,
+    maxCachedWorkflows: maxCachedWorkflows ?? Math.max(v8.getHeapStatistics().total_available_size / GiB - 1, 1) * 200,
     enableSDKTracing: false,
     debugMode: debugMode ?? false,
     interceptors: appendDefaultInterceptors({}),


Vincechen611:
Create a new PR because I was committed (https://github.com/temporalio/sdk-typescript/pull/763) with empty email which block license/cla check.

bergundy (commented):
Doesn't look like `total_available_size` is what we should be using here:

```
> v8.getHeapStatistics().total_available_size
4327370136
> Buffer.from('a'.repeat(10000000))
<Buffer 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 ... 9999950 more bytes>
> v8.getHeapStatistics().total_available_size
4328819472
> Buffer.from('a'.repeat(10000000))
<Buffer 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 61 ... 9999950 more bytes>
> v8.getHeapStatistics().total_available_size
4329471512
```

bergundy:
I looked into it and it seems like what you want is: `v8.getHeapStatistics().heap_size_limit`, the number seems to correlate to the `--max-old-space-size` option.
The docs don't provide any information about these values so I just tried it out.

lorensr:
Perhaps `heap_size_limit - used_heap_size`?

image

`used_heap_size` temporarily goes up after `Buffer.fom`

Vincechen611:
Yes, the [nodejs doc](https://nodejs.org/api/v8.html#v8getheapstatistics) didn't explain how `total_available_size ` be calculated. 

The default `maxCachedWorkflows` value is only calculated when user don't provide their own setting.
Using `total_available_size` can reduce chance of OOM if user don't notice this config, that's the reason why I choose it.

If we want the SDK wring out the resource, maybe we can still use `heap_size_limit` and add some warning in doc?

lorensr:
The issue is that `total_available_size` doesn't appear to be the number of bytes available for the process to use (otherwise it would get smaller, not larger when we use more memory with `Buffer.from`). 

`heap_size_limit` is the total amount the process is allowed to use, and I suggested calculating "true available" by subtracting `used_heap_size`. (If `used_heap_size` accurately represents the bytes already used by the process. Not sure why it decreases—maybe OS is swapping to disk. Let's see what @bergundy thinks)

Vincechen611:
Attach a sample test to check `total_available_size`, not sure if the increase case is due to gc.

```js
var v8 = require('v8');
for (var i = 0; i <= 10; i++) {
    var before = v8.getHeapStatistics().total_available_size
    Buffer.from('a'.repeat(10000));
    console.log(`- before: ${before}, after: `+ v8.getHeapStatistics().total_available_size);
}
```

```shell
root@worker-ts:/usr/src/app/tmp# node  --trace_gc t.js
- before: 953480328, after: 953463128
- before: 953131160, after: 953119904
- before: 953118320, after: 953107064
- before: 953105360, after: 953094104
- before: 953092552, after: 953081296
- before: 952940504, after: 952929056
- before: 952927488, after: 952916216
- before: 952914496, after: 952903224
- before: 952901656, after: 952890384
- before: 952888664, after: 952877392
- before: 952875696, after: 952864424
```

lorensr:
I'm also getting always decreasing now:

image

bergundy:
I'd use the heap size not the currently available size, the calculation already reserves memory for the rest of the process

Vincechen611:
NP, change `total_available_size` to `heap_size_limit`

bergundy:
@Vincechen611 sorry, I haven't gotten around to reviewing this.
I want to take some time understanding the different values.
@mjameswh could you PTAL?

mjameswh:
From what I see, none of the value in `v8.getHeapStatistics()` or `os.*mem()` reflects cgroup's limits on max memory.

See for example this:

**At 128 mb:**
```
jwatkins:~$ docker run --rm -it --memory=128m --memory-swap=128m node bash -c "cat /sys/fs/cgroup/memory/memory.limit_in_bytes ; node -e 'console.log(v8.getHeapStatistics()) ; console.log(os.totalmem()); console.log(os.freemem())'"
134217728
{
  total_heap_size: 5705728,
  total_heap_size_executable: 524288,
  total_physical_size: 5488640,
  total_available_size: 267528168,
  used_heap_size: 4948440,
  heap_size_limit: 271581184,
  malloced_memory: 254104,
  peak_malloced_memory: 99952,
  does_zap_garbage: 0,
  number_of_native_contexts: 1,
  number_of_detached_contexts: 0,
  total_global_handles_size: 8192,
  used_global_handles_size: 2592,
  external_memory: 334256
}
1027469312
193646592
```

**At 256Mb:**
```
jwatkins:~$ docker run --rm -it --memory=256m --memory-swap=256m node bash -c "cat /sys/fs/cgroup/memory/memory.limit_in_bytes ; node -e 'console.log(v8.getHeapStatistics()) ; console.log(os.totalmem()); console.log(os.freemem())'"
268435456
{
  total_heap_size: 5705728,
  total_heap_size_executable: 524288,
  total_physical_size: 5488640,
  total_available_size: 267528168,
  used_heap_size: 4948440,
  heap_size_limit: 271581184,
  malloced_memory: 254104,
  peak_malloced_memory: 99952,
  does_zap_garbage: 0,
  number_of_native_contexts: 1,
  number_of_detached_contexts: 0,
  total_global_handles_size: 8192,
  used_global_handles_size: 2592,
  external_memory: 334256
}
1027469312
185380864
```

**At 512 Mb:**
```
jwatkins:~$ docker run --rm -it --memory=512m --memory-swap=512m node bash -c "cat /sys/fs/cgroup/memory/memory.limit_in_bytes ; node -e 'console.log(v8.getHeapStatistics()) ; console.log(os.totalmem()); console.log(os.freemem())'"
536870912
{
  total_heap_size: 5705728,
  total_heap_size_executable: 524288,
  total_physical_size: 5488640,
  total_available_size: 267528168,
  used_heap_size: 4948440,
  heap_size_limit: 271581184,
  malloced_memory: 254104,
  peak_malloced_memory: 99952,
  does_zap_garbage: 0,
  number_of_native_contexts: 1,
  number_of_detached_contexts: 0,
  total_global_handles_size: 8192,
  used_global_handles_size: 2592,
  external_memory: 334256
}
1027469312
186355712
```

There has been a [PR to add a `os.constrainedmem()` function to nodejs](https://github.com/nodejs/node/pull/27805), but it has been discarded.

mjameswh:
Also, `v8.getHeapStatistics().total_available_size` and `v8.getHeapStatistics().heap_size_limit` are the only values that correlate with `--max-old-space-size` (though neither is exactly equal to it). See for example:

**With `--max-old-space-size=64`:**
```
root@433611c8eacb:/# node --max-old-space-size=64 -e "console.log(v8.getHeapStatistics()) ; console.log(os.totalmem()); console.log(os.freemem()); console.log(process.memoryUsage());"
{
  total_heap_size: 5705728,
  total_heap_size_executable: 524288,
  total_physical_size: 5488640,
  total_available_size: 66200904,
  used_heap_size: 4949080,
  heap_size_limit: 70254592,
  malloced_memory: 254104,
  peak_malloced_memory: 214168,
  does_zap_garbage: 0,
  number_of_native_contexts: 1,
  number_of_detached_contexts: 0,
  total_global_handles_size: 8192,
  used_global_handles_size: 2592,
  external_memory: 334256
}
1027469312
190455808
{
  rss: 39669760,
  heapTotal: 5967872,
  heapUsed: 5291192,
  external: 354937,
  arrayBuffers: 17374
}
```

**With `--max-old-space-size=256`:**
```
root@433611c8eacb:/# node --max-old-space-size=256 -e "console.log(v8.getHeapStatistics()) ; console.log(os.totalmem()); console.log(os.freemem()); console.log(process.memoryUsage());"
{
  total_heap_size: 5705728,
  total_heap_size_executable: 524288,
  total_physical_size: 5488640,
  total_available_size: 267527496,
  used_heap_size: 4949080,
  heap_size_limit: 271581184,
  malloced_memory: 254104,
  peak_malloced_memory: 214168,
  does_zap_garbage: 0,
  number_of_native_contexts: 1,
  number_of_detached_contexts: 0,
  total_global_handles_size: 8192,
  used_global_handles_size: 2592,
  external_memory: 334256
}
1027469312
191877120
{
  rss: 39936000,
  heapTotal: 5967872,
  heapUsed: 5291200,
  external: 354937,
  arrayBuffers: 17374
}
```

mjameswh:
There's definitely more to be done to correctly fix this issue.

As it is, the PR would only help if `--max-old-space-size` is _explicitly_ and _correctly_ defined by the user. We need properly handle the highly common case where the worker is running in a container environment, but where `--max-old-space-size` has not been set.

Also, given that by default, `heap_size_limit` is approximately equal to 25% of `os.totalmem()`, this PR would divide by 4 the value of `maxCachedWorkflows` in the default case. We'll need to adjust the constants, but I wonder if the factor might have to be different depending on either we are in a container vs classic execution environment.

I'm also concerned that the current heuristic doesn't account for the fact that they might be more than one worker running in the same process.

Considering all of this, I think we should discuss a more complete heuristic before we do anything else.

bergundy:
I had similar concerns @mjameswh, I don't think there's a single heuristic for max #workflows that works for every case.
It's been a while since I tracked context resource consumption and it correlates to bundle size (and obviously memory consumed by the running workflow).

I think the best experience would be to accept max allowed memory instead of using a fixed number of workflows to make this more adaptive to different workloads.
To properly address this we'd need to do more research though, we should track and prioritize.
In the short term, I'd like to improve the situation for container based deployments (the most common form).

mjameswh:
Actually, the workflow cache is handled in the rust core, so its memory is outside of node's GC control.

Hence, given that the memory limit constraint on the container remains the same, then available memory for the cache actually decrease when `max-old-space-size` increases. Setting `max-old-space-size` to a high value (in the original issue, it is set at 90% of docker's memory limit) is actually the wrong way to go and _increases_ the probability of an OOM.

mjameswh:
Here is a rough idea of the heuristic I have in mind:

```
// Check if the cgroups file system exists; if so, read the memory limit from it
const cgroupsMemLimit = getCgroupsMemoryLimit() ?? getCgroupsMemoryLimit();
const machineMemory = (cgroupsMemLimit && cgroupsMemLimit > 0) ? cgroupsMemLimit : os.totalmem();

const heapSizeLimit = v8.getHeapStatistics().heap_size_limit();
if (heapSizeLimit > machineMemory / 2) {
    // Report a warning suggesting that the user set '--max-old-space-size` to something between 25% and 50% of machineMemory
}

const memAvailableForWorkflowCache = machineMemory - heapSizeLimit;

// Adjusted the constant to account for the fact that memAvailableForWorkflowCache has been reduced by 25%
const defaultMaxCachedWorkflows = Math.max(memAvailableForWorkflowCache / GiB - 1, 1) * 250;

// Tell the user the value that has been automatically determined, so he can better understand how other adjustment affect this one
console.log(`Default value of maxCachedWorkflow is ${defaultMaxCachedWorkflows}`)
```

bergundy:
@mjameswh that seems like the right direction.
We should consider logging all of the worker options on startup, we do that in the load test worker and it's been super useful.

bergundy:
Replaced with #802

PR #767, opened 2022-07-21
feat(common): Improve ApplicationFailure arguments; add create

### LINKED ISSUE #752: [Feature Request] Add `ApplicationFailure.create`

#562 

Add `create` static method to [ApplicationFailure](https://typescript.temporal.io/api/classes/common.applicationfailure/):

```ts
throw ApplicationFailure.create({
  message: 'my message',
  type: 'old-cookie', 
  nonRetryable: true, 
  cause: e
})
```

lorensr:
## What was changed

- Fixes #752
```ts
ApplicationFailure.create({ cause })
```
- Makes fields optional and allows them to be `undefined | null`, so you can do:

```ts
new ApplicationFailure()
new ApplicationFailure(null, null, null, null, cause)
new ApplicationFailure(undefined, undefined, undefined, undefined, cause)
ApplicationFailure.retryable()
ApplicationFailure.retryable(null, null, ...details)
```

## Why?

DX

### DIFF
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -139,8 +139,8 @@ export class ApplicationFailure extends TemporalFailure {
    * @param details Optional details about the failure. Serialized by the Worker's {@link PayloadConverter}.
    */
   public static retryable(
-    message?: string | undefined | null,
-    type?: string | undefined | null,
+    message?: string | null,
+    type?: string | null,
     ...details: unknown[]
   ): ApplicationFailure {
     return new this(message, type ?? 'Error', false, details);
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -126,7 +126,7 @@ export class ApplicationFailure extends TemporalFailure {
    * By default, will be retryable (unless its `type` is included in {@link RetryPolicy.nonRetryableErrorTypes}).
    */
   public static create(options: ApplicationFailureOptions): ApplicationFailure {
-    const { message, type, nonRetryable = false, details, cause } = options ?? {};
+    const { message, type, nonRetryable = false, details, cause } = options;
     return new this(message, type, nonRetryable, details, cause);
   }
 
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -55,7 +55,7 @@ export class TemporalFailure extends Error {
    */
   public failure?: ProtoFailure;
 
-  constructor(message: string | undefined, public readonly cause?: Error) {
+  constructor(message?: string | undefined | null, public readonly cause?: Error) {
     super(message ?? undefined);
   }
 }
@@ -96,14 +96,22 @@ export class ServerFailure extends TemporalFailure {
 export class ApplicationFailure extends TemporalFailure {
   public readonly name: string = 'ApplicationFailure';
 
+  /**
+   * Alternatively, use {@link create}.
+   */
   constructor(
-    message: string | undefined,
-    public readonly type: string | undefined | null,
-    public readonly nonRetryable: boolean,
-    public readonly details?: unknown[],
+    message?: string | undefined | null,
+    public readonly type?: string | undefined | null,
+    public readonly nonRetryable?: boolean | undefined | null,
+    public readonly details?: unknown[] | undefined | null,
     cause?: Error
   ) {
     super(message, cause);
+  }
+
+  public static create(options?: ApplicationFailureOptions): ApplicationFailure {
+    const { message, type, nonRetryable = false, details, cause } = options ?? {};
+    return new this(message, type, nonRetryable, details, cause);
   }
 
   /**
@@ -114,7 +122,11 @@ export class ApplicationFailure extends TemporalFailure {
    * @param type Optional error type (used by {@link RetryPolicy.nonRetryableErrorTypes})
    * @param details Optional details about the failure. Serialized by the Worker's {@link PayloadConverter}.
    */
-  public static retryable(message: string | undefined, type?: string, ...details: unknown[]): ApplicationFailure {
+  public static retryable(
+    message?: string | undefined | null,
+    type?: string | undefined | null,
+    ...details: unknown[]
+  ): ApplicationFailure {
     return new this(message, type ?? 'Error', false, details);
   }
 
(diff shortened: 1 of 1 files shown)


packages/common/src/failure.ts
```diff
@@ -128,11 +145,44 @@ export class ApplicationFailure extends TemporalFailure {
...
+
+  /**
+   * Details about the failure. Serialized by the Worker's {@link PayloadConverter}.
+   */
+  details?: unknown[];
+
+  /**
+   * Cause of the failure
+   */
+  cause?: Error;
```
bergundy, line 183: This can only be a TemporalFailure for it to be properly serialized.

lorensr: I think a common use case is putting a 3rd party library's `Error` here. The message and stackTrace should get serialized fine: https://github.com/temporalio/sdk-typescript/blob/32c12fd2aaec1acf6866cfe0b47c3893c428cc15/packages/common/src/failure.ts#L337-L344

cretz: :+1: And `TemporalFailure` accepts `cause` as non-Temporal error. Python `from error` which we support, Go https://pkg.go.dev/go.temporal.io/sdk/temporal#NewApplicationErrorWithCause, and Java https://www.javadoc.io/static/io.temporal/temporal-sdk/1.14.0/io/temporal/failure/ApplicationFailure.html#newFailureWithCause(java.lang.String,java.lang.String,java.lang.Throwable,java.lang.Object...) all accept generic causes

bergundy: True, thanks

mjameswh:
Good to see this!

TBH, it is still unclear to me what is the correct way (a.k.a. how developers are expected to do it) to create an `ApplicationFailure` based on some `Error`. The doc gives [an interesting example](https://docs.temporal.io/typescript/handling-failure/#pattern-wrapping-errors-with-interceptors), but that example doesn't capture the error's stack trace...

We regularly face that use case in activities. We generally end up writing something similar to the following, but it feels very awkward...

```
    } catch (err) {
        if (isLimitExceededException(err) || isRequestLimitExceeded(err)) {
            throw ApplicationFailure.retryable(err.message, err.name, [err.stack]);
        }

        throw ApplicationFailure.nonRetryable(
             (err as Error).message ?? 'Unknown error',
             (err as Error).type,
             [(err as Error).stack],
        );
    }
```

lorensr:
I can update the docs example to use `.create` and include `cause` for getting stackTrace.

In your snippet, I'd do:

```
    } catch (err) {
        if (isLimitExceededException(err) || isRequestLimitExceeded(err)) {
            throw ApplicationFailure.create({ message: 'Limit exceeded', type: err.name, cause: err });
        }

        throw ApplicationFailure.create(
             message: 'Unknown error',
             nonRetryable: true,
             cause: err
        );
    }
```

bergundy:
Instead of setting a non-ApplicationFailure as the cause which results in an extra ApplicationFailure in the chain, it might be better to use ensureApplicationFailure(err) and flatten that out

lorensr:
Ah yeah, that works for retryable:

image

The naming feels off for use outside SDK though. Maybe we could add this?

```
throw ApplicationFailure.fromError(error)
throw ApplicationFailure.fromError(error, { nonRetryable: true, type: 'foo', details: [1] })
```

mjameswh:
+1 for `create` and `wrap`:

```
ApplicationFailure.create(options: ApplicationFailureOptions);
ApplicationFailure.wrap(error: Error | unknown, overrides?: ApplicationFailureOptions);
```

@lorensr One detail. Just noticed that in your original proposition, the `options` parameter on `create(...)` was optional. I think that parameter should be mandatory... First, just calling `ApplicationFailure.create()` without any detail feels like a code smell, so we should not encourage it. And if one really wants to do such a thing anyway, he can still just pass an empty object in argument, or call `ApplicationFailure.retryable()`, or call `new ApplicationFailure()`...

> Instead of setting a non-ApplicationFailure as the cause which results in an extra ApplicationFailure in the chain, it might be better to use `ensureApplicationFailure(err)` and flatten that out

Actually, I see legitimate use cases for both approach (flattened vs ApplicationFailure+cause), but I agree that the flattened approach (eg. `ApplicationFailure.wrap()`) is probably the one that should be highlighted in documentation.

cretz:
>> The naming feels off for use outside SDK though. Maybe we could add this?

> I like it, alternative names:
>  wrap, from

:+1: We should add this across the SDKs. None of the other SDKs have it today, but they are basically doing it themselves internally when auto-converting the error.

I do not think we should use `wrap` because it has meaning in at least Go and Java to set it on the chain. You want to convert the error, not wrap it. I like the static `fromError` personally.

One thing we need to make sure here (especially in the other SDK languages where cause/wrap is part of the language) - we need to not convert the chain recursively when in use. We should only convert to proto failure recursively. So cause should remain the actual user-given cause while in use. Yes, when deserialized back from the cause will be application failure, but that's ok.

packages/common/src/failure.ts
```diff
@@ -96,16 +96,40 @@ export class ServerFailure extends TemporalFailure {
...
+    return failure;
+  }
+
+  /**
+   * Create a new `ApplicationFailure`.
+   *
+   * By default, will be retryable (unless its `type` is included in {@link RetryPolicy.nonRetryableErrorTypes}).
+   */
+  public static create(options: ApplicationFailureOptions): ApplicationFailure {
+    const { message, type, nonRetryable = false, details, cause } = options ?? {};
```
bergundy, line 129:
```suggestion
    const { message, type, nonRetryable = false, details, cause } = options;
```

packages/common/src/failure.ts
```diff
@@ -114,7 +138,11 @@ export class ApplicationFailure extends TemporalFailure {
    * @param type Optional error type (used by {@link RetryPolicy.nonRetryableErrorTypes})
    * @param details Optional details about the failure. Serialized by the Worker's {@link PayloadConverter}.
    */
-  public static retryable(message: string | undefined, type?: string, ...details: unknown[]): ApplicationFailure {
+  public static retryable(
+    message?: string | undefined | null,
+    type?: string | undefined | null,
```
bergundy, line 143:
```suggestion
    message?: string | null,
    type?: string | null,
```

packages/common/src/failure.ts
```diff
@@ -128,11 +156,44 @@ export class ApplicationFailure extends TemporalFailure {
    * @param type Optional error type
    * @param details Optional details about the failure. Serialized by the Worker's {@link PayloadConverter}.
    */
-  public static nonRetryable(message: string | undefined, type?: string, ...details: unknown[]): ApplicationFailure {
+  public static nonRetryable(
+    message?: string | undefined | null,
+    type?: string | undefined | null,
```
bergundy, line 161:
```suggestion
    message?: string | null,
    type?: string | null,
```

PR #769, opened 2022-07-21
feat(bundler): Support typescript paths aliasing

mjameswh:
## What was changed?
* Added the TsconfigPathsPlugin in the workflow bundler
* TsconfigPathsPlugin inspects tsconfig.json files, looking up `compilerOptions.paths` entries; if such entries are found, they are automatically added as module aliases to the webpack configuration

## Why?
* TypeScript's `compilerOptions.paths` is a very common strategy to implement module aliases in TypeScript projects.
* Developpers would expect these aliases to be automatically without particular effort on their part.
* Reference: https://temporalio.slack.com/archives/C01DKSMU94L/p1658251242548939

## Checklist
- [ ] ~Add integration test~
- [x] Test locally using verdaccio build
- [ ] Add sample to typescript-example repo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11220,7 +11220,6 @@
       "version": "2.2.1",
       "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.1.tgz",
       "integrity": "sha512-1hqLFMSrGHRHxav9q9gNjJ5EXznIxGVO09xQRrwplcS8qs28pZ8s8hupZAmqDwZUmVZ2Qb2jnyPOWcDH8m8dlA==",
-      "dev": true,
       "bin": {
         "json5": "lib/cli.js"
       },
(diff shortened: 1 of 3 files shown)


bergundy (commented):
I'd test this to verify it's doing what we want

mjameswh:
We already use that same plugin in a custom fork of the workflow bundler, so I know that it does its job at least in some cases. But there's a few corner cases that I want to validate.

bergundy (approved):
Pre-approving this, waiting for tests though

mjameswh:
Tested locally using a verdaccio build on the following configurations:
- Yarn 1 with tsconfig paths aliases
- Yarn 3 PnP with tsconfig paths aliases (I had a concern that the plugin might have issue with PnP, but it succeed)
- npm 8 with pure JavaScript (I had a concern that there might be some error message if there was no tsconfig.json file, turned out ok)

@lorensr @bergundy Our integration tests run on pure JavaScript files (ie. TS files are precompiled to JS before test execution), so I can't easily add integration tests for this, or at least I do some disproportionate work specifically for this. I propose that I instead add a _tsconfig's paths alias_ example to the sample-typescript repo. Is that reasonable to both of you?

bergundy:
SGTM @mjameswh, please open an issue for it, now high priority IMO.

bergundy:
@mjameswh looks like sdk-features is failing with these errors:
https://github.com/temporalio/sdk-typescript/runs/7858682050?check_suite_focus=true

mjameswh:
The errors reported by sdk-features tests are due to the tsconfig.json file injected during test. I'm particularly dubious about [this specific line](https://github.com/temporalio/sdk-features/blob/3cf5ffacde846065de5b9c1e78cc426ec8f436f3/harness/ts/tsconfig.json.tmpl#L11).

mjameswh:
@bergundy This PR now magically works. Looks like recent modifications in `sdk-features` solved the failure we had previously. There's nothing on my side holding a merge. Are you still ok with this PR?

bergundy (approved):
I’m fine with this, only thing I’m concerned about is the added dependencies, maybe instead we can make it a plugin that people can use to amend the webpack config?

Another option is to make these optional dependencies..

Let’s discuss the options and find the best approach.

bergundy:
@mjameswh should we close this PR in favor of a sample?

mjameswh:
Yes, let's close it. Anyway, the complexity of having users add the required `webpackConfigHook` by themselves is essentially the same as what they would have to do if we make this feature optional.

mjameswh:
For those looking on how to add support for TypeScript paths to their workflow bundle, here is how to do this:

1. Add NPM package `tsconfig-paths-webpack-plugin` to your project (the one containing the `bundleWorkflowCode()` or the worker, depending on how your workflows are loaded), as a regular dependency.
2. Add the following code to your bundler’s or worker’s config:
```
import path from 'path';
import TsconfigPathsPlugin from 'tsconfig-paths-webpack-plugin';

// ...

webpackConfigHook: (config) => {
    config.resolve!.plugins = [
        ...(config.resolve!.plugins ?? []),
        new TsconfigPathsPlugin({ baseUrl: path.dirname(config.entry!) }) as unknown as typeof config['resolve']['plugins'][0],
    ];
    return config;
}
```

jdnichollsc:
@mjameswh thanks for sharing that solution, I was getting issues using Temporal with NestJS

nubunto:
@mrsufgi take a look at my [example repo](https://github.com/nubunto/nx-with-temporal/tree/main), maybe it can help.

PR #771, opened 2022-07-22
chore: Add eslint-import-resolver-typescript

lorensr:


.eslintrc
```diff
@@ -23,6 +23,18 @@
...
+    "import/no-unresolved": "error"
+  },
+  "settings": {
+    "import/parsers": {
+      "@typescript-eslint/parser": [".ts"]
+    },
+    "import/resolver": {
+      "typescript": {
+        "alwaysTryTypes": true, // always try to resolve types under `<root>@types` directory even it doesn't contain any source code, like `@types/unist`
+        "project": "packages/*/tsconfig.json"
```
JounQin, line 36:
```suggestion
        "project": ["tsconfig.prune.json", "packages/*/tsconfig.json"]
```

JounQin (changes requested):

JounQin:
You can also use https://github.com/JounQin/eslint-plugin-i for better performance.

```sh
npm install -D eslint-plugin-import@npm:eslint-plugin-i
```

.eslintrc
```diff
@@ -23,6 +23,18 @@
...
+    "import/no-unresolved": "error"
+  },
+  "settings": {
+    "import/parsers": {
+      "@typescript-eslint/parser": [".ts"]
+    },
+    "import/resolver": {
+      "typescript": {
+        "alwaysTryTypes": true, // always try to resolve types under `<root>@types` directory even it doesn't contain any source code, like `@types/unist`
+        "project": ["tsconfig.prune.json", "packages/*/tsconfig.json"]
```
bergundy, line 36: Why do we need tsconfig.prune here?
The entire codebase is covered by the per-project tsconfigs

JounQin: https://github.com/temporalio/sdk-typescript/blob/32c12fd2aaec1acf6866cfe0b47c3893c428cc15/tsconfig.prune.json#L8-L19

These files are failing without this.

bergundy: Can you elaborate? Failing how?

JounQin:
You'll need to run build before lint, because `"@temporalio/*": ["packages/*"]`, or you should use `"@temporalio/*": ["packages/*/src"]` without needing building before lint.

.eslintrc
JounQin: https://github.com/temporalio/sdk-typescript/runs/7462992242?check_suite_focus=true#step:6:12

bergundy: Isn’t that due to not building before running eslint?

JounQin: > Isn’t that due to not building before running eslint?

It could be, and it could be better to not requiring building before lint.

bergundy: We’d still need to build afaict because we reference some js files in lib

JounQin: OK then.

lorensr: image

This didn't change linting errors. I'll add a build step

.eslintrc
```diff
@@ -23,6 +23,18 @@
         "argsIgnorePattern": "^_",
         "varsIgnorePattern": "^_"
       }
-    ]
+    ],
+    "import/no-unresolved": "error"
```
bergundy, line 27: Any reason not to add: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md?

bergundy: I'd go over the the list of rules in this plugin and figure out which ones are important.
Let's also increase the severity to error for all of the rules that are set to warn right now.

.eslintrc
```diff
@@ -23,6 +23,18 @@
...
+  },
+  "settings": {
+    "import/parsers": {
+      "@typescript-eslint/parser": [".ts"]
+    },
+    "import/resolver": {
+      "typescript": {
+        "alwaysTryTypes": true, // always try to resolve types under `<root>@types` directory even it doesn't contain any source code, like `@types/unist`
```
bergundy, line 35: nit: can you please break this line?

packages/worker/src/workflow-log-interceptor.ts
```diff
@@ -74,4 +74,5 @@ export class WorkflowInboundLogInterceptor implements WorkflowInboundCallsInterc
   }
 }
 
+// ts-prune-ignore-next
 export const interceptors: WorkflowInterceptorsFactory = () => ({ inbound: [new WorkflowInboundLogInterceptor()] });
```
bergundy, line 78: Or just add this to the list of entrypoints in the prune tsconfig.. fine either way I guess..

lorensr:
Fixed! 😆 Well, mostly. Left a couple TODOs for later.

image

.eslintrc
```diff
@@ -23,6 +23,43 @@
...
+    "import/no-duplicates": "error",
+    // TODO next PR
+    // "import/order": "error",
+    "import/newline-after-import": "error",
+    "import/no-unassigned-import": "error",
+    "import/no-named-default": "error"
+  },
+  "settings": {
+    "import/parsers": {
+      "@typescript-eslint/parser": [".ts", ".d.ts"]
```
JounQin, line 54: `.ts` should already cover `.d.ts`

.eslintrc
JounQin: So it means `prune` can be removed here?

packages/test/src/test-time.ts
```diff
@@ -1,4 +1,5 @@
 import test from 'ava';
+// eslint-disable-next-line import/no-named-as-default
 import Long from 'long';
```
JounQin, line 2: This can be `import * Long from 'long'`, or you can enable `esModuleInterop` instead.

packages/test/src/workflows/definitions.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-duplicates */
 import { defineSignal } from '@temporalio/workflow';
 // @@@SNIPSTART typescript-logger-sink-interface
 import { Sinks } from '@temporalio/workflow';
```
JounQin, line 3: `import type { Sinks }` would help.

packages/workflow/src/worker-interface.ts
```diff
@@ -130,7 +130,7 @@ export async function initRuntime({ info, randomnessSeed, now, patches }: Workfl
   // webpack doesn't know what to bundle given a dynamic import expression, so we can't do:
   // state.payloadConverter = (await import(payloadConverterPath)).payloadConverter;
   // @ts-expect-error this is a webpack alias to payloadConverterPath
-  const customPayloadConverter = (await import('__temporal_custom_payload_converter')).payloadConverter;
+  const customPayloadConverter = (await import('__temporal_custom_payload_converter')).payloadConverter; // eslint-disable-line import/no-unresolved
```
JounQin, line 133: Or you can use https://github.com/import-js/eslint-plugin-import#importignore instead.

JounQin (changes requested):

packages/test/src/test-time.ts
lorensr: `import * as Long` results in:

image

`esModuleInterop` results in:

image

.eslintrc
lorensr: Removed, thanks

packages/test/src/test-time.ts
JounQin: That seems a bug from plugin-import that it doesn't handle class static properties correctly.

JounQin: But on the other hand, `long` does not use `exports` field with `types` correctly.

https://unpkg.com/browse/long@5.2.0/package.json

lorensr:
Thanks a lot for the suggestions, @JounQin !

packages/create-project/package.json
```diff
@@ -19,7 +19,9 @@
   "author": "Temporal Technologies Inc. <sdk@temporal.io>",
   "license": "MIT",
   "dependencies": {
+    "@types/uuid": "^8.3.4",
```
JounQin, line 22: `@types/` should already be `dev` dependency instead.

JounQin (changes requested):

JounQin (approved):

packages/common/src/type-helpers.ts
```diff
@@ -12,12 +12,11 @@ export function checkExtends<_Orig, _Copy extends _Orig>(): void {
 
 export type Replace<Base, New> = Omit<Base, keyof New> & New;
 
-export type MakeOptional<Base, Keys extends keyof Base> = Omit<Base, Keys> & Partial<Pick<Base, Keys>>;
-
 export function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === 'object' && value !== null;
 }
 
+// ts-prune-ignore-next
```
lorensr, line 19: not sure why `ts-prune` isn't able to find these relative imports like:

image

packages/common/src/type-helpers.ts
```diff
@@ -12,12 +12,11 @@ export function checkExtends<_Orig, _Copy extends _Orig>(): void {
 
 export type Replace<Base, New> = Omit<Base, keyof New> & New;
 
-export type MakeOptional<Base, Keys extends keyof Base> = Omit<Base, Keys> & Partial<Pick<Base, Keys>>;
```
lorensr, line 15: removed, as it was unused. alternatively could add ignore comment if we want to have for future use

package.json
```diff
@@ -26,11 +26,12 @@
     "build.watch": "lerna run --stream build.watch",
     "test": "lerna run --stream test",
     "test.watch": "lerna run --stream test.watch",
+    "ci-build-no-rust": "lerna run --ignore @temporalio/core-bridge --stream build",
```
bergundy, line 29: Don't need this IMHO

packages/client/src/workflow-client.ts
```diff
@@ -872,9 +866,9 @@ export class WorkflowClient {
           },
           // Safe to convert to number, max history length is 50k, which is much less than Number.MAX_SAFE_INTEGER
           historyLength: raw.workflowExecutionInfo!.historyLength!.toNumber(),
-          startTime: tsToDate(raw.workflowExecutionInfo!.startTime!),
-          executionTime: optionalTsToDate(raw.workflowExecutionInfo!.executionTime),
-          closeTime: optionalTsToDate(raw.workflowExecutionInfo!.closeTime),
+          startTime: tsToDate(raw.workflowExecutionInfo!.startTime!), // eslint-disable-line deprecation/deprecation
+          executionTime: optionalTsToDate(raw.workflowExecutionInfo!.executionTime), // eslint-disable-line deprecation/deprecation
+          closeTime: optionalTsToDate(raw.workflowExecutionInfo!.closeTime), // eslint-disable-line deprecation/deprecation
```
bergundy, line 871: What's deprecated here?

packages/client/src/errors.ts
```diff
@@ -1,6 +1,5 @@
 import { ServerErrorResponse } from '@grpc/grpc-js';
 import { RetryState, TemporalFailure } from '@temporalio/common';
-export { WorkflowExecutionAlreadyStartedError } from '@temporalio/common';
```
bergundy, line 3: We should still export this from client

bergundy: Is it exported elsewhere?

packages/common/src/interceptors.ts
```diff
@@ -18,6 +18,7 @@ export type Headers = Record<string, Payload>;
  * @param method the name of the interceptor method to compose
  * @param next the original function to be executed at the end of the interception chain
  */
+// ts-prune-ignore-next (imported via lib/interceptors)
```
bergundy, line 21: Can you explain why this needs to be ignored?

packages/common/src/type-helpers.ts
bergundy: 👍

packages/create-project/package.json
```diff
@@ -20,6 +20,7 @@
   "license": "MIT",
   "dependencies": {
     "async-retry": "^1.3.3",
+    "ava": "^4.3.1",
```
bergundy, line 23: This should be a dev dependency

packages/create-project/package.json
```diff
@@ -39,6 +40,7 @@
     "@types/prompts": "2.0.14",
     "@types/rimraf": "^3.0.2",
     "@types/tar": "^6.1.1",
+    "@types/uuid": "^8.3.4",
```
bergundy, line 43: Does uuid still require `types`?

packages/create-project/src/helpers/samples.ts
```diff
@@ -1,7 +1,7 @@
 // Modified from: https://github.com/vercel/next.js/blob/2425f4703c4c6164cecfdb6aa8f80046213f0cc6/packages/create-next-app/helpers/examples.ts
 import chalk from 'chalk';
 import got from 'got';
-import tar from 'tar';
+import { extract } from 'tar';
```
bergundy, line 4: nit: I'd import `* as tar`

packages/test/package.json
```diff
@@ -25,18 +25,36 @@
   "author": "Temporal Technologies Inc. <sdk@temporal.io>",
   "license": "MIT",
   "dependencies": {
+    "@grpc/grpc-js": "^1.6.8",
```
bergundy, line 28: Did we have to move all of the shared dev dependencies from the root package?

packages/testing/package.json
```diff
@@ -16,7 +16,10 @@
     "@temporalio/activity": "file:../activity",
     "@temporalio/client": "file:../client",
     "@temporalio/common": "file:../common",
+    "@temporalio/core-bridge": "^1.3.0",
+    "@temporalio/proto": "^1.3.0",
```
bergundy, line 20: Should be a path

bergundy (approved):
Overall LGTM.
I'd want to go over this again to make sure we didn't miss anything.
@mjameswh please review too.

Are we adding a sorted imports linter so the imports don't change every time you touch the code? Is that `import/order`?

packages/client/src/errors.ts
lorensr: Yes, I removed things caught by the duplicate export lint rule. This one is covered by `client/index.js`:

```ts
export * from '@temporalio/common/lib/errors';
```

packages/client/src/workflow-client.ts
lorensr: ah, thanks. the time functions. i'll change it to import from `common/lib/time`

.eslintrc
```diff
@@ -23,6 +23,43 @@
...
+    },
+    "import/resolver": {
+      "typescript": {
+        // always try to resolve types under `<root>@types` directory even it doesn't contain any source code, like
+        // `@types/unist`
+        "alwaysTryTypes": true,
+        "project": ["tsconfig.prune.json", "packages/*/tsconfig.json"]
```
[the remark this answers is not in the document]
lorensr: Removed, now just ` "project": ["packages/*/tsconfig.json"]`

packages/common/src/interceptors.ts
lorensr: `ts-prune` doesn't see that this function is used. the compiled version is used (imported via `common/lib/interceptors`). I read the ts-prune readme, but didn't see how to tell it to include `lib/` imports in its search, or this relative import:

image

packages/create-project/package.json
lorensr: tried removing 

image

packages/test/package.json
lorensr: Yes, when I remove, I get: 
image

packages/create-project/package.json
bergundy: Just use crypto randomUUID in tests

packages/test/package.json
```diff
@@ -25,18 +25,36 @@
...
     "@temporalio/client": "file:../client",
     "@temporalio/common": "file:../common",
+    "@temporalio/core-bridge": "^1.0.0-rc.1",
     "@temporalio/interceptors-opentelemetry": "file:../interceptors-opentelemetry",
     "@temporalio/proto": "file:../proto",
     "@temporalio/testing": "file:../testing",
     "@temporalio/worker": "file:../worker",
     "@temporalio/workflow": "file:../workflow",
     "@types/async-retry": "^1.4.4",
+    "@types/dedent": "^0.7.0",
```
mjameswh, line 45: `@types/*` should be `devDependencies`. Note that `ava` has to be kept in `dependencies` for this project.

packages/test/package.json
```diff
@@ -25,18 +25,36 @@
...
+    "@grpc/proto-loader": "^0.7.0",
+    "@opentelemetry/api": "^1.1.0",
+    "@opentelemetry/core": "^1.4.0",
     "@opentelemetry/exporter-trace-otlp-grpc": "^0.29.2",
+    "@opentelemetry/sdk-node": "^0.30.0",
+    "@opentelemetry/semantic-conventions": "^1.4.0",
     "@temporalio/activity": "file:../activity",
     "@temporalio/client": "file:../client",
     "@temporalio/common": "file:../common",
+    "@temporalio/core-bridge": "^1.0.0-rc.1",
```
mjameswh, line 38: Use `file:` instead

mjameswh:
I think enabling the `no-unused-modules` rule would also help. It finds things that are exported from a module but not used elsewhere. A quick try demonstrate that it catches a lot of things that are missed by `ts-prune`.

Try with:
```
    "import/no-unused-modules": [
      "error",
      {
        "unusedExports": true,
        "ignoreExports": [
          "packages/activity/src/index.ts",
          "packages/client/src/index.ts",
          "packages/common/src/index.ts",
          "packages/create-project/src/index.ts",
          "packages/meta/src/index.ts",
          "packages/worker/src/index.ts",
          "packages/worker/src/workflow/workflow-worker-thread.ts",
          "packages/workflow/src/index.ts",
          "packages/workflow/src/worker-interface.ts"
        ]
      }
    ]
```

packages/common/src/type-helpers.ts
mjameswh: I was able to remove the comment on `errorMessage` and `errorCode`.

I still get the error on `hasOwnProperty`. I get the feeling that this false-positive is related to the fact that `hasOwnProperty` is only invoked from type guards functions. But not much we can do about that.

mjameswh:
And btw, just noticed that tsconfig.prune.json might also need a little refresh.

It still refers to `"./packages/workflow-common/src/index.ts"`, which doesn't exists anymore, and it is missing a few packages that have been added recently.

packages/test/package.json
mjameswh: That means several dependencies can now be pruned from the top level package.json... `depcheck` might help with that.

mjameswh: This can be done later, but we will definitely have to find an efficient way of ensuring version coherence between dependencies of the various subprojects. There are tools for this, but haven't used any yet.

packages/common/src/type-helpers.ts
lorensr: removed, thanks!

lorensr:
I tried to figure out why local lint was different from CI, gave up and added ignore comment to `test-failure-converter.ts`.

`npm run lint` and `act` both passed without the comment.

```
act pull_request -W .github/workflows/test.yml --container-architecture linux/amd64
```

```
name: Continuous Integration

on: # rebuild any PRs and main branch changes
  pull_request:

jobs:
  # Run TS linting and ts-prune to find unused code
  testy:
    runs-on: ubuntu-latest
    steps:
      - run: echo 'start'
      - uses: actions/checkout@v2
        with:
          submodules: recursive
      - uses: actions/setup-node@v1
        with:
          node-version: 16
      - run: npm ci --ignore-scripts
      # eslint-import-resolver-typescript requires packages to be built
      - name: Compile all non-rust code
        run: npm run build -- --ignore @temporalio/core-bridge
      - run: npm run lint
      - run: npm run lint.prune
      - run: echo 'done'
```

PR #772, opened 2022-07-23
test: Test ABANDONED child timer can be fast-forwarded

bergundy:
Tried to reproduce an issue reported in community slack and failed, might as well leave this test for reference.

### DIFF
--- a/packages/test/src/test-testenvironment.ts
+++ b/packages/test/src/test-testenvironment.ts
@@ -3,8 +3,10 @@ import { v4 as uuid4 } from 'uuid';
 import { WorkflowFailedError } from '@temporalio/client';
 import { TestWorkflowEnvironment, workflowInterceptorModules } from '@temporalio/testing';
 import { Connection } from '@temporalio/testing/lib/connection';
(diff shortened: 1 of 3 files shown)


packages/test/src/test-testenvironment.ts
```diff
@@ -175,3 +180,32 @@ test.serial('Workflow code can run assertions', async (t) => {
     t.is(err.cause?.message, 'Expected values to be strictly equal:\n\n6 !== 7\n');
   });
 });
+
+test.serial('ABNADONED child timer can be fast-forwarded', async (t) => {
```
lorensr, line 184:
```suggestion
test.serial('ABANDONED child timer can be fast-forwarded', async (t) => {
```

lorensr (approved):

lorensr:
This error is consistent:

image

[build-and-test (14, windows-latest)](https://github.com/temporalio/sdk-typescript/runs/7568517994?check_suite_focus=true#logs)

bergundy:
@lorensr I saw the failure, it was just not much of a priority for me to fix this

PR #784, opened 2022-07-27
fix: Allow `RetryPolicy.maximumAttempts: Number.POSITIVE_INFINITY`

lorensr:
Edit: Allow user to provide `maximumAttempts: Number.POSITIVE_INFINITY` and drop it, since it's the default.

---

In the changelog we suggest using 0 for Infinity, but are currently throwing when 0 is passed. 

image


bergundy (approved):
🙌

lorensr:
Given that this is a ~bug discovered at runtime (so far by 2 users), I'm inclined to publish a patch tomorrow.

PR #786, opened 2022-07-28
fix: Prevent ending a worker span twice

### LINKED ISSUE #781: [Bug] Span ended twice in TypeScript SDK
We're seeing a lot of these messages in our logs:

```
Can not execute the operation on ended Span {traceId: …, spanId: …}
You can only call end() on a span once.
```

Ignoring the double error message for a single error source (that's an issue in the otel SDK), this is the source of the error message:

https://github.com/temporalio/sdk-typescript/blob/ea100209b1f62561d69ab230258dbc4f07565c34/packages/worker/src/tracing.ts#L54

It's clear it can be double ended in the worker:

https://github.com/temporalio/sdk-typescript/blob/ea100209b1f62561d69ab230258dbc4f07565c34/packages/worker/src/worker.ts#L1129

Easy fix, should just require dropping the `.end()` on ☝️ since `instrument` takes care of it 🙌 

### Environment/Versions

- OS and processor: macOS (intel)
- Temporal Version: TypeScript SDK 1.0.0 (and 0.22.0).

thegedge:
## What was changed
Remove a call to `end()` the `workflow.process` span in a worker.

## Why?
Since this span is coming from an `instrument` call, it will be guaranteed to end once the function given to instrument ends, so we should not call `end`.

Ending a span twice results in some logging that is undesirable in production scenarios ([see otel](https://github.com/open-telemetry/opentelemetry-js/blob/cfda625e83a164c9e19745392879c32aedcfe76f/packages/opentelemetry-sdk-trace-base/src/Span.ts#L239)).

## Checklist
Closes #781

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -1126,7 +1126,7 @@ export class Worker {
                     const completion = await this.workflowCodecRunner.encodeCompletion(unencodedCompletion);
                     this.log.trace('Completed activation', workflowLogAttributes(state.info));
 
-                    span.setAttribute('close', close).end();
+                    span.setAttribute('close', close);
                     return { state, output: { close, completion, parentSpan } };
                   } catch (err) {
                     if (err instanceof errors.UnexpectedError) {


bergundy (approved):
Thanks!

PR #787, opened 2022-07-28
Update CHANGELOG for v1.0.1

lorensr:


bergundy:
Let's update core before releasing this version

bergundy (approved):

PR #791, opened 2022-07-29
feat(workflow): List registered queries in error response when a quer…

bergundy:
…y is not found

This is required by the UI to display available queries.
Eventually we should implement a `__get_handlers` internal query but this will do for now.

### DIFF
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -3,6 +3,7 @@ import {
   ActivityFailure,
   ApplicationFailure,
   Connection,
+  QueryNotRegisteredError,
   WorkflowClient,
   WorkflowContinuedAsNewError,
   WorkflowFailedError,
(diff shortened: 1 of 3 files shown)


lorensr (approved):
test?

bergundy:
I was lazy at this late hour, I tested it manually.
Yes, this deserves a test.

packages/test/src/integration-tests.ts
```diff
@@ -391,6 +392,21 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+    const workflow = await client.start(workflows.unblockOrCancel, {
+      taskQueue: 'test',
+      workflowId: uuid4(),
+    });
+    await workflow.signal(workflows.unblockSignal);
+    await workflow.result();
+    await t.throwsAsync(workflow.query('not found'), {
+      instanceOf: QueryNotRegisteredError,
+      message:
+        '3 INVALID_ARGUMENT: Workflow did not register a handler for not found. Registered queries: [__stack_trace isBlocked]',
```
lorensr, line 406: Is this part helpful? `3 INVALID_ARGUMENT: ` Wonder if we should remove

bergundy: I didn't want to mess with the returned message, I guess it's okay to trim.

PR #793, opened 2022-07-29
docs: Add API doc for `bundleWorkflowCode`, fixes #792

### LINKED ISSUE #792: Better Documentation for bundleWorkflowCode in Typescript SDK
Better documentation around `bundleWorkflowCode` here: https://typescript.temporal.io/api/namespaces/worker/#bundleworkflowcode or somewhere else in the Temporal to indicate that interceptors should be passed in through bundleWorkflowCode and is necessary if used with replaying, or non-determinism errors will be thrown.

lorensr:


bergundy (approved):
Let's also add a comment in the WorkerOptions.interceptors and WorkerOptions.workflowBundle docstring saying that the workflow interceptors are not taken from there when using a bundle

PR #798, opened 2022-08-01
feat: Add nyc-test-coverage package

vkarpov15:
## What was changed

Porting the code from https://github.com/temporalio/temporal-pendulum/pull/19/files into a package in sdk-typescript. I'm not 100% convinced this belongs in the sdk-typescript repo, I'm open to suggestions either way. But I figured this was the easiest way to get the ball rolling.

## Why?

Code coverage is very valuable. But it is tricky to get code coverage with Workflows - it took me many hours to figure out how to wrangle istanbul/nyc to pick up Workflow code. With this package, you'll be able to get code coverage on Workflows with 4 lines of code.

## Checklist

1. Closes 

2. How was this tested:

Using `npm pack`, I bundled this package and did `npm install` in the branch for https://github.com/temporalio/temporal-pendulum/pull/19/files . With the below code in `workflows.test.ts`, I'm able to get test coverage for Workflow code:

```ts
import { mergeWorkflowCoverage, sinks } from 'temporalio-nyc-test-coverage';

beforeEach(async () => {
  const { nativeConnection } = testEnv;
  worker = await Worker.create({
    connection: nativeConnection,
    taskQueue,
    workflowsPath: require.resolve("./workflows"),
    activities: undefined,
    // Add sinks and interceptors to pull coverage data from each Workflow execution
    interceptors: {
      workflowModules: [require.resolve("temporalio-nyc-test-coverage/lib/interceptors")]
    },
    sinks,
  });

  runPromise = worker.run();

  handle = await testEnv.workflowClient.start(pendulum, {
    args: [initInfo],
    workflowId: uuid4(),
    taskQueue,
  });
});

after(() => {
  mergeWorkflowCoverage(); // Take the Workflow coverage map and merge it with `global.__coverage__`
});
```

3. Any docs updates needed?

We'll need to add some docs on how to use this. But I'd rather get some feedback on the code setup and API first.


packages/nyc-test-coverage/src/index.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  InjectedSinks,
+} from '@temporalio/worker';
+import { CoverageSinks } from './sinks';
+import libCoverage from 'istanbul-lib-coverage';
+
+export const coverageMap = libCoverage.createCoverageMap();
```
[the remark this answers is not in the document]
bergundy: After seeing that istanbul uses a global coverage object, I guess this isn’t too bad.
Why not merge into `global__coverage__` directly then?

packages/nyc-test-coverage/src/interceptors.ts
```diff
@@ -0,0 +1,59 @@
...
+const { coverage } = proxySinks<CoverageSinks>();
+
+class CoverageInterceptor implements WorkflowInboundCallsInterceptor {
+  public async handleQuery(input: QueryInput, next: any): Promise<unknown> {
+    const ret = await next(input);
+
+    // @ts-ignore
+    const globalCoverage = global.__coverage__;
+    coverage.merge(JSON.parse(JSON.stringify(globalCoverage)));
+    clearCoverage(globalCoverage);
```
bergundy, line 20: Why clear?

packages/nyc-test-coverage/src/interceptors.ts
```diff
@@ -0,0 +1,59 @@
...
+  WorkflowExecuteInput,
+  WorkflowInterceptorsFactory,
+  WorkflowInboundCallsInterceptor,
+  QueryInput
+} from '@temporalio/workflow';
+import { CoverageSinks } from './sinks';
+
+const { coverage } = proxySinks<CoverageSinks>();
+
+class CoverageInterceptor implements WorkflowInboundCallsInterceptor {
```
bergundy, line 13: You would probably be better off using internals interceptors, I’m not a 100% sure but I think you can simplify this.

Also note that if an exception is thrown in these methods you won’t get coverage

bergundy: Thinking about this s bit more, I think you could intercept internals concludeActivation instead of these 3 methods

packages/nyc-test-coverage/src/interceptors.ts
```diff
@@ -0,0 +1,59 @@
...
+    return ret;
+  }
+}
+
+// Export the interceptors
+export const interceptors: WorkflowInterceptorsFactory = () => ({
+  inbound: [new CoverageInterceptor()],
+});
+
+function clearCoverage(coverage: any): void {
```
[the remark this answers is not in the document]
vkarpov15: Since we merge coverage on every query, signal, etc. we need to make sure we're not double counting executions. Think of it this way:

1. A signal comes in. nyc increments the signal handler's execution count to 1. We merge the workflow's current coverage state. 
2. Another instance of the same signal comes in. nyc increments the signal handler's execution count to 2. We merge the workflow's current coverage state, which means the coverage map that stores all the Workflows' coverage data thinks the signal handler executed 3 times (1 + 2), instead of 2 times.

packages/nyc-test-coverage/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@temporalio/nyc-test-coverage",
```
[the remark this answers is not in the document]
vkarpov15: [istanbul](https://npmjs.com/package/istanbul) is technically deprecated, and nyc is the replacement. However, under the hood, nyc uses the same core code coverage logic as istanbul (like istanbul-lib-coverage).

packages/nyc-test-coverage/src/index.ts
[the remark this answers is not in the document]
vkarpov15: Because, weirdly enough, `global.__coverage__` is not a CoverageMap instance. I'm not familiar enough with nyc internals to explain exactly why. But in order to merge coverage in memory, we need to keep a separate CoverageMap instance to allow us to use `merge()`, and then replace `global.__coverage__` with a transformed version of the CoverageMap.

packages/nyc-test-coverage/src/interceptors.ts
vkarpov15: Same as https://github.com/temporalio/sdk-typescript/pull/798#discussion_r934087391 : if we don't clear coverage after merging, we start double-counting line and function executions.

packages/nyc-test-coverage/src/interceptors.ts
vkarpov15: I'll do some research on internals interceptors and see what I can do.

packages/nyc-test-coverage/src/interceptors.ts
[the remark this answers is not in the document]
bergundy: I figured that out already but didn't have a chance to reply.
Thanks for the clarification.

lorensr (approved):
This is great, thanks! Note lint errors

packages/nyc-test-coverage/package.json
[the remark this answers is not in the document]
bergundy: So why are we using istanbul if it's deprecated?
I'm not following.. either call this `istanbul-test-coverage` and use istanbul or call it `nyc-test-coverage` and use `nyc`.
Am I missing something?

packages/nyc-test-coverage/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "1.0.1",
+  "description": "Temporal.io SDK code coverage integration",
+  "main": "lib/index.js",
+  "types": "./lib/index.d.ts",
+  "keywords": [
+    "temporal",
+    "workflow",
+    "testing"
```
bergundy, line 10: coverage?

packages/nyc-test-coverage/src/index.ts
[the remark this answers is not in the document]
bergundy: In any case, I'd avoid yet another global

packages/nyc-test-coverage/src/interceptors.ts
```diff
@@ -0,0 +1,59 @@
...
+  for (const path of Object.keys(coverage)) {
+    for (const index of Object.keys(coverage[path].s)) {
```
bergundy, line 55: nit: could use `Object.entries(coverage)` to avoid accessing `coverage[path]`

packages/nyc-test-coverage/package.json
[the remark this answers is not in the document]
lorensr: Istanbul isn't deprecated. The istanbul CLI is.

image

image

packages/nyc-test-coverage/src/interceptors.ts
vkarpov15: We could, but then TypeScript starts complaining about the value in `Object.entries(coverage)` being of type `unknown` :(

packages/nyc-test-coverage/package.json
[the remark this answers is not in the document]
vkarpov15: @lorensr is right.

"I'm not following.. either call this istanbul-test-coverage and use istanbul or call it nyc-test-coverage and use nyc." <-- we're calling this package nyc-test-coverage because it's meant to be used with nyc.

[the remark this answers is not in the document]
bergundy: Got it, would be nice to mention this in the package README (which is missing BTW).

packages/nyc-test-coverage/package.json
```diff
@@ -0,0 +1,34 @@
...
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT",
+  "dependencies": {
+    "@temporalio/workflow": "file:../workflow",
+    "@temporalio/worker": "file:../worker",
+    "istanbul-lib-coverage": "3.2.0"
+  },
+  "devDependencies": {
+    "@types/istanbul-lib-coverage": "2.0.4"
```
bergundy, line 21: Just noticed this, why use exact versions instead of caret?

packages/nyc-test-coverage/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+            coverageMap.merge(testCoverage);
+          },
+          callDuringReplay: false,
+        },
+      },
+    },
+  };
+}
+
+export function mergeWorkflowCoverage(coverageMap: libCoverage.CoverageMap): void {
```
bergundy, line 23: At this point I'd just create a class (or instance with defined interface if you prefer) with a `sinks` getter and a `mergeWorkflowCoverage` method.

These methods require docs too.

I'd go with:

```ts
class WorkflowCoverage {
  get sinks(): InjectedSinks<CoverageSinks> { ... }
  mergeIntoGlobalCoverage(): void {}
}
```

packages/nyc-test-coverage/src/interceptors.ts
```diff
@@ -0,0 +1,31 @@
...
+        const globalCoverage = global.__coverage__;
+        coverage.merge(JSON.parse(JSON.stringify(globalCoverage)));
+        clearCoverage(globalCoverage);
+
+        return next(input);
+      },
+    },
+  ],
+  inbound: [],
+  outbound: [],
```
bergundy, line 22: nit: You can just omit these

packages/nyc-test-coverage/src/interceptors.ts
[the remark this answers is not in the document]
bergundy: Aren't there type definitions for global.__coverage__ that we can use?

bergundy (approved):
@vkarpov15 nice work putting this all together, it's just a matter of wrapping it now.
Thanks!

packages/nyc-test-coverage/package.json
vkarpov15: Best practice. I'll switch to `^` for consistency with other packages.

vkarpov15:
@bergundy @lorensr take a look at the latest changes, I added a README and did some more cleanup: classes and better typescript types

packages/nyc-test-coverage/src/index.ts
```diff
@@ -0,0 +1,40 @@
+import { InjectedSinks } from '@temporalio/worker';
+import { ICoverageSinks } from './sinks';
+import libCoverage from 'istanbul-lib-coverage';
+
+export class CoverageSinks {
```
bergundy, line 5: I’d call this WorkflowCoverage so it’s clear what its purpose is

bergundy (approved):
Very nice, I had a couple of super minor comment. Approved already

packages/nyc-test-coverage/src/index.ts
```diff
@@ -0,0 +1,40 @@
+import { InjectedSinks } from '@temporalio/worker';
+import { CoverageSinks } from './sinks';
+import libCoverage from 'istanbul-lib-coverage';
+
+export class WorkflowCoverage {
+  coverageMap = libCoverage.createCoverageMap();
+  sinksInternal: InjectedSinks<CoverageSinks> = {
+    coverage: {
+      merge: {
+        fn: (_workflowInfo: any, testCoverage: libCoverage.CoverageMap) => {
```
bergundy, line 10: Typescript can infer this sugnature

packages/nyc-test-coverage/src/index.ts
```diff
@@ -0,0 +1,40 @@
...
+          this.coverageMap.merge(testCoverage);
+        },
+        callDuringReplay: false,
+      },
+    },
+  };
+
+  get sinks(): InjectedSinks<CoverageSinks> {
```
bergundy, line 18: Why even have sinksInternal, you can replace with readonly public member

packages/nyc-test-coverage/src/index.ts
```diff
@@ -0,0 +1,40 @@
+import { InjectedSinks } from '@temporalio/worker';
+import { CoverageSinks } from './sinks';
+import libCoverage from 'istanbul-lib-coverage';
+
+export class WorkflowCoverage {
```
bergundy, line 5: You could add the interceptor modules to this class and maybe even a method to apply to worker options

packages/nyc-test-coverage/src/index.ts
```diff
@@ -0,0 +1,40 @@
...
+        callDuringReplay: false,
+      },
+    },
+  };
+
+  get sinks(): InjectedSinks<CoverageSinks> {
+    return this.sinksInternal;
+  }
+
+  mergeIntoGlobalCoverage(): void {
```
bergundy, line 22: Missing docstring

bergundy (approved):
This is already approved, the rest of my comments, as well as testing this package can be done in a follow up pr or just tracked in an issue for now

packages/nyc-test-coverage/src/index.ts
vkarpov15: I'm not sure how to do this without having Webpack bundle the entire WorkflowCoverage class. It looks like the only way to pass interceptors is as a workflow module: https://github.com/temporalio/sdk-typescript/blob/0310ad6e44ea49220e3ac14fa90c3e8c78ec3647/packages/worker/src/interceptors.ts#L42-L55 , so I would need a file path for the interceptor modules. Am I missing something?

PR #799, opened 2022-08-01
Enhanced stacktrace added

BashkirovN:
## What was changed
Enhanced stacktrace query was added

## Why?
Implementing https://github.com/temporalio/proposals/blob/ehnanced-stack-trace-query/enhanced-stack-trace-queries.md

## How was this tested:
Generic Integration tests were added

### DIFF
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1194,7 +1194,6 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     });
   }
 
-  //RUN_INTEGRATION_TESTS=1 npx ava packages/test/lib/test-integration.js -m 'Enhanced Stack Trace Test'
   test('Enhanced Stack Trace Test', async (t) => {
     const { client } = t.context;
     const workflowId = uuid4();
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1203,17 +1203,11 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       workflowId,
     });
 
-    const key: string = enhStacks[1].stacks[0].locations[0].filepath;
     t.is(enhStacks[0].sdk.name, 'typescript', 'sdk name is not "typescript"');
     t.is(enhStacks[0].stacks.length, 3, 'Number of stacks is different from expected');
     t.is(enhStacks[0].stacks[0].locations.length, 1, 'Number of locations is different from expected');
     t.is(enhStacks[1].stacks.length, 3, 'Number of stacks is different from expected');
     t.is(enhStacks[1].stacks[0].locations.length, 2, 'Number of locations is different from expected');
-    t.is(
-      enhStacks[1].sources[key][0].content.split('\n').length,
-      50,
-      'Number of lines in stack trace is different from expected'
-    );
   });
 
   test('issue-731', async (t) => {
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1204,12 +1204,16 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     });
 
     const key: string = enhStacks[1].stacks[0].locations[0].filepath;
-    t.is(enhStacks[0].sdk.name, 'typescript');
-    t.is(enhStacks[0].stacks[0].locations.length, 3);
-    t.is(enhStacks[1].stacks[0].locations.length, 6);
-    console.log(enhStacks[1].sources[key][0].content.split('\n').length);
-    console.log(enhStacks[1].sources[key][0].content.split('\n'));
-    t.is(enhStacks[1].sources[key][0].content.split('\n').length, 50);
+    t.is(enhStacks[0].sdk.name, 'typescript', 'sdk name is not "typescript"');
+    t.is(enhStacks[0].stacks.length, 3, 'Number of stacks is different from expected');
+    t.is(enhStacks[0].stacks[0].locations.length, 1, 'Number of locations is different from expected');
+    t.is(enhStacks[1].stacks.length, 3, 'Number of stacks is different from expected');
+    t.is(enhStacks[1].stacks[0].locations.length, 2, 'Number of locations is different from expected');
+    t.is(
+      enhStacks[1].sources[key][0].content.split('\n').length,
+      50,
+      'Number of lines in stack trace is different from expected'
+    );
   });
 
   test('issue-731', async (t) => {
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1163,7 +1163,6 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
           .map((s) => cleanOptionalStackTrace(`\n${s}`))
           .join('\n')
       );
-      console.log(JSON.stringify([stack1, stack2]));
       // Can't get the Trigger stack cleaned, this is okay for now
       // NOTE: we check endsWith because under certain conditions we might see Promise.race in the trace
       t.true(
@@ -1203,9 +1202,6 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       taskQueue: 'test',
       workflowId,
     });
-    console.log('Test output:');
-    console.log(enhancedStack);
-    console.log(JSON.stringify(enhancedStack.stacks));
 
     t.is(enhancedStack.sdk.name, 'typescript', 'sdk name is not "typescript"');
     t.is(enhancedStack.stacks.length, 3, 'Number of stacks is different from expected');
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1194,6 +1194,25 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     });
   }
 
+  //RUN_INTEGRATION_TESTS=1 npx ava packages/test/lib/test-integration.js -m 'Enhanced Stack Trace Test'
+  test('Enhanced Stack Trace Test', async (t) => {
+    const { client } = t.context;
+    const workflowId = uuid4();
+
+    const enhStacks = await client.execute(workflows.enhancedStackTracer, {
+      taskQueue: 'test',
+      workflowId,
+    });
+
+    const key: string = enhStacks[1].stacks[0].locations[0].filepath;
+    t.is(enhStacks[0].sdk.name, 'typescript');
+    t.is(enhStacks[0].stacks[0].locations.length, 3);
+    t.is(enhStacks[1].stacks[0].locations.length, 6);
+    console.log(enhStacks[1].sources[key][0].content.split('\n').length);
+    console.log(enhStacks[1].sources[key][0].content.split('\n'));
+    t.is(enhStacks[1].sources[key][0].content.split('\n').length, 50);
+  });
+
   test('issue-731', async (t) => {
     const { client } = t.context;
     const workflowId = uuid4();
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1163,6 +1163,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
           .map((s) => cleanOptionalStackTrace(`\n${s}`))
           .join('\n')
       );
+      console.log(JSON.stringify([stack1, stack2]));
       // Can't get the Trigger stack cleaned, this is okay for now
       // NOTE: we check endsWith because under certain conditions we might see Promise.race in the trace
       t.true(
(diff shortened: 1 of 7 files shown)


packages/test/src/integration-tests.ts
```diff
@@ -1194,6 +1194,25 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     });
   }
 
+  //RUN_INTEGRATION_TESTS=1 npx ava packages/test/lib/test-integration.js -m 'Enhanced Stack Trace Test'
```
bergundy, line 1197: You can delete this comment

packages/workflow/src/internals.ts
```diff
@@ -416,9 +425,56 @@ export class State {
     [
       '__enhanced_stack_trace',
       (): EnhancedStackTrace => {
-        // const stacks = this.getStackTraces();
-        // const sourceMap = this.sourceMap;
-        return { locations: [], sources: {} };
+        const sourceMap = this.sourceMap;
+        const stacks = this.getStackTraces().join('\n\n');
```
bergundy, line 429: Why join?

bergundy: stacks is an array that should map to a [`StackTrace`](https://github.com/temporalio/proposals/pull/61/files#diff-d9cade3a12b6c56306dfe187736032b0cc8d5b2eed522d605155391dfe62f5f0R65) object

bergundy: When you fix this it'll change the implementation of this method.

packages/workflow/src/internals.ts
```diff
@@ -416,9 +425,56 @@ export class State {
     [
       '__enhanced_stack_trace',
       (): EnhancedStackTrace => {
-        // const stacks = this.getStackTraces();
-        // const sourceMap = this.sourceMap;
-        return { locations: [], sources: {} };
+        const sourceMap = this.sourceMap;
```
bergundy, line 428: I know I wrote this original line but you could replace with:
```suggestion
        const { sourceMap } = this;
```

packages/workflow/src/internals.ts
```diff
@@ -416,9 +425,56 @@ export class State {
...
+        const stacks = this.getStackTraces().join('\n\n');
+        const sdkInfo: SDKInfo = { name: 'typescript', version: '' };
```
bergundy, line 430: Can you put a TODO comment here please so we don't forget this?

packages/workflow/src/internals.ts
```diff
@@ -416,9 +425,56 @@ export class State {
...
+        const locationsPaths: FileLocation[] = [];
+        const sourceMapRecord: Record<string, FileSlice[]> = {};
+        const fileSliceSize = 50; //Lines
```
bergundy, line 433: Can you please extract this to a top level constant?

packages/workflow/src/internals.ts
```diff
@@ -416,9 +425,56 @@ export class State {
...
+        const parsedStacks = errorStackParser.parse({ name: '', message: '', stack: stacks });
+
+        parsedStacks.forEach(
+          (stackTraceEntry: { columnNumber: any; lineNumber: any; fileName: string | undefined }) => {
```
bergundy, line 437: Why `any`?

packages/workflow/src/internals.ts
```diff
@@ -416,9 +425,56 @@ export class State {
...
+            if (stackTraceEntry.columnNumber && stackTraceEntry.lineNumber && stackTraceEntry.fileName) {
+              const fileLocation: FileLocation = {
+                column: stackTraceEntry.columnNumber,
+                line: stackTraceEntry.lineNumber,
+                filepath: stackTraceEntry.fileName,
```
bergundy, line 442: We're going to make these optional

bergundy: You can wait for the spec to be approved before changing

packages/workflow/src/internals.ts
BashkirovN: I had to call join() on the StackTrace because the Error constructor takes stacks as a string.
`const parsedStacks = errorStackParser.parse({ name: '', message: '', stack: stacks });`

packages/workflow/src/internals.ts
BashkirovN: Just wanted to make sure whether you meant to have all three constants to be brought on the global level or just
`const fileSliceSize = 50; //Lines`

packages/workflow/src/internals.ts
BashkirovN: Will change it to
`(stackTraceEntry: StackFrame) => {
            if (stackTraceEntry.columnNumber && stackTraceEntry.lineNumber && stackTraceEntry.fileName) {`

packages/workflow/src/internals.ts
bergundy: You need to parse each individual stack not join them all together into a single stack.

packages/workflow/src/internals.ts
bergundy: Just this one, the others are function local

packages/workflow/src/internals.ts
bergundy: Probably don't even need to specify the type, TS should be able to infer it.

packages/workflow/src/internals.ts
```diff
@@ -416,9 +430,55 @@ export class State {
...
-        // const stacks = this.getStackTraces();
-        // const sourceMap = this.sourceMap;
-        return { locations: [], sources: {} };
+        const { sourceMap } = this;
+        const stacks = this.getStackTraces();
+        const sdkInfo: SDKInfo = { name: 'typescript', version: '' }; //TODO: provide version value
+        const sourceMapRecord: Record<string, FileSlice[]> = {};
+        const stackTraces: StackTrace[] = [];
+
+        stacks.forEach((stack) => {
```
bergundy, line 439: nit: prefer `const stackTraces = stacks.map(...`

packages/workflow/src/internals.ts
```diff
@@ -416,9 +430,55 @@ export class State {
...
+                column: parsedStackTraceLine.columnNumber,
+                line: parsedStackTraceLine.lineNumber,
+                filepath: parsedStackTraceLine.fileName,
+                context: parsedStackTraceLine.functionName,
+              };
+              locationsPaths.push(fileLocation);
+
+              const fileSliceLineOffset: number = Math.max(1, fileLocation.line - Math.floor(fileSliceSize / 2));
+              const fileContent: string = sourceMap.sourcesContent[sourceMap.sources.indexOf(fileLocation.filepath)];
+              const fileSliceContentArray: string[] = fileContent.split('\n');
```
bergundy, line 454: Let's not slice for now, as mentioned in the PR this will might up syntax highlighting in the UI.
If we want to optimize later we could start by sending only the top locations in each trace.

packages/workflow/src/internals.ts
```diff
@@ -416,9 +430,55 @@ export class State {
     [
       '__enhanced_stack_trace',
       (): EnhancedStackTrace => {
-        // const stacks = this.getStackTraces();
-        // const sourceMap = this.sourceMap;
-        return { locations: [], sources: {} };
+        const { sourceMap } = this;
+        const stacks = this.getStackTraces();
+        const sdkInfo: SDKInfo = { name: 'typescript', version: '' }; //TODO: provide version value
```
bergundy, line 435: I'll take care of this later

packages/workflow/src/internals.ts
```diff
@@ -377,7 +391,7 @@ export class State {
   /**
    * Source map file for looking up the source files in response to __enhanced_stack_trace
    */
-  public sourceMap: unknown;
+  public sourceMap: any;
```
bergundy, line 394: Wondering if there's a type definition for this in one of our dependencies that we can use and avoid the `any` type.

packages/workflow/src/internals.ts
```diff
@@ -75,6 +84,11 @@ export type ActivationHandler = {
   [P in keyof coresdk.workflow_activation.IWorkflowActivationJob]: ActivationHandlerFunction<P>;
 };
 
+/**
+ * The number of lines of a code snippet to be displayed
+ */
+export const fileSliceSize = 50; //Lines
```
bergundy, line 90: As mentioned below we can implement this optimization later.

bergundy: You can remove this const.

packages/test/src/workflows/stack-tracer.ts
```diff
@@ -24,3 +25,18 @@ export async function stackTracer(): Promise<[string, string]> {
   const second = await queryOwnWf(wf.stackTraceQuery);
   return [first, second];
 }
+
+export async function enhancedStackTracer(): Promise<[EnhancedStackTrace, EnhancedStackTrace]> {
```
bergundy, line 29: This is a duplicate of the workflow above, you can add a parameter there instead to control which query to issue instead of duplicating.

packages/test/src/integration-tests.ts
```diff
@@ -1194,6 +1194,28 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+  test('Enhanced Stack Trace Test', async (t) => {
+    const { client } = t.context;
+    const workflowId = uuid4();
+
+    const enhStacks = await client.execute(workflows.enhancedStackTracer, {
+      taskQueue: 'test',
+      workflowId,
+    });
+
+    const key: string = enhStacks[1].stacks[0].locations[0].filepath;
```
bergundy, line 1206: Can't TS infer that this is a string?

packages/workflow/src/internals.ts
bergundy: I'd still split the files and make sure the line terminator is always `\n`, note that on Windows line terminator is `\r\n` so you'd need to factor that in when you split the contents.

packages/workflow/src/internals.ts
BashkirovN: Maybe this: 
`public sourceMap: SourceMap | undefined;`

packages/test/src/integration-tests.ts
```diff
@@ -1194,6 +1194,64 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+    t.is(enhancedStack.stacks.length, 3, 'Number of stacks is different from expected');
+    t.is(enhancedStack.stacks[0].locations.length, 2, 'Number of locations is different from expected');
+    t.is(enhancedStack.stacks[1].locations.length, 1, 'Number of locations is different from expected');
+    t.is(enhancedStack.stacks[2].locations.length, 2, 'Number of locations is different from expected');
+    t.is(
+      enhancedStack.stacks[0].locations[0].functionName &&
+        enhancedStack.stacks[0].locations[0].functionName.includes('Function.all (<anonymous>)'),
+      true,
+      'functionName doesn`t match'
+    );
```
bergundy, line 1216:
```suggestion
    t.truthy(enhancedStack.stacks[0].locations[0].functionName?.includes('Function.all (<anonymous>)'));
```

packages/test/src/integration-tests.ts
```diff
@@ -1194,6 +1194,64 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+      'Source key doesn`t match'
+    );
+    t.is(
+      Object.keys(enhancedStack.sources)[1].includes('packages/workflow/src/trigger.ts'),
+      true,
+      'Source key doesn`t match'
+    );
+    t.is(enhancedStack.sources[Object.keys(enhancedStack.sources)[0]].length, 1, 'Number of file slices doesn`t match');
+    t.is(enhancedStack.sources[Object.keys(enhancedStack.sources)[1]].length, 1, 'Number of file slices doesn`t match');
+  });
```
bergundy, line 1253: Can shorten all of these too

bergundy: I'd just do a deep equality check instead of these hand written assertions

packages/workflow/src/internals.ts
```diff
@@ -1,4 +1,6 @@
 import { PayloadConverter } from '@temporalio/common';
+import type { RawSourceMap } from 'source-map';
+import errorStackParser = require('error-stack-parser');
```
bergundy, line 3: You can use import from syntax but I wouldn't bother, I'm going to do some refactoring of this code

packages/workflow/src/internals.ts
```diff
@@ -416,9 +418,42 @@ export class State {
...
+            };
+            locationsPaths.push(fileLocation);
+
+            const fileSlice: FileSlice = {
+              content:
+                sourceMap && sourceMap.sourcesContent && fileLocation.filePath
+                  ? sourceMap.sourcesContent[sourceMap.sources.indexOf(fileLocation.filePath)]
+                  : '',
+              lineOffset: 0,
+            };
```
bergundy, line 445: No need to send a fileSlice for a file that doesn't exist

bergundy (approved):
I'll do some refactoring here and push to this branch.
Overall great work!

PR #802, opened 2022-08-03
fix(worker): Improve worker defaults options heuristics

### LINKED ISSUE #762: [Bug] worker client OOM when running as a container in a k8s cluster.

mjameswh:
## What was changed
- When running inside a cgroups controlled environments (eg. inside a Docker container), use cgroups' memory limit and swap limit to determine the default value of `maxCachedWorkflows`
- Add a warning if `heap_size_limit` is over 50% of the memory limit. This either indicate that a) the process is running in a containerized environment and node is incorrectly determining the default value of `max-old-space-size` to 25% of the _machine_'s total memory (rather than what the container is allowed to obtain), or b) the user has manually configured `max-old-space-size` to a large value, which can be a valid strategy with some node-based software, but risks causing OOM due to the fact that the rust core is left with too little to work with (note: the rust core assigns memory out side of node's heap)
- Dump workers options on the console in order to help users understand and diagnose their configuration

## Why?
See #762 and discussion in #765.

Executing workers in docker containers is a very common scenario. This use case should be properly handled.

## Checklist
1. Fixes #762 and replaces PR #765
2. Add documentation explaining how to properly adjust `--max-old-space-size`


packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,109 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+    if (fs.existsSync('/sys/fs/cgroup/memory/memory.limit_in_bytes')) {
+      // Maximum size (in bytes) of memory usage allowed by this container. '9223372036854771712' if no constraint
+      // Examples:
+      //  - 536870912 => 512 Mb
+      //  - 9223372036854771712 => no constraint
+      const memoryMax = tryReadFileSync('/sys/fs/cgroup/memory/memory.limit_in_bytes');
+      if (isNumber(memoryMax)) limits.memory = Math.min(Number(memoryMax), limits.memory);
+
+      // Maximum size (in bytes) of swap usage allowed by this container. The reported size INCLUDES the size of memory itself.
+      // For example, assuming `docker run --memory 512mb --memory-swap 768mb`, memory.swap.max will report 768mb. File might
```
lorensr, line 498: It won't report 1280mb?

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,109 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+function tryReadFileSync(file: string) {
+  try {
+    return fs.readFileSync(file, { encoding: 'ascii' }) as string;
+  } catch (e) {
+    return undefined;
+  }
+}
+
+function isNumber(s: string | undefined) {
+  return s && /^\d+$/.test(s);
```
lorensr, line 521: Don't think this is important, but I wonder if `!isNaN(parseInt(s))` is faster than regex

packages/worker/src/worker-options.ts
```diff
@@ -431,3 +532,25 @@ export function compileWorkerOptions(opts: WorkerOptionsWithDefaults): CompiledW
...
+  Runtime.instance().logger.info(
+    JSON.stringify(
+      {
+        ...options,
+        ...(options.workflowBundle && isCodeBundleOption(options.workflowBundle)
+          ? {
+              // Avoid dumping workflow bundle code to the console
+              workflowBundle: <WorkflowBundleWithSourceMap>{
+                code: `... source code skipped (length = ${options.workflowBundle.code.length}) ...`,
+                sourceMap: `... source map skipped (length = ${options.workflowBundle.sourceMap.length}) ...`,
```
lorensr, line 547:
```suggestion
                code: `<string of length ${options.workflowBundle.code.length}>`,
                sourceMap: `<string of length ${options.workflowBundle.sourceMap.length}>`,
```

lorensr (changes requested):

lorensr:
Updated the title—sentence case and here are scope options: https://github.com/temporalio/sdk-typescript/blob/main/commitlint.config.js#L8

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,109 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+export function calculateDefaultWorkerOptions(): { maxCachedWorkflows: number } {
+  const systemResources = inspectSystemResources();
+
+  const heapSizeLimit = v8.getHeapStatistics().heap_size_limit;
+  if (heapSizeLimit > systemResources.memory * 0.5) {
+    Runtime.instance().logger.warn(
+      `node's heap size limit is over 50% of available memory (heapSizeLimit = ${heapSizeLimit}, systemMemory = ${systemResources.memory})`
+    );
+    Runtime.instance().logger.warn(`This might degrade performances and could result in OOM errors.`);
+    Runtime.instance().logger.warn(
```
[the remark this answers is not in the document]
mjameswh: Yes, for newlines. Ok... I just looked at what we did in a similar case elsewhere. Will fix.

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,109 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+    if (fs.existsSync('/sys/fs/cgroup/cgroup.controllers')) {
+      // Maximum size (in bytes) of memory usage allowed by this container. 'max' if no constraint.
+      // Examples:
+      //  - 536870912 => 512 Mb
+      //  - max => no constraint
+      const memoryMax = tryReadFileSync('/sys/fs/cgroup/memory.max');
+      if (isNumber(memoryMax)) limits.memory = Math.min(Number(memoryMax), limits.memory);
+
+      // Maximum size (in bytes) of swap usage allowed by this container. The reported size EXCLUDES the size of memory itself.
+      // For example, assuming `docker run --memory 512mb --memory-swap 768mb`, memory.swap.max will report 256mb
```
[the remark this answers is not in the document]
mjameswh: Tested on my machine just now:

```
jwatkins:~$ docker run --rm -it --memory 512mb --memory-swap 768mb node bash -c 'cat /sys/fs/cgroup/memory.swap.max'
268435456
```

and 

```
jwatkins:~$ docker run --rm -it --memory 512mb --memory-swap 100mb node bash -c 'cat /sys/fs/cgroup/memory.swap.max'
docker: Error response from daemon: Minimum memoryswap limit should be larger than memory limit, see usage.
See 'docker run --help'.
```

Note that this is with _cgroups v2_. In __cgroups v1_, `memory.memsw.limit_in_bytes` actually report mem+swap (so 768mb in my first example)

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,109 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      // Defaults to one time the size of maximum memory (if constrained), or 'max' if no constraint on memory
+      //
+      // Examples:
+      //  - 268435456 => 256 Mb
+      //  - max => no constraint
+      const swapMax = tryReadFileSync('/sys/fs/cgroup/memory.swap.max');
+      // For heuristics, don't use more than one time the RAM in swap
```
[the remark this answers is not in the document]
mjameswh: ... don't use more than _the equivalent of RAM's size_ in swap ...

My concern here is that if swap is enabled without cgroup constraint, then we get no specific on how much swap the system itself will allow. Most Linux server administration guidelines however recommend that creating swap partitions that are between 1 and 2 times the size of RAM.

packages/worker/src/worker-options.ts
mjameswh: Note to myself: `s/memory.swap.max/memory.memsw.limit_in_bytes/`

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,109 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
     // 4294967295ms is the maximum allowed time
     isolateExecutionTimeout: debugMode ? '4294967295ms' : '5s',
     workflowThreadPoolSize: 8,
-    maxCachedWorkflows: maxCachedWorkflows ?? Math.max(os.totalmem() / GiB - 1, 1) * 200,
+    maxCachedWorkflows: maxCachedWorkflows ?? defaults.maxCachedWorkflows,
```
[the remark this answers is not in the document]
mjameswh: I have some other options in mind:

- `workflowThreadPoolSize` (default should probably be set to something like `cpu counts * 2`)
- `maxConcurrentActivityTaskExecutions` (default should scale based on heap memory)
- `maxConcurrentLocalActivityExecutions` (default should scale based on heap memory)
- `maxConcurrentWorkflowTaskExecutions` (default should scale based on heap memory)

Adding these should be trivial if we can agree on some appropriate factor.

packages/worker/src/worker-options.ts
mjameswh: Don't think this is a big thing, but you are right that I need to parse the numbers anyway. Fixed.

packages/worker/src/worker-options.ts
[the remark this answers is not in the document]
mjameswh: Still, I refactored the code to avoid the `calculateDefaultWorkerOptions()` function.

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,100 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
     // 4294967295ms is the maximum allowed time
     isolateExecutionTimeout: debugMode ? '4294967295ms' : '5s',
     workflowThreadPoolSize: 8,
-    maxCachedWorkflows: maxCachedWorkflows ?? Math.max(os.totalmem() / GiB - 1, 1) * 200,
+    maxCachedWorkflows:
+      maxCachedWorkflows ??
+      Math.floor(Math.max((systemResources.memory + systemResources.swap - systemResources.heap) / GiB - 1, 1) * 250),
```
lorensr, line 421: For each new workflow in the cache, is the memory taken mostly in or out of heap? I think @Sushisource said there's not much incremental memory usage in Rust? Our apidoc says the largest memory footprint is the code bundle size.

^ determines what the numerator should be, and next question is how much should be subtracted (currently 1GB) for non-workflow-cache (keeping in mind if we have a 50% heap recommendation, then much of the rest is also non-workflow-cache)

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,100 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      }
+    }
+  }
+
+  if (resources.heap > resources.memory * 0.5) {
+    Runtime.instance().logger.warn(
+      `node's heap size limit is over 50% of available memory ` +
+        `(heap size limit = ${resources.heap}, system memory = ${resources.memory}) ` +
+        `This might degrade performances and could result in OOM errors. ` +
+        `For optimal performance, try setting '--max-old-space-size' between 25% and 50% of system memory.`
```
lorensr, line 500: I agree with what you said about having a conservative default, but I wonder if we can give a simple single % rec for them to start at, and if most of workflow cache is in heap, then recommending higher than 50%.

packages/worker/src/worker-options.ts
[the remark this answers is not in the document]
lorensr: Thanks!

packages/worker/src/worker-options.ts
```diff
@@ -169,10 +171,10 @@ export interface WorkerOptions {
    * If the Worker is asked to run an uncached Workflow, it will need to replay the entire Workflow history.
    * Use as a dial for trading memory for CPU time.
    *
-   * You should be able to fit about 500 Workflows per GB of memory dependening on your Workflow bundle size.
+   * Most users are able to fit 250 Workflows per GB of available memory (this depends on your Workflow bundle size).
    * For the SDK test Workflows, we managed to fit 750 Workflows per GB.
```
Sushisource, line 175: And, importantly, on the size of allocations in your workflows.

packages/worker/src/worker-options.ts
Sushisource: The incremental memory usage in Rust is very small. Something on the order of a few kB at most. So it can effectively be ignored for the purposes of this calculation I think. There might be a spot or two where having large payloads can change that but, that's not really predictable.

On another note - if `swap` here means what it normally means (not sure exactly what it means in the container context based on the other code), then we might actually want to ignore it for the purposes of this calculation. If we allow caching until normal memory is filled and we start using swap, things are just gonna be super slow

mjameswh: Yeah... I'll remove that. I was working under the incorrect assumption that once a workflow completed its activation, it associated objects in the nodejs heap were immediately disposed, and that the rust core only kept history in the mean time. I read the code more in depth this morning to finally understand that it is core that nodejs resources for a workflow only get disposed once the workflow is removed from `RunCache`.

bergundy, line 421: FTR, nightly tests take roughly a MB per workflow, I think 4 MB per workflow is a bit conservative but should be a good enough sane default.

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,100 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+export function inspectSystemResources(): { memory: number; swap: number; heap: number } {
+  const resources = {
+    memory: os.totalmem(),
+    swap: 0,
+    heap: v8.getHeapStatistics().heap_size_limit,
+  };
+
+  if (process.platform === 'linux') {
+    // Check for v2 style cgroups
+    if (fs.existsSync('/sys/fs/cgroup/cgroup.controllers')) {
```
bergundy, line 448: I'd avoid the extra syscall and just try to open the file, if it doesn't exist you can try the next method

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,100 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      // Maximum size (in bytes) of memory usage allowed by this container. 'max' if no constraint.
+      // Examples:
+      //  - 536870912 => 512 Mb
+      //  - max => no constraint
+      const memoryMax = Number(tryReadFileSync('/sys/fs/cgroup/memory.max'));
```
bergundy, line 453: I like explicitly handling `"max"`, `undefined` and number in code instead of having the explanation in comments but this is fine AFAIC.

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,100 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      const swapMax = tryReadFileSync('/sys/fs/cgroup/memory.swap.max');
+      if (swapMax) {
+        if (swapMax === 'max')
+          resources.swap = resources.memory; // Swap is allowed, but no size was specified. Default to memory size
+        else if (!isNaN(Number(swapMax))) resources.swap = Number(swapMax);
+      }
+    }
+
+    // Check for v1 style cgroups: /sys/fs/cgroup/memory/memory.limit_in_bytes
+    if (fs.existsSync('/sys/fs/cgroup/memory/memory.limit_in_bytes')) {
```
bergundy, line 470: Same here, might as well try to read the file

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,100 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      const swapMax = Number(tryReadFileSync('/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes'));
+      if (!isNaN(swapMax)) {
+        if (swapMax === 9223372036854771712)
+          resources.swap = resources.memory; // Swap is allowed, but no size was specified. Default to memory size
+        else resources.swap = swapMax - resources.memory; // memory.memsw.limit_in_bytes includes mem+swap
+      }
+    }
+  }
+
+  if (resources.heap > resources.memory * 0.5) {
```
bergundy, line 495: Curious why 50%

packages/worker/src/worker-options.ts
bergundy: Probably don't need to consider the heap size limit here, just use the memory available, most of the resources are going to be consumed by the node heap.

packages/worker/src/worker-options.ts
```diff
@@ -411,13 +416,100 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      maxCachedWorkflows ??
+      Math.floor(Math.max((systemResources.memory + systemResources.swap - systemResources.heap) / GiB - 1, 1) * 250),
     enableSDKTracing: false,
     debugMode: debugMode ?? false,
     interceptors: appendDefaultInterceptors({}),
     sinks: defaultSinks(),
     ...rest,
   };
+
+  reportWorkerOptions(optionsWithDefaults);
```
bergundy, line 429: I'd report during worker construction and use the worker's logger (even though it's the same as the runtime logger), when you call `Runtime.instance()` you're effectively instantiating a runtime which would prevent the user from installing a runtime with their own settings.

packages/worker/src/worker-options.ts
```diff
@@ -431,3 +523,25 @@ export function compileWorkerOptions(opts: WorkerOptionsWithDefaults): CompiledW
     loadedDataConverter: loadDataConverter(opts.dataConverter),
   };
 }
+
+function reportWorkerOptions(options: WorkerOptionsWithDefaults) {
+  Runtime.instance().logger.info(`Worker options:`);
+  Runtime.instance().logger.info(
+    JSON.stringify(
```
bergundy, line 530: Hmm.. why JSON.stringify instead of letting the logger take care of that?

bergundy: Could replace these 2 lines with: `logger.info('Worker created with options', options)` and move this to `Worker.create`

packages/test/src/test-worker-config-heuristics.ts
```diff
@@ -0,0 +1,134 @@
+import test, { ExecutionContext } from 'ava';
+import * as sinon from 'sinon';
+import { addDefaultWorkerOptions } from '@temporalio/worker/lib/worker-options';
+import { GiB } from '@temporalio/worker/lib/utils';
+import type * as V8Type from 'v8';
+import type * as OsType from 'os';
+import type * as FsType from 'fs';
+import type * as ProcessType from 'process';
+
+test.serial('maxCachedWorkflows heuristic without cgroups', (t) => {
```
bergundy, line 10: I'll review the tests after you fix the implementation

packages/worker/src/worker-options.ts
mjameswh: I checked node's official recommandations on this, and they recommand setting `--max-old-space-size` at 75% of total memory, which brings `heap_size_limit` at approximately 80%. I wouldn't mind increasing the warning threashold at 80%. This should be ok in most case, but we risk having containers getting killed from times to times due to OOM. That depends a lot on actual usage patterns.

mjameswh: The problem is not what is _inside_ the heap, but rather what is _outside_ of the heap...

Let's break it down... Essentially, we have following equations:

```
container memory usage = node process memory usage + other processes memory usage

node process memory usage = node heap memory + non-heap memory
node heap memory = young spaces + mid spaces + old space + compiled code + large objects + other heap stuff
node non-heap memory = node stacks + memory allocated from Buffer + memory allocated from native code + other non-heap stuff
```

If `container memory usage` exceeds the (hard) memory limit set on the container, the container immediately gets killed by the OS. No warning and no chance to run GC.

So the question here is essentially: how much space do we reasonably need to set asside for all non-heap to be _safe_.

packages/test/src/test-worker-config-heuristics.ts
```diff
@@ -0,0 +1,36 @@
+import test, { ExecutionContext } from 'ava';
+import * as sinon from 'sinon';
+import { addDefaultWorkerOptions } from '@temporalio/worker/lib/worker-options';
+import { GiB } from '@temporalio/worker/lib/utils';
+import type * as V8Type from 'v8';
+import type * as OsType from 'os';
+
+test.serial('maxCachedWorkflows heuristic', (t) => {
```
mjameswh, line 8: At this point, I am considering removing the unit tests. `maxCachedWorkflows`'s heuristic is once again very simple, so there's little value.

packages/worker/src/worker.ts
```diff
@@ -403,6 +403,18 @@ export class Worker {
   public static async create(options: WorkerOptions): Promise<Worker> {
     const nativeWorkerCtor: WorkerConstructor = this.nativeWorkerCtor;
     const compiledOptions = compileWorkerOptions(addDefaultWorkerOptions(options));
+    Runtime.instance().logger.info(`Worker options:`, {
```
mjameswh, line 406: @bergundy You mentioned a concern about using `Runtime.instance().logger`, which might prevent user form installing its own logger... Not sure about it, but if you still think this is a valid concern, then we'd have to also fix the `getOrCreateBundle(...)` call just bellow.

packages/worker/src/worker-options.ts
```diff
@@ -169,10 +171,12 @@ export interface WorkerOptions {
    * If the Worker is asked to run an uncached Workflow, it will need to replay the entire Workflow history.
    * Use as a dial for trading memory for CPU time.
    *
-   * You should be able to fit about 500 Workflows per GB of memory dependening on your Workflow bundle size.
+   * Most users are able to fit at least 250 Workflows per GB of available memory.
+   * The major factors contributing to a workflow's memory weight are the size of allocations made
+   * by the workflow itself and the size of the workflow bundle (code and source map).
```
lorensr, line 176:
```suggestion
   * The major factors contributing to a Workflow's memory weight are the size of allocations made
   * by the Workflow itself and the size of the Workflow bundle (code and source map).
```

packages/worker/src/worker.ts
bergundy: Here it's not an issue, I'd just avoid using the logger before a call to create a worker because that would instantiate a runtime with the default options.

bergundy:
```suggestion
    Runtime.instance().logger.info('Creating worker', { options: {
```

packages/test/src/test-worker-config-heuristics.ts
bergundy: 👍

packages/worker/src/worker-options.ts
```diff
@@ -420,6 +426,47 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
   };
 }
 
+/**
+ * Inspect the execution environment and return information about available system resources,
+ * taking into account applicable constraints that can be discovered.
+ * At present, it accounts for CGroups v1 and v2 constraints on memory and swap.
+ */
+export function inspectSystemResources(): { memory: number; heap: number } {
```
bergundy, line 434: At this point I'd consider just documenting this instead of issuing a warning but I don't have a strong preference either way.

bergundy (approved):

packages/worker/src/worker-options.ts
mjameswh: I agree that we can disregard the case where `--max-old-space-size` has been explicitly set to a somewhat high but _valid_ value.

However, I think the warning is desirable if the program is running under memory constraints and `--max-old-space-size` has not been set at all. This is _never_ a good situation. OOM is inevitable if the process lives long enough. Yet, most people out there using node and docker won't know about this fact.

So here is what I propose... I remove the percentage and rewrite the message to address specifically the situation I described above. Also, I refer user to the docs for more advice. Does that seems better?

```
    const hasMemoryConstraints = cgroupv1MemoryMax < resources.memory || cgroupv2MemoryMax < resources.memory;
    if (hasMemoryConstraints && resources.heap > resources.memory) {
      const memInMb = toMB(resources.memory, 0);
      const suggestedOldSpaceSizeInMb = toMB(resources.memory * 0.75, 0);

      Runtime.instance().logger.warn(
        `This program is running inside a containerized environment with a memory constraint ` +
          `(eg. docker --memory ${memInMb}m). node itself does not consider such memory constraint ` +
          `in how it manage its heap memory. There is consequently a high probability that ` +
          `the process might face OOM errors. To increase reliability, we recommend adding ` +
          `'--max-old-space-size=${suggestedOldSpaceSizeInMb}' to your node's arguments. ` +
          `Refer to https://docs.temporal.io/application-development/worker-performance for more ` +
          `advices on how to properly tune your workers.`
      );
    }
  }
```

bergundy: This is totally fine but TBH I'd be fine with documenting this too.
Your call, you already did most of the work so I guess it wouldn't hurt.

packages/worker/src/worker-options.ts
```diff
@@ -420,6 +426,54 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
```
lorensr, line 434:
```suggestion
function inspectSystemResources(): { memory: number; heap: number } {
```

packages/worker/src/worker-options.ts
```diff
@@ -420,6 +426,54 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      const suggestedOldSpaceSizeInMb = toMB(resources.memory * 0.75, 0);
+
+      Runtime.instance().logger.warn(
+        `This program is running inside a containerized environment with a memory constraint ` +
+          `(eg. docker --memory ${memInMb}m). node itself does not consider such memory constraint ` +
+          `in how it manage its heap memory. There is consequently a high probability that ` +
+          `the process might face OOM errors. To increase reliability, we recommend adding ` +
+          `'--max-old-space-size=${suggestedOldSpaceSizeInMb}' to your node's arguments. ` +
+          `Refer to https://docs.temporal.io/application-development/worker-performance for more ` +
+          `advices on how to properly tune your workers.`
```
lorensr, line 461:
```suggestion
          `(eg. docker --memory ${memInMb}m). Node itself does not consider this memory constraint ` +
          `in how it manages its heap memory. There is consequently a high probability that ` +
          `the process will crash due to running out of memory. To increase reliability, we recommend adding ` +
          `'--max-old-space-size=${suggestedOldSpaceSizeInMb}' to your node arguments. ` +
          `Refer to https://docs.temporal.io/application-development/worker-performance for more ` +
          `advice on tuning your workers.`
```

lorensr (approved):

packages/worker/src/worker-options.ts
```diff
@@ -420,6 +426,54 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+    heap: v8.getHeapStatistics().heap_size_limit,
+  };
+
+  if (process.platform === 'linux') {
+    // v2 style cgroup
+    const cgroupv2MemoryMax = Number(tryReadFileSync('/sys/fs/cgroup/memory.max'));
+    if (!isNaN(cgroupv2MemoryMax)) resources.memory = Math.min(cgroupv2MemoryMax, resources.memory);
+
+    // v1 style cgroup
+    const cgroupv1MemoryMax = Number(tryReadFileSync('/sys/fs/cgroup/memory/memory.limit_in_bytes'));
```
bergundy, line 446: Let's avoid reading this file if we read the cgroups v2 file successsfully.

mjameswh: Right... Also realized that the warning will show up once per worker. Fix and some refactoring coming in...

packages/worker/src/worker-options.ts
```diff
@@ -420,6 +426,39 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+          `in how it manages its heap memory. There is consequently a high probability that ` +
+          `the process will crash due to running out of memory. To increase reliability, we recommend ` +
+          `adding '--max-old-space-size=${suggestedOldSpaceSizeInMb}' to your node arguments. ` +
+          `Refer to https://docs.temporal.io/application-development/worker-performance for more ` +
+          `advice on tuning your workers.`
+      );
+    }
+  }
+}
+checkHeapSizeLimit();
```
bergundy, line 452: Move this to the Runtime constructor so it isn't automatically initialized when its logger is used

PR #805, opened 2022-08-05
fix: Avoid using dynamic import in /testing

lorensr:
## What was changed

Removes use of dynamic import by downgrading to pre-ESM-only version of `get-port`.

## Why?

So that users don't have to support dynamic imports in their test build system.


bergundy (approved):

bergundy:
Force merged since stress tests are broken

PR #808, opened 2022-08-06
fix(worker): Don't start activity poller if no activities registered

bergundy:


packages/worker/src/worker.ts
```diff
@@ -1498,8 +1498,11 @@ export class Worker {
   }
 
   protected activity$(): Observable<void> {
-    // Note that we poll on activities even if there are no activities registered.
-    // This is so workflows invoking activities on this task queue get a non-retryable error.
+    // This Worker did not register any activities, return early
+    if (this.workflowCreator === undefined) {
+      this.activityPollerStateSubject.next('SHUTDOWN');
+      return EMPTY;
+    }
```
Sushisource, line 1505: You should log a warning in this situation. We might somewhere else already I think, but users can easily get confused.

Sushisource (approved):

packages/worker/src/worker.ts
bergundy: Yes, that's a good idea

lorensr (approved):

bergundy:
I was hoping that the downgrade of docker-compose would fix the stress tests, it didn't..
I'll dig into this when I get more time.

PR #812, opened 2022-08-12
docs: Surface missing core-bridge exports

lorensr:
Definitions of types in `core-bridge/index.d.ts` like `MetricsExplorer` are missing from API ref:

image

I tried adding `core-bridge` to meta package, but no change in docs.


bergundy (approved):

bergundy:
I approved but after thinking about it some more, let's re-export the relevant interfaces from `@temporalio/worker` core-bridge is an internal package.

PR #814, opened 2022-08-12
fix: Help protobufjs find 'long' in Yarn PnP (#810)

### LINKED ISSUE #810: [Bug] Issue with Long dependency being listed as devDependency

Trying to get worker deployments to work correctly in a container that uses yarn to only install dependencies, not devDependencies.

### Describe the bug

Running the workers in Docker containers that are built with `yarn workspaces focus --production @packages/temporal-workflows` results in workers that won't execute with the following stack trace:
```
2022-08-04T03:32:56.766Z [ERROR] Failed to activate workflow {
  runId: 'b2fd073b-42a6-40ce-b719-1468474d7f2c',
  error: TypeError: (seconds || long_1.default.UZERO).mul is not a function
      at tsToMs (/opt/app-root/src/.yarn/cache/@temporalio-internal-workflow-common-npm-1.0.0-rc.1-b66f54d7d6-4272b669aa.zip/node_modules/@temporalio/internal-workflow-common/lib/time.js:30:10)
      at /opt/app-root/src/.yarn/cache/@temporalio-worker-npm-1.0.0-rc.1-2998f0d7a7-147819cf17.zip/node_modules/@temporalio/worker/lib/worker.js:664:90
      at async /opt/app-root/src/.yarn/cache/@temporalio-worker-npm-1.0.0-rc.1-2998f0d7a7-147819cf17.zip/node_modules/@temporalio/worker/lib/tracing.js:65:20
      at async /opt/app-root/src/.yarn/cache/@temporalio-worker-npm-1.0.0-rc.1-2998f0d7a7-147819cf17.zip/node_modules/@temporalio/worker/lib/worker.js:593:28,
  workflowExists: false
}
```

However, if we build this instead with `yarn workspaces focus --all @packages/temporal-workflows`, the worker/container works, which implies that some devDependency needs to be listed as a dependency (likely the `long` package).

### Minimal Reproduction

I'm having issues figuring out how to reproduce this or test this to be honest. I tried testing this solution by forking this repository to see if making the suggested changes would fix it but run into errors trying to use the forked repository as my `temporalio` version. I get the following error (https://community.temporal.io/t/fork-temporal-sdk-typescript/5630).

### Environment/Versions

Containers have been built on both M1Macs and Linux machines, using the latest Temporal SDK version. We're using yarn pnp and 3.2.0.

### Additional context

Related to [this forum post](https://community.temporal.io/t/worker-error-with-mul-is-not-a-function/5573)

mjameswh:
## What was changed

Forcibly inject `long` into `protobufs` objects.

Reference: https://github.com/protobufjs/protobuf.js/issues/1745#issuecomment-1200319399

## Why?
`protobufjs/light` delegates loading of `long` to the `@protobufjs/inquire` package. Now, `@protobufjs/inquire` itself does not declare a dependency on long in its `package.json`.

This pattern works in most environment, thanks to the way traditional package managers lay out their `node_modules` directory, but is forbidden by Yarn 3 and PNPM, though the extend to which they goes in applying that restriction depend on various configuration item... Yarn 3's PnP is essentially the strictest of all.

In practice, that means that in some environement, `protobufjs` fails to load `long`. It fallbacks to create number objects that are very similar to the objects created by `long`, but without any method on them. This in turn cause an error message such as `TypeError: (seconds || long_1.default.UZERO).mul is not a function`.

## Checklist
1. Closes #810

2. How was this tested:
Verdaccio build on the hello world sample, reconfigured for Yarn 3 PnP.

### DIFF
--- a/packages/testing/scripts/compile-proto.mjs
+++ b/packages/testing/scripts/compile-proto.mjs
@@ -47,6 +47,20 @@ async function compileProtos(protoPath, jsOutputFile, dtsOutputFile, ...args) {
   ];
   await promisify(pbjs.main)(pbjsArgs);
 
+  // Workaround an issue that prevents protobufjs from loading 'long' in Yarn 3 PnP
+  // https://github.com/protobufjs/protobuf.js/issues/1745#issuecomment-1200319399
+  const pbjsOutput = readFileSync(jsOutputFile, 'utf8');
+  writeFileSync(
+    jsOutputFile,
+    pbjsOutput.replace(
+      /(require\("protobufjs\/minimal"\);)$/m,
+      `$1
+       $protobuf.util.Long = require('long');
+       $protobuf.configure();
+    `
+    )
+  );
+
   console.log(`Creating protobuf TS definitions from ${protoPath}`);
   await promisify(pbts.main)(['--out', dtsOutputFile, jsOutputFile]);
 
--- a/packages/proto/protos/root.js
+++ b/packages/proto/protos/root.js
@@ -1,3 +1,9 @@
+// Workaround an issue that prevents protobufjs from loading 'long' in Yarn 3 PnP
+// https://github.com/protobufjs/protobuf.js/issues/1745#issuecomment-1200319399
+const $protobuf = require('protobufjs/light');
+$protobuf.util.Long = require('long');
+$protobuf.configure();
+
 const { patchProtobufRoot } = require('../lib/patch-protobuf-root');
 const unpatchedRoot = require('./json-module');
 module.exports = patchProtobufRoot(unpatchedRoot);


packages/testing/scripts/compile-proto.mjs
```diff
@@ -47,6 +47,20 @@ async function compileProtos(protoPath, jsOutputFile, dtsOutputFile, ...args) {
   ];
   await promisify(pbjs.main)(pbjsArgs);
 
+  // Workaround an issue that prevents protobufjs from loading 'long' in Yarn 3 PnP
+  // https://github.com/protobufjs/protobuf.js/issues/1745#issuecomment-1200319399
+  const pbjsOutput = readFileSync(jsOutputFile, 'utf8');
```
bergundy, line 52: You'e have to fix this in the protos build script, not everyone uses the testing package.

This specific script will go away once we take the java test server protos from core (which now embeds it).

bergundy (commented):
Thanks @mjameswh!

packages/testing/scripts/compile-proto.mjs
mjameswh: Ehh… not sure what you mean here… I _did_ fixed _both_ proto and testing… In proto, I injected in root.js… seems cleaner and is just as effective. Do you mean that we should not fix testing at all?

lorensr (approved):

packages/testing/scripts/compile-proto.mjs
bergundy: Sorry, I missed that for some reason.

PR #815, opened 2022-08-15
feat(worker): Add webpack configuration, closes #537

### LINKED ISSUE #537: [Feature Request] Allow webpack configuration

Users needing or wanting to configure webpack in various ways.

@mjameswh listed a couple ideas in https://github.com/temporalio/sdk-typescript/issues/516#issue-1162103439

> Allow users to apply more generic overrides to the Webpack config. For example, an optional property webpackConfigOverrides could be added to BundleOptions. If it is there, that property would be merged with the default/calculated Webpack configuration. Alternatively, it could be an override function, which receive the default/calculated Webpack configuration, and returns the modified configuration (this is much more flexible, and easier from our point of view, more harder to work with for developers). This is for example [how Gatsby works](https://www.gatsbyjs.com/docs/how-to/custom-configuration/add-custom-webpack-config/). This approach would not only solve the current issue, but also provide a way to deal with many other very custom needs that developers may have (think of.. Hey guys! We would like to use [define](https://webpack.js.org/plugins/define-plugin/) for our workflows 😄 ) However, this would be more complex for developers to use, and arguably, there are not that many things that can be customized in Temporal's Webpack build process without risking breaking things.

lorensr:
#537


lorensr:
*sigh*

Using GITHUB_TOKEN env var for downloading from GitHub
(https://github.com/temporalio/sdk-typescript/runs/7842899143?check_suite_focus=true#step:6:20)
HTTPError: Response code 403 (rate limit exceeded)

packages/test/src/test-webpack.ts
```diff
@@ -0,0 +1,27 @@
...
+
+if (RUN_INTEGRATION_TESTS) {
+  test('WorkerOptions.bundlerOptions.configureWebpack works', async (t) => {
+    const taskQueue = `${t.title}-${uuid4()}`;
+    await t.throwsAsync(
+      Worker.create({
+        taskQueue,
+        workflowsPath: require.resolve('./workflows'),
+        bundlerOptions: {
+          configureWebpack: (config) => {
```
bergundy, line 14: I'd use the declarative form to match the rest of the options (`webpackConfigHook`?)

packages/worker/src/worker-options.ts
```diff
@@ -252,6 +255,12 @@ export interface WorkerOptions {
   debugMode?: boolean;
 
   bundlerOptions?: {
+    /**
+     * Before we bundle the Workflow code with Webpack, we call `configureWebpack`, passing the Webpack
```
bergundy, line 259: Don't use "we" in API reference, say "Before Workflow code is bundled with Webpack, ..."

packages/worker/src/worker.ts
```diff
@@ -543,17 +543,26 @@ export class Worker {
     logger: Logger
   ): Promise<WorkflowBundleWithSourceMap | undefined> {
     if (compiledOptions.workflowsPath) {
+      if (compiledOptions.workflowBundle) {
+        logger.info('WorkerOptions.workflowsPath is set, so workflowBundle option is ignored');
```
bergundy, line 547: This should throw an error instead of logging in info level which is often disabled

packages/worker/src/worker.ts
```diff
@@ -543,17 +543,26 @@ export class Worker {
...
         payloadConverterPath: compiledOptions.dataConverter?.payloadConverterPath,
         ignoreModules: compiledOptions.bundlerOptions?.ignoreModules,
+        configureWebpack: compiledOptions.bundlerOptions?.configureWebpack,
       });
       const bundle = await bundler.createBundle();
       logger.info('Workflow bundle created', { size: `${toMB(bundle.code.length)}MB` });
       return bundle;
     } else if (compiledOptions.workflowBundle) {
+      if (compiledOptions.bundlerOptions) {
+        logger.info('WorkerOptions.workflowsBundle is set, so bundlerOptions is ignored');
```
bergundy, line 563: Same

packages/worker/src/workflow/bundler.ts
```diff
@@ -315,6 +320,12 @@ export interface BundleOptions {
    * > NOTE: This is an advanced option that should be used with care.
    */
   ignoreModules?: string[];
+
+  /**
+   * Before we bundle the Workflow code with Webpack, we call `configureWebpack`, passing the Webpack
+   * {@link https://webpack.js.org/configuration/ | configuration} object so you can modify it.
+   */
+  configureWebpack?: (config: webpack.Configuration) => webpack.Configuration;
```
bergundy, line 328: I'd just make this return `void` and explicitly document that the config should be mutated by this function.

lorensr: I think `=> void` and "so you can modify it." suffices?

mjameswh: I disagree on this one.

It should be possible to do something such as:

```
function myCustomizer(input: Webpack.Configuration) {
    return {
         ...input,
         someExtraField: "whatever",
    }
}
```
That pattern actually creates a new object.

Similarly, there are some utility packages that exists to help with customizing an existing webpack config (that can sometime be a non-trivial task). Take [this one](https://www.npmjs.com/package/webpack-merge) for example. Some of these packages actually returns a _new_ object, rather than modify the existing object.

packages/test/src/test-webpack.ts
```diff
@@ -0,0 +1,27 @@
+import { Worker } from '@temporalio/worker';
```
mjameswh, line 1: I think `test-bundler.ts` might be more appropriate.

Also considering that it might be better for future maintainability if we merge this with `test-worker-from-bundle.ts`... That's where most bundler-related tests are at the moment, including some tests that are not related to workers.

packages/worker/src/workflow/bundler.ts
```diff
@@ -315,6 +320,12 @@ export interface BundleOptions {
    * > NOTE: This is an advanced option that should be used with care.
    */
   ignoreModules?: string[];
+
+  /**
+   * Before Workflow code is bundled with Webpack, `webpackConfigHook` is called with the Webpack
+   * {@link https://webpack.js.org/configuration/ | configuration} object so you can modify it.
+   */
+  webpackConfigHook?: (config: webpack.Configuration) => void;
```
mjameswh, line 328: As it is now, `webpackConfigHook` is defined to return `void`, but everywhere, the code expects it to actually return the configuration.

mjameswh (changes requested):
Need to decide on correct return type for `webpackConfigHook` and make sure it is coherent everywhere.

bergundy:
I'd be fine with mutating the object or returning a modified object.
I guess I went with mutating because it's easier for users, I personally prefer not to mutate but I don't see much value in requesting not to mutate the object in this case.

packages/worker/src/workflow/bundler.ts
lorensr: Fixed

packages/worker/src/workflow/bundler.ts
lorensr: I prefer the flexibility of returning—you can mutate & return, or create new—will change back

packages/test/src/test-webpack.ts
lorensr: good idea, thanks

mjameswh (approved):

PR #816, opened 2022-08-15
fix(worker): Remove unnecessary `ReplayWorkerOptions`

lorensr:
Added the ones @cretz suggested as well as `stickyQueueScheduleToStartTimeout` and `maxCachedWorkflows`

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -546,7 +546,7 @@ export class Worker {
     if (compiledOptions.workflowsPath) {
       if (compiledOptions.workflowBundle) {
         throw new ValueError(
-          'You cannot set both WorkerOptions.workflowsPath and .workflowBundle: only one can be used.'
+          'You cannot set both WorkerOptions.workflowsPath and .workflowBundle: only one can be used'
         );
       }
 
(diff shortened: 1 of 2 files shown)


bergundy (approved):

packages/worker/src/worker.ts
```diff
@@ -563,9 +563,7 @@ export class Worker {
       return bundle;
     } else if (compiledOptions.workflowBundle) {
       if (compiledOptions.bundlerOptions) {
-        throw new ValueError(
-          `You cannot set both WorkerOptions.workflowBundle and .bundlerOptions: if you're providing the bundle, bundlerOptions are not used.`
-        );
+        throw new ValueError(`You cannot set both WorkerOptions.workflowBundle and .bundlerOptions`);
```
bergundy, line 566: We should add a lint rule to disallow string templates with no args, definitely not blocking this PR

bergundy:
windows test hangs again :( force merged

packages/worker/src/worker.ts
lorensr: I also use when there's a `'`

PR #818, opened 2022-08-16
refactor(client)!: Remove Long

lorensr:
Followup to https://github.com/temporalio/sdk-typescript/pull/735#pullrequestreview-1071907277


bergundy (approved):
Make sure to note that this is a breaking change

PR #819, opened 2022-08-16
fix(workflow)!: Make breakpoints work inside workflow isolate context

### LINKED ISSUE #795: [Bug] Breakpoints do not work for workflow code
Due to bundling of workflow code, it's impossible to set breakpoints on code that runs in workflow context.

There used to be a workaround for this before stack trace query was implemented by setting a `debugger` statement at the top of a workflow function but since that change we no longer inline the source-maps making it impossible to locate the source of the bundled code.

@mjameswh discovered that by setting a file name on the vm script we use to compile the code and having the file and source map available (on disk and possibly HTTP URL) the debugger can map back to the source and breakpoint work again.

Our proposed solution is to start an HTTP server in the worker when running in debug mode to serve the source map.
We'd debated whether to use http URLs or save the source map to a temporary directory and decided to go with HTTP since that will enable access when debugging remotely (for example when running a worker in a docker container).
The security implications of running an HTTP server in debug mode are minimal since a server has to be started regardless for the debugger to connect to the process.
The port to listen on should be configurable to enable remote access, if it is not provided, the server will listen on a random port.

mjameswh:
## Objective
Makes it possible to actually use breakpoints in workflow code.

## What was changed
- Use inline source maps
- Set script filename to be an absolute path, matching the original filename stored in the source map
- Make sure different workflow bundles get different filename
- Preload and cache the script object

## Checklist
1. Closes #795
2. How was this tested:
- Breakpoints during tests execution (vscode)
- Verdaccio build on local examples (vscode)
- Verdaccio build on local examples (chrome's debugger)

## Other notes
- Chrome's debugger don't know much about nodejs worker threads. It however works properly when running a worker in single thread mode (ie. `debugMode = true`)

### DIFF
--- a/packages/test/src/test-workflows.ts
+++ b/packages/test/src/test-workflows.ts
@@ -10,6 +10,7 @@ import { msToTs } from '@temporalio/internal-workflow-common';
 import { coresdk } from '@temporalio/proto';
 import { WorkflowCodeBundler } from '@temporalio/worker/lib/workflow/bundler';
 import { VMWorkflow, VMWorkflowCreator } from '@temporalio/worker/lib/workflow/vm';
+import { parseWorkflowCode } from '@temporalio/worker/lib/worker';
 import { WorkflowInfo } from '@temporalio/workflow';
 import anyTest, { ExecutionContext, TestInterface } from 'ava';
 import dedent from 'dedent';
@@ -46,8 +47,8 @@ const test = anyTest as TestInterface<Context>;
 test.before(async (t) => {
   const workflowsPath = path.join(__dirname, 'workflows');
   const bundler = new WorkflowCodeBundler({ workflowsPath });
-  const { code, sourceMap } = await bundler.createBundle();
-  t.context.workflowCreator = await TestVMWorkflowCreator.create(code, sourceMap, 100);
+  const workflowBundle = parseWorkflowCode((await bundler.createBundle()).code);
+  t.context.workflowCreator = await TestVMWorkflowCreator.create(workflowBundle, 100);
 });
 
 test.after.always(async (t) => {
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -9,26 +9,25 @@ import { Runtime } from './runtime';
 import { InjectedSinks } from './sinks';
 import { GiB } from './utils';
 import { LoggerSinks } from './workflow-log-interceptor';
-import { WorkflowBundleWithSourceMap } from './workflow/bundler';
+import { WorkflowBundle } from './workflow/bundler';
 import * as v8 from 'v8';
 import * as os from 'os';
 
 export type { WebpackConfiguration };
 
-export interface WorkflowBundlePathWithSourceMap {
+export interface WorkflowBundlePath {
   codePath: string;
-  sourceMapPath: string;
 }
-export type WorkflowBundleOption = WorkflowBundleWithSourceMap | WorkflowBundlePathWithSourceMap;
+export type WorkflowBundleOption = WorkflowBundle | WorkflowBundlePath;
 
-export function isCodeBundleOption(bundleOpt: WorkflowBundleOption): bundleOpt is WorkflowBundleWithSourceMap {
+export function isCodeBundleOption(bundleOpt: WorkflowBundleOption): bundleOpt is WorkflowBundle {
   const opt = bundleOpt as any; // Cast to access properties without TS complaining
-  return typeof opt.code === 'string' && typeof opt.sourceMap === 'string';
+  return typeof opt.code === 'string';
 }
 
-export function isPathBundleOption(bundleOpt: WorkflowBundleOption): bundleOpt is WorkflowBundlePathWithSourceMap {
+export function isPathBundleOption(bundleOpt: WorkflowBundleOption): bundleOpt is WorkflowBundlePath {
   const opt = bundleOpt as any; // Cast to access properties without TS complaining
-  return typeof opt.codePath === 'string' && opt.sourceMapPath;
+  return typeof opt.codePath === 'string';
 }
 
 /**
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -75,10 +75,13 @@ import {
   WorkerOptions,
 } from './worker-options';
 import { WorkflowCodecRunner } from './workflow-codec-runner';
-import { WorkflowBundleWithSourceMap, WorkflowCodeBundler } from './workflow/bundler';
+import { WorkflowBundle, WorkflowCodeBundler } from './workflow/bundler';
 import { Workflow, WorkflowCreator } from './workflow/interface';
 import { ThreadedVMWorkflowCreator } from './workflow/threaded-vm';
 import { VMWorkflowCreator } from './workflow/vm';
+import type { RawSourceMap } from 'source-map';
+import * as vm from 'node:vm';
+import * as path from 'node:path';
 
 import IWorkflowActivationJob = coresdk.workflow_activation.IWorkflowActivationJob;
 import EvictionReason = coresdk.workflow_activation.RemoveFromCache.EvictionReason;
@@ -145,13 +148,9 @@ export interface WorkerConstructor {
   create(
     connection: NativeConnection,
     options: CompiledWorkerOptions,
-    bundle?: WorkflowBundleWithSourceMap
+    bundle?: WorkflowBundle
   ): Promise<NativeWorkerLike>;
-  createReplay(
-    options: CompiledWorkerOptions,
-    history: History,
-    bundle: WorkflowBundleWithSourceMap
-  ): Promise<NativeWorkerLike>;
+  createReplay(options: CompiledWorkerOptions, history: History, bundle: WorkflowBundle): Promise<NativeWorkerLike>;
 }
 
 function isOptionsWithBuildId<T extends CompiledWorkerOptions>(options: T): options is T & { buildId: string } {
@@ -180,7 +179,7 @@ export class NativeWorker implements NativeWorkerLike {
   public static async create(
     connection: NativeConnection,
     options: CompiledWorkerOptions,
-    bundle?: WorkflowBundleWithSourceMap
+    bundle?: WorkflowBundle
   ): Promise<NativeWorkerLike> {
     const runtime = Runtime.instance();
     const nativeWorker = await runtime.registerWorker(
@@ -193,7 +192,7 @@ export class NativeWorker implements NativeWorkerLike {
   public static async createReplay(
     options: CompiledWorkerOptions,
     history: History,
-    bundle: WorkflowBundleWithSourceMap
+    bundle: WorkflowBundle
   ): Promise<NativeWorkerLike> {
     const runtime = Runtime.instance();
     const nativeWorker = await runtime.createReplayWorker(addBuildIdIfMissing(options, bundle.code), history);
@@ -410,9 +409,8 @@ export class Worker {
         ...(compiledOptions.workflowBundle && isCodeBundleOption(compiledOptions.workflowBundle)
           ? {
               // Avoid dumping workflow bundle code to the console
-              workflowBundle: <WorkflowBundleWithSourceMap>{
+              workflowBundle: <WorkflowBundle>{
                 code: `<string of length ${compiledOptions.workflowBundle.code.length}>`,
-                sourceMap: `<string of length ${compiledOptions.workflowBundle.sourceMap.length}>`,
               },
             }
           : {}),
@@ -441,14 +439,14 @@ export class Worker {
   }
 
   protected static async createWorkflowCreator(
-    bundle: WorkflowBundleWithSourceMap,
+    workflowBundle: WorkflowBundleWithSourceMap,
     compiledOptions: CompiledWorkerOptions
   ): Promise<WorkflowCreator> {
     if (compiledOptions.debugMode) {
-      return await VMWorkflowCreator.create(bundle.code, bundle.sourceMap, compiledOptions.isolateExecutionTimeoutMs);
+      return await VMWorkflowCreator.create(workflowBundle, compiledOptions.isolateExecutionTimeoutMs);
     } else {
       return await ThreadedVMWorkflowCreator.create({
-        ...bundle,
+        workflowBundle,
         threadPoolSize: compiledOptions.workflowThreadPoolSize,
         isolateExecutionTimeoutMs: compiledOptions.isolateExecutionTimeoutMs,
       });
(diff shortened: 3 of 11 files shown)


packages/worker/src/workflow/bundler.ts
```diff
@@ -95,16 +94,15 @@ export class WorkflowCodeBundler {
     // Cast because the type definitions are inaccurate
     const memoryFs = memfs.createFsFromVolume(vol);
     ufs.use(memoryFs as any).use({ ...realFS, readdir: readdir as any });
-    const distDir = '/dist';
+    const distDir = '/dist'; // FIXME: Make it possible for user to specify destination path
```
bergundy, line 97: This would require writing to the real FS as well

packages/worker/src/worker.ts
```diff
@@ -441,11 +440,11 @@ export class Worker {
   }
 
   protected static async createWorkflowCreator(
-    bundle: WorkflowBundleWithSourceMap,
+    bundle: WorkflowBundleAndCodePath,
     compiledOptions: CompiledWorkerOptions
   ): Promise<WorkflowCreator> {
     if (compiledOptions.debugMode) {
-      return await VMWorkflowCreator.create(bundle.code, bundle.sourceMap, compiledOptions.isolateExecutionTimeoutMs);
+      return await VMWorkflowCreator.create(bundle.code, bundle.codePath, compiledOptions.isolateExecutionTimeoutMs);
```
bergundy, line 447: Wouldn't the path be encoded in the bundle's source map?

packages/worker/src/worker-options.ts
```diff
@@ -9,26 +9,25 @@ import { Runtime } from './runtime';
...
-  sourceMapPath: string;
 }
-export type WorkflowBundleOption = WorkflowBundleWithSourceMap | WorkflowBundlePathWithSourceMap;
+export type WorkflowBundleOption = WorkflowBundle | WorkflowBundlePath;
 
-export function isCodeBundleOption(bundleOpt: WorkflowBundleOption): bundleOpt is WorkflowBundleWithSourceMap {
+export function isCodeBundleOption(bundleOpt: WorkflowBundleOption): bundleOpt is WorkflowBundle {
   const opt = bundleOpt as any; // Cast to access properties without TS complaining
-  return typeof opt.code === 'string' && typeof opt.sourceMap === 'string';
+  return typeof opt.code === 'string';
```
bergundy, line 25: How do we extract the path from this?

packages/worker/src/workflow/vm.ts
```diff
@@ -158,11 +158,10 @@ export class VMWorkflowCreator implements WorkflowCreator {
   public static async create<T extends typeof VMWorkflowCreator>(
     this: T,
     code: string,
-    sourceMap: string,
+    codePath: string | undefined,
```
bergundy, line 161: Doesn't look like this is used

packages/worker/src/workflow/vm.ts
```diff
@@ -256,7 +255,27 @@ export class VMWorkflowCreator implements WorkflowCreator {
       });
     }
 
-    return new this(script, sourceMap, isolateExecutionTimeoutMs) as InstanceType<T>;
+    return new this(script, isolateExecutionTimeoutMs) as InstanceType<T>;
+  }
+
+  private static async makeScriptWithSourceMap(code: string) {
```
bergundy, line 261: Might as well make this protected in case anyone wants to extend this (granted we won't provide API safety for protected members)

packages/worker/src/workflow/vm.ts
```diff
@@ -256,7 +255,27 @@ export class VMWorkflowCreator implements WorkflowCreator {
...
+    const sourceMappingUrlDataRegex = /\s*\n[/][/][#]\s+sourceMappingURL=data:(?:[^,]*),([^\s]+)\s*$/m;
```
bergundy, line 262: You could probably restrict data to base64 only characters.

packages/worker/src/workflow/vm.ts
```diff
@@ -256,7 +255,27 @@ export class VMWorkflowCreator implements WorkflowCreator {
...
+
+      const filename = `/temporal/${sourceMapParsed.file}`;
+      const script = new vm.Script(code, { filename });
+      const sourceMapConsumer = await new SourceMapConsumer(sourceMapParsed);
+
+      return { script, filename, sourceMapConsumer };
+    } else {
+      const filename = `/temporal/workflowbundle-${uuid4()}.js`;
+      const script = new vm.Script(code, { filename });
+      return { script, filename, sourceMapConsumer: undefined };
```
bergundy, line 277: When would you expect this to happen? Wondering if we should emit an error here.

packages/worker/src/workflow/vm.ts
bergundy: Could use some comments in this function

packages/worker/src/workflow/vm.ts
```diff
@@ -256,7 +255,27 @@ export class VMWorkflowCreator implements WorkflowCreator {
...
+
+  private static async makeScriptWithSourceMap(code: string) {
+    const sourceMappingUrlDataRegex = /\s*\n[/][/][#]\s+sourceMappingURL=data:(?:[^,]*),([^\s]+)\s*$/m;
+    const sourceMapMatcher = code.match(sourceMappingUrlDataRegex);
+
+    if (sourceMapMatcher) {
+      const sourceMap = Buffer.from(sourceMapMatcher[1], 'base64').toString();
+      const sourceMapParsed: RawSourceMap = JSON.parse(sourceMap);
+
+      const filename = `/temporal/${sourceMapParsed.file}`;
```
bergundy, line 269: Why `/temporal`?

packages/worker/src/workflow/vm.ts
mjameswh: I don’t expect this to happen if the user has correctly followed the bundling process, except for bundles that will have been produced with 1.0.1 and will be run on 1.0.2 without being rebundled (Which isn’t safe anyway…)

But I still need to deal with this case somehow. I’d be willing though to go with a throw new Error(…) if you are ok with it

packages/worker/src/workflow/bundler.ts
```diff
@@ -221,11 +219,11 @@ export { api };
       },
       entry: [entry],
       mode: 'development',
-      devtool: 'source-map',
+      devtool: 'inline-source-map',
       output: {
+        pathinfo: true,
```
mjameswh, line 224: Note to myself: not sure if this is rquired

packages/worker/src/workflow/bundler.ts
mjameswh: I removed the comment. The use case for this isn't clear enough. It would complexify the API and yet, there would be no guarantee that the file would actually be loaded from the same path location as it was created initially. For the same reason, I removed all references to `codePath` in other files.

packages/worker/src/worker.ts
mjameswh: Same

packages/worker/src/workflow/vm.ts
bergundy: In that case it could make sense to extract this in the main Node.js thread so the error could be caught by the user (would also save some (insignificant amount of?) work because right now we run this code once per worker thread.

packages/worker/src/worker.ts
```diff
@@ -569,11 +564,8 @@ export class Worker {
       if (isCodeBundleOption(compiledOptions.workflowBundle)) {
         return compiledOptions.workflowBundle;
       } else if (isPathBundleOption(compiledOptions.workflowBundle)) {
-        const [code, sourceMap] = await Promise.all([
-          fs.readFile(compiledOptions.workflowBundle.codePath, 'utf8'),
-          fs.readFile(compiledOptions.workflowBundle.sourceMapPath, 'utf8'),
-        ]);
-        return { code, sourceMap };
+        const code = await fs.readFile(compiledOptions.workflowBundle.codePath, 'utf8');
```
bergundy, line 567: Does it matter if we use the path here as name of the vm script?

packages/worker/src/workflow/vm.ts
bergundy: I think this will be a problem on windows.

packages/worker/src/workflow/vm.ts
mjameswh: That's actually something I was considering... I'd probably move this to `getOrCreateBundle()`. What I'm unsure is what exactly should then be transferred to threads (ie. through `postMessage`)... Would it be reasonable (performance, memory...) to pass `sourceMapParsed` through the messaging port?

bergundy: The raw source map is transferrable, it's just a plain JS object.

packages/worker/src/workflow/vm.ts
mjameswh: After more testing, it is much more reliable if I use `cwd()`. That also resolve the Windows concern.

packages/worker/src/worker.ts
mjameswh: At this point, I don't think supporting this use case is worth delaying a release.
But should we want to do it, this would probably require rewriting the content of the `sourceMapURL` so that the filename matches the actual file.

packages/test/src/test-workflows.ts
```diff
@@ -10,6 +10,7 @@ import { msToTs } from '@temporalio/internal-workflow-common';
 import { coresdk } from '@temporalio/proto';
 import { WorkflowCodeBundler } from '@temporalio/worker/lib/workflow/bundler';
 import { VMWorkflow, VMWorkflowCreator } from '@temporalio/worker/lib/workflow/vm';
+import { parseWorkflowCode } from '@temporalio/worker/src/worker';
```
bergundy, line 13: We typically import from lib, I guess for tests it doesn’t matter though

packages/worker/src/worker.ts
bergundy: Sounds like it shouldn’t matter

packages/worker/src/worker.ts
```diff
@@ -1648,6 +1643,38 @@ export class Worker {
...
+    // Context has not been properly configured, so eventual errors are possible. Just ignore at this point
+  }
+  preloadedScriptAndContext = [script, context];
+
+  return { code, sourceMap, filename };
+}
+
+// Keep these objects from GC. Otherwise the debugger risks source mapping resolution errors
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+let preloadedScriptAndContext: [vm.Script, unknown];
```
bergundy, line 1676: What happens if multiple bundles are loaded?
Can we store this in a worker attribute?

packages/worker/src/workflow/bundler.ts
bergundy: Want to check?

bergundy (approved):
Approving but I would like to see if there’s a way to get rid of the global script and context, maybe just store on the worker for now?
Wondering why it’s needed and if there’s a way around it.

packages/worker/src/worker.ts
mjameswh: > I would like to see if there’s a way to get rid of the global script and context, maybe just store on the worker for now?
> Wondering why it’s needed and if there’s a way around it.

The issue is that parsing of the source map is done by the debugger (ie. in VS Code's process), concurrently with the execution of the workflow code (ie. node's process). Parsing the source map takes some time, and is associated with the vm context in which the script was discovered. Once parsing is completed, the debugger informs node about file mappings that it has discovered, _with the associated script id and context id_. Now, if that context has already been garbaged collected at that point (which is very likely with workflows since activations are very short lived and workflow bundles tends to be rather large), then node throws an error, ignoring file mappings provided by the debugger.

Adding preload and caching of the script and context has decreased breakpoints misses from over 9 out of 10 in some test scenarios to no miss at all. Workflow breakpoints are simply not reliable without this.

We can indeed attach the objects to the worker instead of keeping it in a global variable, and that was actually my original intent. Another possibility I'm considering at the moment is to attach the objects to a 10 seconds timer. That is more than enough to let the debugger completes parsing. I will have to look at vscode's debugger trace log however to determine if this cause errors at some later point.

I also observed that preloading is more reliable if performed _before_ worker threads are started, rather that once in worker threads. For example, I had cases where only one or two threads (over 8) would hit breakpoints, until disable and reenable all breakpoints. Looking at a trace from vscode's debugger, it looks like the debugger attaches some listeners to newly created threads, so my hypothesis is that the debugger might miss some type of events if they happens very early after thread creation.

mjameswh:
> I would like to see if there’s a way to get rid of the global script and context, maybe just store on the worker for now?
> Wondering why it’s needed and if there’s a way around it.

The issue is that parsing of the source map is done by the debugger (ie. in VS Code's process), concurrently with the execution of the workflow code (ie. node's process). Parsing the source map takes some time, and is associated with the vm context in which the script was discovered. Once parsing is completed, the debugger informs node about file mappings that it has discovered, _with the associated script id and context id_. Now, if that context has already been garbaged collected at that point (which is very likely with workflows since activations are very short lived and workflow bundles tends to be rather large), then node throws an error, ignoring file mappings provided by the debugger.

Adding preload and caching of the script and context has decreased breakpoints misses from over 9 out of 10 in some test scenarios to no miss at all. Workflow breakpoints are simply not reliable without this.

We can indeed attach the objects to the worker instead of keeping it in a global variable, and that was actually my original intent. Another possibility I'm considering at the moment is to attach the objects to a 10 seconds timer. That is more than enough to let the debugger completes parsing. I will have to look at vscode's debugger trace log however to determine if this cause errors at some later point.

I also observed that preloading is more reliable if performed _before_ worker threads are started, rather that once in worker threads. For example, I had cases where only one or two threads (over 8) would hit breakpoints, until disable and reenable all breakpoints. Looking at a trace from vscode's debugger, it looks like the debugger attaches some listeners to newly created threads, so my hypothesis is that the debugger might miss some type of events if they happens very early after thread creation.

packages/worker/src/worker.ts
mjameswh: My intuition also tells me that it should not matter (after all, the code being executed gave that source map, so why would we care about the name of the file), but experimentation tells me that setting breakpoints from the UI doesn't work unless the file name stored inside the source map exactly matches the script's filename.

TBH, that's not a big thing to fix. I'm just concerned that it might delay a release. Your call.

mjameswh: Done.

packages/worker/src/worker.ts
mjameswh: Moved this to a 10 seconds timer. That's well enough for the debugger to parse the source and fully attach to the node process.

packages/worker/src/worker.ts
```diff
@@ -1649,6 +1644,46 @@ export class Worker {
...
+  if (codePath) {
+    sourceMap.file = filename;
+    const patchedSourceMapJson = Buffer.from(JSON.stringify(sourceMap)).toString('base64');
+    const fixedSourceMappingUrl = `\n//# sourceMappingURL=data:application/json;base64,${patchedSourceMapJson}`;
+    code = code.slice(0, -sourceMapMatcher[1].length) + fixedSourceMappingUrl;
+  }
+
+  // Preloading the script makes breakpoints significantly more reliable and more responsive
+  // Keep these objects from GC long enough for debugger to complete parsing the source map and reporting locations
+  // to the node process. Otherwise, the debugger risks source mapping resolution errors, meaning breakings wont work.
```
lorensr, line 1671:
```suggestion
  // to the node process. Otherwise, the debugger risks source mapping resolution errors, meaning breakpoints won't work.
```

packages/worker/src/worker.ts
```diff
@@ -1649,6 +1644,46 @@ export class Worker {
...
+
+  if (codePath) {
+    sourceMap.file = filename;
+    const patchedSourceMapJson = Buffer.from(JSON.stringify(sourceMap)).toString('base64');
+    const fixedSourceMappingUrl = `\n//# sourceMappingURL=data:application/json;base64,${patchedSourceMapJson}`;
+    code = code.slice(0, -sourceMapMatcher[1].length) + fixedSourceMappingUrl;
+  }
+
+  // Preloading the script makes breakpoints significantly more reliable and more responsive
+  // Keep these objects from GC long enough for debugger to complete parsing the source map and reporting locations
```
lorensr, line 1670: move this comment down to `setTimeout`?

packages/worker/src/worker.ts
```diff
@@ -1649,6 +1644,46 @@ export class Worker {
   }
 }
 
+export interface WorkflowBundleWithSourceMap {
+  code: string;
+  sourceMap: RawSourceMap;
+  filename: string;
+}
+
+export function parseWorkflowCode(code: string, codePath?: string): WorkflowBundleWithSourceMap {
```
lorensr, line 1653: Could add automated test for this fn

lorensr (approved):

bergundy (approved):

PR #820, opened 2022-08-17
Implement enhanced stack traces

bergundy:
Thanks @BashkirovN for coming up with the initial version.
Rewritten to avoid unnecessarily parsing stack traces, propagated version, and made tests more comprehensive.
Replaces #799

### DIFF
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1178,6 +1178,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
           .map((s) => cleanOptionalStackTrace(`\n${s}`))
           .join('\n')
       );
+      console.log(JSON.stringify([stack1, stack2]));
       // Can't get the Trigger stack cleaned, this is okay for now
       // NOTE: we check endsWith because under certain conditions we might see Promise.race in the trace
       t.true(
@@ -1213,16 +1214,61 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     const { client } = t.context;
     const workflowId = uuid4();
 
-    const enhStacks = await client.execute(workflows.enhancedStackTracer, {
+    const enhancedStack = await client.execute(workflows.enhancedStackTracer, {
       taskQueue: 'test',
       workflowId,
     });
+    console.log('Test output:');
+    console.log(enhancedStack);
+    console.log(JSON.stringify(enhancedStack.stacks));
 
-    t.is(enhStacks[0].sdk.name, 'typescript', 'sdk name is not "typescript"');
-    t.is(enhStacks[0].stacks.length, 3, 'Number of stacks is different from expected');
-    t.is(enhStacks[0].stacks[0].locations.length, 1, 'Number of locations is different from expected');
-    t.is(enhStacks[1].stacks.length, 3, 'Number of stacks is different from expected');
-    t.is(enhStacks[1].stacks[0].locations.length, 2, 'Number of locations is different from expected');
+    t.is(enhancedStack.sdk.name, 'typescript', 'sdk name is not "typescript"');
+    t.is(enhancedStack.stacks.length, 3, 'Number of stacks is different from expected');
+    t.is(enhancedStack.stacks[0].locations.length, 2, 'Number of locations is different from expected');
+    t.is(enhancedStack.stacks[1].locations.length, 1, 'Number of locations is different from expected');
+    t.is(enhancedStack.stacks[2].locations.length, 2, 'Number of locations is different from expected');
+    t.is(
+      enhancedStack.stacks[0].locations[0].functionName &&
+        enhancedStack.stacks[0].locations[0].functionName.includes('Function.all (<anonymous>)'),
+      true,
+      'functionName doesn`t match'
+    );
+    t.is(
+      enhancedStack.stacks[0].locations[1].filePath &&
+        enhancedStack.stacks[0].locations[1].filePath.includes('packages/test/src/workflows/stack-tracer.ts'),
+      true,
+      'filePath doesn`t match'
+    );
+    t.is(
+      enhancedStack.stacks[1].locations[0].filePath &&
+        enhancedStack.stacks[1].locations[0].filePath.includes('packages/test/src/workflows/stack-tracer.ts'),
+      true,
+      'filePath doesn`t match'
+    );
+    t.is(
+      enhancedStack.stacks[2].locations[0].functionName &&
+        enhancedStack.stacks[2].locations[0].functionName.includes('Promise.then (<anonymous>)'),
+      true,
+      'functionName doesn`t match'
+    );
+    t.is(
+      enhancedStack.stacks[2].locations[1].filePath &&
+        enhancedStack.stacks[2].locations[1].filePath.includes('packages/workflow/src/trigger.ts'),
+      true,
+      'filePath doesn`t match'
+    );
+    t.is(
+      Object.keys(enhancedStack.sources)[0].includes('packages/test/src/workflows/stack-tracer.ts'),
+      true,
+      'Source key doesn`t match'
+    );
+    t.is(
+      Object.keys(enhancedStack.sources)[1].includes('packages/workflow/src/trigger.ts'),
+      true,
+      'Source key doesn`t match'
+    );
+    t.is(enhancedStack.sources[Object.keys(enhancedStack.sources)[0]].length, 1, 'Number of file slices doesn`t match');
+    t.is(enhancedStack.sources[Object.keys(enhancedStack.sources)[1]].length, 1, 'Number of file slices doesn`t match');
   });
 
   test('issue-731', async (t) => {
--- a/packages/test/src/integration-tests.ts
+++ b/packages/test/src/integration-tests.ts
@@ -1218,17 +1218,11 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       workflowId,
     });
 
-    const key: string = enhStacks[1].stacks[0].locations[0].filepath;
     t.is(enhStacks[0].sdk.name, 'typescript', 'sdk name is not "typescript"');
     t.is(enhStacks[0].stacks.length, 3, 'Number of stacks is different from expected');
     t.is(enhStacks[0].stacks[0].locations.length, 1, 'Number of locations is different from expected');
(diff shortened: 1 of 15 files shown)


packages/test/src/integration-tests.ts
```diff
@@ -1207,6 +1211,68 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+
+      const enhancedStack = await client.execute(workflows.enhancedStackTracer, {
+        taskQueue: 'test',
+        workflowId,
+      });
+
+      const stacks = enhancedStack.stacks.map((s) => ({
+        locations: s.locations.map((l) => ({
+          ...l,
+          ...(l.filePath ? { filePath: l.filePath.replace(path.resolve(__dirname, '../../../'), '') } : undefined),
```
mjameswh, line 1227: I'd make sure that the normalization done here is identical to what we have in `helper.ts#cleanStackTrace()`.

packages/test/src/integration-tests.ts
```diff
@@ -1207,6 +1211,68 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+      t.deepEqual(stacks, [
+        {
+          locations: [
+            {
+              functionName: 'Function.all',
+            },
+            {
+              filePath: '/packages/test/src/workflows/stack-tracer.ts',
+              functionName: 'enhancedStackTracer',
+              line: 32,
```
mjameswh, line 1241: Validating that line and column fields are set seems like a good thing, but having actual line number and column numbers here will make this test pretty fragile. Not sure how to make it better though.

Could we simply confirm that both fields are integer numbers greater than 1?
Or round the numbers to the next multiple of 20? Not pretty, but it would at least means that the values would rarely change...

packages/test/src/integration-tests.ts
```diff
@@ -1207,6 +1211,68 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+            {
+              filePath: '/packages/workflow/src/trigger.ts',
+              functionName: 'Trigger.then',
+              line: 47,
+              column: 24,
+            },
+          ],
+        },
+      ]);
+      const expectedSources = ['../src/workflows/stack-tracer.ts', '../../workflow/src/trigger.ts'].map((p) => [
```
mjameswh, line 1270: Having .. and ../.. in these files paths makes it more difficult to read and compare to the values in the trace above.

I'd suggest writing these as forward-only paths relative to the packages directory. Then, simply rewrite the `path.resolve`s below to something like `path.resolve(__dirname, '../..', p)`

packages/test/src/workflows/stack-tracer.ts
```diff
@@ -3,6 +3,7 @@
  * @module
  */
 import * as wf from '@temporalio/workflow';
+import { EnhancedStackTrace } from '@temporalio/workflow/src/interfaces';
```
mjameswh, line 6: `type`

mjameswh, line 6:
```suggestion
import type { EnhancedStackTrace } from '@temporalio/workflow/src/interfaces';
```

packages/workflow/src/internals.ts
```diff
@@ -377,35 +384,75 @@ export class State {
...
+
+  protected getStackTraces(): Stack[] {
+    const { childToParent, promiseToStack } = (globalThis as any).__TEMPORAL__.promiseStackStore as PromiseStackStore;
+    const internalNodes = new Set(
+      [...childToParent.values()].reduce((acc, curr) => {
+        for (const p of curr) {
+          acc.add(p);
+        }
+        return acc;
+      }, new Set())
```
mjameswh, line 405: There are two `new Set()` in this. Looks like an incomplete refactor.

packages/workflow/src/internals.ts
```diff
@@ -377,35 +384,75 @@ export class State {
...
+        const sources: Record<string, FileSlice[]> = {};
+        if (this.showStackTraceSources) {
+          for (const { locations } of stacks) {
+            for (const { filePath } of locations) {
+              if (!filePath) continue;
+              const content = sourceMap?.sourcesContent?.[sourceMap?.sources.indexOf(filePath)];
+              if (!content) continue;
+              sources[filePath] = [
+                {
+                  content,
```
mjameswh, line 448: What about `\r\n` to `\n` normalization?

mjameswh: I wonder if we could somehow make sure that normalization is done in webpack. That would be more efficient but not sure how difficult it might be...

lorensr (approved):

packages/worker/src/workflow/vm.ts
```diff
@@ -19,13 +19,51 @@ interface ActivationContext {
 // Not present in @types/node for some reason
 const { promiseHooks } = v8 as any;
 
+/**
+ * Variant of {@link cutoffStackTrace} that works with FileLocation, keep this in sync with the original implementation
```
mjameswh, line 23: There's more and more source map and stack traces manipulations sparkled all over the place... It will be difficult to keep everything in sync and it shadows `vm.ts`'s actual responsibilities. Not high priority, but it be pertinent to at least gather some part of this logic to some central place.

packages/test/src/integration-tests.ts
```diff
@@ -1207,6 +1211,68 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+            {
+              filePath: '/packages/test/src/workflows/stack-tracer.ts',
+              functionName: 'enhancedStackTracer',
+              line: 32,
+              column: 35,
+            },
+          ],
+        },
+        {
+          locations: [
```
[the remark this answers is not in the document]
mjameswh: There might be [other properties](https://github.com/temporalio/proposals/pull/61#discussion_r932748665) on `StackTrace` in other SDK, and some might be added even in TS eventually.

packages/test/src/integration-tests.ts
```diff
@@ -1207,6 +1211,68 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+          ...(l.filePath ? { filePath: l.filePath.replace(path.resolve(__dirname, '../../../'), '') } : undefined),
+        })),
+      }));
+      t.is(enhancedStack.sdk.name, 'typescript');
+      t.is(enhancedStack.sdk.version, pkg.version); // Expect workflow and worker versions to match
+      t.deepEqual(stacks, [
+        {
+          locations: [
+            {
+              functionName: 'Function.all',
```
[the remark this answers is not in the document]
bergundy: Nope, that's not what you get in the trace because it's a static method

mjameswh (approved):

packages/test/src/integration-tests.ts
bergundy: I did remove the line numbers in the basic stack trace test, I guess it's fine to leave them here.
I don't expect this file to change much.
Don't mind removing them either.

packages/test/src/integration-tests.ts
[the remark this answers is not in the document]
bergundy: 👍 @mjameswh

packages/test/src/integration-tests.ts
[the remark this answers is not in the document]
mjameswh: Might be worth a comment on this line though... I too was unsure about this when I read it

packages/test/src/integration-tests.ts
bergundy: Don't think this is critical TBH

packages/test/src/workflows/stack-tracer.ts
bergundy: I think there's an eslint rule that can verify this, don't think it's critical but don't mind changing either.

packages/worker/src/workflow/vm.ts
bergundy: Agree 👍

bergundy: I didn't want to put this in common since I didn't want to export these functions to users so just embedded it here.
Tests should protect us from desync.

packages/worker/src/workflow/vm.ts
```diff
@@ -19,13 +19,51 @@ interface ActivationContext {
...
+ * Internal helper to format callsite "name" portion in stack trace
+ */
+function formatCallsiteName(callsite: NodeJS.CallSite) {
+  const typeName = callsite.getTypeName();
+  const methodName = callsite.getMethodName();
+  const functionName = callsite.getFunctionName();
+  const isConstructor = callsite.isConstructor();
+
+  return typeName && methodName
+    ? `${callsite.getTypeName()}.${callsite.getMethodName()}`
```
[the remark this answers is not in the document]
bergundy: Thanks

packages/test/src/workflows/stack-tracer.ts
mjameswh: Right. It might be worth adding the linter for this. Definitely not critical in this case.

packages/test/src/integration-tests.ts
mjameswh: No strong opinion on this.

packages/workflow/src/internals.ts
bergundy: Hmm.. I missed this, will fix, thanks!

packages/workflow/src/internals.ts
bergundy: TBH I thought about this some more and I don't think it really matters, the UI can handle both line endings.
Didn't bother to update the proposal.

packages/test/src/integration-tests.ts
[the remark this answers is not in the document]
lorensr: I'm proposing we change the trace so that it's easier to line up with code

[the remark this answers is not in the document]
bergundy: This is what we get from node (v8), I don't know if we want to rewrite the native stack traces.
That being said if I run an `mjs` file containing:
```js
await Promise.all([
  (async () => {
    throw new Error("bla");
  })(),
]);
```

I do see `    at async Promise.all (index 0)` in the stack trace, it looks different with commonjs though.
Could be worth exploring but don't think it's a high priority.

PR #826, opened 2022-08-19
docs: Improve API reference

lorensr:


bergundy (commented):
This PR is titled `docs:` but it upgrades core with a breaking change..

lorensr:
It didn't upgrade, it was on `15ea5d3` from 7 weeks ago, when I created this branch. I need to remember to change my workflow from:

```
git checkout main
git pull
git checkout branch
git merge main
[resolve conflicts]
git add .
git commit
```

After `git merge main` I need to notice:

```
 modified:   packages/core-bridge/sdk-core (new commits)
```

And do `git submodule update` before `git add .`

packages/activity/src/index.ts
```diff
@@ -15,7 +15,13 @@
  *
  * ### Cancellation
  *
- * Activities may be cancelled only if they {@link Context.heartbeat | emit heartbeats}.
+ * Activity Cancellation serves three purposes:
+ *
+ * - It lets an Activity know it doesn't need to keep doing work.
+ * - It gives the Activity time to clean up any resources it has created.
+ * - If the Activity accepts Cancellation, the Server will know to retry the Activity.
```
bergundy, line 22: A cancelled activity doesn't get retried, this is incorrect.

packages/activity/src/index.ts
```diff
@@ -15,7 +15,13 @@
...
+ *
+ * Activities may be Cancelled only if they {@link Context.heartbeat | emit heartbeats}.
```
bergundy, line 24: You might want to note that this doesn't apply to local activities.

packages/worker/src/worker.ts
```diff
@@ -658,10 +658,13 @@ export class Worker {
   }
 
   /**
-   * Start shutting down the Worker. Immediately transitions {@link state} to `'STOPPING'` and asks Core to shut down.
-   * Once Core has confirmed that it's shutting down, the Worker enters `'DRAINING'` state unless the Worker has already
-   * been `'DRAINED'`. All currently running Activities are Cancelled. Once all currently running Activities and
-   * Workflow Tasks have completed, the Worker transitions to `'STOPPED'`.
+   * Start shutting down the Worker. The Worker stops polling for new tasks and sends
+   * {@link https://typescript.temporal.io/api/namespaces/activity#cancellation | cancellation} to running Activities.
```
bergundy, line 662: I'd mention that these activities don't end up cancelled, they end up failed in case cancellation is respected.

bergundy: Also might want to note that we the reason for cancellation in this case will be WORKER_SHUTDOWN (please double check the exact reason).

packages/activity/src/index.ts
lorensr: Can LAs call `Context.current().heartbeat()`? 

Can LAs receive Cancellation?

packages/worker/src/worker.ts
lorensr: Checked. How's this?

The Worker stops polling for new tasks and sends
   * {@link https://typescript.temporal.io/api/namespaces/activity#cancellation | cancellation} (via a
   * {@link CancelledFailure} with `message` set to `'WORKER_SHUTDOWN'`) to running Activities. Note: if the Activity
   * accepts cancellation (i.e. re-throws or allows the `CancelledFailure` to be thrown out of the Activity function),
   * the Activity Task will be marked as failed, not cancelled. It's helpful for the Activity Task to be marked failed
   * during shutdown because the Server will retry the Activity sooner (than if the Server had to wait for the Activity
   * Task to time out).

packages/activity/src/index.ts
bergundy: > Can LAs call `Context.current().heartbeat()`?

No (I think we ignore that for now but we might want to throw there, that's what Python does).

> Can LAs receive Cancellation?

Only in core based SDKs.

packages/worker/src/worker.ts
bergundy: LGTM

bergundy (approved):
Pre-approving

PR #827, opened 2022-08-19
fix: Revert all breaking changes from 1.0.0, deprecate sourceMap

bergundy:


packages/worker/src/workflow/bundler.ts
```diff
@@ -26,7 +26,12 @@ export function moduleMatches(userModule: string, modules: string[]): boolean {
   return modules.some((module) => userModule === module || userModule.startsWith(`${module}/`));
 }
 
-export interface WorkflowBundle {
+export interface WorkflowBundleWithSourceMap {
+  /**
+   * Source maps are genreated inline, this is no longer used
```
lorensr, line 31:
```suggestion
   * Source maps are generated inline—this is no longer used
```

packages/worker/src/workflow/bundler.ts
```diff
@@ -102,6 +107,7 @@ export class WorkflowCodeBundler {
 
     // Cast because the type definitions are inaccurate
     return {
+      sourceMap: '',
```
lorensr, line 110:
```suggestion
      sourceMap: 'this field is deprecated',
```

I'm envisioning devs logging this value, seeing it empty, and wondering what they did wrong.

lorensr (changes requested):
this PR:
```
export interface WorkflowBundlePath {
  codePath: string;
}
export type WorkflowBundleOption = WorkflowBundle | WorkflowBundlePath;
```

`1.0.1`:
```
export interface WorkflowBundlePathWithSourceMap {
    codePath: string;
    sourceMapPath: string;
}
export declare type WorkflowBundleOption = WorkflowBundleWithSourceMap | WorkflowBundlePathWithSourceMap;
```

I think it's okay for `WorkflowBundleWithSourceMap` to change name to `WorkflowBundle` because it has the same fields, but think we need to add `sourceMapPath: string;`?

packages/worker/src/worker.ts
```diff
@@ -426,7 +422,7 @@ export class Worker {
     const connection = options.connection ?? (await InternalNativeConnection.connect());
     let nativeWorker: NativeWorkerLike;
     try {
-      nativeWorker = await nativeWorkerCtor.create(connection, compiledOptions);
+      nativeWorker = await nativeWorkerCtor.create(connection, compiledOptions, bundle?.code);
```
mjameswh, line 425: I had miss that previously, but we should not need to pass code to native worker. Not blocking though.

bergundy: Yeah, I missed this too TBH, it's for generating the buildId, I think we should do this in the worker factory.

packages/worker/src/workflow/bundler.ts
bergundy: Good suggestion.

bergundy:
I'll rename the second interface, will drop the internal type alias to verify the docs aren't referencing it accidentally.

packages/worker/src/worker.ts
bergundy: Refactored this a bit.

packages/worker/src/workflow/bundler.ts
```diff
@@ -26,7 +26,12 @@ export function moduleMatches(userModule: string, modules: string[]): boolean {
   return modules.some((module) => userModule === module || userModule.startsWith(`${module}/`));
 }
 
-export interface WorkflowBundle {
+export interface WorkflowBundleWithSourceMap {
+  /**
+   * Source maps are genreated inline - this is no longer used
```
lorensr, line 31:
```suggestion
   * Source maps are generated inline - this is no longer used
```

packages/worker/src/worker-options.ts
```diff
@@ -18,7 +18,27 @@ export type { WebpackConfiguration };
...
+
+/**
+ * Note this no longer contains a source map.
+ * The name was preserved to avoid breaking backwards compatibility.
+ *
+ * @deprecated
+ */
+export interface WorkflowBundlePathWithSourceMap {
+  codePath: string;
+  sourceMap: string;
```
lorensr, line 30:
```suggestion
  sourceMapPath: string;
```

lorensr (approved):

packages/worker/src/worker-options.ts
bergundy: Thanks!

PR #833, opened 2022-08-20
fix: Warn instead of throwing when getting `workflowBundle` with `workflowsPath` and `bundlerOptions`.

bergundy:
NOTE: We now prefer taking the bundle over workflowsPath which is the correct behavior and what users *should* expect.

Also warning that workflow interceptors are ignored when using a bundle.

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -64,7 +64,7 @@ import pkg from './pkg';
 import { History, Runtime } from './runtime';
 import { closeableGroupBy, mapWithState, mergeMapWithState } from './rxutils';
 import { childSpan, getTracer, instrument } from './tracing';
-import { byteArrayToBuffer, convertToParentWorkflowType, toMB } from './utils';
+import { byteArrayToBuffer, convertToParentWorkflowType } from './utils';
 import {
   addDefaultWorkerOptions,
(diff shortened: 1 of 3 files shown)


packages/worker/src/worker.ts
```diff
@@ -535,13 +535,29 @@ export class Worker {
...
+      if (compiledOptions.workflowsPath) {
+        logger.warn('Ignoring WorkerOptions.workflowsPath because WorkerOptions.workflowBundle is set');
+      }
+      if (compiledOptions.bundlerOptions) {
+        logger.warn('Ignoring WorkerOptions.bundlerOptions because WorkerOptions.workflowBundle is set');
+      }
+      if (compiledOptions.interceptors?.workflowModules) {
+        logger.warn(
+          'Ignoring compiledOptions.interceptors?.workflowModules because WorkerOptions.workflowBundle is set.\n' +
+            'To use workflow interceptors with a workflowBundle, pass them in the call to bundleWorkflowCode.'
```
bergundy, line 548: @lorensr multiple sentences, I decided to use punctuation :)

bergundy:
@lorensr this should eliminate any backwards incompatible behavior for the release.

lorensr (approved):
Thanks! ☺️

PR #836, opened 2022-08-22
chore: Upgrade ubuntu CI runner `18.04 -> 22.04`

lorensr:
18.04 is deprecated, and outages start today: https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,9 +119,9 @@ jobs:
       fail-fast: true
       matrix:
         include:
-          - os: ubuntu-18.04
+          - os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


bergundy:
@lorensr this will cause users to require newer libc.
We intentionally downgraded from `latest` for this.
If we use newer ubuntu, we'll have to make sure to link to older libc when we build the bridge.

mjameswh:
TIL: Deno's CI produces its Linux builds [inside a sysroot](https://github.com/denoland/deno/blob/main/.github/workflows/ci.yml#L170) for this exact reason (their actual build environment is Ubuntu 16.4).

bergundy:
@mjameswh is taking care of this

PR #847, opened 2022-08-24
fix(client): Only require `signalArgs` when needed

lorensr:
when calling `signalWithStart`

### DIFF
--- a/packages/client/src/workflow-options.ts
+++ b/packages/client/src/workflow-options.ts
@@ -37,15 +37,26 @@ export interface WorkflowOptions extends CommonWorkflowOptions {
   followRuns?: boolean;
 }
 
-export interface WorkflowSignalWithStartOptions<SA extends any[] = []> extends WorkflowOptions {
+export type WorkflowSignalWithStartOptions<SignalArgs extends any[] = []> = SignalArgs extends []
+  ? WorkflowSignalWithStartOptionsWithoutArgs
+  : WorkflowSignalWithStartOptionsWithArgs<SignalArgs>;
+
+export interface WorkflowSignalWithStartOptionsWithoutArgs extends WorkflowOptions {
   /**
    * SignalDefinition or name of signal
    */
-  signal: SignalDefinition<SA> | string;
+  signal: SignalDefinition | string;
+}
+
+export interface WorkflowSignalWithStartOptionsWithArgs<SignalArgs extends any[]> extends WorkflowOptions {
+  /**
+   * SignalDefinition or name of signal
+   */
+  signal: SignalDefinition<SignalArgs> | string;
   /**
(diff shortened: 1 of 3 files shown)


packages/client/src/workflow-options.ts
```diff
@@ -37,15 +37,26 @@ export interface WorkflowOptions extends CommonWorkflowOptions {
...
   /**
    * SignalDefinition or name of signal
    */
-  signal: SignalDefinition<SA> | string;
+  signal: SignalDefinition<SignalArgs> | string;
   /**
    * Arguments to invoke the signal handler with
    */
-  signalArgs: SA;
+  signalArgs: SignalArgs;
```
lorensr, line 59: If there's a way for an interface to have a conditional property, I couldn't find it 😄

packages/client/src/workflow-options.ts
```diff
@@ -37,15 +37,26 @@ export interface WorkflowOptions extends CommonWorkflowOptions {
   followRuns?: boolean;
 }
 
-export interface WorkflowSignalWithStartOptions<SA extends any[] = []> extends WorkflowOptions {
+export type WorkflowSignalWithStartOptions<SignalArgs extends any[] = []> = SignalArgs extends []
```
bergundy, line 40: You want the condition to be the other way around.
See https://github.com/temporalio/sdk-typescript/blob/a7be2cd12bf535df607a53035373e33d74c28e25/packages/internal-workflow-common/src/workflow-options.ts#L87

lorensr: Ah, I thought `[]` would just be empty array. I've changed to an array with at least one item:`[any, ...any[]]`

packages/client/src/workflow-options.ts
```diff
@@ -37,15 +37,26 @@ export interface WorkflowOptions extends CommonWorkflowOptions {
...
-export interface WorkflowSignalWithStartOptions<SA extends any[] = []> extends WorkflowOptions {
+export type WorkflowSignalWithStartOptions<SignalArgs extends any[] = []> = SignalArgs extends [any, ...any[]]
+  ? WorkflowSignalWithStartOptionsWithArgs<SignalArgs>
+  : WorkflowSignalWithStartOptionsWithoutArgs;
+
+export interface WorkflowSignalWithStartOptionsWithoutArgs extends WorkflowOptions {
+  /**
+   * SignalDefinition or name of signal
+   */
+  signal: SignalDefinition | string;
```
bergundy, line 48: I'd support optional empty args here, same way we do for regular workflow `args`

lorensr: Thanks, updated

bergundy (approved):
LGTM, thanks

PR #848, opened 2022-08-24
feat: Implement entrypoint for debug replayer

bergundy:
Will be used in the VS Code debugger extension.

See [`.vscode/settings.json`](https://github.com/bergundy/sdk-typescript/blob/0205f0b11a815e6a3aff0fc4c25fe6b651a95db5/.vscode/settings.json) and [`debug-replayer.ts`](https://github.com/bergundy/sdk-typescript/blob/0205f0b11a815e6a3aff0fc4c25fe6b651a95db5/packages/test/src/debug-replayer.ts) for reference.

### DIFF
--- a/packages/worker/src/debug-replayer/index.ts
+++ b/packages/worker/src/debug-replayer/index.ts
@@ -1,0 +1,75 @@
+import worker_threads from 'node:worker_threads';
+import { temporal } from '@temporalio/proto';
+import { ReplayWorkerOptions } from '../worker-options';
+import { Worker } from '../worker';
+import { Client } from './client';
+
+async function run(options: ReplayWorkerOptions): Promise<void> {
+  const baseUrl = process.env.TEMPORAL_DEBUGGER_PLUGIN_URL;
+  if (!baseUrl) {
+    throw new Error('Missing TEMPORAL_DEBUGGER_PLUGIN_URL environment variable');
+  }
+  const client = new Client({ baseUrl });
+  const response = await client.get('history');
+  const history = temporal.api.history.v1.History.decode(
+    await Client.readAll(response),
+    Client.contentLength(response)
+  );
+
+  const thread = new worker_threads.Worker(require.resolve('./worker-thread'));
+
+  const rejectedPromise = new Promise<void>((_, reject) => {
+    // Set the global notifyRunner runner function that can be used from the workflow interceptors.
+    // The function makes an HTTP request to the runner in a separate thread and blocks the current thread until a
+    // response is received.
+    (globalThis as any).notifyRunner = (wftStartEventId: number) => {
+      const sab = new SharedArrayBuffer(4);
+      const responseBuffer = new Int32Array(sab);
+      thread.postMessage({ eventId: wftStartEventId, responseBuffer });
+      Atomics.wait(responseBuffer, 0, 0);
+      if (responseBuffer[0] === 2) {
+        // Error occurred (logged by worker thread)
+        reject(new Error('Failed to call runner back'));
+      }
+    };
+  });
+
+  await Promise.race([
+    rejectedPromise,
+    Worker.runReplayHistory(
+      {
+        ...options,
+        interceptors: {
+          ...options.interceptors,
+          workflowModules: [
+            // Inbound goes first so user can set breakpoints in own inbound interceptors.
+            require.resolve('./debug-inbound-interceptor'),
+            ...(options.interceptors?.workflowModules ?? []),
+            // Outbound goes last - notifies the runner in the finally block, user-provided outbound interceptors are
+            // resumed after the interceptor methods resolve.
+            require.resolve('./debug-outbound-interceptor'),
+          ],
+        },
+      },
+      history
+    ),
+  ]);
+}
+
+/**
+ * Start a replayer for debugging purposes.
+ *
+ * Use this method to integrate the replayer with external debuggers like the Temporal VS Code debbuger extension.
+ */
+export function startDebugReplayer(options: ReplayWorkerOptions): void {
+  run(options).then(
+    () => {
+      console.log('Replay completed successfully');
+      process.exit(0);
+    },
+    (err) => {
+      console.error('Replay failed', err);
+      process.exit(1);
+    }
+  );
+}
--- a/packages/worker/src/debug-replayer/index.ts
+++ b/packages/worker/src/debug-replayer/index.ts
@@ -4,6 +4,8 @@ import { ReplayWorkerOptions } from '../worker-options';
 import { Worker } from '../worker';
 import { Client } from './client';
 
+let thread: worker_threads.Worker | undefined = undefined;
+
 async function run(options: ReplayWorkerOptions): Promise<void> {
   const baseUrl = process.env.TEMPORAL_DEBUGGER_PLUGIN_URL;
   if (!baseUrl) {
@@ -16,7 +18,9 @@ async function run(options: ReplayWorkerOptions): Promise<void> {
     Client.contentLength(response)
   );
 
-  const thread = new worker_threads.Worker(require.resolve('./worker-thread'));
+  // Only create one per process.
+  // Not caring about globals here for to get simpler DX, this isn't meant for production use cases.
+  thread = thread || new worker_threads.Worker(require.resolve('./worker-thread'));
 
   const rejectedPromise = new Promise<void>((_, reject) => {
     // Set the global notifyRunner runner function that can be used from the workflow interceptors.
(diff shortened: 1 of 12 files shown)


packages/worker/src/debug-replayer/worker-thread.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Worker thread entrypoint used by ./index.ts to synchronously make an HTTP call to the "runner".
```
lorensr, line 2: Why a thread? I think a `docs/vs-code-debugger.md` architecture description would be helpful

lorensr (approved):

packages/worker/src/debug-replayer/worker-thread.ts
bergundy: Definitely agree a diagram will help here.
There's another comment saying why we use a thread.

packages/worker/src/debug-replayer/worker-thread.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+import { isMainThread, parentPort as parentPortOrNull } from 'node:worker_threads';
+import { Client } from './client';
+
+export interface Request {
+  type: 'wft-started';
```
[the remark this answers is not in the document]
bergundy: I can add more comments in this context.

packages/worker/src/debug-replayer/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+  const thread = new worker_threads.Worker(require.resolve('./worker-thread'));
+
+  const rejectedPromise = new Promise<void>((_, reject) => {
+    // Set the global notifyRunner runner function that can be used from the workflow interceptors.
+    // The function makes an HTTP request to the runner in a separate thread and blocks the current thread until a
+    // response is received.
+    (globalThis as any).notifyRunner = (wftStartEventId: number) => {
+      const sab = new SharedArrayBuffer(4);
```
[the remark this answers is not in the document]
bergundy: int32 is 4 bytes.

packages/worker/src/worker.ts
```diff
@@ -432,6 +432,7 @@ export class Worker {
     workflowBundle: WorkflowBundleWithSourceMapAndFilename,
     compiledOptions: CompiledWorkerOptions
   ): Promise<WorkflowCreator> {
+    // TODO: this isn't required, consider running everything in the threaded vm workflow runtime
```
mjameswh, line 435: I had came up to the same conclusion. We only need `isolateExecutionTimeout` to be set to some high values...

... except if we want to support debugging using Chrome's Dev Tools. Apparently, Chrome Dev Tools doesn't support debugging nodejs threads. But there are some other draw bakcs to Dev Tools. I don't think we should keep that debug target in mind.

packages/worker/src/debug-replayer/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  await Promise.race([
+    rejectedPromise,
+    Worker.runReplayHistory(
+      {
+        ...options,
+        interceptors: {
+          ...options.interceptors,
+          workflowModules: [
+            // Inbound goes first so user can set breakpoints in own inbound interceptors.
+            require.resolve('./debug-inbound-interceptor'),
```
mjameswh, line 46: Filename does match that of commited files... => `inbound-interceptor.ts`?

packages/worker/src/debug-replayer/index.ts
```diff
@@ -0,0 +1,75 @@
...
+            ...(options.interceptors?.workflowModules ?? []),
+            // Outbound goes last - notifies the runner in the finally block, user-provided outbound interceptors are
+            // resumed after the interceptor methods resolve.
+            require.resolve('./debug-outbound-interceptor'),
```
mjameswh, line 50: Same => `./outbound-interceptor.ts`

packages/worker/src/debug-replayer/workflow-notifier.ts
```diff
@@ -0,0 +1,40 @@
...
+    return this._instance;
+  }
+
+  private constructor() {}
+
+  lastNotifiedStartEvent = -1;
+
+  notifyRunner(): void {
+    const eventId = workflowInfo().historyLength;
+    if (this.lastNotifiedStartEvent >= eventId) return;
```
mjameswh, line 22: Why this condition? Do you mean to avoid notifying multiple times during per workflow task?

Assuming you are, why call `notifyRunner()` from outbound-interceptor?

packages/worker/src/debug-replayer/worker-thread.ts
```diff
@@ -0,0 +1,45 @@
+/**
```
mjameswh, line 1: Not a big deal, but I'd rename this to workflow-debugging-worker-thread.ts or something else that clearly set it appart from workflow-worker-thread.ts.

packages/worker/src/debug-replayer/index.ts
```diff
@@ -0,0 +1,75 @@
...
+    throw new Error('Missing TEMPORAL_DEBUGGER_PLUGIN_URL environment variable');
+  }
+  const client = new Client({ baseUrl });
+  const response = await client.get('history');
+  const history = temporal.api.history.v1.History.decode(
+    await Client.readAll(response),
+    Client.contentLength(response)
+  );
+
+  const thread = new worker_threads.Worker(require.resolve('./worker-thread'));
```
mjameswh, line 19: Not sure about the intended execution model of the debug-replayer, but there might be a risk of leaking threads here...

packages/worker/src/debug-replayer/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  const rejectedPromise = new Promise<void>((_, reject) => {
+    // Set the global notifyRunner runner function that can be used from the workflow interceptors.
+    // The function makes an HTTP request to the runner in a separate thread and blocks the current thread until a
+    // response is received.
+    (globalThis as any).notifyRunner = (wftStartEventId: number) => {
+      const sab = new SharedArrayBuffer(4);
+      const responseBuffer = new Int32Array(sab);
+      thread.postMessage({ eventId: wftStartEventId, responseBuffer });
+      Atomics.wait(responseBuffer, 0, 0);
```
mjameswh, line 29: I understand that this to be the exact point at which the workflow thread will be waiting while the debugger plugin is setting up breakpoints.. Right?

mjameswh (approved):
SGTM. A few comments, nothing blocking.

packages/worker/src/worker.ts
bergundy: I guess it doesn't hurt but I'm fine keeping this.
I'm relying on this to access a global variable from workflow context.

packages/worker/src/debug-replayer/index.ts
bergundy: Thanks... it worked locally because I had compiled versions of these files before the rename.

packages/worker/src/debug-replayer/workflow-notifier.ts
bergundy: This condition is meant to avoid notifying for every operation.
We call from outbound interceptor after an operation is resolved (activity completes for example), that might be the first event to process in a workflow task.

packages/worker/src/debug-replayer/worker-thread.ts
bergundy: Don't mind changing this.

packages/worker/src/debug-replayer/index.ts
bergundy: It's meant to be running once per process but I definitely get your concern. I'll make the thread global.

packages/worker/src/debug-replayer/index.ts
bergundy: The debugger plugin will pause the execution if a breakpoint is set on this workflow task.

packages/worker/src/debug-replayer/index.ts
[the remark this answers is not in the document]
lorensr: Oops that's the next line

packages/worker/src/debug-replayer/worker-thread.ts
lorensr: because you need to block the main thread?

packages/worker/src/debug-replayer/workflow-notifier.ts
mjameswh: Ok, it's clear now. I had missed the fact that in outgoing, the notification would be sent only _after_ the completion of a command promise. Clever.

PR #854, opened 2022-08-30
fix: Don't trigger conditions for query jobs

bergundy:


mjameswh (approved):

Sushisource (approved):

packages/test/src/integration-tests.ts
```diff
@@ -1301,4 +1301,16 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
+  test('Query does not cause condition to be triggered', async (t) => {
+    const { client } = t.context;
+    const workflowId = uuid4();
+    const handle = await client.start(workflows.queryAndCondition, {
+      taskQueue: 'test',
+      workflowId,
+    });
+    await handle.query(workflows.mutateWorkflowStateQuery);
+    // Worker did not crash
+    t.pass();
```
lorensr, line 1314: Why would it crash? Vs the workflow just completing

PR #855, opened 2022-08-30
feat(client): Enable gRPC keep-alive by default

bergundy:


Sushisource (approved):

packages/client/src/connection.ts
```diff
@@ -79,7 +79,11 @@ export function defaultConnectionOpts(): ConnectionOptionsWithDefaults {
   return {
     address: LOCAL_TARGET,
     credentials: grpc.credentials.createInsecure(),
-    channelArgs: {},
+    channelArgs: {
+      'grpc.keepalive_permit_without_calls': 1,
+      'grpc.keepalive_time_ms': 30_000,
+      'grpc.keepalive_timeout_ms': 15_000,
```
cretz, line 85: So to confirm what this is saying...it's saying "Send a keep alive every 30s and wait 15s for a response and if we don't get it, kill the connection"? That seems reasonable to me.

cretz (approved):

packages/client/src/connection.ts
bergundy: Yes, this is what it is.

PR #858, opened 2022-08-31
feat: Add Webpack rule to auto instrument Workflows for code coverage, add augmentWorkerOptions()

### LINKED ISSUE #852: Unable to get @temporalio/nyc-test-coverage working in jest

Trying to get `@temporalio/nyc-test-coverage` working with our typescript temporal jest tests. 

### Describe the bug

After following the instructions in the [package readme](https://github.com/temporalio/sdk-typescript/blob/main/packages/nyc-test-coverage/README.md), no coverage is generated.

### Minimal Reproduction

I've generated a minimal repro draft PR: https://github.com/temporalio/samples-typescript/pull/192

### Environment/Versions

- OS and processor: Linux x86_64
- Temporal Version: 1.1.0
- using temporal test server

vkarpov15:
## What was changed

I added a `webpackConfigHook()` method that adds a 'istanbul-instrumenter-loader' rule, which will automatically instrument Workflow code for code coverage when it is bundled. That removes the need for the `nyc instrument` step, and also makes this work with Jest (as long as it is using 'babel' `coverageProvider`).

## Why?

The manual instrumentation step is cumbersome and causes confusion, see https://github.com/temporalio/samples-typescript/pull/192 . Plus the manual instrumentation is not an option with Jest.

## Checklist

1. Closes #852

2. How was this tested:

Got a separate PR for samples-typescript that I'll put in shortly.

3. Any docs updates needed?


packages/nyc-test-coverage/src/index.ts
```diff
@@ -1,10 +1,51 @@
...
 
+  /**
+   * Add all necessary coverage-specific logic to Worker config:
+   * interceptors, sinks, and Webpack config hook.
+   */
+
+  augmentWorkerOptions(workerOptions: WorkerOptions): WorkerOptions {
+    if (!workerOptions.workflowsPath) {
+      throw new TypeError('Cannot automatically instrument coverage without specifying `workflowsPath`');
+    }
```
lorensr, line 19: Could catch at compile time with:

```ts
type MakeRequired<Type, Key extends keyof Type> = Type & {
  [Property in Key]-?: Type[Property];
};

workerOptions: MakeRequired<WorkerOptions, 'workflowsPath'>
```

packages/nyc-test-coverage/src/index.ts
```diff
@@ -28,6 +69,26 @@ export class WorkflowCoverage {
     };
   }
 
+  /**
+   * Modify the given Worker config to auto instrument Workflow
+   * code using istanbul-instrumenter-loader
+   */
+
```
lorensr, line 76:
```suggestion
```

packages/nyc-test-coverage/src/index.ts
```diff
@@ -28,6 +69,26 @@ export class WorkflowCoverage {
...
+  webpackConfigHook(workflowsPath: string, config: WebpackConfigHookType): WebpackConfigHookType {
```
lorensr, line 77: Calling this `webpackConfigHook` confuses me with `WorkerOptions.webpackConfigHook`. What about `addInstrumenterRule`?

lorensr (approved):

packages/nyc-test-coverage/src/index.ts
```diff
@@ -1,10 +1,51 @@
...
+    if (!workerOptions.workflowsPath) {
+      throw new TypeError('Cannot automatically instrument coverage without specifying `workflowsPath`');
+    }
+
+    const workflowsPath = workerOptions.workflowsPath;
+
+    // Interceptors
+    workerOptions.interceptors = workerOptions.interceptors || {};
+    workerOptions.interceptors.workflowModules = workerOptions.interceptors.workflowModules || [];
+    workerOptions.interceptors.workflowModules.push(this.interceptorModule);
```
bergundy, line 26: Please don't mutate the original object.

bergundy: (Applies to this entire implementation).

bergundy (commented):
Nice @vkarpov15, apart from the comment about mutation, LGTM.

packages/nyc-test-coverage/src/index.ts
vkarpov15: We'll do this as well. I'd still prefer to keep the runtime check for JS devs.

packages/nyc-test-coverage/src/index.ts
vkarpov15: That's a fair point. I called this `webpackConfigHook` originally because you were supposed to add this method to the Webpack config yourself. But then I needed to add `workflowsPath`. It makes sense to rename now that this method is no longer the same function signature as the `webpackConfigHook` option

vkarpov15:
@bergundy @lorensr I addressed your code review comments, please let me know what you think :+1:

packages/nyc-test-coverage/src/index.ts
```diff
@@ -1,10 +1,52 @@
...
+type WebpackConfigType = ReturnType<NonNullable<BundleOptions['webpackConfigHook']>>;
+
 export class WorkflowCoverage {
   coverageMap = libCoverage.createCoverageMap();
 
+  /**
+   * Add all necessary coverage-specific logic to Worker config:
+   * interceptors, sinks, and Webpack config hook.
+   */
+  augmentWorkerOptions(
```
mjameswh, line 15: Should we also have a similar `augmentBundleOptions()` method to support the `bundleWorkflowCode()` case? I doubt this will be a very common use case, I see a few reasons someone might want to go that way, and it seems trivial to add...

bergundy: +1

fetahokey (commented):
‏‏‎ ‎

PR #861, opened 2022-09-01
chore: In-process verdaccio server

mjameswh:
## What was changed
- Extract docs build out of main build job
- [publish-to-verdacio]: Use an in-process verdaccio server rather than starting a new one using npx
- [init-to-verdacio]: Configure npm to go through verdaccio only for `@temporalio` packages
- Make the whole process more verbose

## Why
- Fixes some observed CI test flakes
- Make CI tests a little bit faster


scripts/registry.js
```diff
@@ -1,88 +1,94 @@
...
-            contents = `Error ${e}`;
-          }
-          reject(new Error(`Timed out waiting for verdaccio - ${contents}`));
-        }, 60 * 1000);
-      });
-    } finally {
-      tail.unwatch();
-    }
+
+    app.listen(4873);
```
bergundy, line 79: Is there a way to wait for the server to listen?

scripts/utils.js
```diff
@@ -41,31 +41,18 @@ async function kill(child, signal = 'SIGINT') {
...
-
 function spawnNpxSync(args, opts) {
-  let cmd = 'npx';
+  let fullCommand = ['npm', 'exec', ...args];
 
   // NPX is a .cmd on Windows
   if (process.platform == 'win32') {
-    args = ['/C', cmd, ...args];
-    cmd = 'cmd';
+    fullCommand = ['cmd', '/C', fullCommand];
```
bergundy, line 49:
```suggestion
    fullCommand = ['cmd', '/C', ...fullCommand];
```

scripts/registry.js
mjameswh: I think this should do it. `app` is actually an instance of nodejs' `http`.

```
await new Promise((resolve, reject) => {
    try {
        app.listen(4873, resolve)
    } catch (e) {
        reject(e)
    }
})
```

bergundy: Yeah, that's what I usually do, you might need to add `app.on('error', reject)`.

bergundy:
We might need to wait on the server to listen here, see the CI logs: https://github.com/temporalio/sdk-typescript/runs/8139539160?check_suite_focus=true#step:6:15

packages/core-bridge/common.js
```diff
@@ -29,7 +29,7 @@ class PrebuildError extends Error {
   }
 }
 
-function getPrebuiltPath() {
+function getPrebuiltTargetName() {
```
bergundy, line 32: Thank you for this change! Much better!

scripts/publish-to-verdaccio.js
```diff
@@ -1,17 +1,17 @@
...
-      {
+    try {
+      await spawnNpx(['lerna', 'publish', 'from-package', '--yes', '--registry', 'http://localhost:4873/'], {
         stdio: 'inherit',
-      }
-    );
-    if (status !== 0) {
-      console.error(error);
+        stdout: 'inherit',
+        stderr: 'inherit',
```
bergundy, line 11: Yes please :)

bergundy (approved):

scripts/init-from-verdaccio.js
```diff
@@ -1,28 +1,36 @@
...
+      await spawnNpx(
+        ['@temporalio/create', 'example', '--no-git-init', '--temporalio-version', 'latest', ...initArgs],
+        {
+          stdio: 'inherit',
+          stdout: 'inherit',
+          stderr: 'inherit',
+          cwd: registryDir,
+          env: {
+            ...process.env,
+            NPM_CONFIG_USERCONFIG: npmConfigFile,
```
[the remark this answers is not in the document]
mjameswh: The environment variable means that _every_ packages must be requested through Verdaccio. The custom config file makes it possible to only request `@temporalio` packages through Verdaccio.

Requesting a package through Verdaccio is significantly slower than requesting it from NPM... On my local machine, I measure ~12 seconds drop on `init-from-verdaccio` time, when I enable this specific change.

PR #865, opened 2022-09-12
feat!: Bunch of minor features and fixes

### LINKED ISSUE #834: [Bug] test-npm-init rate limited in CI
This PR did not fix the rate limit failure:

https://github.com/temporalio/sdk-typescript/pull/721

I can't tell why. If we put a Cloudflare domain in front of `api.github.com`, we can switch both create and test packages to that domain (related: #782).

### LINKED ISSUE #844: [Bug] Remove `@types/long` stub types definition from `@temporalio/client`
## Describe the bug

Deprecated `@types/long` is listed in https://github.com/temporalio/sdk-typescript/blob/main/packages/client/package.json#L21

Apart from the fact that `@types/*` packages generally should be moved to `devDependencies` to not end up being installed for the consumer, it can be removed since v5 of `long`.

This issue started popping up when upgrading `@temporalio/client` from 1.0.1 to 1.1.0

### Environment/Versions

- OS and processor: Manjaro Linux
- Temporal Version: 1.1.0

bergundy:
- Updated ci to use sdk-ci namespace for testing with cloud.
- Add health service and generate testservice and health in proto package
- Use ephemeral server from Core (supports both time skipping and temporalite)
- Test server is now only downloaded on first use
- Removed some unused dependencies
- Refactored core bridge into multiple files
- Closes #834 
- Closes #844
- Breaking change for the testing package in some of the more advanced and rarely used options.
  - No longer accepting `runInNormalTime` when waiting for workflow result
  - `TestWorkflowEnvironmentOptions` is completely redone


bergundy:
Stress tests failed with latest version of temporalite, so at least we know the problem isn't server specific.
I still can't reproduce this locally, will keep trying.

bergundy:
This shouldn't be merged yet, I'll probably wait for https://github.com/temporalio/sdk-core/pull/389 to be merged so I can remove the testing dependencies that cause `packages/testing/node_modules` to be created and break CI with node 14.

bergundy:
Ended up removing the entire download script for the test server, we'll download when starting the test environment instead of on installation using the soon-to-be-merged Rust downloader in a follow up PR.

lorensr (approved):

bergundy:
I'm going to need to disable the test server related tests before merging this.

packages/test/src/test-ephemeral-server.ts
```diff
@@ -0,0 +1,57 @@
...
+test.todo('TestEnvironment sets up test server with extra args');
+test.todo('TestEnvironment sets up test server with specified port');
+test.todo('TestEnvironment sets up test server with latest version');
+test.todo('TestEnvironment sets up test server from executable path');
+
+test.todo('TestEnvironment sets up temporalite with extra args');
+test.todo('TestEnvironment sets up temporalite with latest version');
+test.todo('TestEnvironment sets up temporalite from executable path');
+test.todo('TestEnvironment sets up temporalite with custom log level');
+test.todo('TestEnvironment sets up temporalite with custom namespace, IP, db filename, and UI');
```
bergundy, line 57: I'll do these in a followup PR, don't want to block anymore

packages/worker/src/runtime.ts
```diff
@@ -283,7 +282,7 @@ export class Runtime {
    * @hidden
    */
   public async closeNativeClient(client: native.Client): Promise<void> {
-    this.registeredClients.delete(client);
+    this.backRefs.delete(client);
```
mjameswh, line 285: Not sure if there is actually a risk here, but removing `client` from `backrefs` before making the call to `clientClose` means that there is no protection against tokio runtime destruction while in the `clientClose()` function (line 326). Compare this to line 317 and 463.

packages/worker/src/runtime.ts
```diff
@@ -435,4 +434,35 @@ export class Runtime {
...
     }
   }
+
+  /**
+   * Create an ephemeral Temporal server.
+   *
+   * Hidden since it is meant to be used internally by the testing framework.
+   * @hidden
+   */
+  public async createEphemeralServer(options: native.EphemeralServerConfig): Promise<native.EphemeralServer> {
```
mjameswh, line 444: [Minor/Style] This should be moved closer to other create/close resources methods higher in the file. They share some important semantic (`pendingCalls`, `backRefs`) which would be easier to maintain if all these are gathered.

packages/testing/package.json
```diff
@@ -38,7 +27,6 @@
   "files": [
     "lib",
     "src",
-    "generated-protos",
     "scripts"
```
mjameswh, line 30: Can remove the `"scrips"` directory too, it doesn't exists anymore.

bergundy: True, thanks

packages/worker/src/runtime.ts
bergundy: I'm refactoring this whole class

packages/worker/src/runtime.ts
bergundy: There's no risk because it's sync JS and therefore atomic, but I'll change this for consistency

mjameswh (approved):

PR #871, opened 2022-09-17
docs: Improve API reference

lorensr:


packages/activity/src/index.ts
```diff
@@ -15,23 +15,37 @@
...
+ *
  * 1. await on {@link Context.cancelled | `Context.current().cancelled`} or
  *    {@link Context.sleep | `Context.current().sleep()`}, which each throw a {@link CancelledFailure}.
  * 1. Pass the context's {@link https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal | `AbortSignal`} at
  *    {@link Context.cancellationSignal | `Context.current().cancellationSignal`} to a library that supports it.
  *
  * ### Examples
  *
- * #### An Activity that sends progress heartbeats and can be cancelled
+ * #### An Activity that sends progress heartbeats and can be Cancelled
```
bergundy, line 48: Why?

packages/activity/src/index.ts
```diff
@@ -15,23 +15,37 @@
...
+ * - The Activity was manually Cancelled (it was called inside a {@link CancellationScope} that was Cancelled via
+ *   {@link CancellationScope.cancel}).
+ * - The Worker has started to shut down. Shutdown is initiated by either:
+ *   - One of the {@link RuntimeOptions.shutdownSignals} was sent to the process.
+ *   - {@link Worker.shutdown | `Worker.shutdown()`} was called.
+ * - The Activity was considered failed by the Server because the Server didn't receive a heartbeat within the
+ *   {@link ActivityOptions.heartbeatTimeout}. The {@link CancellationFailure} will have `message: 'TIMED_OUT'`.
+ * - The Workflow was terminated, in which case the {@link CancellationFailure} will have `message: 'NOT_FOUND'`.
+ *
+ * There are two ways to handle Activity Cancellation:
```
bergundy, line 39: Why is Cancellation capitalized?

packages/activity/src/index.ts
```diff
@@ -15,23 +15,37 @@
...
+ * - The Workflow was Cancelled (via {@link WorkflowHandle.cancel}) and {@link ActivityOptions.cancellationType} was
+ *   **not** set to {@link ActivityCancellationType.ABANDON}.
+ * - The Activity was manually Cancelled (it was called inside a {@link CancellationScope} that was Cancelled via
+ *   {@link CancellationScope.cancel}).
+ * - The Worker has started to shut down. Shutdown is initiated by either:
+ *   - One of the {@link RuntimeOptions.shutdownSignals} was sent to the process.
+ *   - {@link Worker.shutdown | `Worker.shutdown()`} was called.
+ * - The Activity was considered failed by the Server because the Server didn't receive a heartbeat within the
+ *   {@link ActivityOptions.heartbeatTimeout}. The {@link CancellationFailure} will have `message: 'TIMED_OUT'`.
+ * - The Workflow was terminated, in which case the {@link CancellationFailure} will have `message: 'NOT_FOUND'`.
```
bergundy, line 37: This can be generalized to say the workflow *run* has reached a "final" state completed, terminated, failed, and so forth.

packages/activity/src/index.ts
```diff
@@ -15,23 +15,37 @@
...
+ * Activities can only receive Cancellation if they {@link Context.heartbeat | emit heartbeats} or are Local Activities
  * (which can't heartbeat but receive Cancellation anyway).
  *
- * There are two ways to handle Activity cancellation:
+ * An Activity may receive Cancellation if:
+ *
+ * - The Workflow was Cancelled (via {@link WorkflowHandle.cancel}) and {@link ActivityOptions.cancellationType} was
+ *   **not** set to {@link ActivityCancellationType.ABANDON}.
+ * - The Activity was manually Cancelled (it was called inside a {@link CancellationScope} that was Cancelled via
+ *   {@link CancellationScope.cancel}).
```
bergundy, line 31: I'd just say the workflow scope containing the activity was requested to be cancelled (either the root one - e.g. the entire workflow via a handle or a manually created scope).
ABANDON controls whether or not cancellation is sent as an explicit command from the workflow.

packages/activity/src/index.ts
bergundy: We might also cancel an activity if it sends heartbeat details that cannot be converter by the worker's data converter.

bergundy: I'd also mention that the cancel reason is provided in the CancelledFailure thrown in the activity and is accessible via the abort signal (don't remember exactly how to do the latter).

packages/client/src/types.ts
```diff
@@ -34,6 +34,9 @@ export interface WorkflowExecutionDescription {
   runId: string;
   taskQueue: string;
   status: { code: proto.temporal.api.enums.v1.WorkflowExecutionStatus; name: WorkflowExecutionStatusName };
+  /**
+   * This can safely be converted to a number with `.toNumber()`, as the maximum value is 50,000.
+   */
   historyLength: Long;
```
bergundy, line 40: This is stupid, let's just fix this, it's a bug, should've been a number in the first place.

bergundy (approved):
I'd just fix the comment about activity cancellation but otherwise LGTM

packages/activity/src/index.ts
lorensr: https://github.com/temporalio/documentation/blob/master/STYLE.md#capitalization-of-core-terms

PR #875, opened 2022-09-20
chore: Add v1.3.0 to CHANGELOG

lorensr:
[rendered](https://github.com/temporalio/sdk-typescript/blob/release-v1.3.0/CHANGELOG.md#130---2022-09-20)


bergundy (approved):
I’d maybe show an example of how to use the new client and say that we recommend using it instead of the more specific clients

PR #878, opened 2022-09-20
feat: Delete search attributes with empty array values in describe() response

bergundy:
As of 1.18 the server deletes search attributes with empty array values, mimic that behavior when getting the `describe` response from older servers and update CI to use 1.18.

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -877,10 +877,14 @@ export class WorkflowClient {
           executionTime: optionalTsToDate(raw.workflowExecutionInfo!.executionTime),
           closeTime: optionalTsToDate(raw.workflowExecutionInfo!.closeTime),
           memo: await decodeMapFromPayloads(this.client.dataConverter, raw.workflowExecutionInfo!.memo?.fields),
-          searchAttributes: mapFromPayloads(
-            searchAttributePayloadConverter,
-            raw.workflowExecutionInfo!.searchAttributes?.indexedFields ?? {}
-          ) as SearchAttributes,
+          searchAttributes: Object.fromEntries(
+            Object.entries(
+              mapFromPayloads(
+                searchAttributePayloadConverter,
+                raw.workflowExecutionInfo!.searchAttributes?.indexedFields ?? {}
+              ) as SearchAttributes
+            ).filter(([_, v]) => v && v.length > 0) // Filter out empty arrays returned by pre 1.18 servers
+          ),
           parentExecution: raw.workflowExecutionInfo?.parentExecution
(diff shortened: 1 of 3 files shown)


packages/test/src/integration-tests.ts
```diff
@@ -668,7 +668,6 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     const { BinaryChecksums, ...rest } = searchAttributes;
     t.deepEqual(rest, {
       CustomBoolField: [true],
-      CustomIntField: [], // clear
```
lorensr, line 671: This makes the client describe inconsistent with `workflowInfo().searchAttributes`, which still has empty array?

packages/client/src/workflow-client.ts
```diff
@@ -877,10 +877,14 @@ export class WorkflowClient {
...
-            raw.workflowExecutionInfo!.searchAttributes?.indexedFields ?? {}
-          ) as SearchAttributes,
+          searchAttributes: Object.fromEntries(
+            Object.entries(
+              mapFromPayloads(
+                searchAttributePayloadConverter,
+                raw.workflowExecutionInfo!.searchAttributes?.indexedFields ?? {}
+              ) as SearchAttributes
+            ).filter(([_, v]) => v && v.length > 0) // Filter out empty arrays returned by pre 1.18 servers
+          ),
```
lorensr, line 887: This breaks clients that expect empty array.

If we leave as is / don't break our side, then people can see it noted as a breaking change in server 1.18 before they upgrade?

bergundy: I'd categorize this as a bug fix, the new behavior is expected.

packages/test/src/integration-tests.ts
bergundy: I considered this but wasn't sure that's what we'd want since if you set an empty array in the workflow you'd probably expect to see an empty array when you read it back.
We need to do some more changes for search attributes to support standard persistence, let's tackle this when we do that batch of changes.

packages/client/src/workflow-client.ts
lorensr: I'd say it's nice/preferable, but not expected, given past behavior and `workflowInfo()`. Would at least add a bang to title

lorensr (approved):

packages/client/src/workflow-client.ts
bergundy: Still, if users upgrade the server to 1.18 this is what they'll see.
I'd rather keep the SDK behavior consistent regardless of server version.
I wouldn't consider this breaking.

PR #879, opened 2022-09-21
feat(client)!: Make client gRPC retry more configurable

bergundy:
Also backs off more on RESOURCE_EXHAUSTED errors (https://github.com/temporalio/sdk-features/issues/96)

Technically a breaking change but affects only really advanced options.

### DIFF
--- a/packages/client/src/grpc-retry.ts
+++ b/packages/client/src/grpc-retry.ts
@@ -1,37 +1,85 @@
-import {
-  InterceptingCall,
-  Interceptor,
-  ListenerBuilder,
-  Metadata,
-  RequesterBuilder,
-  StatusObject,
-} from '@grpc/grpc-js';
+import { InterceptingCall, Interceptor, ListenerBuilder, RequesterBuilder, StatusObject } from '@grpc/grpc-js';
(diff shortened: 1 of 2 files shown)


packages/client/src/grpc-retry.ts
```diff
@@ -1,37 +1,85 @@
...
 }
 
+/**
+ * Generates the default retry behavior based on given backoff options
+ */
+export function defaultGrpcRetryOptions(options: Partial<BackoffOptions> = {}): GrpcRetryOptions {
+  const { maxAttempts, factor, maxJitter, initialIntervalMs } = withDefaultBackoffOptions(options);
+  return {
+    delayFunction(attempt, status) {
+      return factor ** attempt * initialIntervalMs(status) * jitter(maxJitter);
```
cretz, line 72: I wish it were this easy in Go. Neither in the gRPC interceptor nor in the poll throttler are you allowed to differentiate error types. This is all hidden, so we have to copy/paste a bunch of third party code or have a bad algorithm. Not sure which will be chosen yet.

cretz (approved):

PR #882, opened 2022-09-21
feat(workflow): Add `workflowInfo().unsafe.now()`

### LINKED ISSUE #796: Add WorkflowInfo.unsafe.now()

I'm trying to write a query that will return the time left on a timer.

### Describe the bug

If a query depends on `Date.now()`, it will calculate `Date.now()` once and not update it it again on successive queries.

For example:
```
setHandler(timeLeftMsQuery, () => timer.getDeadline() - Date.now());
```

If I query this, I keep getting the same number back even though I would expect to keep getting smaller numbers.

### Minimal Reproduction

```
async function wf() {
  setHandler(timeLeftMsQuery, () => Date.now());
  await sleep('5 hours');
}

const timeLeftMsQuery = defineQuery<number>('timeLeftMs');
```

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]

lorensr:
- Fixes #796


packages/worker/src/worker.ts
```diff
@@ -1116,6 +1116,7 @@ export class Worker {
                         cronScheduleToScheduleInterval: optionalTsToMs(cronScheduleToScheduleInterval) || undefined,
                         historyLength: activation.historyLength,
                         unsafe: {
+                          now: () => -1, // set in initRuntime
```
lorensr, line 1119: Alternatively, I could make a new type that omits this field?

mjameswh: Any reason not to set it to `Date.now` here instead of doing it in `initRuntime`?

lorensr: I didn't try, but it looked like this is outside the worker thread and a that function wouldn't make it through serialization https://nodejs.org/api/worker_threads.html#portpostmessagevalue-transferlist

packages/test/src/test-workflows.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
bergundy, line 1: We should probably delete most of this file, it doesn't bring that much value

packages/test/src/integration-tests.ts
```diff
@@ -708,7 +708,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       workflowType: 'returnWorkflowInfo',
       workflowId,
       historyLength: 3,
-      unsafe: { isReplaying: false },
+      unsafe: { isReplaying: false, now: Date.now },
```
bergundy, line 711: `Date.now` is a function

packages/test/src/workflows/unsafe-now.ts
```diff
@@ -0,0 +1,10 @@
+import { workflowInfo } from '@temporalio/workflow';
+
+export async function unsafeNow(): Promise<number> {
+  const start = workflowInfo().unsafe.now();
+  for (let i = 0; i < 100000000; i++) {
+    // empty
+  }
```
bergundy, line 7: Let's replace with a local activity instead of this tight loop

packages/worker/src/worker.ts
bergundy: You don't need this here

lorensr: @bergundy Build fails without it. Do you prefer my type suggestion?

packages/test/src/integration-tests.ts
bergundy, line 711: How does this work? the function isn't going to be serialized

packages/test/src/test-workflows.ts
```diff
@@ -1830,3 +1825,11 @@ test('scopeCancelledWhileWaitingOnExternalWorkflowCancellation', async (t) => {
     );
   }
 });
+
+test('unsafeNow', async (t) => {
```
bergundy, line 1829: Just forget about the unit test, there's not much value in it

packages/worker/src/worker.ts
bergundy: I'd just make it `Date.now()` just in case anyone tries to access this in their sinks

packages/test/src/integration-tests.ts
lorensr: I don't imagine this test passes, will update

packages/workflow/src/interfaces.ts
```diff
@@ -135,6 +135,11 @@ export interface WorkflowInfo {
  * Never rely on this information in Workflow logic as it will cause non-deterministic behavior.
  */
 export interface UnsafeWorkflowInfo {
+  /**
+   * Current system time in milliseconds
```
bergundy, line 139: I'd explain what the difference between this and `Date.now()` is.
Also wondering if we just want to expose the raw `Date` object instead of just `now()`.

bergundy (approved):
Overall lgtm, just some minor comments

packages/workflow/src/interfaces.ts
lorensr: Adding: The safe version of time is `new Date()` and `Date.now()`, which are set on the first invocation of a Workflow Task

> Also wondering if we just want to expose the raw Date object instead of just now().

`new workflowInfo().unsafe.Date()` and `workflowInfo().unsafe.Date.now()`

I like the simplicity of just `unsafe.now`, but unsure whether the Date flexiblity is important / if there are things people would need Date for that they can't do `new Date(workflowInfo().unsafe.now)`

mjameswh: >Also wondering if we just want to expose the raw Date object instead of just now().

I doubt there is a need for this.

However, if we _really_ want to go there, I think I'd prefer a global `UnsafeDate` class rather than to have something like `new workflowInfo().unsafe.Date()`.

bergundy: Agree, let's keep just `now()`

packages/worker/src/worker.ts
bergundy: You can clear before transferring to the worker thread and recreate it there or in the workflow context.

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -180,6 +180,7 @@ export class VMWorkflowThreadProxy implements Workflow {
     workerThreadClient: WorkerThreadClient,
     options: WorkflowCreateOptions
   ): Promise<VMWorkflowThreadProxy> {
+    delete (options.info.unsafe as any).now;
```
lorensr, line 183: @bergundy how's this?

packages/test/src/integration-tests.ts
```diff
@@ -708,7 +709,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       workflowType: 'returnWorkflowInfo',
       workflowId,
       historyLength: 3,
-      unsafe: { isReplaying: false },
+      unsafe: { isReplaying: false } as UnsafeWorkflowInfo,
```
bergundy, line 712: Please leave a comment saying why the case is required.

packages/worker/src/workflow/threaded-vm.ts
bergundy: 👍 just please add a comment explaining the cast

packages/workflow/src/interfaces.ts
```diff
@@ -135,6 +135,14 @@ export interface WorkflowInfo {
...
+   *
+   * The safe version of time is `new Date()` and `Date.now()`, which are set on the first invocation of a Workflow
+   * Task.
+   */
+  now: () => number;
```
bergundy, line 144: nit:
```suggestion
  now(): number;
```

bergundy (approved):
Had a few minor comments, LGTM

PR #893, opened 2022-09-28
chore: Deprecate internal functions that should have never been exported

bergundy:


packages/common/src/converter/protobuf-payload-converters.ts
```diff
@@ -1,21 +1,22 @@
...
 } from './payload-converter';
-import { Payload } from '../interfaces';
 
-import { EncodingType, encodingTypes, METADATA_ENCODING_KEY, METADATA_MESSAGE_TYPE_KEY, str, u8 } from './types';
+import { encodingTypes, METADATA_ENCODING_KEY, METADATA_MESSAGE_TYPE_KEY } from './types';
 
 abstract class ProtobufPayloadConverter implements PayloadConverterWithEncoding {
   protected readonly root: Root | undefined;
-  public abstract encodingType: EncodingType;
+  public abstract encodingType: string;
```
lorensr, line 19: so that users can use their own encoding type?

lorensr (approved):

packages/common/src/converter/protobuf-payload-converters.ts
bergundy: Yes, this type should have never existed, we intentionally don't limit the types of encoding.

PR #896, opened 2022-09-29
fix: Export and hide `errorMessage` and `errorCode`

lorensr:
These used to be re-exported from `@temporalio/common`, but were moved in: https://github.com/temporalio/sdk-typescript/pull/881/files#diff-b3f127f417ba115081f7791c1ab78f6a143a3287346c90fdfacc22ddf281ebed

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -3,23 +3,22 @@
  *
  * @module
  */
-export { Headers, Next } from './interceptors';
-
 export * from './activity-options';
 export * from './converter/data-converter';
 export * from './converter/failure-converter';
 export * from './converter/payload-codec';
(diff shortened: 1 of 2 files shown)


bergundy (changes requested):
These are internal utilities that should not be exported

lorensr:
Alternatively, we could update the CHANGELOG to say it's a breaking bugfix?

bergundy:
I guess we can wrap them and export them as deprecated and hidden like we did with some of the other internals

bergundy:
@lorensr I think we can close this.
If you feel it's important enough, let's wrap these functions with deprecated, internal wrappers.

lorensr:
Closing in favor of #897

PR #899, opened 2022-10-01
Improve regex for extracting source map

bergundy:
User complained in slack that the previous regex was causing `RangeError: Maximum call stack size exceeded`.

I tried a couple alternatives, the chosen one and splitting the lines first, ran 10,000 iterations and got these results:

- Split first: 12.371s
- New regex: 4.797s
- Original regex: 36.535s

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -291,7 +291,7 @@ jobs:
 
   # Runs the sdk features repo tests with this repo's current SDK code
   sdk-features-tests:
-    uses: temporalio/sdk-features/.github/workflows/typescript.yaml@protoc
+    uses: temporalio/sdk-features/.github/workflows/typescript.yaml@main
     with:
       typescript-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       typescript-repo-ref: ${{github.event.pull_request.head.ref}}
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -1659,7 +1659,7 @@ export interface WorkflowBundleWithSourceMapAndFilename {
 }
 
 export function parseWorkflowCode(code: string, codePath?: string): WorkflowBundleWithSourceMapAndFilename {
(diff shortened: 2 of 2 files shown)


lorensr (approved):

mjameswh:
@bergundy Here are some interesting results I got. To make the test easier to reproduce, I used ./node_modules/@verdaccio/ui-theme/static/vendors.178789ed84145d39f4e7.js, which is a 1.5 mb file containing inlined sourcemap, with 10000 iterations.

### 1. Original regex
```
/\s*\n[/][/][#]\s+sourceMappingURL=data:(?:[^,]*;)base64,([0-9A-Za-z+/=]+)\s*$/;
```
84107 ms

### 2. Your suggestion
```
/^\/\/#\s+sourceMappingURL=data:(?:[^,]*;)base64,([0-9A-Za-z+/=]+)$/m;
```
3313 ms

### 3. Original with the initial `\s*` removed and some character classes replaced by escaped character literal
```
/\n\/\/\#\ssourceMappingURL=data:(?:[^,]*;)base64,([0-9A-Za-z+/=]+)\s*$/;
```
2634 ms

### 4. Use `lastIndexOf` to locate the `sourceMappingURL=data` substring, then regex to split the rest
```
	let index = source.lastIndexOf('//# sourceMappingURL=data:');
	matcher = (source.substring(index - 1)).match(/^[\r\n]\/\/\#\ sourceMappingURL=data:(?:[^,]*;)base64,([0-9A-Za-z+/=]+)\s*$/);
```
247 ms

I'm a big fan of regex, but sometime, like with any declarative language, we get disappointed by the inability of a specific engine to properly optimise a specific use case... Definitely true in this case.

lorensr:
Ooh nice! 🙌😊

mjameswh:
This one gets it down to 87 ms for 10000 iterations:

```
	let index = verdaccioSrc.lastIndexOf('//# sourceMappingURL=data:');
	let secondIndex = verdaccioSrc.indexOf('base64,', index);
	let base64 = verdaccioSrc.substring(secondIndex + 'base64,'.length);
```

Look mom! No regex! 🤣

bergundy:
Awesome! Want to update the PR?

PR #902, opened 2022-10-06
fix(client): Handle test server absent history when waiting for workflow result

bergundy:
Tested manually but this fixes an issue reported on slack where the client throws when the test server returns a result with absent history.


lorensr (approved):
It's normal/valid for test server to have no history?

bergundy:
It happens if the workflow doesn't complete in 20 seconds. Since all fields are optional in proto this is fine.

PR #906, opened 2022-10-06
fix(client): Fix default gRPC keepalive settings not applied

bergundy:


lorensr (approved):

packages/client/src/connection.ts
```diff
@@ -75,18 +85,21 @@ export type ConnectionOptionsWithDefaults = Required<Omit<ConnectionOptions, 'tl
...
     address: LOCAL_TARGET,
     credentials: grpc.credentials.createInsecure(),
     channelArgs: {
       'grpc.keepalive_permit_without_calls': 1,
       'grpc.keepalive_time_ms': 30_000,
       'grpc.keepalive_timeout_ms': 15_000,
+      ...channelArgs,
     },
-    interceptors: [makeGrpcRetryInterceptor(defaultGrpcRetryOptions())],
+    interceptors: interceptors ?? [makeGrpcRetryInterceptor(defaultGrpcRetryOptions())],
```
mjameswh, line 99: So if users add interceptors, they are expected to themselves add the `grpc retry interceptor`, with proper options, to there interceptors array? Have we documented this anywhere?

lorensr: That's by design and documented in `ConnectionOptions.interceptors`. Added this to help:

https://github.com/temporalio/sdk-typescript/pull/908

PR #910, opened 2022-10-10
fix: Delay checking for __coverage__ until helpers run because __coverage__ is undefined when Jest starts

vkarpov15:
Re: temporalio/temporal-pendulum#21

## What was changed

Don't check for `__coverage__` when the module loads, but when the `augment()` methods run.

## Why?

Jest quirks: `__coverage__` isn't set when the module first loads.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/nyc-test-coverage/src/index.ts
+++ b/packages/nyc-test-coverage/src/index.ts
@@ -5,12 +5,14 @@ import libCoverage from 'istanbul-lib-coverage';
 // Pull `webpack.Configuration` type without needing to import Webpack
 type WebpackConfigType = ReturnType<NonNullable<BundleOptions['webpackConfigHook']>>;
 
-// Check if running through nyc or some other Istanbul-based tool.
-// If not, any `workflowCoverage()` tools are a no-op.
-const hasCoverageGlobal = '__coverage__' in global;
-
 export class WorkflowCoverage {
   coverageMap = libCoverage.createCoverageMap();
+
+  // Check if running through nyc or some other Istanbul-based tool.
+  // If not, any `workflowCoverage()` tools are a no-op.
+  hasCoverageGlobal() {
+    return '__coverage__' in global;
+  }
 
   /**
    * Add all necessary coverage-specific logic to Worker config:
@@ -23,7 +25,7 @@ export class WorkflowCoverage {
       throw new TypeError('Cannot automatically instrument coverage without specifying `workflowsPath`');
     }
 
-    if (!hasCoverageGlobal) {
+    if (!this.hasCoverageGlobal()) {
       return workerOptions;
     }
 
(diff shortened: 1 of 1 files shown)


packages/nyc-test-coverage/src/index.ts
```diff
@@ -5,13 +5,15 @@ import libCoverage from 'istanbul-lib-coverage';
...
-// Check if running through nyc or some other Istanbul-based tool.
-// If not, any `workflowCoverage()` tools are a no-op.
-const hasCoverageGlobal = '__coverage__' in global;
-
 export class WorkflowCoverage {
   coverageMap = libCoverage.createCoverageMap();
 
+  // Check if running through nyc or some other Istanbul-based tool.
+  // If not, any `workflowCoverage()` tools are a no-op.
+  hasCoverageGlobal() {
```
bergundy, line 13: Make this private

packages/nyc-test-coverage/src/index.ts
```diff
@@ -23,7 +25,7 @@ export class WorkflowCoverage {
       throw new TypeError('Cannot automatically instrument coverage without specifying `workflowsPath`');
     }
 
-    if (!hasCoverageGlobal) {
+    if (!this.hasCoverageGlobal()) {
```
bergundy, line 28: Why is this necessary?

vkarpov15: This avoids instrumenting Workflow code if coverage isn't enabled. I think it is beneficial to avoid piping the Workflow code through istanbul-instrumenter-loader if coverage isn't enabled. I don't expect it to cause any problems, but I'd prefer to err on the side of caution there.

bergundy: I'd throw an error here instead then instead of silently failing.

vkarpov15: Do you want to put the responsibility on the user to avoid calling this function if code coverage is disabled?

bergundy: Do you expect this to be a common scenario?

joebowbeer: I think this is very common, IIUC. The user doesn't want to have to rip out their workflowCoverage augmentation code just because they disabled coverage on a specific test run.

bergundy: SGTM, please ignore this comment then.
I would maybe consider issuing a warning to let the user know that coverage is disabled.

vkarpov15: I agree with @joebowbeer . It isn't uncommon to run tests with coverage on vs coverage off. And that means every user would have to implement their own check for whether they're running with coverage or not.

bergundy (approved):

PR #918, opened 2022-10-12
fix(worker): avoid a race in shutdown hooks

### LINKED ISSUE #909: [Bug] Ignored shutdown signal causing stale workers during development

While developing using TS SDK, we run `npm run start.watch`. When we change the code in `src`, we want the worker to always restart. When we interrupt this task with `Ctrl+C`, we want the worker to always stop.

### Describe the bug

1. If we interrupt with `Ctrl+C`, sometimes the worker will continue to run as an orphaned process.
2. Sometimes when we change the code in `src`, the worker ignores it and continues running.

This causes issues during development because old workers with old code are unexpectedly still running.

### Minimal Reproduction

Reproduction 1:
Run `npx @temporalio/create@latest temporal-hello-world  --sample hello-world`
Run `cd temporal-hello-world`
Run `npm run start.watch`
Hit `Ctrl+C` after "[INFO] Creating worker" appears in the logs
Run `ps -a | grep src/worker.js`
Sometimes the worker will continue running as an orphaned process

Video: https://www.loom.com/share/ff77941fd3d14940ba644759b1da0cd6

Reproduction 2:
Run `npx @temporalio/create@latest temporal-hello-world  --sample hello-world`
Run `npm run start.watch`
Change `src/activities.ts` and save (for example `return "Hello 1!"`)
Wait for the "[INFO] Creating worker" message in the logs
Immediately change `src/activities.ts` again and save (for example `return "Hello 2!"`)
Run `npm run workflow`
Sometimes the old activities code will run (`"Hello 1!"`)

It may be easier to reproduce with a larger codebase, etc.

### Environment/Versions

- OS and processor: Intel Mac
- Temporal Version: Temporal Server 1.15.0, TypeScript SDK 1.4.2
- Are you using Docker or Kubernetes or building Temporal from source? Temporal Server in Docker

### Additional context

nodemon: 2.0.20
ts-node: 10.9.1

One cause seems to be related to worker `Runtime` listening to process signals in `setupShutdownHook` without any shutdown signal callbacks registered:

https://github.com/temporalio/sdk-typescript/blob/cb1029537976c37270a61bc9744a566618c55e26/packages/worker/src/runtime.ts#L114-L124

https://github.com/temporalio/sdk-typescript/blob/cb1029537976c37270a61bc9744a566618c55e26/packages/worker/src/runtime.ts#L397-L404

When a listener to a process signal is added, the signal no longer causes the default behavior, so the signal is ignored. The shutdown signal callbacks are only added after the worker is started: 

https://github.com/temporalio/sdk-typescript/blob/cb1029537976c37270a61bc9744a566618c55e26/packages/worker/src/worker.ts#L1597-L1600

If the signal occurs before this, it may be ignored. There could be other issues possibly related to core, nodemon, or ts-node.

Workarounds could be:
- Patch `Runtime` to only add listener to process signal when there is a callback
- Add a listener for process signals that always exits for development only

mjameswh:
## What was changed
In `Runtime`, deny adding more shutdown hooks once a shutdown signal has been received.

## Why?
Avoid a race situation where a process might be sent a shutdown signal, but some worker has not yet registered its shutdown hook, causing the worker to never terminate.

## Checklist
1. Closes #909

### DIFF
--- a/packages/test/src/test-worker-lifecycle.ts
+++ b/packages/test/src/test-worker-lifecycle.ts
@@ -5,14 +5,60 @@
  * @module
  */
 import { Worker } from '@temporalio/worker';
+import { Runtime } from '@temporalio/worker';
 import test from 'ava';
 import { RUN_INTEGRATION_TESTS } from './helpers';
 import { defaultOptions, isolateFreeWorker } from './mock-native-worker';
 
 if (RUN_INTEGRATION_TESTS) {
   test.serial('Worker shuts down gracefully', async (t) => {
-    const worker = await Worker.create({
-      ...defaultOptions,
+    try {
+      const worker = await Worker.create({
+        ...defaultOptions,
+        shutdownGraceTime: '500ms',
+        taskQueue: 'shutdown-test',
+      });
+      t.is(worker.getState(), 'INITIALIZED');
+      const p = worker.run();
+      t.is(worker.getState(), 'RUNNING');
+      process.emit('SIGINT', 'SIGINT');
+      // Shutdown callback is enqueued as a microtask
+      await new Promise((resolve) => process.nextTick(resolve));
+      t.is(worker.getState(), 'STOPPING');
+      await p;
+      t.is(worker.getState(), 'STOPPED');
+      await t.throwsAsync(worker.run(), { message: 'Poller was already started' });
+    } finally {
+      // Make sure Runtime is reset after this test
+      if (Runtime._instance) await Runtime._instance.shutdown();
+      Runtime.instance();
+    }
+  });
+
+  test.serial('Worker shuts down gracefully if interrupted before running', async (t) => {
+    try {
+      const worker = await Worker.create({
+        ...defaultOptions,
+        shutdownGraceTime: '500ms',
+        taskQueue: 'shutdown-test',
+      });
+      t.is(worker.getState(), 'INITIALIZED');
+      process.emit('SIGINT', 'SIGINT');
+      const p = worker.run();
+      t.is(worker.getState(), 'STOPPING');
+      await p;
+      t.is(worker.getState(), 'STOPPED');
+    } finally {
+      // Make sure Runtime is reset after this test
+      if (Runtime._instance) await Runtime._instance.shutdown();
+      Runtime.instance();
+    }
+  });
+}
+
+test.serial('Mocked run shuts down gracefully', async (t) => {
+  try {
+    const worker = isolateFreeWorker({
       shutdownGraceTime: '500ms',
       taskQueue: 'shutdown-test',
     });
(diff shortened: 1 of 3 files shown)


packages/worker/src/runtime.ts
```diff
@@ -380,8 +381,13 @@ export class Runtime {
    * Hidden in the docs because it is only meant to be used internally by the Worker.
    * @hidden
    */
-  public registerShutdownSignalCallback(callback: () => void): void {
-    this.shutdownSignalCallbacks.add(callback);
+  public registerShutdownSignalCallback(callback: () => void): boolean {
```
Sushisource, line 384: Docstring should be updated w/ meaning of return value

Sushisource (approved):
Makes sense to me

packages/test/src/test-worker-lifecycle.ts
```diff
@@ -5,45 +5,98 @@
...
+      // Shutdown callback is enqueued as a microtask
+      await new Promise((resolve) => process.nextTick(resolve));
+      t.is(worker.getState(), 'STOPPING');
+      await p;
+      t.is(worker.getState(), 'STOPPED');
+      await t.throwsAsync(worker.run(), { message: 'Poller was already started' });
+    } finally {
+      // Make sure Runtime is reset after this test
+      if (Runtime._instance) await Runtime._instance.shutdown();
+      Runtime.instance();
```
bergundy, line 34: Why create a new instance in this test?
Shouldn't the runtime automatically be shutdown after this test?

packages/worker/src/worker.ts
```diff
@@ -1596,10 +1596,16 @@ export class Worker {
     }
     this.state = 'RUNNING';
 
-    const shutdownCallback = () => this.shutdown();
-    Runtime.instance().registerShutdownSignalCallback(shutdownCallback);
-
     try {
+      const shutdownCallback = () => this.shutdown();
```
bergundy, line 1600: There's some more teardown happening below that's missed here.

Note that below the worker transitions to "STOPPED" state before it finalizes shutdown.
We could change that so we transition only after finalizeShutdown is called below and reuse the same teardown code.

packages/worker/src/runtime.ts
```diff
@@ -377,11 +378,20 @@ export class Runtime {
...
+   * Returns false if a shutdown has already been requested. In that case, the Worker should
+   * immediately initiate its shutdown.
+   *
    * Hidden in the docs because it is only meant to be used internally by the Worker.
+   * @returns true if callback was registered, or false if shutdown has already been requested
    * @hidden
    */
-  public registerShutdownSignalCallback(callback: () => void): void {
-    this.shutdownSignalCallbacks.add(callback);
+  public registerShutdownSignalCallback(callback: () => void): boolean {
```
bergundy, line 388: What's the benefit of eagerly returning `false` here?

bergundy: It seems like it would save some complexity on the worker side to always use a callback.

packages/test/src/test-worker-lifecycle.ts
mjameswh: Actually, it is required only for the two mock tests. It automatically reset after native tests.

The thing is that the mock native worker doesn't register itself with Runtime, and obviously doesn't unregister either. Consequently, `Runtime.shutdownIfIdle()` doesn't gets called at the end of the mock tests, and therefore, the Runtime instance never get discarded.

mjameswh:
@bergundy Pushed a second attempt, addressing most of your comments.

Note that at this point, I have kept the `Runtime._instance.shutdown()` calls at the end of the two mock workers shutdown tests. Removing these would be possible but would require that `MockNativeWorker::finalizeShutdown` actually do something on `Runtime`, for example calling `shutdownIfIdle()`. I don't feel like this is a good thing.

packages/worker/src/runtime.ts
```diff
@@ -101,6 +101,7 @@ export class Runtime {
   protected readonly logPollPromise: Promise<void>;
   public readonly logger: Logger;
   protected readonly shutdownSignalCallbacks = new Set<() => void>();
+  protected state: 'running' | 'shuttingdown' = 'running';
```
bergundy, line 104: nit:

```suggestion
  protected state: 'RUNNING' | 'SHUTTING_DOWN' = 'RUNNING';
```

bergundy (approved):
LGTM, had a single nit.
Thanks!

PR #920, opened 2022-10-13
Multiple history downloading & replay

Sushisource:
This is more-or-less ready to go. There's some minor cleanup to do, and we should decide if parallel history downloading really matters in this first cut, it's a bit annoying to implement w/o adding many deps to the client.

## What was changed
* Support multiple-history replay
* Add history downloading functionality to workflow handle
* Add a history downloader to the client

## Why?
Make it easy to download & replay lots of histories at once!

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/775

2. How was this tested:
Added tests

3. Any docs updates needed?


packages/common/src/interfaces.ts
```diff
@@ -62,3 +62,11 @@ export type ActivityInterface = Record<string, ActivityFunction>;
  * Mapping of Activity name to function
  */
 export type UntypedActivities = Record<string, ActivityFunction>;
+
+/**
+ * A workflow history with it's ID. Useful for replay.
```
lorensr, line 67:
```suggestion
 * A workflow's history and ID. Useful for replay.
```

packages/common/src/interfaces.ts
```diff
@@ -62,3 +62,11 @@ export type ActivityInterface = Record<string, ActivityFunction>;
...
+ */
+export interface HistoryAndWorkflowID {
+  workflowID: string;
+  history: temporal.api.history.v1.History | unknown;
```
lorensr, line 71: could we get one of these working instead?

```
  history?: temporal.api.history.v1.History;
```

or

```
  history: temporal.api.history.v1.History | undefined;
```

packages/worker/src/worker.ts
```diff
@@ -451,35 +456,67 @@ export class Worker {
...
+   * If `options.failFast` is set, the function will throw upon the first encountered error.
+   * If not, all histories will be replayed and the returned results object will contain information
+   * about any failures.
+   *
+   * @throws {@link DeterminismViolationError} if in fail fast mode and there is some incompatible
+   * history.
+   */
+  public static async runReplayHistories(
+    options: ReplayWorkerOptions,
+    histories: AsyncIterable<HistoryAndWorkflowID>
```
lorensr, line 478: maybe:
```suggestion
    options: ReplayHistoriesWorkerOptions,
```

```
type ReplayHistoriesWorkerOptions = ReplayWorkerOptions & {
  failFast: 
  histories: AsyncIterable<HistoryAndWorkflowID>
```

packages/client/src/types.ts
```diff
@@ -43,6 +43,10 @@ export interface WorkflowExecutionDescription {
   raw: DescribeWorkflowExecutionResponse;
 }
 
+export interface FetchedHistory {
+  history: proto.temporal.api.history.v1.History;
+}
```
lorensr, line 48: might we want more fields on this in future? if not, can we just use `proto.temporal.api.history.v1.History`?

packages/client/src/workflow-client.ts
```diff
@@ -265,6 +272,15 @@ export interface WorkflowResultOptions {
   followRuns?: boolean;
 }
 
+export interface DownloadLazilyOptions {
```
cretz, line 275: Remove interface if unused

packages/client/src/workflow-client.ts
```diff
@@ -138,6 +140,11 @@ export interface WorkflowHandle<T extends Workflow = Workflow> extends BaseWorkf
    */
   describe(): Promise<WorkflowExecutionDescription>;
 
+  /**
+   * Download the entire workflow history
+   */
+  fetchHistory(): Promise<FetchedHistory>;
```
cretz, line 146: Might as well have this include the ID in it. I think we can just call it `History` and have it contain `workflowId: string` and `raw: proto.temporal.api.history.v1.History`.

packages/client/src/workflow-client.ts
```diff
@@ -909,6 +950,25 @@ export class WorkflowClient {
           headers: {},
         }) as Promise<Ret>;
       },
+      async fetchHistory(): Promise<FetchedHistory> {
+        // TODO: Interceptors? Bleh
```
cretz, line 954: Yeah, it's a tough call. We originally only did it on things that might want interceptors, but then I was told that was wrong the other day when we added interceptors for "workflow describe". So now basically the rule seems to be "interceptors for all high-level client calls".

packages/client/src/workflow-client.ts
```diff
@@ -594,6 +610,31 @@ export class WorkflowClient {
     }
   }
 
+  /**
+   * Lazily downloads the histories of the workflows indicated by the passed in iterator. These
+   * histories may be used in conjunction with replay functionality to verify changes to workflow
+   * code.
+   */
+  public downloadLazily(
```
cretz, line 618: If we change `fetchHistory` to return the workflow ID and run ID, I think we can remove this helper from the top-level client. Just accept either an iterable of workflows or an iterable of histories in the replayer and if you get the former, map there.

I understand there may be options later for async fetch, but since there's not now, let's remove this method.

If we don't remove this method, we should at least call it `lazyFetchHistories` or something that makes it clear it's about histories.

packages/common/src/interfaces.ts
cretz, line 71:
```suggestion
export interface WorkflowHistory {
  workflowId: string;
  history: temporal.api.history.v1.History;
```

Suggestions:

* Change `workflowID` to `workflowId`
* Call this `WorkflowHistory` and use it as the public-facing wrapper around history everywhere. Meaning, change `fetchHistory` to return `Promise<WorkflowHistory>` and remove `FetchedHistory` interface
* This interface should never be allowed to have an unset history
* Potentially expose some kind of helpers to serialize to/from JSON (but accept workflow ID on deserialization since the JSON format likely needs to be UI/tctl compatible)

I will probably be doing similar in Python (though I will probably be putting `WorkflowHistory` in the `client` package, not a common area, but the surface area is much smaller on that project).

packages/worker/src/worker.ts
cretz: `runReplayHistory` explicitly chose to separate options from the history provided, so this is consistent with that

packages/worker/src/worker.ts
```diff
@@ -451,35 +456,67 @@ export class Worker {
    * @throws {@link DeterminismViolationError} if the workflow code is not compatible with the history.
    */
   public static async runReplayHistory(options: ReplayWorkerOptions, history: History | unknown): Promise<void> {
```
cretz, line 458: Can this not just be a one-liner to `runReplayHistories`?

packages/client/src/workflow-client.ts
```diff
@@ -265,6 +272,15 @@ export interface WorkflowResultOptions {
   followRuns?: boolean;
 }
 
+export interface DownloadLazilyOptions {
+  /**
+   * Maximum number of workflows that will be fetched concurrently.
+   *
+   * @default 5
```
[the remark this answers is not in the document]
Sushisource: Need some number. 5 is a pleasant one. :shrug:

packages/common/src/interfaces.ts
Sushisource: No, it needs more ambiguity since it can be either the proto or some weird json mess.

packages/client/src/types.ts
Sushisource: Yeah, I think this could potentially end up with more stuff which is why I added it

packages/common/src/interfaces.ts
Sushisource: The two types being separate is desirable b/c I don't want to force users to break out a `fromJson` kinda step separately when I can handle that nicely for them w/o issue.

I have passed through the id attached to `FetchedHistory` though.

packages/client/src/workflow-client.ts
Sushisource: Sounds like we don't need to if there are no headers according to @bergundy and I'll take that answer and run with it

packages/worker/src/runtime.ts
```diff
@@ -63,7 +63,7 @@ export interface CompiledRuntimeOptions {
 
 function defaultTelemetryOptions(): TelemetryOptions {
   return {
-    tracingFilter: 'temporal_sdk_core=WARN',
+    tracingFilter: 'temporal_sdk_core=INFO',
```
cretz, line 66: Is this change by intention?

cretz (approved):

packages/worker/src/runtime.ts
Sushisource: Yes, it is, meant to call attention to that. The context variables are emitted only at info level. So, default log level needs to be up to there or even warning events won't have context in the console output. This is... semi unfortunate but just the way it is. There's very few things emitted at info level so I think it's a reasonable change to default.

packages/client/src/types.ts
```diff
@@ -43,6 +43,11 @@ export interface WorkflowExecutionDescription {
   raw: DescribeWorkflowExecutionResponse;
 }
 
+export interface FetchedHistory {
```
bergundy, line 46: Docstring please

packages/core-bridge/index.d.ts
```diff
@@ -348,10 +356,19 @@ export interface Client {
 export interface EphemeralServer {
   type: 'EphemeralServer';
 }
+export interface HistoryPusher {
+  type: 'HistoryPusher';
+}
+export interface ReplayWorker {
+  type: 'ReplayWorker';
+  worker: Worker;
+  pusher: HistoryPusher;
```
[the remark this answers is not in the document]
bergundy: Not worth blocking this PR but I feel like this approach is needlessly adding complexity to every lang SDK.

packages/test/src/test-replay.ts
```diff
@@ -94,3 +109,59 @@ test('workflow-task-failure-fails-replay', async (t) => {
...
+  const hist2 = await getHist('cancel_fake_progress_history.json');
+  hist1.events[0].workflowExecutionStartedEventAttributes!.workflowType!.name = 'http';
+  hist2.events[0].workflowExecutionStartedEventAttributes!.workflowType!.name = 'http';
+  const hists = historator([hist1, hist2], true);
+
+  await t.throwsAsync(
+    Worker.runReplayHistories(
+      {
+        workflowsPath: require.resolve('./workflows'),
+        replayName: t.title,
```
bergundy, line 142: I thought `replayName` is now optional, if it isn't we should make it optional.

packages/worker/src/replay.ts
```diff
@@ -0,0 +1,10 @@
+export interface ReplayResults {
+  // True if any workflow failed replay
+  readonly hadAnyFailure: boolean;
+  // Maps run id to information about the replay failure
+  readonly failureDetails: Map<string, Error>;
```
bergundy, line 5: Is runId going to be enough to debug the error source? We should probably add the workflowId here too.
I'd change this to:
```
interface WorkflowRunError { workflowId: string; runId: string; error: Error }
readonly failureDetails: Array<WorkflowRunError>
```

packages/worker/src/worker.ts
```diff
@@ -444,42 +470,104 @@ export class Worker {
     }
   }
 
+  /**
+   * Lazily downloads the histories of the workflows indicated by the passed in iterator. These
+   * histories may be used in conjunction with replay functionality to verify changes to workflow
+   * code.
+   */
+  public static fetchHistories(
```
bergundy, line 478: Let's make this private, no reason for the worker to expose this method.

packages/worker/src/worker.ts
```diff
@@ -444,42 +470,104 @@ export class Worker {
...
+   * contain a workflow id, so it must be provided separately if your workflow depends on it.
    * @throws {@link DeterminismViolationError} if the workflow code is not compatible with the history.
    */
-  public static async runReplayHistory(options: ReplayWorkerOptions, history: History | unknown): Promise<void> {
+  public static async runReplayHistory(
+    options: ReplayWorkerOptions,
+    history: History | unknown,
+    workflowId?: string
+  ): Promise<void> {
+    const hist = this.validateHistory(history);
```
bergundy, line 505:
```suggestion
    const validatedHistory = this.validateHistory(history);
```

bergundy (commented):
I haven't finished reviewing yet.
I'll push a minor fix to prevent a memory leak if replayer is slower than concurrent downloader.

packages/core-bridge/index.d.ts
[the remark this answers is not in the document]
Sushisource: No, because then there need to be different worker types. There is no replay worker at the moment, which is nice. Initially I did have two different types and that made other things harder. We can go that direction if more divergent needs crop up.

The amount of complexity it adds is I think pretty minor. It's even easier in pyo3. IMO Sharing a `feed_history` implementation for sync task dispatch wouldn't make much sense because the semantics are entirely different for replay vs. pushing new history.

packages/worker/src/runtime.ts
```diff
@@ -284,13 +286,36 @@ export class Runtime {
...
+  }
+
+  /**
+   * Push history to a replay worker's history pusher stream.
+   *
+   * Hidden in the docs because it is only meant to be used internally by the Worker.
+   *
+   * @hidden
+   */
+  public async pushHistory(pusher: native.HistoryPusher, workflowId: string, history: History): Promise<void> {
```
mjameswh, line 305: This seems a little bit out of place... Why bring this here rather than doing it directly in Worker?

packages/core-bridge/index.d.ts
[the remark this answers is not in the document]
mjameswh: I too feel that there is some unjustified complexity added in this PR, mostly in worker.ts and runtime.ts.

I think what really bothers me is the fact that, in Worker.ts, regular workers and replay workers follow distinct initialization/registration paths, making it much less obvious who owns which resource at which moment, and the fact that we have that `historyPublisher` object that needs to be passed over, in parallel with the corresponding worker instance.

I'd say however that this can be fixed with some small refactoring on the TS side only, without impacting core-bridge. Not blocking IMHO.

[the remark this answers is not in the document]
bergundy: Not sure what you're suggesting @mjameswh but I'm open to hear it, it's not too late.

packages/worker/src/replay.ts
```diff
@@ -3,11 +3,37 @@ import { Client } from '@temporalio/client';
...
+    super(message);
+  }
+}
+
 export interface ReplayResults {
-  /** True if any workflow failed replay */
-  readonly hadAnyFailure: boolean;
   /** Maps run id to information about the replay failure */
-  readonly failureDetails: Map<string, Error>;
+  readonly errors: ReplayError[];
```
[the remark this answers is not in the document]
bergundy: I don't see much value here but I'll add it.

PR #925, opened 2022-10-14
Test #924 from a fork

lorensr:
Deploy docs should be skipped

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,7 +102,8 @@ jobs:
         if: ${{ github.ref == 'refs/heads/main' }}
         run: npx vercel deploy packages/docs/build -t ${{ secrets.VERCEL_TOKEN }} --name typescript --scope temporal --prod --yes
       - name: Deploy draft docs
-        if: ${{ github.ref != 'refs/heads/main' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -102,7 +102,8 @@ jobs:
         if: ${{ github.ref == 'refs/heads/main' }}
         run: npx vercel deploy packages/docs/build -t ${{ secrets.VERCEL_TOKEN }} --name typescript --scope temporal --prod --yes
       - name: Deploy draft docs
-        if: ${{ github.ref != 'refs/heads/main' }}
+        # Don't run on forks, since secrets won't be available, and command will fail
+        if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository && github.ref != 'refs/heads/main' }}
```
bergundy, line 106: Can probably check if the token is available but whatever works

bergundy (approved):

.github/workflows/ci.yml
lorensr: don't have access to secrets in expressions, but could put secret in env var and then check that. either works

PR #932, opened 2022-10-19
fix(workflow): Don't fail workflow task if query handler not found

### LINKED ISSUE #931: [Bug] Queries targeting a nonexistent handler fail WFTs on non-legacy path
### Describe the bug

If you query a workflow with a nonexistent query type, and the query is issued as part of a WFT (not a legacy query), then that causes the SDK to reject the workflow task, where it should instead complete the activation successfully and include a failed query result command.

### Minimal Reproduction

You have to get lucky and have server send the query the nonlegacy way, so inconsistently repros. Would be easy to write a UT for.
In sdk-features:

```typescript
export const finishSignal = wf.defineSignal('finish');
export const query = wf.defineQuery<string>('nonexistent');

export async function workflow(): Promise<void> {
  await new Promise((resolve) => wf.setHandler(finishSignal, () => resolve(null)));
}

export const feature = new Feature({
  workflow,
  checkResult: async (runner, handle) => {

    await assert.rejects(async () => {
      await handle.query(query);
    }, QueryNotRegisteredError);

    await handle.signal(finishSignal);
    await runner.waitForRunResult(handle);
  },
});
```

This miiiiight only happen if the query is part of the first start workflow activation, too. (Something to do with bufferedQueries)

bergundy:
Closes #931

### DIFF
--- a/packages/test/src/test-workflows.ts
+++ b/packages/test/src/test-workflows.ts
@@ -208,20 +208,20 @@ function makeNotifyHasPatchJob(patchId: string): coresdk.workflow_activation.IWo
   };
 }
 
-async function makeQueryWorkflow(
+function makeQueryWorkflow(
   queryId: string,
   queryType: string,
   queryArgs: any[],
   timestamp: number = Date.now()
-): Promise<coresdk.workflow_activation.IWorkflowActivation> {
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
Those sneaky promises

PR #934, opened 2022-10-24
docs: Update publish runbook

lorensr:


CONTRIBUTING.md
```diff
@@ -193,10 +193,24 @@ for f in ~/Downloads/packages-*.zip; do mkdir "$HOME/Downloads/$(basename -s .zi
 ls packages/core-bridge/releases/
 
 npx lerna version patch --force-publish='*' # or major|minor|etc, or leave out to be prompted. either way, you get a confirmation dialog.
+```
+
+Since `lerna version --exact` doesn't work with local path deps like `file:../common`, do this:
 
-# to only publish tilde deps, see section `# Tilde deps` instead of doing next step
-npx lerna publish from-git # add `--dist-tag next` for pre-release versions
+- Find & replace `file:../.*"` with eg `~1.5.0"` in `packages/`.
```
bergundy, line 200: Wanna write a script for this?

bergundy (approved):

PR #935, opened 2022-10-24
chore: Use large GH hosted runner for stress test

bergundy:


.github/workflows/stress.yml
```diff
@@ -17,7 +17,7 @@ env:
 
 jobs:
   stress-test:
-    runs-on: [self-hosted, Linux, X64]
+    runs-on: [ubuntu-latest-4-cores, linux]
```
lorensr, line 20: so we're going to have 2 org runner groups with these names that this repo is given access to? why the second?

https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#overview-of-larger-runners

lorensr (approved):

.github/workflows/stress.yml
bergundy: Not sure I understand the question.
This is just requiring 2 labels, the `linux` label might be redundant.

Sushisource (approved):
9 min is not bad

bergundy:
@Sushisource it's not 9 minutes, the runner isn't picking up the job for some reason.

lorensr:
Is there a runner in the `temporalio` org with those 2 tags that this repo has access to?

bergundy:
There's supposed to be one, I'm trying to understand why it's not starting.

Sushisource:
15 minutes, less great, lol.  Still fine considering our check-in rate though I think.

PR #937, opened 2022-10-25
feat: Add Schedule API

mjameswh:
## What was changed
- Add Schedule API

## Why?
- Add Schedule API

### DIFF
--- a/packages/client/src/interceptors.ts
+++ b/packages/client/src/interceptors.ts
@@ -6,6 +6,7 @@
 
 import { Headers, Next } from '@temporalio/common';
 import { temporal } from '@temporalio/proto';
+import { CompiledScheduleOptions } from './schedule-types';
 import {
   DescribeWorkflowExecutionResponse,
   RequestCancelWorkflowExecutionResponse,
@@ -131,11 +132,43 @@ export interface WorkflowClientInterceptors {
   calls?: WorkflowClientCallsInterceptorFactory[];
 }
 
+export interface ScheduleClientCallsInterceptor {
+  /**
+   * Intercept a service call to CreateSchedule
+   */
+  create?: (input: CreateScheduleInput, next: Next<this, 'create'>) => Promise<Uint8Array /* conflictToken */>;
+}
+
+/** Input for {@link ScheduleClientCallsInterceptor.create} */
+export interface CreateScheduleInput {
+  readonly headers: Headers;
+  readonly options: CompiledScheduleOptions;
+}
+
+interface ScheduleClientCallsInterceptorFactoryInput {
+  scheduleId: string;
+}
+
+/**
+ * A function that takes a {@link ScheduleClientCallsInterceptorFactoryInput} and returns an interceptor
+ */
+export interface ScheduleClientCallsInterceptorFactory {
+  (input: ScheduleClientCallsInterceptorFactoryInput): ScheduleClientCallsInterceptor;
+}
+
+/**
+ * A mapping of interceptor type of a list of factory functions
+ */
+export interface ScheduleClientInterceptors {
+  calls?: ScheduleClientCallsInterceptorFactory[];
+}
+
 /**
  * Interceptors for any high-level SDK client.
  *
- * NOTE: Currently only for {@link WorkflowClient}. More will be added later as needed.
+ * NOTE: Currently only for {@link WorkflowClient} and {@link ScheduleClient}. More will be added later as needed.
  */
 export interface ClientInterceptors {
   workflow?: WorkflowClientInterceptors;
+  schedule?: ScheduleClientInterceptors;
 }
--- a/packages/client/src/interceptors.ts
+++ b/packages/client/src/interceptors.ts
@@ -141,7 +141,7 @@ export interface ScheduleClientInterceptor {
   /**
    * Intercept a service call to CreateSchedule
    */
-  create?: (input: CreateScheduleInput, next: Next<this, 'create'>) => Promise<Uint8Array /* conflictToken */>;
+  create?: (input: CreateScheduleInput, next: Next<this, 'create'>) => Promise<CreateScheduleOutput>;
 }
 
 /**
@@ -153,6 +153,10 @@ export interface CreateScheduleInput {
   readonly headers: Headers;
   readonly options: CompiledScheduleOptions;
 }
+
+export type CreateScheduleOutput = {
+  readonly conflictToken: Uint8Array;
+};
 
 /**
  * Interceptors for any high-level SDK client.
--- a/packages/client/src/interceptors.ts
+++ b/packages/client/src/interceptors.ts
@@ -161,5 +161,9 @@ export interface CreateScheduleInput {
  */
 export interface ClientInterceptors {
   workflow?: WorkflowClientInterceptors;
+
+  /**
+   * @experimental
+   */
   schedule?: ScheduleClientInterceptor[];
 }
--- a/packages/client/src/interceptors.ts
+++ b/packages/client/src/interceptors.ts
@@ -132,35 +132,26 @@ export interface WorkflowClientInterceptors {
   calls?: WorkflowClientCallsInterceptorFactory[];
 }
 
-export interface ScheduleClientCallsInterceptor {
+/**
+ * Implement any of these methods to intercept ScheduleClient outbound calls
+ *
+ * @experimental
+ */
+export interface ScheduleClientInterceptor {
   /**
    * Intercept a service call to CreateSchedule
    */
   create?: (input: CreateScheduleInput, next: Next<this, 'create'>) => Promise<Uint8Array /* conflictToken */>;
 }
 
-/** Input for {@link ScheduleClientCallsInterceptor.create} */
+/**
+ * Input for {@link ScheduleClientInterceptor.create}
+ *
+ * @experimental
+ */
 export interface CreateScheduleInput {
   readonly headers: Headers;
   readonly options: CompiledScheduleOptions;
-}
-
-interface ScheduleClientCallsInterceptorFactoryInput {
-  scheduleId: string;
-}
-
-/**
- * A function that takes a {@link ScheduleClientCallsInterceptorFactoryInput} and returns an interceptor
- */
-export interface ScheduleClientCallsInterceptorFactory {
-  (input: ScheduleClientCallsInterceptorFactoryInput): ScheduleClientCallsInterceptor;
-}
-
-/**
- * A mapping of interceptor type of a list of factory functions
- */
-export interface ScheduleClientInterceptors {
-  calls?: ScheduleClientCallsInterceptorFactory[];
 }
 
 /**
@@ -170,5 +161,5 @@ export interface ScheduleClientInterceptors {
  */
 export interface ClientInterceptors {
   workflow?: WorkflowClientInterceptors;
-  schedule?: ScheduleClientInterceptors;
+  schedule?: ScheduleClientInterceptor[];
 }
--- a/packages/client/src/schedule-client.ts
+++ b/packages/client/src/schedule-client.ts
@@ -1,0 +1,910 @@
+import { status as grpcStatus } from '@grpc/grpc-js';
+import {
+  compileRetryPolicy,
+  DataConverter,
+  LoadedDataConverter,
+  mapFromPayloads,
+  mapToPayloads,
+  searchAttributePayloadConverter,
+  SearchAttributes,
+  Workflow,
+} from '@temporalio/common';
+import { composeInterceptors, Headers } from '@temporalio/common/lib/interceptors';
+import {
+  encodeMapToPayloads,
+  loadDataConverter,
+  isLoadedDataConverter,
+  encodeToPayloads,
+  filterNullAndUndefined,
+  decodeMapFromPayloads,
+  decodeArrayFromPayloads,
+} from '@temporalio/common/lib/internal-non-workflow';
+import os from 'os';
+import { Connection } from './connection';
+import { CreateScheduleInput, ScheduleClientCallsInterceptor, ScheduleClientInterceptors } from './interceptors';
+import { ConnectionLike, Metadata, WorkflowService } from './types';
+import { v4 as uuid4 } from 'uuid';
+import { isServerErrorResponse, ServiceError } from './errors';
+import {
+  Backfill,
+  CalendarSpec,
+  CalendarSpecDescription,
+  CompiledScheduleOptions,
+  CompiledUpdatedSchedule,
+  compileScheduleOptions,
+  compileUpdatedScheduleOptions,
+  dayOfWeekNameToNumber,
+  dayOfWeekNumberToName,
+  IntervalSpecDescription,
+  ListScheduleEntry,
+  ListScheduleOptions,
+  monthNameToNumber,
+  monthNumberToName,
+  Range,
+  Schedule,
+  ScheduleActionType,
+  ScheduleOptions,
+  ScheduleOverlapPolicy,
+  UpdatedSchedule,
+  UpdateScheduleOptions,
+} from './schedule-types';
+import { Replace } from '@temporalio/common/lib/type-helpers';
+import { temporal } from '@temporalio/proto';
+import {
+  msOptionalToTs,
+  msToTs,
+  optionalDateToTs,
+  optionalTsToDate,
+  optionalTsToMs,
+  tsToDate,
+} from '@temporalio/common/lib/time';
+
+type CreateScheduleRequest = temporal.api.workflowservice.v1.ICreateScheduleRequest;
+// type CreateScheduleResponse = temporal.api.workflowservice.v1.ICreateScheduleResponse;
+type DescribeScheduleRequest = temporal.api.workflowservice.v1.IDescribeScheduleRequest;
+type DescribeScheduleResponse = temporal.api.workflowservice.v1.IDescribeScheduleResponse;
+type DeleteScheduleRequest = temporal.api.workflowservice.v1.IDeleteScheduleRequest;
+type DeleteScheduleResponse = temporal.api.workflowservice.v1.IDeleteScheduleResponse;
+type UpdateScheduleRequest = temporal.api.workflowservice.v1.IUpdateScheduleRequest;
+type UpdateScheduleResponse = temporal.api.workflowservice.v1.IUpdateScheduleResponse;
+type PatchScheduleRequest = temporal.api.workflowservice.v1.IPatchScheduleRequest;
+type PatchScheduleResponse = temporal.api.workflowservice.v1.IPatchScheduleResponse;
+type ListSchedulesRequest = temporal.api.workflowservice.v1.IListSchedulesRequest;
+type ListSchedulesResponse = temporal.api.workflowservice.v1.IListSchedulesResponse;
+
+// FIXME: Determine if we want to support these
+type ListScheduleMatchingTimesRequest = temporal.api.workflowservice.v1.IListScheduleMatchingTimesRequest;
+type ListScheduleMatchingTimesResponse = temporal.api.workflowservice.v1.IListScheduleMatchingTimesResponse;
+
+/**
+ * Handle to a single Schedule
+ */
+export interface ScheduleHandle {
+  /**
+   * This Schedule's identifier
+   */
+  readonly scheduleId: string;
+
+  /**
+   * Fetch the Schedule's description from the Server
+   */
+  describe(): Promise<Schedule>;
+
+  /**
+   * Update the Schedule
+   *
+   * Apply a user update function to the current schedule definition, and then try to update the
+   * schedule definition on the server. Note that, in the future, the user provided function might
+   * be invoked multiple time, with identical or different input.
+   *
+   * WARNING: At this moment, the server provides no garantee that update was applied sucessfully.
+   */
+  update<Action extends Workflow = Workflow>(updateFn: (previous: Schedule) => UpdatedSchedule<Action>): Promise<void>;
+
+  /**
+   * Delete the Schedule
+   */
+  delete(): Promise<void>;
+
+  /**
+   * Trigger an Action to be taken immediately
+   *
+   * @param overlap Override the Overlap Policy for this one trigger. Defaults to {@link ScheduleOverlapPolicy.ALLOW_ALL}.
+   */
+  trigger(overlap?: ScheduleOverlapPolicy): Promise<void>;
+
+  /**
+   * Run though the specified time period(s) and take Actions as if that time passed by right now, all at once. The
+   * Overlap Policy can be overridden for the scope of the Backfill.
+   */
+  backfill(options: Backfill | Backfill[]): Promise<void>;
+
+  /**
+   * Pause the Schedule
+   *
+   * @param note A new {@link Schedule.note}. Defaults to `"Paused via TypeScript SDK"`
+   * @throws {@link ValueError} if empty string is passed
+   */
+  pause(note?: string): Promise<void>;
+
+  /**
+   * Unpause the Schedule
+   *
+   * @param note A new {@link Schedule.note}. Defaults to `"Unpaused via TypeScript SDK"`
+   * @throws {@link ValueError} if empty string is passed
+   */
+  unpause(note?: string): Promise<void>;
+
+  /**
+   * Readonly accessor to the underlying ScheduleClient
+   */
+  readonly client: ScheduleClient;
+}
+
+export interface ScheduleClientOptions {
+  /**
+   * {@link DataConverter} to use for serializing and deserializing payloads
+   */
+  dataConverter?: DataConverter | LoadedDataConverter;
+
+  /**
+   * Used to override and extend default Connection functionality
+   *
+   * Useful for injecting auth headers and tracing Workflow executions
+   */
+  interceptors?: ScheduleClientInterceptors;
+
+  /**
+   * Identity to report to the server
+   *
+   * @default `${process.pid}@${os.hostname()}`
+   */
+  identity?: string;
+
+  /**
+   * Connection to use to communicate with the server.
+   *
+   * By default `ScheduleClient` connects to localhost.
+   *
+   * Connections are expensive to construct and should be reused.
+   */
+  connection?: ConnectionLike;
+
+  /**
+   * Server namespace
+   *
+   * @default default
+   */
+  namespace?: string;
+}
+
+export type ScheduleClientOptionsWithDefaults = Replace<
+  Required<ScheduleClientOptions>,
+  {
+    connection?: ConnectionLike;
+  }
+>;
+export type LoadedScheduleClientOptions = ScheduleClientOptionsWithDefaults & {
+  loadedDataConverter: LoadedDataConverter;
+};
+
+export function defaultScheduleClientOptions(): ScheduleClientOptionsWithDefaults {
+  return {
+    dataConverter: {},
+    // The equivalent in Java is ManagementFactory.getRuntimeMXBean().getName()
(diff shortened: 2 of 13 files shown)


packages/client/src/interceptors.ts
```diff
@@ -131,11 +132,43 @@ export interface WorkflowClientInterceptors {
   calls?: WorkflowClientCallsInterceptorFactory[];
 }
 
+export interface ScheduleClientCallsInterceptor {
+  /**
+   * Intercept a service call to CreateSchedule
+   */
+  create?: (input: CreateScheduleInput, next: Next<this, 'create'>) => Promise<Uint8Array /* conflictToken */>;
```
mjameswh, line 139: @bergundy @lorensr 
1. Proposal specified a the return value as being the `conflictToken`, in line with `WorkflowClientCallsInterceptor.start()` and `.signalWithStart()`, which returns `runId`. I however see a risk that this might eventually cause breaking changes if we need to bubble up other fields from the grpc response. I'm considering that we should instead return the gRPC response here, or some `CreateScheduleOutput` type. Opinions?
2. Should we also intercept `update` (which can also modify headers)?
3. What about other calls? One frequent use case for interceptors is the ability to log things from there. `WorkflowClientCallsInterceptor` can intercept things like `describe` and `query`...

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+type DeleteScheduleResponse = temporal.api.workflowservice.v1.IDeleteScheduleResponse;
+type UpdateScheduleRequest = temporal.api.workflowservice.v1.IUpdateScheduleRequest;
+type UpdateScheduleResponse = temporal.api.workflowservice.v1.IUpdateScheduleResponse;
+type PatchScheduleRequest = temporal.api.workflowservice.v1.IPatchScheduleRequest;
+type PatchScheduleResponse = temporal.api.workflowservice.v1.IPatchScheduleResponse;
+type ListSchedulesRequest = temporal.api.workflowservice.v1.IListSchedulesRequest;
+type ListSchedulesResponse = temporal.api.workflowservice.v1.IListSchedulesResponse;
+
+// FIXME: Determine if we want to support these
+type ListScheduleMatchingTimesRequest = temporal.api.workflowservice.v1.IListScheduleMatchingTimesRequest;
```
mjameswh, line 76: @lorensr High level API for this request were left out of the proposal. What's the status? Is that in scope for this PR?

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+
+  /**
+   * Fetch the Schedule's description from the Server
+   */
+  describe(): Promise<Schedule>;
+
+  /**
+   * Update the Schedule
+   *
+   * Apply a user update function to the current schedule definition, and then try to update the
```
mjameswh, line 96: @lorensr Comfortable with the doc comment?

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+  runningWorkflows: WorkflowExecutionWithFirstExecutionRunId[];
+
+  createdAt: Date;
+  lastUpdatedAt: Date | undefined;
+
+  raw: temporal.api.workflowservice.v1.IDescribeScheduleResponse;
+};
+
+export type UpdatedSchedule<Action extends Workflow> = Pick<
+  ScheduleOptions<Action>,
```
Quinn-With-Two-Ns, line 138: I think your missing `LimitActions` and `RemainingActions` from the proto.

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+  return {
+    dataConverter: {},
+    // The equivalent in Java is ManagementFactory.getRuntimeMXBean().getName()
+    identity: `${process.pid}@${os.hostname()}`,
+    interceptors: {},
+    namespace: 'default',
+  };
+}
+
+// FIXME: assertRequiredScheduleOptions
```
mjameswh, line 201: @mjameswh

packages/client/src/schedule-client.ts
lorensr: @dnr has list matching been finalized yet?

image

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+    const interceptors = (this.options.interceptors.calls ?? []).map((ctor) => ctor({ scheduleId }));
+
+    await this._createSchedule(options, interceptors);
+    return this._createScheduleHandle({ scheduleId, interceptors });
+  }
+
+  /**
+   * Create a new Schedule.
+   *
+   * @returns ???
```
mjameswh, line 274: @mjameswh Review doc on all methods

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+  /**
+   * Update the Schedule
+   *
+   * Apply a user update function to the current schedule definition, and then try to update the
+   * schedule definition on the server. Note that, in the future, the user provided function might
+   * be invoked multiple time, with identical or different input.
+   *
+   * WARNING: At this moment, the server provides no garantee that update was applied sucessfully.
+   */
+  update<Action extends Workflow = Workflow>(updateFn: (previous: Schedule) => UpdatedSchedule<Action>): Promise<void>;
```
lorensr, line 102:
```suggestion
  /**
   * Update the Schedule
   *
   * This function calls `.describe()`, provides the `Schedule` to the provided `updateFn`, and sends the returned `UpdatedSchedule` to the Server to update the Schedule definition. Note that, in the future, `updateFn` might
   * be invoked multiple times, with identical or different input.
   *
   * WARNING: At this moment, the server provides no guarantee that an update is applied successfully.
   */
  update<Action extends Workflow = Workflow>(updateFn: (previous: Schedule) => UpdatedSchedule<Action>): Promise<void>;
```

Are we including `conflict_token` in `UpdateScheduleRequest`? If not, I'd remove the WARNING, since the update will always be applied.

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+
+  /**
+   * Server namespace
+   *
+   * @default default
+   */
+  namespace?: string;
+}
+
+export type ScheduleClientOptionsWithDefaults = Replace<
```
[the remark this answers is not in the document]
lorensr: If user never needs to call `defaultScheduleClientOptions`, fine to not export. The 2 reasons I can think of for exporting `LoadedWorkflowClientOptions` are:

- You can click on it here: https://typescript.temporal.io/api/classes/client.WorkflowClient#options
- User can import it and do `const options: LoadedScheduleClientOptions = workflowClient.options`

packages/client/src/schedule-client.ts
dnr: The first API is implemented and works. The problem is that I just don't think it's particularly useful at all. The UI ended up not needing it. So I was considering removing it, but haven't done that yet. Do you (any of you) think it would be useful?

The second (with a separate spec) could be more useful, but isn't implemented yet. It would probably be more useful to the UI than to SDK clients.

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+  takenAt: Date;
+
+  /** If action was {@link StartWorkflowAction}. */
+  workflow?: WorkflowExecutionWithFirstExecutionRunId;
+}
+
+/**
+ * Policy for overlapping Actions.
+ */
+export enum ScheduleOverlapPolicy {
```
bergundy, line 173: We should probably export the original proto enums here to be consistent with the rest of the SDK.

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+
+  /**
+   * The step to take between each value.
+   *
+   * @default 1
+   */
+  step?: never;
+}
+
+interface RangeStartEndSkip<Unit> extends Omit<RangeStartOnly<Unit>, 'end' | 'step'> {
```
bergundy, line 268: Why do we need this variant?

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+  /**
+   * The step to take between each value.
+   *
+   * @default 1
+   */
+  step?: number;
+}
+
+export type NumberSpec = number | Range<number> | (Range<number> | number)[] | '*';
+export type NumberSpecDescription = Range<number>[];
```
bergundy, line 285: This type name is unclear to me, why not just use the range array as the type?

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+   * @format number of milliseconds
+   */
+  jitter?: number;
+};
+
+export type StartWorkflowAction<Action extends Workflow> = Omit<
+  WorkflowStartOptions<Action>,
+  'workflowIdReusePolicy' | 'cronSchedule' | 'followRuns'
+> & {
+  // This is most convenient for TS typing. Other SDKs may want to implement this differently, for example nesting:
```
bergundy, line 639: This comment can be deleted

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+  /**
+   * All times will be incremented by a random value from 0 to this amount of jitter.
+   *
+   * @default 1 second
+   * @format number of milliseconds
+   */
+  jitter?: number;
+};
+
+export type StartWorkflowAction<Action extends Workflow> = Omit<
```
bergundy, line 635: Replace Action with W or WorkflowFunction

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+   */
+  runningWorkflows: WorkflowExecutionWithFirstExecutionRunId[];
+
+  createdAt: Date;
+  lastUpdatedAt: Date | undefined;
+
+  raw: temporal.api.workflowservice.v1.IDescribeScheduleResponse;
+};
+
+export type UpdatedSchedule<Action extends Workflow> = Pick<
```
bergundy, line 137: Let’s make this `Omit<ScheduleOptions<Action>, ‘state’>` and make sure that anything that can be updated is not part of the state.

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+  nextActionTimes: Date[];
+}
+
+/**
+ * The current Schedule details. They may not match the Schedule as created because:
+ * - some fields in the state are modified automatically
+ * - the schedule may have been modified by {@link ScheduleHandle.update} or
+ *   {@link ScheduleHandle.pause}/{@link ScheduleHandle.unpause}
+ */
+export type Schedule = ListScheduleEntry & {
```
bergundy, line 71: I want to go over this and the update and create requests to see what’s overlapping.
I get why they’re all separate types, but I just need another pass on those and it’s way passed my bedtime.

bergundy (commented):
Overall LGTM.
All of these new APIs need to be marked experimental.
I reviewed this very late at night so I may have missed things or asked trivial questions.

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+   */
+  pauseOnFailure: boolean;
+
+  /**
+   * The Actions remaining in this Schedule. Once this number hits `0`, no further Actions are taken (unless {@link
+   * ScheduleHandle.trigger} is called).
+   *
+   * @default undefined (unlimited)
+   */
+  remainingActions?: number;
```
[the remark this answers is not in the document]
dnr: In the protos, `ScheduleInfo` is the stuff that can't be updated. It is not part of `Schedule`, which is everything that can be updated (everything else). `ScheduleState` is part of `Schedule` and can be updated. It's essentially the things that are expected to change and might change without an explicit update, as opposed to `SchedulePolicies`, which are not expected to change (but can be changed by the user).

I think at least using the same terms will help everyone avoid getting confused

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+  workflowId: string;
+
+  /**
+   * The Run Id of the original execution that was started by the Schedule. If the Workflow retried, did
+   * Continue-As-New, or was Reset, the following runs would have different Run Ids.
+   */
+  firstExecutionRunId: string;
+}
+
+export interface ScheduleAction {
```
dnr, line 159: I understand that we might want to tweak some names at this level, but it does seem confusing to have a `ScheduleAction` that corresponds to a `ScheduleActionResult` in the proto, while `ScheduleAction` in the proto is something else

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+   *
+   * This number is decremented after each Action is taken, and Actions are not
+   * taken when the number is `0` (unless {@link ScheduleHandle.trigger} is called).
+   *
+   * @default unlimited
+   */
+  remainingActions?: number;
+
+  /**
+   * Trigger one Action immediately.
```
dnr, line 741: might want to note that this is "Trigger one Action immediately on schedule creation" (same for backfill)

packages/client/src/interceptors.ts
```diff
@@ -131,11 +132,43 @@ export interface WorkflowClientInterceptors {
...
+}
+
+/** Input for {@link ScheduleClientCallsInterceptor.create} */
+export interface CreateScheduleInput {
+  readonly headers: Headers;
+  readonly options: CompiledScheduleOptions;
+}
+
+interface ScheduleClientCallsInterceptorFactoryInput {
+  scheduleId: string;
```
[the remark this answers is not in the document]
mjameswh: I agree.

In that case, I don't think we need interceptor factories, do we?

If we can avoid these, then we can safely deprecate the previous approach in workflow client by defining:

```
export interface WorkflowClientInterceptors {
  calls?: (WorkflowClientCallsInterceptorFactory|WorkflowClientCallsInterceptor)[];
}
```

and then type checking either `calls` items are typeof object (new way) or function (deprecated way). We will have to continue supporting the previous approach for some minor versions.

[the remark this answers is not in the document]
mjameswh: Regarding Schedule interceptors, I will immediately remove `factory` and make 

```
export interface ScheduleClientInterceptors {
  calls?: ScheduleClientCallsInterceptor[];
}
```

Makes sense?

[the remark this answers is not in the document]
bergundy: Probably get rid of the `calls` attribute and flatten it out:

```ts
export interface WorkflowClientInterceptor {
  // ... rename from WorkflowClientCallsInterceptor
}

// Maintain the export alias
export type WorkflowClientCallsInterceptor = WorkflowClientInterceptor;

export interface WorkflowClientInterceptors = {
  /** @deprecated */
  calls?: WorkflowClientCallsInterceptorFactory[];
}

interface ScheduleClientInterceptor {
}

export interface ClientInterceptors {
  workflow?: WorkflowClientInterceptors | WorkflowClientInterceptor[];
  schedule?: ScheduleClientInterceptor[];
}
```

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+  }
+
+  protected get dataConverter(): LoadedDataConverter {
+    return this.options.loadedDataConverter;
+  }
+
+  /**
+   * Set the deadline for any service requests executed in `fn`'s scope.
+   */
+  async withDeadline<R>(deadline: number | Date, fn: () => Promise<R>): Promise<R> {
```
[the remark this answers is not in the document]
bergundy: I'd say yes

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+   * Whether Schedule is currently paused.
+   *
+   * @default false
+   */
+  paused: boolean;
+
+  /**
+   * Present if action is a {@link StartWorkflowAction}.
+   */
+  workflowType?: string;
```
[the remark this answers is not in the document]
bergundy: I'm fine with not having the full action, I don't think it's needed.
I would want to structure this differently in TS:

```ts
interface WorkflowScheduleListAction {
  type: 'workflow';
  workflowType: string;
}

// In the future this will be a union type with other actions
type ScheduleListAction = WorkflowScheduleListAction;

// This has to map to something and can be switched on
action: ScheduleListAction;
```

packages/client/src/interceptors.ts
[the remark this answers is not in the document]
mjameswh: Pretty much what I had in mind. I'll do the WorkflowClient refactor in a distinct PR though.

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+    }
+  }
+
+  /**
+   * Get a handle to a Schedule
+   *
+   * This method does not validate `scheduleId`. If there is no Schedule with the given `scheduleId`, handle
+   * methods like `handle.describe()` will throw a {@link ScheduleNotFoundError} error.
+   */
+  public getHandle(scheduleId: string): ScheduleHandle {
```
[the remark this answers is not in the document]
mjameswh: Ehhh... What do you mean?

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,830 @@
...
+export interface ListScheduleEntry {
+  /**
+   * Schedule Id
+   *
+   * We recommend using a meaningful business identifier.
+   */
+  scheduleId: string;
+
+  /** When Actions are taken */
+  spec: RequireAtLeastOne<ScheduleSpecDescription, 'calendars' | 'intervals'>;
```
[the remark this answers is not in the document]
mjameswh: It really is _at last one_. Rules in calendars and rules in intervals cumulate as alternatives (ie. the schedule trigger when any of the rule match). I doubt that mixing calendars and intervals will be a common usage pattern though.

packages/client/src/schedule-types.ts
mjameswh: Definitely agreed. That name is also confusing in regard to other types in the TypeScript's schedule API.

I'm considering this:
```
export interface ScheduleExecutionStartWorkflowActionResult {
  /** Time that the Action was scheduled for, including jitter */
  scheduledAt: Date;

  /** Time that the Action was actually taken */
  takenAt: Date;

  /** The Action that was taken */
  action: ScheduleExecutionActionResult;
}

export type ScheduleExecutionActionResult = ScheduleExecutionStartWorkflowActionResult;

export interface ScheduleExecutionStartWorkflowActionResult {
  type: 'startWorkflow';
  workflow: WorkflowExecutionWithFirstExecutionRunId;
}
```

packages/client/src/schedule-types.ts
mjameswh: Because we want to support:

```
 * { start: 2 } ➡️ 2
 * { start: 2, end: 4 } ➡️ 2, 3, 4
 * { start: 2, end: 10, step: 3 } ➡️ 2, 5, 8
```

but not:

```
 * { start: 2, step: 1 }
```

So both `end` and `step` are optional, but `step` can't be there unless `end` is also present. I can however shorten this to only the `type` (ie. remove the two useless interfaces).

mjameswh:
```
export type Range<Unit> =
  | {
      /** Start of range (inclusive) */
      start: Unit;

      /**
       * End of range (inclusive)
       * @default `start`
       */
      end?: Unit;

      /**
       * The step to take between each value.
       * @default 1
       */
      step?: never;
    }
  | {
      /** Start of range (inclusive) */
      start: Unit;

      /** End of range (inclusive) */
      end: Unit;

      /** The step to take between each value. */
      step: number;
    };
```

bergundy: Does it hurt if we have just `start` and `step`?
Isn't it the exact same as just `start`?

mjameswh: I think that would be confusing... Intuitively, I would read `{ start: 3, step: 1 }` as meaning an open range starting at 3 and including every other values after... I think it better disallow this form entirely.

packages/client/src/schedule-client.ts
```diff
@@ -0,0 +1,910 @@
...
+        const step = x.step! > 0 ? x.step! : 1;
+        return { start, end, step };
+      }) ?? undefined
+    );
+  }
+
+  return { encoder, decoder };
+}
+
+function encodeOptionalStructuredCalendarSpecs(
```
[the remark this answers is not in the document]
mjameswh: Are you sure about that?

AFAICS `compileRetryPolicy`/`decompileRetryPolicy` seems to be the only one following that naming pattern for TS<->Proto conversions. `encode`/`decode` seems to be more prevalent.

No preference personally. Your call.

packages/client/src/schedule-types.ts
mjameswh: Coming up with a big refactor of Schedule types. Should be easier to review.

packages/client/src/schedule-types.ts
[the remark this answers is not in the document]
mjameswh: For the moment, I'll mocking the proto's naming structure ('state' and 'info'). I'm open to discussion on this however. I feel that 'state' will be hard to understand for users.

packages/client/src/schedule-types.ts
mjameswh: That will definitely make things much more easier.

I suppose that the proposal used this form instead to provide simpler names (ie. no 'SCHEDULE_OVERLAP_POLICY_*' prefixes) and user documentation.

@lorensr Do you want to chime in before I proceed with Roey's suggestion?

lorensr: I prefer ease of use over consistency here, don't think latter is that impt

packages/client/src/schedule-types.ts
lorensr: This API hides all protos, right? So I don't see the issue with name conflicts. And generally prefer shorter terms, like:

```ts
export interface ScheduleAction {
  /** Time that the Action was scheduled for, including jitter */
  scheduledAt: Date;

  /** Time that the Action was actually taken */
  takenAt: Date;

  /** The result of the Action that was taken */
  result: ScheduleActionResult;
}

export type ScheduleActionResult = StartWorkflowActionResult;

export interface StartWorkflowActionResult {
  type: 'startWorkflow';
  workflow: WorkflowExecutionWithFirstExecutionRunId;
}
```

mjameswh: @lorensr I obviously prefer sorter names.

The problem here is that Describe, List, Create/Update, CompiledCreate/CompiledUpdate, recentActions, and runningActions all have distinct variants of the `ScheduleAction*` types. At this moment, we could deal with this by `Omit<>`ing or `Pick<>`ing some fields, but that would not work once we add other types of actions.

packages/client/src/schedule-client.ts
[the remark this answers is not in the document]
mjameswh: I think this might be useful, but will consider it for a distinct PR.

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,845 @@
...
+ *
+ * @experimental
+ */
+export type LooseRange<Unit> =
+  | Range<Unit>
+  | { start: Range<Unit>['start']; end?: Range<Unit>['end']; step?: never }
+  | Unit;
+
+/** @experimental */
+export const Months = [
```
bergundy, line 688: nit: Use CONST case or camelCase for constants.

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,845 @@
...
+    args: unknown[];
+  }
+>;
+
+/**
+ * Policy for overlapping Actions.
+ *
+ * @experimental
+ */
+export enum ScheduleOverlapPolicy {
```
bergundy, line 775: Still not liking the inconsistency here with the rest of the SDK.
I'm not blocking on this but it deserves a short discussion IMHO.

packages/test/src/test-schedules.ts
```diff
@@ -0,0 +1,610 @@
...
+};
+
+test.before(async (t) => {
+  const testEnv = await TestWorkflowEnvironment.createLocal();
+  t.context = {
+    testEnv,
+    client: testEnv.client,
+  };
+  // In case we're running with temporalite or other non default server.
+  // NOTE: at the time this was added temporalite did not expose the grpc OperatorService.
```
bergundy, line 42: It actually exposes it and if you start it with `system.forceSearchAttributesCacheRefreshOnRead=true` it works pretty well for testing.

bergundy (approved):
This looks great, I don't have a lot of comments at this point and I would merge soon.
I'd follow up with a cleanup PR that adds an internal base class to all of the different SDK clients.

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,845 @@
...
+   * Schedule Id
+   *
+   * We recommend using a meaningful business identifier.
+   */
+  scheduleId: string;
+
+  /**
+   * When Actions should be taken
+   */
+  spec: RequireAtLeastOne<ScheduleSpec, 'calendars' | 'intervals' | 'cronExpressions'>;
```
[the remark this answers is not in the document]
mjameswh: That's right, but the schedule is then a no-op. Seems like a reasonable DX addition...

packages/test/src/test-schedules.ts
```diff
@@ -0,0 +1,610 @@
...
+      maxTimeout: 1000,
+    }
+  );
+});
+
+test.after.always(async (t) => {
+  await t.context.testEnv?.teardown();
+});
+
+if (RUN_INTEGRATION_TESTS) {
```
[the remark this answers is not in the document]
mjameswh: Not applicable: no longer using test env, because Temporalite is missing some schedule features.

packages/client/src/schedule-types.ts
lorensr: Here's the proposal discussion: https://github.com/temporalio/proposals/pull/62#discussion_r942844591

packages/client/src/schedule-types.ts
```diff
@@ -0,0 +1,845 @@
...
+    action: CompiledScheduleAction;
+  }
+>;
+
+/**
+ * The specification of an updated Schedule, as expected by {@link ScheduleHandle.update}.
+ *
+ * @experimental
+ */
+export type ScheduleUpdateOptions = Omit<ScheduleOptions, 'scheduleId' | 'memo' | 'searchAttributes'>;
```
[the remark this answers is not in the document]
mjameswh: Actually, removed `triggerImmediately` and `backfill` from ScheduleUpdateOptions.state as they are valid on create, but have to be updated through a patch rather than through `Update`

packages/test/src/test-schedules.ts
```diff
@@ -0,0 +1,610 @@
...
+    const { client } = t.context;
+    const scheduleId = `can-create-schedule-with-inteval-${randomUUID()}`;
+    const handle = await client.schedule.create({
+      scheduleId,
+      spec: {
+        intervals: [{ every: '1h', offset: '5m' }],
+      },
+      action: {
+        type: 'startWorkflow',
+        workflowId: `${scheduleId}-workflow`,
```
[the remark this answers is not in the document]
mjameswh: That makes sense to me.

packages/client/src/schedule-types.ts
mjameswh: The only thing that bothers me with reexporting the proto's enum directly is that it doesn't provide type doc. The values in this enum are not trivial to understand for users, so that doc is particularly important.

PR #940, opened 2022-10-28
chore: Use `process.stderr.write` for the default log function

bergundy:
This will prevent jest from complaining about use of `console.error`.


lorensr (approved):
Did you test colors? Wondering if you have to use a format string like with `%O` to get coloring

image


https://nodejs.org/api/util.html#customizing-utilinspect-colors

mjameswh (approved):

bergundy:
I tested the colors @lorensr

PR #941, opened 2022-10-28
chore: Use latest docker-compose in stress tests

bergundy:


mjameswh (approved):

mjameswh:
FTR: this line had been added in #808 as an attempt to get rid of stress test flakes, but didn't worked. No reason to keep it.

PR #942, opened 2022-10-28
feat(client): Implement basic list API

bergundy:


packages/client/src/workflow-client.ts
```diff
@@ -946,6 +924,31 @@ export class WorkflowClient {
...
+  public async *list(query: string, options?: ListOptions): AsyncIterable<WorkflowExecutionInfo> {
+    let nextPageToken: Uint8Array = Buffer.alloc(0);
+    for (;;) {
+      const respose = await this.workflowService.listWorkflowExecutions({
+        namespace: this.options.namespace,
+        query,
+        nextPageToken,
+        pageSize: options?.pageSize,
+      });
+      // Not decoding memo payloads concurrently even though we could have to keep the lazy nature of this iterator.
```
Sushisource, line 944: Any particular reason?

Sushisource (approved):
Cool. Straightforward, works for me.

packages/client/src/workflow-client.ts
bergundy: It's not worth it IMHO for now, memos tend to generally be small.

packages/client/src/workflow-client.ts
```diff
@@ -946,6 +926,31 @@ export class WorkflowClient {
...
+      for (const raw of respose.executions) {
+        yield await executionInfoFromRaw(raw, this.dataConverter);
+      }
+      nextPageToken = respose.nextPageToken;
```
lorensr, line 950:
```suggestion
      const response = await this.workflowService.listWorkflowExecutions({
        namespace: this.options.namespace,
        query,
        nextPageToken,
        pageSize: options?.pageSize,
      });
      // Not decoding memo payloads concurrently even though we could have to keep the lazy nature of this iterator.
      for (const raw of response.executions) {
        yield await executionInfoFromRaw(raw, this.dataConverter);
      }
      nextPageToken = response.nextPageToken;
```

lorensr (approved):

packages/client/src/workflow-client.ts
```diff
@@ -946,6 +926,31 @@ export class WorkflowClient {
...
   }
+
+  /**
+   * List workflows by given `query`.
+   *
+   * See more info on how to build the query in the Temporal docs: https://docs.temporal.io/visibility
+   *
+   * This API requires Elasticsearch based visibility as of server version 1.18.
+   */
+  public async *list(query: string, options?: ListOptions): AsyncIterable<WorkflowExecutionInfo> {
```
cretz, line 937: Should query be a list option or at least optional?

packages/client/src/workflow-client.ts
bergundy: Me and my typos :(

packages/client/src/workflow-client.ts
```diff
@@ -946,6 +926,31 @@ export class WorkflowClient {
       followRuns: options?.followRuns ?? true,
     });
   }
+
+  /**
+   * List workflows by given `query`.
+   *
+   * See more info on how to build the query in the Temporal docs: https://docs.temporal.io/visibility
+   *
+   * This API requires Elasticsearch based visibility as of server version 1.18.
```
cretz, line 935: Not exactly true as simple queries, e.g. listing open workflows, work w/out elastic search

packages/client/src/workflow-client.ts
cretz, line 937: If describe is interceptable, this should be right?

packages/client/src/workflow-client.ts
bergundy: I didn't know that, I thought this entire API only works with ES right now.
I'll fix.

packages/client/src/workflow-client.ts
bergundy: Good question... is it optional on the server?

bergundy: I don't see any point making either of them interceptable.

lorensr: think so

image

cretz: I support either approach, but the one-is-one-isn't seems a little strange. Today describe is interceptable.

bergundy: Looks like it's optional on the server..

packages/client/src/workflow-client.ts
bergundy: Thanks, fixed.

packages/client/src/workflow-client.ts
bergundy: Made `query` optional.

bergundy: Tracking this in a separate issue.
I want to clean up client interceptors a bit before adding this.

PR #943, opened 2022-10-31
fix(worker): Wait for worker shutdown if runUntil promise throws

bergundy:


packages/worker/src/worker.ts
```diff
@@ -1644,21 +1671,43 @@ export class Worker {
...
    */
   async runUntil<R>(fnOrPromise: Promise<R> | (() => Promise<R>)): Promise<R> {
-    const runAndShutdown = async () => {
-      try {
-        if (typeof fnOrPromise === 'function') {
-          return await fnOrPromise();
-        } else {
-          return await fnOrPromise;
+    const workerRunPromise = this.run();
+    const innerPromise = typeof fnOrPromise === 'function' ? fnOrPromise() : fnOrPromise;
```
mjameswh, line 1685: In the function case, `fnOrPromise()` could throw an exception directly (rather than wrapped in a promise). In that case, the worker would not be shutdown.

packages/worker/src/worker.ts
```diff
@@ -1644,21 +1671,43 @@ export class Worker {
...
+    const [innerResult, workerRunResult] = await Promise.allSettled([
+      innerPromise.finally(() => {
+        if (this.state === 'RUNNING') {
```
mjameswh, line 1688: Wondering if there is a risk of race condition here... What if worker's state is one of `'STOPPING' | 'DRAINING' | 'DRAINED'` at the moment the user's promise throws? Should we wait for the worker to actually reach one of `'STOPPED' | 'FAILED'` state?

packages/worker/src/worker.ts
```diff
@@ -140,6 +140,33 @@ export type ActivityTaskWithContext = ContextAware<{
...
+ * Error thrown by {@link Worker.runUntil}
+ */
+export class CombinedWorkerRunError extends Error {
+  public readonly name = 'CombinedWorkerRunError';
+  public readonly cause: CombinedWorkerRunErrorCause;
+
+  constructor(message: string, { cause }: { cause: CombinedWorkerRunErrorCause }) {
+    super(message);
+    this.cause = cause;
+  }
```
mjameswh, line 167: It might be useful to customize the `toString()` output on this error to include details about both exceptions. I doubt most devs will actually expect and properly decode the CombinedWorkerRunError exception.

packages/worker/src/worker.ts
```diff
@@ -140,6 +140,33 @@ export type ActivityTaskWithContext = ContextAware<{
...
+   */
+  innerError: unknown;
+}
+
+/**
+ * Error thrown by {@link Worker.runUntil}
+ */
+export class CombinedWorkerRunError extends Error {
+  public readonly name = 'CombinedWorkerRunError';
+  public readonly cause: CombinedWorkerRunErrorCause;
```
mjameswh, line 162: `Error.cause` is a pretty common property, and most devs expect this one to be an error, not an array. It might be safer in our case to name that property `causes` (plural).

packages/worker/src/worker.ts
mjameswh: You can probably just reorder these two lines...

mjameswh (approved):
Overall LGTM. Approving, though at least one thing that must be fixed, and a few other things for you to consider.

packages/worker/src/worker.ts
bergundy: True, if it's not an async function.
I can wrap that though, good point.

packages/worker/src/worker.ts
bergundy: That's the point of this PR, to make sure the worker is fully stopped before returning from `runUntil`.

packages/worker/src/worker.ts
bergundy: Node 16.9 and above will auto format cause so that's not needed.

packages/worker/src/worker.ts
bergundy: According to MDN, you can put whatever you want in `cause`.

PR #944, opened 2022-11-01
chore: Refactor module scoped `state` variable to use globals

bergundy:
Also merged the `State` class into the `Activator` class, there was no point in keeping both of those.

We get a few things out of this:
1. Allows manipulating the workflow context externally so it can be reused (in a follow up PR)
2. Support using multiple versions of `@temporalio/workflow` (e.g. with library dependencies)
3. Clean up some of the code and make some attributes which were nullable (like workflow info) not nullable

### DIFF
--- a/packages/workflow/src/workflow.ts
+++ b/packages/workflow/src/workflow.ts
@@ -35,7 +35,7 @@ import {
   EnhancedStackTrace,
   WorkflowInfo,
 } from './interfaces';
-import { LocalActivityDoBackoff, getState } from './internals';
+import { LocalActivityDoBackoff, getActivator } from './internals';
 import { Sinks } from './sinks';
 import { untrackPromise } from './stack-helpers';
 import { ChildWorkflowHandle, ExternalWorkflowHandle } from './workflow-handle';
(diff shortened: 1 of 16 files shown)


packages/workflow/src/workflow.ts
```diff
@@ -832,18 +836,15 @@ export async function executeChild<T extends Workflow>(
...
+  return getActivator().info;
 }
 
 /**
  * Returns whether or not code is executing in workflow context
  */
 export function inWorkflowContext(): boolean {
   try {
-    workflowInfo();
+    getActivator();
```
mjameswh, line 847: Seems simple enough to avoid depending on an exception here...

```
  return (globalThis as any).__TEMPORAL__?.activator !== undefined;
```

I can't find any up-to-date performance data on this subject, but I generally assume a performance penalty on throwing/catching exceptions...

mjameswh (approved):
LGTM

packages/workflow/src/workflow.ts
bergundy: Hmm.. I could probably add a `maybeGetActivator()` helper.
I'm trying to minimize the use of the untyped global object.

bergundy: I might do this in a later PR, would rather merge for now.

PR #946, opened 2022-11-01
chore: Use require instead of import to make worker-interface methods sync

bergundy:
While this could be considered a step in the wrong direction, it simplifies our code a bit.
It will pay off when we get to context reuse.

Tested this manually with the ESM sample, it works.

Broke the experimental `internal` interceptors `dispose` API, it's no longer async.


packages/workflow/src/worker-interface.ts
```diff
@@ -103,14 +103,12 @@ export async function initRuntime(options: WorkflowCreateOptionsWithSourceMap):
 
   // TODO(bergundy): check if we can use `require` with the ESM sample and make all of the runtime methods sync.
 
-  // @ts-expect-error this is a webpack alias to payloadConverterPath
```
mjameswh, line 106: The prose in this comment still seems useful...

mjameswh (approved):
Why would that be the wrong direction? I see no problem with require in this context.

bergundy:
Looks like I need to fix a few things here.
I'll look into it.

packages/workflow/src/worker-interface.ts
bergundy: Thanks, I'll add it back.

packages/workflow/src/worker-interface.ts
```diff
@@ -103,14 +103,12 @@ export async function initRuntime(options: WorkflowCreateOptionsWithSourceMap):
 
   // TODO(bergundy): check if we can use `require` with the ESM sample and make all of the runtime methods sync.
```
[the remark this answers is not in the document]
bergundy: Nope, thanks for catching this

PR #951, opened 2022-11-04
feat(workflow): Reuse a single v8 context

bergundy:
This PR includes a major experimental optimization to the workflow runtime.
Without context reuse, the worker uses a separate v8 context for each workflow instance spread across 8 worker threads (isolates).
With the optimization on, I was able to eliminate the need for the extra worker threads and reuse a single v8 context to run all of the workflow instances, the optimization preserves the safety of the deterministic sandbox.

Removing the worker threads is a big part of the optimization as it saves the serialization of activations and completions.
The major downside of removing the workflow threads is that we now have to set 3 global (process-wide) handlers:
- A promise hook to capture stack traces in workflows
- A `Error.prepareStackTrace` handler to map the stack traces from the bundled code to the original sources
- An unhandled rejection handler to catch unhandled promise rejections in workflows and turn them into activation (workflow task) failures

The first global handler is reasonable, the second is suboptimal and the third is pretty critical IMHO, I will try to get rid of it.

Initial perf tests have shown memory reduction of about 2 thirds and CPU reduction of 50%.

TODOs to address before marking this ready for review:

- [x] Freeze the few shared global variables (`__TEMPORAL__`, `Date`, etc...)
- [x] Better support for multiple workers in the same process using the global unhandled rejection handler
- [ ] Publish some benchmark results


docs/activation-with-context-reuse.mermaid
```diff
@@ -0,0 +1,29 @@
+%% Activation diagram for Worker with `reuseV8Context` option on
```
bergundy, line 1: This is no longer relevant, I ended up putting the threads back in place.

packages/test/src/integration-tests.ts
```diff
@@ -82,6 +76,8 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
   }
 
   _test.before(async (t) => {
+    // We don't want AVA to whine about unhandled rejections thrown by workflows
+    process.removeAllListeners('unhandledRejection');
```
bergundy, line 80: This can be removed

packages/test/src/integration-tests.ts
```diff
@@ -259,8 +255,8 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     t.is(
       cleanOptionalStackTrace(err.cause.cause.stack),
       dedent`
-    Error: Fail me
-        at Activity.throwAnError (test/src/activities/index.ts)
+      Error: Fail me
+          at Activity.throwAnError (test/src/activities/index.ts)
```
bergundy, line 259: Can reset this change (and the one below)

packages/test/src/test-sinks.ts
```diff
@@ -106,15 +106,29 @@ if (RUN_INTEGRATION_TESTS) {
...
-      { info, fn: 'error.throwSync', counter: 2 },
-      { info, fn: 'error.throwAsync', counter: 3 },
-    ]);
+    /**
+     * Remove the `now()` function from WorkflowInfo.unsafe since it conditionally appears when running with the
+     * `reuseV8Context` worker option.
+     */
+    function sanitizeInfoUnsafe(info: WorkflowInfo) {
+      return { ...info, unsafe: { isReplaying: info.unsafe.isReplaying } };
+    }
```
bergundy, line 115: No need for this anymore

packages/test/src/test-workflow-unhandled-rejection-crash.ts
```diff
@@ -1,6 +1,11 @@
+/**
+ * This is not testable with reusable contexts since the unhandled rejection handler is set on the main Node.js thread.
```
bergundy, line 2: Not relevant anymore

packages/worker/src/worker-options.ts
```diff
@@ -284,6 +284,22 @@ export interface WorkerOptions {
    */
   debugMode?: boolean;
 
+  /**
+   * Toggle whether to reuse a single V8 contexts for the workflow sandbox.
```
bergundy, line 288:
```suggestion
   * Toggle whether to reuse a single V8 context for the workflow sandbox.
```

packages/worker/src/worker-options.ts
```diff
@@ -284,6 +284,22 @@ export interface WorkerOptions {
...
+   *
+   * Context reuse significantly decreases the amount of resources taken up by workflows.
+   * From running basic stress tests we've observed 2/3 reduction in memory usage and 1/3 to 1/2 in CPU usage with this
+   * feature turned on.
+   *
+   * This feature is still considered experimental and requires further testing before we can consider making it stable
```
bergundy, line 294:
```suggestion
   * This feature is considered experimental and requires further testing before we can consider making it stable
```

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -0,0 +1,269 @@
...
+  return Object.freeze(object);
+}
+
+/**
+ * Best effort to catch unhandled rejections from workflow code.
+ * Crash the process if the rejection did not originate from a workflow.
+ * TODO(bergundy): Clean this up a bit, it uses a problematic static `ReusableVMWorkflowCreator.workflowByRunId`.
+ * Running multiple workers in the same process will be hard with this approach.
+ */
+export function setUnhandledRejectionHandler(): void {
```
bergundy, line 41: TODO: can use a common implementation now that we run in a worker thread again

packages/workflow/src/internals.ts
```diff
@@ -695,8 +696,12 @@ export class Activator implements ActivationHandler {
   }
 }
 
+export function maybeGetActivator(): Activator | undefined {
+  return (globalThis as any).__TEMPORAL_ACTIVATOR__;
+}
```
bergundy, line 701: This can be refactored to avoid referencing `__TEMPORAL_ACTIVATOR__` in some of the other places.

cretz:
Note, I am leaving this review up to others, but let me know explicitly if my review is required.

.github/workflows/ci.yml
```diff
@@ -15,8 +15,14 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
+        reuse-v8-context: [false]
```
Sushisource, line 18: Why bother parameterizing this bit if it's only false?

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -0,0 +1,250 @@
...
+   *
+   * Overridable for test purposes.
+   */
+  protected injectConsole(): void {
+    this.context.console = {
+      log: (...args: any[]) => {
+        const { info } = this.context.__TEMPORAL_ACTIVATOR__;
+        if (info.isReplaying) return;
+        console.log(`[${info.workflowType}(${info.workflowId})]`, ...args);
+      },
```
Sushisource, line 192: Could just wrap this in a little closure that takes the method to call on `console` since everything else is identical

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  return Object.freeze(object);
+}
+
+/**
+ * A WorkflowCreator that creates VMWorkflows in the current isolate
+ */
+export class ReusableVMWorkflowCreator implements WorkflowCreator {
+  /**
+   * TODO(bergundy): Get rid of this static state somehow
```
Sushisource, line 39: I dunno how you could ever get rid of `unhandledRejectionHandlerHasBeenSet` since it's for a process-wide handler.

The map, maybe.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -0,0 +1,364 @@
...
+    if (!this.installed) {
+      this.overridePrepareStackTrace();
+      this.setPromiseHook();
+      this.installed = true;
+    }
+  }
+
+  /**
+   * Unset all installed global hooks
+   * TODO(bergundy): Where should this be called?
```
Sushisource, line 119: Refcount workers/runtimes?

The process-wide handlers I think are another good example of why we're really more like a process that hosts user code than a library they plug into their app. But, obviously that's not gonna change in this PR.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -0,0 +1,364 @@
...
+      });
+      return `${err}\n${converted.join('\n')}`;
+    };
+  }
+
+  private setPromiseHook(): void {
+    // Track Promise aggregators like `race` and `all` to link their internally created promises
+    let currentAggregation: Promise<unknown> | undefined = undefined;
+
+    // This also is set globally for the isolate which unless the worker is run in debug mode is insignificant
```
Sushisource, line 181:
```suggestion
    // This also is set globally for the isolate, which is insignificant unless the worker is run in debug mode 
```

Sushisource (approved):

.github/workflows/ci.yml
mjameswh: Because of the `include` block at line 21... If this was not set to false explicitly first, then reuse-v8-context would always be set to true for the ubuntu-latest/16 case. Now we get both ubuntu-latest/16/false AND ubuntu-latest/16/true.

packages/worker/src/worker.ts
```diff
@@ -90,6 +90,7 @@ import { workflowLogAttributes } from './workflow-log-interceptor';
 import { defaultWorflowInterceptorModules, WorkflowCodeBundler } from './workflow/bundler';
 import { Workflow, WorkflowCreator } from './workflow/interface';
 import { ThreadedVMWorkflowCreator } from './workflow/threaded-vm';
+import { ReusableVMWorkflowCreator as ReusableVMWorkflowCreator } from './workflow/reusable-vm';
```
mjameswh, line 93: nit: Dont alias to same name

packages/worker/src/worker.ts
```diff
@@ -475,12 +476,16 @@ export class Worker {
     // This isn't required for vscode, only for Chrome Dev Tools which doesn't support debugging worker threads.
     // We also rely on this in debug-replayer where we inject a global variable to be read from workflow context.
     if (compiledOptions.debugMode) {
```
mjameswh, line 478: This condition is not new, but I'm hesitant about it.

1. I dont like that we bind this condition to `debugMode` directly. `debugMode` is like an "higher abstraction level" option, implicitely affecting multiple lower (or internal) options, but it should not be exposed/used after `compileWorkerOptions()`. Shouldn't we avoid the Threaded VM model whenever `workflowThreadPoolSize = 1`? Or should we support some other magic value, let say `workflowThreadPoolSize = 0`, that disable the use of the Threaded VM model?
2. We have almost no tests for `debugMode = true`, and absolutely none coverring the `debugMode && reuseV8Context` case. We should definitely try to cover these sooner than later.

packages/worker/src/worker-options.ts
```diff
@@ -337,7 +353,7 @@ export type WorkerOptionsWithDefaults = WorkerOptions &
      *
      * This value is not exposed at the moment.
      *
-     * @default 8
+     * @default 1 for reuseV8Context, otherwise 8
```
mjameswh, line 356: Is `workflowThreadPoolSize > 1` valid with `reuseV8Context = true`? Might it be pertinent to set default to something else than 1 for `reuseV8Context = true`?

packages/workflow/src/cancellation-scope.ts
```diff
@@ -185,7 +185,14 @@ export class CancellationScope {
 /**
  * This is exported so it can be disposed in the worker interface
```
mjameswh, line 186: Outdated comment?

packages/workflow/src/workflow.ts
```diff
@@ -843,18 +843,7 @@ export function workflowInfo(): WorkflowInfo {
...
-  } catch (err: any) {
-    // Use string comparison in case multiple versions of @temporalio/common are
-    // installed in which case an instanceof check would fail.
-    if (err.name === 'IllegalStateError') {
-      return false;
-    } else {
-      throw err;
-    }
-  }
+  return maybeGetActivator() !== undefined;
```
mjameswh, line 846: Great!

packages/workflow/src/internals.ts
```diff
@@ -100,6 +101,7 @@ export const LATEST_INTERNAL_PATCH_NUMBER = 1;
  * Implements handlers for all workflow activation jobs.
  */
 export class Activator implements ActivationHandler {
+  readonly moduleCache = new Map<string, unknown>();
```
mjameswh, line 104: Comment on this one would be useful, especially given that it is used only in `reusable-vm.ts`... Also, could we reasonably (ie. without going out of our way) to avoid initializing a Map unless we actually needs it? Maybe a lazy getter?

packages/worker/src/workflow/vm.ts
```diff
@@ -156,10 +85,11 @@ export class VMWorkflowCreator implements WorkflowCreator {
       throw new IllegalStateError('Isolate context provider was destroyed');
     }
     let context;
+    const globals = { AsyncLocalStorage, assert, __webpack_module_cache__: {} };
     if (this.hasSeparateMicrotaskQueue) {
```
mjameswh, line 89: nit: At this point, we can actually remove this condition, since is <14.18 is no longer supported anyway.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -0,0 +1,250 @@
...
+      {},
+      {
+        get: (_, p) => {
+          // Try the shared modules first
+          const sharedModule = sharedModules.get(p);
+          if (sharedModule) {
+            return sharedModule;
+          }
+          const moduleCache = this.context.__TEMPORAL_ACTIVATOR__?.moduleCache;
+          if (moduleCache != null) return moduleCache.get(p);
```
mjameswh, line 82: Maybe `return moduleCache?.get(p)`?

packages/worker/src/workflow/reusable-vm.ts
mjameswh: I'd move this to vm-shared.ts and use it for regular vm too. Then we can remove `ActivationContext` everywhere.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -0,0 +1,364 @@
...
+  bundleFilenameToSourceMapConsumer = new Map<string, SourceMapConsumer>();
+  origPrepareStackTrace = Error.prepareStackTrace;
+  private stopPromiseHook = () => {
+    // noop
+  };
+  installed = false;
+
+  async addWorkflowBundle(workflowBundle: WorkflowBundleWithSourceMapAndFilename): Promise<void> {
+    const sourceMapConsumer = await new SourceMapConsumer(workflowBundle.sourceMap);
+    this.bundleFilenameToSourceMapConsumer.set(workflowBundle.filename, sourceMapConsumer);
```
mjameswh, line 99: Is there a risk that we might register the same bundle filename multiple time? And more important... a risk that we could remove a filename multiple time? (haven't yet completed review of this class, so unsure if this can happen or not)

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -0,0 +1,364 @@
...
+      return `${err}\n${converted.join('\n')}`;
+    };
+  }
+
+  private setPromiseHook(): void {
+    // Track Promise aggregators like `race` and `all` to link their internally created promises
+    let currentAggregation: Promise<unknown> | undefined = undefined;
+
+    // This also is set globally for the isolate which unless the worker is run in debug mode is insignificant
+    if (promiseHooks) {
```
mjameswh, line 182: This condition is technically no longer required since we do not support <14.18

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -0,0 +1,364 @@
...
+        init: (promise: Promise<unknown>, parent: Promise<unknown>) => {
+          // Only run in workflow context
+          const activator = getActivator(promise);
+          if (!activator) return;
+          const store = activator.promiseStackStore;
+          // TODO: hide this somehow: defineProperty + symbol
+          (promise as any).runId = activator.info.runId;
+          // Reset currentStackTrace just in case (it will be set in `prepareStackTrace` above)
+          this.currentStackTrace = undefined;
+          // To see the full stack replace with commented line
```
mjameswh, line 194: Comment not pertinent at this line

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  private setPromiseHook(): void {
+    // Track Promise aggregators like `race` and `all` to link their internally created promises
+    let currentAggregation: Promise<unknown> | undefined = undefined;
+
+    // This also is set globally for the isolate which unless the worker is run in debug mode is insignificant
+    if (promiseHooks) {
+      // Node >=16.14 only
+      this.stopPromiseHook = promiseHooks.createHook({
```
mjameswh, line 184: I'd really like to measure the memory/performance penalty we get from this. Seems like it would be possible to delay some of the string processing until we actually needs it, if ever.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -0,0 +1,250 @@
...
+    public readonly isolateExecutionTimeoutMs: number
+  ) {
+    if (!ReusableVMWorkflowCreator.unhandledRejectionHandlerHasBeenSet) {
+      setUnhandledRejectionHandler((runId) => ReusableVMWorkflowCreator.workflowByRunId.get(runId));
+      ReusableVMWorkflowCreator.unhandledRejectionHandlerHasBeenSet = true;
+    }
+
+    // https://nodejs.org/api/vm.html#vmcreatecontextcontextobject-options
+    // microtaskMode=afterEvaluate was added in 14.6.0
+    this.hasSeparateMicrotaskQueue = gte(process.versions.node, '14.6.0');
```
mjameswh, line 69: Can remove, <14.18 is no longer supported

mjameswh:
Neat! Left several suggestions, but no show stopper.

I can actually take it from there if you prefer.

mjameswh (approved):

packages/worker/src/worker-options.ts
bergundy: There's not much value in it AFAICT.
It's better to create less threads until we see a need for more, it'll make startup faster.

packages/worker/src/worker.ts
bergundy: Probably due to somre refactor, will fix.

packages/worker/src/worker.ts
bergundy: We have test-worker-debug-mode that tests this path.
We can't avoid the worker thread, we need it for isolation of the process global hooks.

bergundy: For now, I'd rather not expose these advanced options to users and support debugging in chrome devtools.

packages/worker/src/workflow/reusable-vm.ts
bergundy: We should probably error on unsupported node versions.

packages/worker/src/workflow/reusable-vm.ts
bergundy: Sure.

packages/worker/src/workflow/reusable-vm.ts
bergundy: Yeah.. I can dedupe this a bit.

packages/worker/src/workflow/vm-shared.ts
bergundy: The filename contain a sha sum, I don't think there's much concern here.

packages/worker/src/workflow/vm-shared.ts
bergundy: It's not super critical since we typically install these in a separate thread which is cleaned up after worker shutdown.
The only case that isn't covered is debug mode where it's insignificant.
I'll add a comment in the code.

packages/worker/src/workflow/vm-shared.ts
bergundy: Let's remove all of these in another PR where we add assertions for minimum node version.

packages/worker/src/workflow/vm-shared.ts
bergundy: Sure, I'll leave that up to you.
In any case I feel like it's not super critical ATM.

packages/workflow/src/cancellation-scope.ts
bergundy: Thanks

packages/workflow/src/internals.ts
bergundy: I don't think the initialization here is critical perf wise.
I'm sure there are many optimizations we can make before this becomes an issue.

I'll add a comment.

bergundy:
We need to update the branch protection rules to enforce the new test names

PR #955, opened 2022-11-09
fix(workflow): do encode memos on continueAsNew

mjameswh:
## What changed

- Changed the type of `ContinueAsNewOptions.memo` from `Record<string, any>` to `Record<string, unknown>`, so that it matches `BaseWorkflowOptions.memo`.
- Make sure that `makeContinueAsNewFunc()` properly encode `memo` if provided.

## Why

- `makeContinueAsNewFunc()` was previously copying new `memo` field as-is, without encoding them to Payload, resulting in a failure due to invalid data in request to core.

## Checklist

- Added integration tests


bergundy (approved):
Thanks!

PR #960, opened 2022-11-10
fix(client): Add missing field in schedules

mjameswh:
## What changed

- Handle field `timezoneName` in schedule Proto to TypeScript decoding

## Why

- Missed that field in original PR


bergundy (approved):
I'm blindly approving this one, I trust that you know exactly what was missing.

PR #961, opened 2022-11-10
Use https submodule url for easier clones in CI

Sushisource:
Git url is hard to make work in workflows ci


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "sdk-core"]
 	path = packages/core-bridge/sdk-core
-	url = git@github.com:temporalio/sdk-core.git
+	url = ../sdk-core
```
cretz, line 3: I think this should be HTTP. I changed to HTTP in SDK Python so people had better experience.

bergundy (approved):

PR #963, opened 2022-11-10
chore: Update sdk-core

mjameswh:
## What changed

- Updated sdk-core to latest
- Introduce `TelemetryOptions.logging.filter` and `TelemetryOptions.tracing.filter`, matching similar changes in core-sdk. Backward compatibility for legacy options `TelemetryOptions.logging.forward.level` and `TelemetryOptions.tracingFilter` is provided, though uasge of these properties is now deprecated.
- Set `maxCachedWorkflows` to 0 in some tests to avoid waiting for sticky execution timeout in worker stop/start sequences

## Checklist

- DEPRECATIONS: the `TelemetryOptions` interface has changed quite a bit, matching similar changes in core-sdk. Using the appropriate new options is highly recommended. Backward compatibility for legacy options is provided, to the extent possible, but these legacy options have been deprecated. As these are specialized APIs, there are probably few affected end users.
- Doc: need to update docs and samples to use new `TelemetryOptions` format.


packages/core-bridge/index.d.ts
```diff
@@ -63,12 +63,13 @@ export interface ConsoleLogger {
...
 export interface ForwardLogger {
   forward: {
+    // FIXME: This is no longer supported in core
     /**
      * What level, if any, logs should be forwarded from core at
      *
      */
     // These strings should match the log::LevelFilter enum in rust
-    level: LogLevel;
+    // level: LogLevel;
```
Sushisource, line 72: FYI this can be supported the way it looks still, by just constructing an appropriate filter string.

Then we could deprecate it and add a filter string field.

mjameswh: Definitely. TBH I'd personally go for what @cretz did in Python (only support level, not a filter string), but that might be a more breaking change.

Sushisource: I think you flipped what you meant there, but yeah I agree. I suspect very few people use/care about this, so switching to filter string makes sense to me.

packages/core-bridge/src/conversions.rs
```diff
@@ -176,29 +176,19 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
...
-        if let Some(tf) = js_optional_value_getter!(cx, self, "tracingFilter", JsString) {
-            telemetry_opts.tracing_filter(tf);
-        }
         telemetry_opts.no_temporal_prefix_for_metrics(
             js_optional_value_getter!(cx, self, "noTemporalPrefixForMetrics", JsBoolean)
                 .unwrap_or_default(),
         );
+
+        // FIXME: Default value?
+        let tracing_filter = js_optional_value_getter!(cx, self, "tracingFilter", JsString).unwrap_or("temporal_sdk_core=WARN".to_owned());
```
bergundy, line 185: I think we typically set the default from JS.

packages/core-bridge/src/conversions.rs
```diff
@@ -264,7 +254,10 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
                     } else {
                         Default::default()
                     };
-                telemetry_opts.tracing(TraceExporter::Otel(OtelCollectorOptions { url, headers }));
+                telemetry_opts.tracing(TraceExportConfig {
+                    filter: "".to_string(),
```
bergundy, line 258: Should this be set?

packages/test/src/test-patch-and-condition.ts
```diff
@@ -14,6 +14,7 @@ if (RUN_INTEGRATION_TESTS) {
     const worker1 = await Worker.create({
       taskQueue: 'patch-in-condition',
       workflowsPath: require.resolve('./workflows/patch-and-condition-pre-patch'),
+      maxCachedWorkflows: 0,
```
bergundy, line 17: Curious why you set this..

bergundy (approved):
Let's bring the log level back but otherwise lgtm

packages/test/src/test-patch-and-condition.ts
mjameswh: This fixes a very subtile failure introduced by this [apparently innocuous PR](https://github.com/temporalio/sdk-core/commit/c72322f4ae317f0745f65e89fa4c5912ba0e2498)...

Essentially, once worker1 get stopped, WFT will still be sent to worker1's sticky queue. Before, worker2 would register to the very same sticky queue and thus, would _immediately_ be able to accept the WFT, without delay. But now that sticky queue names are randomized, it the WFT will have to wait 10 seconds in the sticky queue before it gets moved to the regular queue so that it can be picked up by worker2. And then the same goes for worker2 to worker3.

Now, the workflow use a 5 second timer, and then dies. That timer still tick even while the WFT is waiting in the queue, so the workflow will now completes in worker2, rather than staying alive until worker3.

I could have avoid this by increasing the sleep time, but that would have increased the total execution time of this test. By disabling caching, we both resolve the issue without side effect on execution time.

packages/core-bridge/index.d.ts
mjameswh: The thing is that unless I'm missing something, the rust side now accept only a filter string, not tracingFilter+level. We can easily build a filter string from a level, but then, we would have to ignore `telemetryOptions.tracingFilter` in the forward case.

mjameswh: Essentially, we now need a filter string on the logger (either Console or Forward) and a filter on the trace exporter.

mjameswh: If had to design an API for this at this moment, _without considering backward compatibility_, it would looks something like this:

```
export interface TelemetryOptions {
  logging: {
    filter: string,
    mode: 'console' | 'forward'
  },

  tracing: {
    filter: string;
    ...
  }
}

function makeFilterString(sdkLevel: Level, otherLevel: Level = 'WARN') { ... }
```

That is roughly how [Python defines this](https://github.com/temporalio/sdk-python/pull/191/files#diff-0d990372bfa478206aff7cbb03297d8f4423f41304104e26474d159eef870c65R27-R43).

bergundy: I agree, it's better than what we have now.
I wish I'd used `mode` instead of having the nested `console` and `forward` objects.

mjameswh: But given the current API, I would propose this instead: 

```
export interface TelemetryOptions {
  /** @deprecated */
  tracingFilter?: string

  logging: {
    // If undefined, then default to these values, in that order:
    //   1. makeFilterString(forward.level, forward.level)  [if forward.level exists]
    //   2. TelemetryOptions.tracingFilter  [if it exists]
    //   3. makeFilterString('INFO', 'INFO')  [if everything else is undefined]
    filter?: string,
    console: {}    // Either 'console' or 'forward', not both
    forward: {
      /** @deprecated */
      level: Level;
    }
  },

  tracing: {
    // If undefined, then default to these values, in that order:
    //   1. TelemetryOptions.tracingFilter  [if it exists]
    //   2. makeFilterString('INFO', 'INFO')  [if everything else is undefined]
    filter?: string;
    ...
  }
}

function makeFilterString(sdkLevel: Level, otherLevel: Level = 'WARN') { ... }
```

mjameswh: It's complicated though. Depends on how far we need to go in backward compatibility...

mjameswh: And in any cases, these backward compatibility things would be handled in TS. Bridge would accept a `CompiledTelemetryOptions`, expecting something more in line with my first proposition.

packages/core-bridge/index.d.ts
```diff
@@ -129,9 +131,10 @@ export interface TelemetryOptions {
    * A string in the env filter format specified here:
    * https://docs.rs/tracing-subscriber/0.2.20/tracing_subscriber/struct.EnvFilter.html
    *
-   * Which determines what tracing data is collected in the Core SDK
+   * Which determines what tracing data is collected in the Core SDK.
    *
-   * @default `temporal_sdk_core=WARN`
+   * @default `makeTracingFilterString('INFO', 'INFO')`
+   * @deprecated Use either `logging.filter` or `traccing.filter` instead
```
Sushisource, line 137:
```suggestion
   * @deprecated Use either `logging.filter` or `tracing.filter` instead
```

packages/core-bridge/src/runtime.rs
```diff
@@ -211,12 +211,18 @@ pub fn start_bridge_loop(channel: Arc<Channel>, receiver: &mut UnboundedReceiver
...
-                    let worker = init_worker(config, client.into_inner());
-                    let (tx, rx) = unbounded_channel();
-                    tokio::spawn(start_worker_loop(worker, rx, channel.clone()));
-                    send_result(channel.clone(), callback, |cx| {
-                        Ok(cx.boxed(RefCell::new(Some(WorkerHandle { sender: tx }))))
-                    });
+                    match init_worker(&runtime, config, client.into_inner()) {
+                        Ok(worker) => {
+                            let (tx, rx) = unbounded_channel();
+                            tokio::spawn(start_worker_loop(worker, rx, channel.clone()));
```
Sushisource, line 217: Probably not strictly necessary, but for consistency you might want to use `runtime.handle().spawn()` here

Sushisource (approved):
I would just double check that when you turn on debug logging from core you see lots of stuff from inside core's internals (in console / forward mode) just to make sure the tracing settings are applied properly.

packages/core-bridge/index.d.ts
```diff
@@ -145,17 +148,48 @@ export interface TelemetryOptions {
...
+         * which determines the verboseness of logging output.
+         *
+         * You can use {@link Runtime.makeTracingFilterString()} to easily build a correctly formatted filter
+         * string based on desired log level for Core SDK and other native packages.
+         *
+         * @default `console`
+         */
+        mode?: 'console' | 'forward';
+      }
+    | Logger;
```
bergundy, line 176: Looks like `Logger` is deprecated but isn't marked as such.
I'm considering marking all of the telemetry options experimental now but it might be too late for that.

packages/core-bridge/index.d.ts
```diff
@@ -145,17 +148,48 @@ export interface TelemetryOptions {
...
+        /**
+         * A string in (env filter format)[https://docs.rs/tracing-subscriber/0.2.20/tracing_subscriber/struct.EnvFilter.html]
+         * which determines the verboseness of logging output.
+         *
+         * You can use {@link Runtime.makeTracingFilterString()} to easily build a correctly formatted filter
+         * string based on desired log level for Core SDK and other native packages.
+         *
+         * @default `console`
+         */
+        mode?: 'console' | 'forward';
```
bergundy, line 174: I don't love that we're now inconsistent with `tracing` and `metrics`.
Here's we're using `mode` and in the other options, we're using a union type.

I'd make these consistent, probably preserving the structure we had previously even if it means we'll have empty options in `console` and `forward`.

packages/core-bridge/index.d.ts
bergundy: Based on my comment above, I'm leaning towards preserving the structure we had before instead of redefining logging altogether.

packages/worker/src/runtime.ts
```diff
@@ -56,17 +55,29 @@ export interface RuntimeOptions {
...
+ *     // ...
+ *    },
+ *    tracing: {
+ *     filter: makeTracingFilterString('DEBUG', 'INFO');
+ *     // ...
+ *    }
+ *  }
+ * ```
+ */
+export function makeTracingFilterString(coreLevel: native.LogLevel, otherLevel: native.LogLevel = 'WARN'): string {
```
bergundy, line 79: nit: use a single options arg instead of 2 separate args so it's easier to understand what each level refers to.

packages/core-bridge/index.d.ts
mjameswh: >Looks like Logger is deprecated but isn't marked as such.

I intentionally deprecated the _properties_ inside *Logger interfaces (`console` and `forward`), rather than the _interfaces_ themselves, because I'm worried that otherwise, the compiler might incorrectly match `logger: {}` to the deprecated `ConsoleLogger` type and report deprecation.

mjameswh: >I'm considering marking all of the telemetry options experimental now but it might be too late for that.

I don't think it is too late for that, and I see the value of marking TelemetryOptions as such. Will do.

packages/core-bridge/index.d.ts
mjameswh: Good catch

packages/core-bridge/index.d.ts
mjameswh: Reverted to `console: {} | forward: {}` instead of `mode: 'console' | 'forward'`

packages/core-bridge/src/runtime.rs
```diff
@@ -147,7 +147,7 @@ pub fn start_bridge_loop(channel: Arc<Channel>, receiver: &mut UnboundedReceiver
                     // returned client directly at the moment, when we repurpose the client to be
                     // used by a Worker, `init_worker` will attach the correct metrics meter for
                     // us.
-                    tokio::spawn(async move {
+                    core_runtime.tokio_handle().spawn(async move {
```
bergundy, line 150: Can we not use `tokio::spawn` anymore? Doesn't this call reference the "current" tokio runtime?

packages/core-bridge/src/conversions.rs
```diff
@@ -176,38 +177,25 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
...
-                    Err(err) => cx.throw_type_error(format!(
-                        "Invalid telemetryOptions.logging.forward.level: {}",
-                        err
-                    ))?,
-                }
+            let filter = js_value_getter!(cx, logging, "filter", JsString);
+            if get_optional(cx, logging, "console").is_some() {
+                telemetry_opts.logging(Logger::Console { filter });
+            } else if get_optional(cx, logging, "forward").is_some() {
+                telemetry_opts.logging(Logger::Forward { filter });
```
bergundy, line 189: We don't support the deprecated `level` option anymore, please fix.

packages/core-bridge/src/conversions.rs
```diff
@@ -176,38 +177,25 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
 
     fn as_telemetry_options(&self, cx: &mut FunctionContext) -> NeonResult<TelemetryOptions> {
         let mut telemetry_opts = TelemetryOptionsBuilder::default();
-
-        if let Some(tf) = js_optional_value_getter!(cx, self, "tracingFilter", JsString) {
-            telemetry_opts.tracing_filter(tf);
-        }
```
bergundy, line 182: Didn't we deprecate `tracingFilter`? Looks like it is ignored here.

bergundy (approved):
Nice that we managed to do all of this without breaking changes.

I left a couple of comments that should be addressed and tested before merging but other than that LGTM.

packages/core-bridge/src/conversions.rs
mjameswh: Backward compatibility is dealt in TypeScript land (runtime.ts). Same for deprecated level option.

packages/core-bridge/src/runtime.rs
Sushisource: This is more consistent. It'd be possible after a refactor for a different tokio runtime to be the current context for the thread. The code being like this makes it clear that the same runtime that lives in `CoreRuntime` and is used for the overall loop should be used for the operations within it.

mjameswh: > Can we not use tokio::spawn anymore? Doesn't this call reference the "current" tokio runtime?

Yes, we can... But I agree with Spencer that explicitly using the tokio runtime handle for this makes the code easier to understand for Tokio neophytes.

packages/core-bridge/src/conversions.rs
mjameswh: Backward compatibility and default values are handled in [runtime.ts](https://github.com/temporalio/sdk-typescript/pull/963/files#diff-b144ddafebfd1e9ede8c1761ec8b821fd2611f017a5f0d88d1faac7362752e3eR216-R217).

If present, the `level` property is converted to a filter string. The rust code only sees see compiled options, matching the new format.

mjameswh: Note that since we now have a single filter for logging, if both `forward.level` and `tracingFilter` were previously defined, then the `tracingFilter` will be silently ignored. AFAIK, we can't do any better there.

PR #965, opened 2022-11-13
chore: Use `import/order` eslint rule

lorensr:
Roey suggested this automated solution in lieu of running VS Code's `Organize imports` command when working in a file.

Left the default order, but it's customizable if we want: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

ESLint recommends not using VS Code's `Organize imports` command (which works differently): https://github.com/import-js/eslint-plugin-import/issues/1742#issuecomment-968376099


package.json
```diff
@@ -31,7 +31,7 @@
     "wait-namespace": "node ./scripts/wait-on-temporal.mjs",
     "lint": "eslint packages/*/src --ext .ts --no-error-on-unmatched-pattern && prettier --end-of-line auto --check .",
     "lint.prune": "ts-prune --error -p tsconfig.prune.json --ignore \"used in module\" --skip \".d.ts\"",
-    "format": "prettier --write .",
+    "format": "eslint packages/*/src --ext .ts --no-error-on-unmatched-pattern --fix && prettier --write .",
```
bergundy, line 34: nit: not related to formatting

bergundy (approved):

package.json
lorensr: maybe this?

https://github.com/temporalio/sdk-typescript/pull/965/commits/6e02630cb50f97d0f77661327969cc3f91cd91d7

looking for one command to run locally before committing

mjameswh:
@loren Can we try adding `@temporalio/*` as "internal packages", and have them appears after external packages?

I think that would be something like this:

```
# .eslintrc.yml
settings:
  "import/internal-regex": ^@temporalio/
  "import/order": [
    "error",
    {
      "groups": [
        "builtin",
        "external",
        "internal",
        "parent",
        "sibling",
        "index"
      ]
    }
  ]
 ```

mjameswh:
I added definition of '@temporalio/*' as "internal packages".

lorensr:
Thanks! ☺️

PR #968, opened 2022-11-14
Log errors for VS Code debugger

laurakwhit:
## What was changed

Get data from responses in order to log error messages.

## Why?

https://github.com/temporalio/vscode-debugger-plugin/pull/22#discussion_r1019718787

## Checklist

1. Closes: n/a 

2. How was this tested: n/a

3. Any docs updates needed? n/a

### DIFF
--- a/packages/worker/src/debug-replayer/client.ts
+++ b/packages/worker/src/debug-replayer/client.ts
@@ -44,9 +44,9 @@ export class Client {
       request.once('response', resolve);
     });
     if (response.statusCode !== 200) {
-      const responseBody = await Client.readAll(response);
       let message = response.statusMessage;
       try {
+        const responseBody = await Client.readAll(response);
         message = JSON.parse(responseBody.toString())?.error ?? message;
       } catch {
         // ignore
@@ -70,9 +70,9 @@ export class Client {
       request.once('response', resolve);
     });
     if (response.statusCode !== 200) {
-      const responseBody = await Client.readAll(response);
       let message = response.statusMessage;
       try {
+        const responseBody = await Client.readAll(response);
         message = JSON.parse(responseBody.toString())?.error ?? message;
       } catch {
         // ignore
--- a/packages/worker/src/debug-replayer/client.ts
+++ b/packages/worker/src/debug-replayer/client.ts
@@ -45,7 +45,12 @@ export class Client {
     });
     if (response.statusCode !== 200) {
       const responseBody = await Client.readAll(response);
-      const message = JSON.parse(responseBody.toString())?.error ?? response.statusMessage;
+      let message = response.statusMessage;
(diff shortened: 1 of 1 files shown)


packages/worker/src/debug-replayer/client.ts
```diff
@@ -41,10 +41,19 @@ export class Client {
...
+      request.once('response', (res) => {
+        let body = [];
+        res.on('data', function (chunk) {
+          body.push(chunk);
+        });
+        res.on('end', function () {
+          res.body = JSON.parse(body.toString());
+          resolve(res);
+        });
+      });
```
bergundy, line 53: You can use the response as an async iterator if you need the body.
In some cases we don't want to consume the body as JSON, I would only consume it on error.

laurakwhit: That makes total sense. Updated ✅

packages/worker/src/debug-replayer/client.ts
```diff
@@ -44,7 +44,9 @@ export class Client {
       request.once('response', resolve);
     });
     if (response.statusCode !== 200) {
-      throw new Error(`Bad response code from VS Code: ${response.statusCode}`);
+      const responseBody = await Client.readAll(response);
+      const message = JSON.parse(responseBody.toString())?.error ?? response.statusMessage;
```
bergundy, line 48: Oh I didn't realize we already had this `readAll` helper..

One last thing here.. Wrap the `JSON.parse` statement in `try .. catch`.
It's not safe to assume that all response bodies would contain valid JSON.

laurakwhit: Ah, good point. I think we can just ignore the error in that case? e.g.

```js
      let message = response.statusMessage;
      try {
        message = JSON.parse(responseBody.toString())?.error ?? message;
      } catch {
        // ignore
      }
```

packages/worker/src/debug-replayer/client.ts
```diff
@@ -44,7 +44,14 @@ export class Client {
       request.once('response', resolve);
     });
     if (response.statusCode !== 200) {
-      throw new Error(`Bad response code from VS Code: ${response.statusCode}`);
+      const responseBody = await Client.readAll(response);
```
bergundy, line 47: I hate to do request another change but can you please move reading the body to the try block?
Not super critical, I'm approving the PR and leaving it up to you.

bergundy (approved):

packages/worker/src/debug-replayer/client.ts
laurakwhit: Done, thank you ✅

bergundy:
Thanks @laurakwhit!

PR #974, opened 2022-11-18
chore: Refactor mass history replayer API (Part 1)

mjameswh:
## What changed

- In client package, introduce `WorkflowHandle.fetchHistory()` which returns the complete history for a workflow. That function might be useful in testing and replaces function `fetchWorkflowHistory`.
- The iterable returned by `WorkflowClient.list()` can now be easily transformed to an iterable of Workflow histories, using `WorkflowClient.list().intoHistories()`. Multiple Workflow histories are fetched in parallel (with a concurency of 5 by default).
- `WorkflowClient.list().intoHistories()` is now the proper way to provide histories to `Worker.runReplayHistories()` when those histories have to be fetched from an actual server. The previous experimental interface `ReplayExecutions` is no longer accepted.

### DIFF
--- a/packages/client/src/iterators-utils.ts
+++ b/packages/client/src/iterators-utils.ts
@@ -1,5 +1,6 @@
 import { EventEmitter, on, once } from 'node:events';
 import { AbortController } from 'abort-controller';
+
 export interface MapAsyncOptions {
   /**
    * How many items to map concurrently. If set to less than 2 (or not set), then items are not mapped concurrently.
--- a/packages/client/src/iterators-utils.ts
+++ b/packages/client/src/iterators-utils.ts
@@ -1,0 +1,56 @@
+import { EventEmitter, on } from 'node:events';
+
+export async function* mapAsyncIterable<A, B>(
+  source: AsyncIterable<A>,
+  func: (val: A) => Promise<B>
+): AsyncIterable<B> {
+  for await (const x of source) {
+    yield func(x);
+  }
+}
+
+export async function* raceMapAsyncIterable<A, B>(
+  source: AsyncIterable<A>,
+  mapFn: (val: A) => Promise<B>,
+  concurrency: number
+): AsyncIterable<B> {
+  const emiter = new EventEmitter();
+  const sourceIterator = source[Symbol.asyncIterator]();
+  let sourceIteratorDone = false;
+  let pendingPromisesCount = 0;
+  let pendingResultsCount = 0;
+
+  async function maybeStartTasks() {
+    while (!sourceIteratorDone && pendingPromisesCount + pendingResultsCount < concurrency) {
+      const val = await sourceIterator.next();
+      if (val.done) {
+        sourceIteratorDone = true;
+        return pendingPromisesCount > 0;
+      }
+
+      pendingPromisesCount++;
+      (async () => {
+        try {
+          emiter.emit('result', await mapFn(val.value));
+          pendingResultsCount++;
+        } finally {
+          pendingPromisesCount--;
+        }
+        return maybeStartTasks();
+      })().catch((e) => console.error(e)); // FIXME-JWH: Deal with exceptions
+    }
+
+    return !sourceIteratorDone || pendingPromisesCount > 0;
+  }
+
+  // The listener must be registered before we start emiting events
+  // through the emiter; otherwise, these events get silently discarded.
+  const emiterEventsIterable = on(emiter, 'result');
+
+  if (!(await maybeStartTasks())) return;
+  for await (const [res] of emiterEventsIterable) {
+    pendingResultsCount--;
+    yield res as B;
+    if (!(await maybeStartTasks())) return;
+  }
+}
--- a/packages/client/src/iterators-utils.ts
+++ b/packages/client/src/iterators-utils.ts
@@ -1,5 +1,5 @@
 import { EventEmitter, on, once } from 'node:events';
-
+import { AbortController } from 'abort-controller';
 export interface MapAsyncOptions {
   /**
    * How many items to map concurrently. If set to less than 2 (or not set), then items are not mapped concurrently.
--- a/packages/client/src/iterators-utils.ts
+++ b/packages/client/src/iterators-utils.ts
@@ -1,14 +1,32 @@
 import { EventEmitter, on } from 'node:events';
 
+/**
+ * Return an async iterable that transforms items from a source iterable by mapping each item
+ * through a mapping function.
+ *
+ * @param source the source async iterable
+ * @param mapFn a mapping function to apply on every item of the source iterable
+ */
 export async function* mapAsyncIterable<A, B>(
   source: AsyncIterable<A>,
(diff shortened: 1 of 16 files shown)


packages/client/src/iterators-utils.ts
```diff
@@ -0,0 +1,74 @@
+import { EventEmitter, on } from 'node:events';
+
+/**
+ * Return an async iterable that transforms items from a source iterable by mapping each item
+ * through a mapping function.
+ *
+ * @param source the source async iterable
+ * @param mapFn a mapping function to apply on every item of the source iterable
+ */
+export async function* mapAsyncIterable<A, B>(
```
bergundy, line 10: I definitely wrote this function in a past life..

packages/client/src/iterators-utils.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    return !sourceIteratorDone || pendingPromisesCount > 0;
+  }
+
+  // The listener must be registered before we start emiting events
+  // through the emiter; otherwise, these events get silently discarded.
+  const emiterEventsIterable = on(emiter, 'result');
+
+  if (!(await maybeStartTasks())) return;
```
bergundy, line 68: I would do this a bit differently.

Spawn `concurrency` coroutines each running:

```
for (;;) {
  await sourceIterator.next();
  ...
}
```

And create an `AbortController` (or simple promise) that you can use to signal the coroutines to stop on first error (in your code they keep consuming the iterator).

This way you can get rid of the spawning logic and keeping track of `sourceIteratorDone, pendingPromisesCount, pendingResultsCount`.

My version looks roughly like:

```ts
  const emitter = new EventEmitter();
  const controller = new AbortController();
  const emitterEventsIterable = on(emitter, 'result', { signal: controller.signal });
  const emitterError = once(emitter, 'not-an-event'); // so we can break on error.

  const consume = async () => {
    for (;;) {
      const val = await Promise.race([sourceIterator.next(), emitterError]);
      if (val.done) return;

      emitter.emit('result', await mapFn(val.value));
    }
  };

  const coroutines = Array(concurrency)
    .fill(consume)
    .map((f) => f());
  Promise.all(coroutines).then(
    () => controller.abort(),
    (err) => emitter.emit('error', err)
  );

  try {
    for await (const [res] of emitterEventsIterable) {
      yield res;
    }
  } catch (err) {
    if (err.name === 'AbortError') {
      return;
    }
    throw err;
  }

```

packages/client/src/iterators-utils.ts
```diff
@@ -0,0 +1,74 @@
...
+ * each item through a mapping function.
+ *
+ * Mapped items are returned by the resulting iterator in the order they complete processing,
+ * not the order in which the corresponding source items were obtained from the source iterator.
+ *
+ * @param source the source async iterable
+ * @param mapFn a mapping function to apply on every item of the source iterable
+ * @param concurrency maximum number of items to map concurrently
+ */
+export async function* raceMapAsyncIterable<A, B>(
```
bergundy, line 30: I'd reuse `mapAsyncIterable` and change the behavior if concurrency > 1.

bergundy (commented):
Managed to get through a couple files.
I'll review the rest next week.

packages/client/src/iterators-utils.ts
mjameswh: I like that

packages/client/src/workflow-client.ts
```diff
@@ -245,6 +253,19 @@ interface WorkflowHandleOptions extends GetWorkflowHandleOptions {
   runIdForResult?: string;
 }
 
+/**
+ * An iterable list of WorkflowExecution, as returned by {@link WorkflowClient.list}.
+ */
+interface AsyncWorkflowListIterable extends AsyncIterable<WorkflowExecutionInfo> {
+  /**
+   * Return an iterable list of histories corresponding to this iterable's WorkflowExecution.
```
bergundy, line 261:
```suggestion
   * Return an iterable of histories corresponding to this iterable's WorkflowExecutions.
```

packages/client/src/iterators-utils.ts
mjameswh: Yeah, I had considered something similar to this at some point, and it definitely look simpler...

My concern however is that I think we really need backpressure control (which, at this point, is provided by the `pendingPromisesCount + pendingResultsCount < concurrency` condition), at least for "fail fast" scenarios. Otherwise, the iterator risk consuming a very large quantity of data from the server, only to discard that data at the first error...

But let me think about that a little more, it might not be that difficult to introduce backpressure control in the pattern you suggest.

packages/client/src/workflow-client.ts
```diff
@@ -799,6 +832,22 @@ export class WorkflowClient extends BaseClient {
...
+        for (;;) {
+          const response: temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryResponse =
+            await this.client.workflowService.getWorkflowExecutionHistory({
+              nextPageToken,
+              namespace: this.client.options.namespace,
+              execution: { workflowId, runId },
+            });
+          history.push(...(response.history?.events ?? []));
+          nextPageToken = response.nextPageToken;
+          if (nextPageToken == null || nextPageToken.length == 0) break;
```
bergundy, line 847: nit.

We should probably have a lint rule for this.
```suggestion
          if (nextPageToken == null || nextPageToken.length === 0) break;
```

packages/client/src/workflow-client.ts
```diff
@@ -799,6 +832,22 @@ export class WorkflowClient extends BaseClient {
         (info as unknown as WorkflowExecutionDescription).raw = raw;
         return info;
       },
+      async fetchHistory() {
+        let nextPageToken: Uint8Array | undefined = undefined;
+        const history = Array<temporal.api.history.v1.IHistoryEvent>();
```
bergundy, line 837: We should probably call this `events`, not critical

bergundy (approved):
Nice!

packages/client/src/workflow-client.ts
mjameswh: Right. Actually copied that code elsewhere without paying too much attention, so I'll definitely add a lint rule.

packages/client/src/iterators-utils.ts
bergundy: I see your concern, definitely a valid one.
You could implement a semaphore and release the permits on consumption (which is more or less what you did here).

mjameswh: Done

bergundy: I saw the last commit.
LGTM with the exception that I'd move the semaphore implementation out of this function for better readability.
Looks like you need `Queue` and `Semaphore` classes.

mjameswh: > I'd move the semaphore implementation out of this function for better readability. Looks like you need Queue and Semaphore classes.

Will do at some later point.

PR #977, opened 2022-11-24
feat(worker): Make OTel export period configurable

mjameswh:
## What changed

- Make metrics export period for OTel configurable ([reference](https://github.com/temporalio/sdk-core/pull/442))

### DIFF
--- a/packages/core-bridge/index.d.ts
+++ b/packages/core-bridge/index.d.ts
@@ -3,6 +3,20 @@ import type { TLSConfig } from '@temporalio/common/lib/internal-non-workflow';
 
 export { TLSConfig };
 
+type Shadow<Base, New> = {
+  [K in keyof Base | keyof New]: K extends keyof Base
+    ? K extends keyof New
+      ? Base[K] extends number | string | boolean | never
+        ? New[K] extends number | string | boolean | never
+          ? New[K]
+          : New[K]
+        : Shadow<Base[K], New[K]>
+      : Base[K]
+    : K extends keyof New
+    ? New[K]
+    : never;
+};
+
 export interface RetryOptions {
   /** Initial wait time before the first retry. */
   initialInterval: number;
@@ -98,8 +112,17 @@ export interface OtelCollectorExporter {
      * Optional set of HTTP request headers to send to Collector (e.g. for authentication)
      */
     headers?: Record<string, string>;
+    /**
+     * Specify how frequently in metrics should be exported.
+     *
+     * @format number of milliseconds or {@link https://www.npmjs.com/package/ms | ms-formatted string}
+     * @defaults 1 second
+     */
+    metricPeriodicity?: string | number;
   };
 }
+
+export type CompiledOtelCollectorExporter = Shadow<OtelCollectorExporter, { otel: { metricPeriodicity: number } }>;
 
 /**
  * Prometheus metrics exporter options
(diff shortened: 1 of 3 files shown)


packages/core-bridge/index.d.ts
```diff
@@ -3,6 +3,20 @@ import type { TLSConfig } from '@temporalio/common/lib/internal-non-workflow';
 
 export { TLSConfig };
 
+type Shadow<Base, New> = {
```
bergundy, line 6: That's quite the type..

packages/core-bridge/index.d.ts
```diff
@@ -98,9 +112,18 @@ export interface OtelCollectorExporter {
      * Optional set of HTTP request headers to send to Collector (e.g. for authentication)
      */
     headers?: Record<string, string>;
+    /**
+     * Specify how frequently in metrics should be exported.
+     *
+     * @format number of milliseconds or {@link https://www.npmjs.com/package/ms | ms-formatted string}
+     * @defaults 1 second
+     */
+    metricPeriodicity?: string | number;
```
bergundy, line 121: Maybe use plural?

```suggestion
    metricsPeriodicity?: string | number;
```

bergundy: TBH I'd name it `metricsExportInterval`, it reads better than Periodicity - a term I encountered here for the first time.

packages/core-bridge/index.d.ts
```diff
@@ -210,9 +220,19 @@ export interface TelemetryOptions {
    *
    * Turned off by default
    */
-  metrics?: MetricsExporter;
+  metrics?: RequireExactlyOne<
```
bergundy, line 223: I don't see `RequireExactlyOne` defined or imported here.

bergundy: Can we keep the types a bit simpler?
It'll be very hard to understand the shape of these types with all of these transformations.

bergundy: We also require either prom or otel and temporality is optional, so the name RequireExactlyOnce is confusing here.

packages/worker/src/runtime.ts
```diff
@@ -232,9 +239,27 @@ export class Runtime {
         tracing: tracing?.otel && {
           filter: defaultFilter,
           otel: {
-            ...filterNullAndUndefined(tracing.otel),
+            url: tracing.otel.url,
+            headers: tracing.otel.headers,
+            metricPeriodicity: msToNumber(tracing.otel.metricPeriodicity ?? '1s'),
```
bergundy, line 244: Don't think this belongs in tracing config, only metrics AFAICT.

bergundy (approved):
Overall lgtm, left some comments inline.

packages/core-bridge/index.d.ts
mjameswh: I originaly intended to name it `RecursiveReplace`, but we already have a `NestedReplace` elsewhere, which is something completely different; that would have been confusing. But I believe it could makes things much easier than `Replace<Base, New>` in many cases.

packages/core-bridge/index.d.ts
mjameswh: Removed usage RequiredExactlyOnce

PR #979, opened 2022-11-29
chore: Docs improvements

bergundy:


README.md
```diff
@@ -9,20 +9,17 @@
   </a>
 </div>
 
-[Temporal](https://temporal.io) is a microservice orchestration platform that enables developers to build scalable applications without sacrificing productivity or reliability. Temporal Server executes units of application logic—Workflows—in a resilient manner that automatically handles intermittent failures and retries failed operations.
+[Temporal](https://temporal.io) is a distributed, scalable, durable, and highly available orchestration engine used to execute asynchronous, long-running business logic in a scalable and resilient way.
```
mjameswh, line 12: nit: I would avoid mentioning 'scalable' twice. It fits better as a quality of the engine (ie. early in the sentence) than as a quality of the execution (ie. later in the sentence)

packages/worker/src/workflow/bundler.ts
```diff
@@ -269,14 +269,15 @@ exports.importInterceptors = function importInterceptors() {
 
             if (this.foundProblematicModules.size) {
               const err = new Error(
-                `Your Workflow code (or a library used by your Workflow code) is importing the following built-in Node modules:\n` +
+                `Your Workflow code (or a library used by your Workflow code) is importing the following black-listed modules:\n` +
```
mjameswh, line 272:
```suggestion
                `Your Workflow code (or a library used by your Workflow code) is importing the following disallowed modules:\n` +
```

packages/worker/src/workflow/bundler.ts
```diff
@@ -269,14 +269,15 @@ exports.importInterceptors = function importInterceptors() {
...
                   Array.from(this.foundProblematicModules)
                     .map((module) => `  - '${module}'\n`)
                     .join('') +
                   `Workflow code doesn't have access to built-in Node modules (in order to help enforce determinism). If you are certain ` +
```
mjameswh, line 276: I think this whole paragraph should be rewritten as a list of things to check instead...

For example:

```suggestion
                  `These modules can't be used in workflow context as they might break determinism.` +
                  `HINT: Consider the following options:\n` +
                  ` • Make sure that activity code is not imported from workflow code. Use \`import type\` to import activity function signature.\n` +
                  ` • Move code that has non-deterministic behaviour to activities.\n` +
                  ` • If you know for sure that a disallowed module will not be used at runtime, add its name to 'WorkerOptions.bundlerOptions.ignoreModules' in order to dismiss this warning.\n` +

```

README.md
```diff
@@ -9,20 +9,17 @@
...
 
-Temporal is a mature technology—it's a fork of Uber's Cadence. Temporal is being developed by Temporal Technologies, a startup by the creators of Cadence.
+"Temporal TypeScript SDK" is the framework for authoring workflows and activities using either the TypeScript or JavaScript programming languages.
 
-## Documentation
-
-The documentation is divided into several sections:
+For getting started and general documentation, please refer to the following resources:
```
mjameswh, line 16: Had to read the "For getting started" part several times before I got to parse the correct meaning of this sentence. Not sure how to best fix, but you can consider this:

```suggestion
For "Getting started instructions" and General documentation, please refer to the following resources:
```

packages/worker/src/workflow/bundler.ts
mjameswh: Also, what would really help here is to give the trace back on problematic imports... We definitely need to get there.

README.md
lorensr, line 12:
```suggestion
[Temporal](https://temporal.io) is a durable execution system that transparently makes your code durable, fault-tolerant, and simple.
```

lorensr: candace + oxford comma

README.md
lorensr, line 16:
```suggestion
To get started, see:
```

packages/test/src/test-bundler.ts
```diff
@@ -86,7 +86,7 @@ if (RUN_INTEGRATION_TESTS) {
       }),
       {
         instanceOf: Error,
-        message: /is importing the following built-in Node modules.*dns/s,
+        message: /is importing the following black-listed modules.*dns/s,
```
lorensr, line 89:
```suggestion
        message: /is importing the following disallowed modules.*dns/s,
```

lorensr (approved):

README.md
bergundy: I was going to update after my discussion on Slack with Candace.
Thanks.

README.md
bergundy: This is nice but the resources are not just for getting started.

bergundy: Let me think about this one.

packages/worker/src/workflow/bundler.ts
bergundy: Yeah, I've been meaning to increase the webpack verbosity on error, never got to that.

mjameswh (approved):
LGTM

Issue #928, opened 2022-10-17
[Bug] Jest detects open file handles


bergundy:
Getting this warning when running a simple test with the test environment:

```
 $ ./node_modules/.bin/jest --detectOpenHandles

...
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  neon threadsafe function
...
```

Code for reproduction:

```ts
import { TestWorkflowEnvironment } from '@temporalio/testing';
import { Worker, Runtime, DefaultLogger, LogEntry } from '@temporalio/worker';
import { someWorkflow } from './workflows';

let testEnv: TestWorkflowEnvironment;

beforeAll(async () => {
  // Use console.log instead of console.error to avoid red output
  // Filter INFO log messages for clearer test output
  Runtime.install({
    logger: new DefaultLogger('WARN', (entry: LogEntry) => console.log(`[${entry.level}]`, entry.message)),
  });

  testEnv = await TestWorkflowEnvironment.createTimeSkipping();
});

afterAll(async () => {
  await testEnv?.teardown();
});

test('some random test', async () => {
  const { client, nativeConnection } = testEnv;
  const worker = await Worker.create({
    connection: nativeConnection,
    taskQueue: 'test',
    workflowsPath: require.resolve('./workflows'),
    activities: {
      // ...
    },
  });

  await worker.runUntil(async () => {
    const result = await client.workflow.execute(someWorkflow, {
      args: [],
      workflowId: 'money-transfer-test-workflow',
      taskQueue: 'test',
    });

    expect(result).toEqual('test');
  });
});
```

lorensr:
Val did some investigating in https://github.com/temporalio/samples-typescript/issues/189

mjameswh:
Test case can be reduced to this:

```
test(`core-bridge registerErrors() doesn't leave open handles`, async () => {
    const native = require("@temporalio/core-bridge");

    native.registerErrors({
        IllegalStateError: () => ({}),
        ShutdownError: () => ({}),
        TransportError: () => ({}),
        UnexpectedError: () => ({}),
        GracefulShutdownPeriodExpiredError: () => ({}),
    });
});
```

The issue is specifically with `registerErrors()`. For example, that one doesn't trigger the error message:

```
test(`core-bridge getTimeOfDay() doesn't leave open handles`, async () => {
    const native = require("@temporalio/core-bridge");
    native.getTimeOfDay();
}
```

mjameswh:
[...]

PR #983, opened 2022-12-01
fix(core): Do not retain static references to error constructors in rust bridge

mjameswh:
## What changed

Previously, rust bridge would instantiate custom JavaScript errors (ie. `TransportError`, `ShutdownError`...) directly. This was done by registering error constructors at runtime initialization, then holding these constructors in `static OnceCell<Root>` variables in rust code.

Now, rust throw raw instances of JavaScript's `Error` class, and set the `name` property to reflect the type of error. These raw errors are then converted to custom errors thrugh a JavaScript-side wrapper.

## Why

- https://github.com/temporalio/sdk-typescript/issues/928
- https://github.com/temporalio/sdk-typescript/issues/876


packages/core-bridge/index.js
```diff
@@ -1,8 +1,27 @@
...
+const typescriptExports = require('./lib/index');
+const { convertFromNamedError } = require('./lib/errors');
+
+function wrapErrors(fn) {
+  return (...args) => {
+    try {
+      if (typeof args[args.length - 1] === 'function') {
+        const callback = args[args.length - 1];
+        args[args.length - 1] = (e, x) => callback(convertFromNamedError(e), x);
+      }
```
Sushisource, line 11: Comment describing what this is for would be helpful

packages/core-bridge/src/errors.rs
```diff
@@ -1,96 +1,38 @@
...
-        self.construct_from_string(cx, format!("{:?}", err))
-    }
+pub static TRANSPORT_ERROR: &str = "TransportError";
+// /// Thrown after shutdown was requested as a response to a poll function, JS should stop polling
+// /// once this error is encountered
+pub static SHUTDOWN_ERROR: &str = "ShutdownError";
+// /// Something unexpected happened, considered fatal
+pub static UNEXPECTED_ERROR: &str = "UnexpectedError";
+// /// Used in different parts of the project to signal that something unexpected has happened
+pub static ILLEGAL_STATE_ERROR: &str = "IllegalStateError";
```
Sushisource, line 11: Weird extra comments making these docstrings invalid

packages/core-bridge/src/errors.ts
```diff
@@ -0,0 +1,40 @@
...
+export function convertFromNamedError(e: unknown): unknown {
+  if (e instanceof Error) {
+    switch (e.name) {
+      case 'TransportError':
+        return new TransportError(e.message);
+      case 'IllegalStateError':
+        return new IllegalStateError(e.message);
+      case 'ShutdownError':
+        return new ShutdownError(e.message);
+      case 'UnexpectedError':
```
Sushisource, line 35: This should really just be the default case

Sushisource (approved):
LGTM just some small things

packages/core-bridge/src/errors.ts
mjameswh: 🤔 Interesting idea... Will do.

packages/core-bridge/src/errors.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Something unexpected happened, considered fatal
+ */
+export class UnexpectedError extends Error {
+  public readonly name = 'UnexpectedError';
+}
+export { IllegalStateError };
+
+export function convertFromNamedError(e: unknown): unknown {
+  if (e instanceof Error) {
```
bergundy, line 27: You should probably remove the stack trace from these errors because it won't be right.

packages/worker/src/errors.ts
```diff
@@ -1,27 +1,3 @@
...
-}
-
-/**
- * Something unexpected happened, considered fatal
- */
-export class UnexpectedError extends Error {
-  public readonly name = 'UnexpectedError';
-}
-
 /**
```
bergundy, line 1: Is it worth keeping this file around for 1 error class?

packages/worker/src/index.ts
```diff
@@ -33,7 +35,6 @@ export {
   CombinedWorkerRunErrorCause,
   DataConverter,
   defaultPayloadConverter,
-  errors,
```
bergundy, line 36: This is a breaking change, can we deprecate this export instead of removing it?

bergundy (approved):
Had a couple of minor comments

bergundy:
> One thing I'm hesitant here is mixing ts and rust files in the src directory... Opinions?

I don't love this, can you make a `ts` directory instead?

packages/core-bridge/src/errors.ts
mjameswh: It should be correct up to the point of entering core bridge, at which point it would have been a native call anyway. That should provide enough context to help users understand issues from their perspective.

But at some point, I thought about copying the stacktrace from the original error, but ended up leaving that aside without giving it a try. Will do some test with this.

packages/worker/src/errors.ts
mjameswh: Keeping this file a little bit longer as a way to structure the deprecated `errors` public export.

packages/core-bridge/src/errors.ts
mjameswh: After experimenting with this, I copied the original stack trace in non-callback exceptions, as it contains meaningful stack trace in user code. Setting to blank in the case of callbacks, as there is nothing useful to extract from there.

PR #985, opened 2022-12-05
fix(workflow): Correctly handle condition with timeout of 0

mjameswh:
## What changed

- In workflows, `condition(fn, 0)` was incorrectly handled the same as `condition(fn, undefined)`, ie. the function would block indefinitely and would return nothing once `fn` evaluates to `true`. It now behaves the same as `condition(fn, 1)`, ie. the function will sleep (at most) for a very short time, then return `true` if `fn` evaluated to `true`, or `false` if `timeout` reached its expiration.
- Compatibility with existing workflow histories is maintained, through an internal patch.
- Introduced a SDK internal patch mechanism, which will make it easier and more efficient to support similar bug fixes in the future.

### DIFF
--- a/packages/workflow/src/internals.ts
+++ b/packages/workflow/src/internals.ts
@@ -85,6 +85,16 @@ export type ActivationHandler = {
 };
 
 /**
+ * SDK Internal Patches are created by the SDK to avoid breaking history when behaviour
+ * of existing API need to be modified. This is the patch number supported by the current
+ * version of the SDK.
+ *
+ * History:
+ * 1: Fix `condition(..., 0)` is not the same as `condition(..., undefined)`
+ */
+export const LATEST_INTERNAL_PATCH_NUMBER = 1;
+
+/**
  * Keeps all of the Workflow runtime state like pending completions for activities and timers.
  *
  * Implements handlers for all workflow activation jobs.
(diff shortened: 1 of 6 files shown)


bergundy (approved):

packages/workflow/src/internals.ts
```diff
@@ -84,6 +84,16 @@ export type ActivationHandler = {
   [P in keyof coresdk.workflow_activation.IWorkflowActivationJob]: ActivationHandlerFunction<P>;
 };
 
+/**
+ * SDK Internal Patches are created by the SDK to avoid breaking history when behaviour
+ * of existing API need to be modified. This is the patch number supported by the current
+ * version of the SDK.
+ *
+ * History:
```
Sushisource, line 92: I'm a little worried this could be easy to accidentally forget to update.

Could we instead have `checkInternalPatchAtLeast` be `checkInternalPatchExists` which accepts some identifier of the patch (`cond-timeout-0` or something in this case). Also makes the markers easier to decipher..

mjameswh: I understand your point, but three things...
1. The sequential number means we introduce less polution in workflow history. Having an internal patch marker X implies that you actually support anything in the range [0...X], no need for further patch. At this moment, we plan to introduce two more similar internal patches in the near future, and we don't know what else after that. Let's not forget that we will likely have to support those patches until the next major release, since we can't know when there there is no more workflows out there needing these legacy behaviour.
2. The sequential number also makes it possible for the worker to refuse working with an history targeted at a later version of the worker (similar to a bad binary, but on based on patch).

Sushisource: Right, good points, makes sense to me.

A bit of color on point 2 - I was actually just discussing with @Spikhalskiy how there are certainly types of changes where it is effectively impossible for us to be forward compatible (an old SDK works with new SDK's history). One example being adding the timer for LA backoff he was working on. An old worker encountering the history has no possible way to know how to deal with the extra command (timer) that exists in the history generated by a new worker. We'd have to do something elaborate like include information about how to skip over the command in the marker metadata, which is :vomiting_face: .

PR #990, opened 2022-12-06
Update docusaurus and other deps with known vaulnerabilities

mjameswh:
## Why

- Should considerably reduce the number of packages reported in #746
- It is known that certain transitive deps will still be reported, until some middle  package fix their own dependencies.


mjameswh:
Causing docs build errors. Postponing after 1.5.0 release.

cretz (approved):

PR #993, opened 2022-12-08
fix(core): Use correct property name for OTel metrics export interval

mjameswh:


Sushisource (approved):
Doesn't have to be in this PR but, seems like there needs to be a test that sets this property (anything converted this way, really) just to make sure we never have a name mismatch again like this.

mjameswh:
> Doesn't have to be in this PR but, seems like there needs to be a test that sets this property (anything converted this way, really) just to make sure we never have a name mismatch again like this.

Definitely. We have some testing blind spots in opentelemetry and prometheus configs, because our current testing environment doesn't include such components. We'll have to figure out a way to correctly validate this does not happen again.

bergundy (approved):

PR #1012, opened 2023-01-03
feat(client): Increase default gRPC retry policy maxAttempts to 13

lorensr:
With the current 10 attempts, the cumulative time spent waiting could be as small as 9.2 seconds. If we raise it to 13, the client will wait at least 38 seconds before giving up. 

image


mjameswh:
Thanks for sharing this observation, but we won't merge at this point. We might however come back to this suggestion at some later point. Closing in the mean time.

PR #1018, opened 2023-01-10
fix: Remove `.post-create` before committing

lorensr:
## What was changed

Currently it's deleted after the init/commit step, so user starts out dirty:

```
$ git status
On branch main
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    .post-create
```

## Checklist

2. How was this tested:

`packages/create-project/cli.js`

image

### DIFF
--- a/packages/create-project/src/create-project.ts
+++ b/packages/create-project/src/create-project.ts
@@ -187,21 +187,22 @@ export async function createApp({
 
   console.log();
 
-  if (await tryGitInit(root, gitInit)) {
-    console.log('Initialized a git repository.');
-  }
-
   const messageFile = path.join(root, '.post-create');
-
-  console.log();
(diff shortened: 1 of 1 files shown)


bergundy (approved):

lorensr:
This is a persistent CI failure, not seeing why: 

image

bergundy:
@lorensr you can print stdout and the error code here to find out why it failed: https://github.com/temporalio/sdk-typescript/blob/main/scripts/test-example.js#L35

mjameswh:
@lorensr Actually, don't worry too much with the ESM failure if you don't have time. I'm facing the very same issue on my ci-refactor branch, so I'll investigate that soon anyway.

PR #1028, opened 2023-01-13
fix(client): Fix args on create/update schedules

### LINKED ISSUE #1022: [Bug] Schedule client does not infer args based on workflow type
### Describe the bug

image

This behaviour is observed both on schedule `create` and `update`.

mjameswh:
Fix #1022


bergundy (approved):
I pulled and verified, LGTM.
Can you fix the lint warnings please?

PR #1031, opened 2023-01-20
feat: Add startTime to WorkflowInfo, and more

### LINKED ISSUE #375: Improve error message: 'foo' is not a function
### Problem

When a Worker doesn't have the right Workflow code, it returns a result with an error message that says eg [`'progress' is not a function`](https://user-images.githubusercontent.com/251288/140634770-d4c6d5bc-adc2-4d5e-bdaa-9214231e3836.png) from https://github.com/temporalio/samples-typescript/issues/56.

### Solution

Instead, say something like:

```
The Worker process that picked up this task doesn't have a Workflow function named 'progress' in its bundle. For more information and possible causes, see: 

[docs URL]
```

so that in the future we can easily update the wiki page, and users of old versions of the SDK will see the updated page.

The docs page could say:

```
The Worker process that picked up this task doesn't have a Workflow function named 'foo' in its bundle. 
The issue might be:

- There's a typo in the Workflow function name.
- The Workflow function isn't exported.
- The entry point (`workflowsPath`) is incorrect.
- You have more Worker processes than you realize polling on the 'tutorial' queue. For instance, a process in another shell or in the background that has an old version of the Workflow code bundle or that's from a different Temporal project. You can check with `ps` or `tctl tq desc --tq tutorial`.
```

Migrated from #359

### LINKED ISSUE #659: [Bug] instanceof doesn't work with WorkflowInfo fields

Determine types of `WorkflowInfo` fields inside Workflow.

### Describe the bug

Things like this are `false`, since the info object is created outside the vm:

```
// searchAttributes: { CustomIntField: [1,2] }
workflowInfo().searchAttributes.CustomIntField instanceof Array
workflowInfo().executionExpirationTime instanceof Date
```

### LINKED ISSUE #694: [Feature Request] Add `WorkflowInfo.runStartTime`
- Go: `WorkflowStartTime` https://pkg.go.dev/go.temporal.io/sdk@v1.15.0/internal#WorkflowInfo
- Java: [getRunStartedTimestampMillis](https://www.javadoc.io/static/io.temporal/temporal-sdk/1.0.2/io/temporal/workflow/WorkflowInfo.html#getRunStartedTimestampMillis--)()

### LINKED ISSUE #890: [Feature Request] Don't warn about no activities during `Worker.runReplayHistory`

Users see this in output and think something is wrong:

```
2022-09-27T16:09:43.579Z [WARN] No activities registered, not polling for activity tasks
```

### LINKED ISSUE #1005: [Feature Request] Add `workflowInfo().startTime`
https://github.com/temporalio/sdk-typescript/blob/bf4ccdf3c7baae544297d19d66cf047957eedbdb/packages/workflow/src/interfaces.ts#L84-L88

Exposed in https://github.com/temporalio/sdk-core/pull/416

### LINKED ISSUE #1020: [Feature Request] Start workers in debug mode if `TEMPORAL_DEBUG` env var is set
I don't love that the use of env vars in libraries but I'd rather support this for consistency with the other SDKs.

mjameswh:
## What changed

- Can now start workers in debug mode by setting environment variable `TEMPORAL_DEBUG=true` (fix #1020)
- No longer warn about no activities when running a replay-only Worker (fix #890)
- Improved error message when requested workflow type is not an exported function (fix #375)
- `instanceof` on `WorkflowInfo` fields now works as expected (fix #659)
- Added `startTime` and `runStartTime` on `WorkflowInfo` (fix #1005, fix #694)

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -1146,9 +1146,7 @@ export class Worker {
                         )
                       ) {
                         throw new TypeError(
-                          `Expected StartWorkflow with workflowId, workflowType, randomnessSeed, firstExecutionRunId, attempt and startTime. Got ${JSON.stringify(
-                            maybeStartWorkflow
-                          )}`
+                          `Malformed StartWorkflow activation: ${JSON.stringify(maybeStartWorkflow)}`
                         );
                       }
                       if (activation.timestamp == null) {
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -50,7 +50,7 @@ import {
   TASK_TOKEN_ATTR_KEY,
 } from '@temporalio/common/lib/otel';
 import { historyFromJSON } from '@temporalio/common/lib/proto-utils';
-import { optionalTsToDate, optionalTsToMs, tsToMs } from '@temporalio/common/lib/time';
+import { optionalTsToDate, optionalTsToMs, tsToDate, tsToMs } from '@temporalio/common/lib/time';
 import { errorMessage } from '@temporalio/common/lib/type-helpers';
 import * as native from '@temporalio/core-bridge';
 import { UnexpectedError } from '@temporalio/core-bridge';
@@ -254,7 +254,7 @@ function formatTaskToken(taskToken: Uint8Array) {
 /**
  * Notify that an activity has started, used as input to {@link Worker.activityHeartbeatSubject}
  *
- * Used to detect rouge activities.
+ * Used to detect rogue activities.
  */
 interface HeartbeatCreateNotification {
   type: 'create';
@@ -575,7 +575,10 @@ export class Worker {
     }
     const workflowCreator = await this.createWorkflowCreator(bundle, compiledOptions);
     const replayHandle = await nativeWorkerCtor.createReplay(addBuildIdIfMissing(compiledOptions, bundle.code));
-    return [new this(replayHandle.worker, workflowCreator, compiledOptions), replayHandle.historyPusher];
+    return [
+      new this(replayHandle.worker, workflowCreator, compiledOptions, undefined, true),
+      replayHandle.historyPusher,
+    ];
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/worker/src/worker.ts
```diff
@@ -254,7 +254,7 @@ function formatTaskToken(taskToken: Uint8Array) {
 /**
  * Notify that an activity has started, used as input to {@link Worker.activityHeartbeatSubject}
  *
- * Used to detect rouge activities.
+ * Used to detect rogue activities.
```
bergundy, line 257: It's to detect "moulin rouge" activities :-p

packages/worker/src/worker.ts
```diff
@@ -1133,13 +1137,16 @@ export class Worker {
...
+                          startWorkflow.attempt != null &&
+                          startWorkflow.startTime != null
                         )
                       ) {
                         throw new TypeError(
-                          `Expected StartWorkflow with workflowId, workflowType and randomnessSeed, got ${JSON.stringify(
+                          `Expected StartWorkflow with workflowId, workflowType, randomnessSeed, firstExecutionRunId, attempt and startTime. Got ${JSON.stringify(
                             maybeStartWorkflow
                           )}`
                         );
```
bergundy, line 1152: nit: At this point just say:

```suggestion
                        throw new TypeError(`Malformed StartWorkflow: ${JSON.stringify(maybeStartWorkflow)}`);
```

packages/workflow/src/interfaces.ts
```diff
@@ -81,11 +81,15 @@ export interface WorkflowInfo {
    */
   continuedFromExecutionRunId?: string;
 
-  // TODO expose from Core
   /**
-   * Time at which the Workflow Run started
+   * Time at which this Workflow Execution Chain was started
```
bergundy, line 85: Do we have a stable link from our docs to the concept of run vs chain?

bergundy (approved):

packages/workflow/src/interfaces.ts
mjameswh:
```suggestion
   * Time at which this [Workflow Execution Chain](https://docs.temporal.io/workflows#workflow-execution-chain) was started
```

PR #1032, opened 2023-01-24
fix(client): Schedules specs can be empty

### LINKED ISSUE #967: Context deadline exceeded on test `schedules > Can update schedule action`
[build-and-test (14, ubuntu-latest)](https://github.com/temporalio/sdk-typescript/actions/runs/3457023156/jobs/5770175142#logs)

```
@temporalio/test:   schedules › Can update schedule action
@temporalio/test:   Rejected promise returned by test. Reason:
@temporalio/test:   ServiceError {
@temporalio/test:     cause: Error {
@temporalio/test:       code: 4,
@temporalio/test:       details: 'context deadline exceeded',
@temporalio/test:       metadata: Metadata {
@temporalio/test:         internalRepr: Map { … },
@temporalio/test:         options: {},
@temporalio/test:       },
@temporalio/test:       message: '4 DEADLINE_EXCEEDED: context deadline exceeded',
@temporalio/test:     },
@temporalio/test:     message: 'Failed to describe schedule',
@temporalio/test:   }
@temporalio/test:   › ScheduleClient.rethrowGrpcError (/home/runner/work/sdk-typescript/sdk-typescript/packages/client/src/schedule-client.ts:508:13)
@temporalio/test:   › ScheduleClient._describeSchedule (/home/runner/work/sdk-typescript/sdk-typescript/packages/client/src/schedule-client.ts:278:12)
@temporalio/test:   › Object.describe (/home/runner/work/sdk-typescript/sdk-typescript/packages/client/src/schedule-client.ts:413:21)
@temporalio/test:   › Object.update (/home/runner/work/sdk-typescript/sdk-typescript/packages/client/src/schedule-client.ts:450:25)
@temporalio/test:   › src/test-schedules.ts:324:7
```

### LINKED ISSUE #1013: [Bug] ScheduleClient.list() fails to return any schedules

List schedules with the `ScheduleClient.list()` API

### Describe the bug

No schedules are returned when I can verify there are schedules through `tctl`, the UI and go-sdk.

### Minimal Reproduction

See the `basic` typescript test here https://github.com/temporalio/sdk-features/pull/192

### Environment/Versions

- OS and processor:M1 Mac
- Temporal Version: tried temporalite 1.17 and 1.19

mjameswh:
## What changed

- In `create` and `update` schedules, `spec` is still required, but can be empty. It is no longer required that the a schedule has at least one calandar, one interval or one cron spec.
- In `list`, all schedule properties (except `scheduleId`) are now optional, with a comment indicating that this structure is eventual consistent. This might change on server at some point in the future, but this matches the server behaviour for 1.18.x and 1.19.x.
- Duplicated fields from `ScheduleSummary` to `ScheduleDescription` to avoid further complexification of doc.

## Why

- Fix #1013
- Fix #967

### DIFF
--- a/packages/client/src/schedule-client.ts
+++ b/packages/client/src/schedule-client.ts
@@ -143,9 +143,6 @@ function assertRequiredScheduleOptions(
   if (action === 'CREATE' && !(opts as ScheduleOptions).scheduleId) {
     throw new TypeError(`Missing ${structureName}.scheduleId`);
   }
-  if (!(opts.spec.calendars?.length || opts.spec.intervals?.length || opts.spec.cronExpressions?.length)) {
-    throw new TypeError(`At least one ${structureName}.spec.calendars, .intervals or .cronExpressions is required`);
-  }
   switch (opts.action.type) {
     case 'startWorkflow':
       if (!opts.action.taskQueue) {
@@ -377,24 +374,22 @@ export class ScheduleClient extends BaseClient {
       );
 
       for (const raw of response.schedules ?? []) {
-        if (!raw.info?.spec) continue;
-
         yield <ScheduleSummary>{
           scheduleId: raw.scheduleId,
 
-          spec: decodeScheduleSpec(raw.info.spec),
-          action: {
+          spec: decodeScheduleSpec(raw.info?.spec ?? {}),
+          action: raw.info?.workflowType?.name && {
             type: 'startWorkflow',
-            workflowType: raw.info.workflowType?.name,
+            workflowType: raw.info.workflowType.name,
           },
           memo: await decodeMapFromPayloads(this.dataConverter, raw.memo?.fields),
           searchAttributes: decodeSearchAttributes(raw.searchAttributes),
           state: {
-            paused: raw.info.paused === true,
-            note: raw.info.notes ?? undefined,
+            paused: raw.info?.paused === true,
+            note: raw.info?.notes ?? undefined,
           },
           info: {
-            recentActions: decodeScheduleRecentActions(raw.info.recentActions),
+            recentActions: decodeScheduleRecentActions(raw.info?.recentActions),
             nextActionTimes: raw.info?.futureActionTimes?.map(tsToDate) ?? [],
           },
         };
--- a/packages/client/src/schedule-types.ts
+++ b/packages/client/src/schedule-types.ts
@@ -1,4 +1,4 @@
-import { checkExtends, Replace, RequireAtLeastOne } from '@temporalio/common/lib/type-helpers';
+import { checkExtends, Replace } from '@temporalio/common/lib/type-helpers';
 import { SearchAttributes, Workflow } from '@temporalio/common';
 import type { temporal } from '@temporalio/proto';
 import { WorkflowStartOptions } from './workflow-options';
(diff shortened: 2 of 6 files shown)


packages/client/src/schedule-client.ts
```diff
@@ -377,24 +374,22 @@ export class ScheduleClient extends BaseClient {
...
       for (const raw of response.schedules ?? []) {
-        if (!raw.info?.spec) continue;
-
         yield <ScheduleSummary>{
           scheduleId: raw.scheduleId,
 
-          spec: decodeScheduleSpec(raw.info.spec),
-          action: {
+          spec: decodeScheduleSpec(raw.info?.spec ?? {}),
+          action: raw.info?.workflowType?.name && {
```
bergundy, line 381: We need to fix this on the server side and make the action non-nullable.

packages/client/src/schedule-client.ts
```diff
@@ -377,24 +374,22 @@ export class ScheduleClient extends BaseClient {
...
 
       for (const raw of response.schedules ?? []) {
-        if (!raw.info?.spec) continue;
-
         yield <ScheduleSummary>{
           scheduleId: raw.scheduleId,
 
-          spec: decodeScheduleSpec(raw.info.spec),
-          action: {
+          spec: decodeScheduleSpec(raw.info?.spec ?? {}),
```
bergundy, line 380: Probably better to return undefined instead of an empty spec.

packages/client/src/schedule-types.ts
```diff
@@ -19,7 +19,7 @@ export interface ScheduleOptions<A extends ScheduleOptionsAction = ScheduleOptio
   /**
    * When Actions should be taken
    */
-  spec: RequireAtLeastOne<ScheduleSpec, 'calendars' | 'intervals' | 'cronExpressions'>;
+  spec: ScheduleSpec;
```
bergundy, line 22: I'd keep the old definition and make it optional.

packages/client/src/schedule-client.ts
mjameswh: Doing that means that using and updating a schedule become significantly more complex. For example

```
      await handle.update((x: ScheduleUpdateOptions) => {
        x.spec.intervals = [{ every: '3h' }];
        return x;
      });
```

would now have to be written as

```
      await handle.update((x: ScheduleUpdateOptions) => {
        x.spec = x.spec ?? {};
        x.spec.intervals = [{ every: '3h' }];
        return x;
      });
```

mjameswh: Making `spec` optional was actually my first attempt. I really didn't like the consequences of that.

packages/client/src/schedule-client.ts
mjameswh: I agree, but in the mean time, I think we're better provide compatibility with 1.18.x and 1.19.x, which are people are currently using. Making optional properties required on that interface at a later point will be non breaking, as that interface is meant to be a return value from our API.

packages/client/src/schedule-types.ts
mjameswh: This has to match `ScheduleDescription` to maintain the `update` callback signature. See above comment on why I think it is better that spec be non-optional.

packages/client/src/schedule-client.ts
mjameswh: Also, the server might actually return a schedule with a spec that has no calandar and no intervals.

bergundy (approved):

PR #1037, opened 2023-01-26
fix(worker): Avoid webpack error on "node:" modules in workflows

### LINKED ISSUE #1036: How to add module to disallowedModules?
We need to exclude a module from being bundled.

Contrary to the documentation, `ignoreModules` does not achieve that.

mjameswh:
## What changed

- In the workflow bundler, ignored modules get replaced by an empty object from Webpack's `externals` hook, in addition to `aliases`.

## Why

- If workflow code add an import on some `node:` prefixed module, Webpack would fail to build, because that namespace is not supported for our target.
- Fixes #1036

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -68,7 +68,7 @@ export class WorkflowCodeBundler {
     this.payloadConverterPath = payloadConverterPath;
     this.failureConverterPath = failureConverterPath;
     this.workflowInterceptorModules = workflowInterceptorModules ?? defaultWorflowInterceptorModules;
-    this.ignoreModules = ignoreModules ?? [];
+    this.ignoreModules = (ignoreModules ?? []).map((x) => (x.startsWith('node:') ? x.substring('node:'.length) : x));
     this.webpackConfigHook = webpackConfigHook ?? ((config) => config);
   }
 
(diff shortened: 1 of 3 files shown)


bergundy (approved):

packages/worker/src/workflow/bundler.ts
```diff
@@ -183,14 +183,20 @@ exports.importInterceptors = function importInterceptors() {
...
         ? data.request.slice('node:'.length)
         : data.request ?? '';
 
-      if (moduleMatches(module, disallowedModules) && !moduleMatches(module, this.ignoreModules)) {
-        this.foundProblematicModules.add(module);
+      if (moduleMatches(module, disallowedModules)) {
+        if (!moduleMatches(module, this.ignoreModules)) {
+          this.foundProblematicModules.add(module);
+        }
+        return 'var {}';
```
bergundy, line 199: Should we change to return an empty object? Is this better experience in your opinion?

mjameswh: Not sure what you mean... Do you mean `var false` vs `var {}`?

bergundy: We used to return `undefined` here.

mjameswh: Oh, I see... Actually, THIS is the fix.

By returning `var {}`, we are telling Webpack to replace imports of that module by an empty object. This is essentially the same as what we do with `alias`, but `externals` is called much earlier in the process, so that we can still handle invalid namespaces such as `node:`.

mjameswh:
Putting this one on hold.

After some more consideration, I think that the currently proposed solution will make it significantly more difficult for users to load build-time polyfills.

For this reason, it might be preferable that we simply rewire `node:*` to `*` from `externals`, then let `alias` behave as before, thus making it possible for users to load polyfill by overwriting the corresponding entry in `alias`.

Issue #1015, opened 2023-01-05
[Feature Request] Support dynamic workflows, activities, signals, queries, and updates


cretz:
### Is your feature request related to a problem? Please describe.

Need opt-in catch-alls for workflows, activities, signals, queries, and updates.

### Describe the solution you'd like

Should be able to register only one workflow as "dynamic" and it is given the name as a string as the first arg, then args as an array as the second arg. In other SDKs we give the second arg as still-serialized payloads because they need the type to deserialize, but since JS doesn't, can already be deserialized.

Need to do the same for activity.

For signal handler, should allow user to register only one catch-all that not only drains all unhandled-yet-buffered when registered, but then gets each unhandled signal after that. The function should get the signal name and the args.

Need to do the same for query and update.

Seems an early proposal on this is already at #973. Incorporate as needed.

### Checklist

- [x] Workflows: #1038
- [x] Signals: #1038 
- [x] Queries: #1639
- [ ] Updates: not implemented yet
- [ ] Activities: not implemented yet

mjameswh:
Also related: #846

PR #1038, opened 2023-01-26
feat(workflow): Add support for default workflow handler and default signal handler

### LINKED ISSUE #846: [Feature Request] Catch-all signal handler

With multiple handlers, each handler processes its own signals in order. In order to get in-order delivery of all signals, you need to either:

- only send a single signal type (["fat handler"](https://docs.temporal.io/typescript/workflows#define-signals-and-queries-dynamically))
- have a catch-all handler that is called when any signal arrives, regardless of type

Something like:

```ts
import { setGlobalSignalHandler } from '@temporalio/workflow'

setGlobalSignalHandler((signalType: string, arg1: Type1, arg2: Type2) => ...)
```

When both a global and individual handler is defined, do we only deliver to one of them, or to both?

mjameswh:
## What changed

- A workflow bundle can now do `export default async function (workflowType: string, ...args: unknown[]): Promise<unknown> { ... }` to define a default workflow handler.
- A workflow function can now do `setDefaultSignalHandler((signalName: string, ...args: unknown[]) => { ... })` to set a function that will handle signals sent for non-registered signal names.

## Why
- Partially covers #1015
- Resolves #973
- Resolves #846

### DIFF
--- a/packages/test/src/test-default-workflow.ts
+++ b/packages/test/src/test-default-workflow.ts
@@ -8,7 +8,7 @@ import { RUN_INTEGRATION_TESTS, Worker } from './helpers';
 import { existing } from './workflows/default-workflow-function';
 
 if (RUN_INTEGRATION_TESTS) {
-  test('Default workflow handler is used if requested workflow does not exists', async (t) => {
+  test('Default workflow handler is used if requested workflow does not exist', async (t) => {
     const taskQueue = `${t.title}-${uuid4()}`;
     const worker = await Worker.create({
       taskQueue,
@@ -23,7 +23,7 @@ if (RUN_INTEGRATION_TESTS) {
     });
   });
 
-  test('Default workflow handler is not used if requested workflow does exists', async (t) => {
+  test('Default workflow handler is not used if requested workflow exists', async (t) => {
     const taskQueue = `${t.title}-${uuid4()}`;
     const worker = await Worker.create({
       taskQueue,
--- a/packages/test/src/test-default-workflow.ts
+++ b/packages/test/src/test-default-workflow.ts
@@ -3,37 +3,53 @@
  */
 import test from 'ava';
 import { v4 as uuid4 } from 'uuid';
-import { WorkflowClient } from '@temporalio/client';
-import { RUN_INTEGRATION_TESTS, Worker } from './helpers';
+import { TestWorkflowEnvironment } from '@temporalio/testing';
+import { Worker } from './helpers';
 import { existing } from './workflows/default-workflow-function';
 
-if (RUN_INTEGRATION_TESTS) {
-  test('Default workflow handler is used if requested workflow does not exist', async (t) => {
+test('Default workflow handler is used if requested workflow does not exist', async (t) => {
+  const env = await TestWorkflowEnvironment.createLocal();
+  try {
     const taskQueue = `${t.title}-${uuid4()}`;
     const worker = await Worker.create({
+      connection: env.nativeConnection,
       taskQueue,
       workflowsPath: require.resolve('./workflows/default-workflow-function'),
     });
-    const client = new WorkflowClient();
     await worker.runUntil(async () => {
-      const result = client.execute('non-existing', { taskQueue, workflowId: uuid4(), args: ['test', 'foo', 'bar'] });
+      const result = env.client.workflow.execute('non-existing', {
+        taskQueue,
+        workflowId: uuid4(),
+        args: ['test', 'foo', 'bar'],
+      });
       t.is((await result).handler, 'default');
       t.is((await result).workflowType, 'non-existing');
       t.deepEqual((await result).args, ['test', 'foo', 'bar']);
     });
-  });
+  } finally {
+    await env.teardown();
+  }
+});
 
-  test('Default workflow handler is not used if requested workflow exists', async (t) => {
+test('Default workflow handler is not used if requested workflow exists', async (t) => {
+  const env = await TestWorkflowEnvironment.createLocal();
+  try {
     const taskQueue = `${t.title}-${uuid4()}`;
     const worker = await Worker.create({
+      connection: env.nativeConnection,
       taskQueue,
       workflowsPath: require.resolve('./workflows/default-workflow-function'),
     });
-    const client = new WorkflowClient();
     await worker.runUntil(async () => {
-      const result = client.execute(existing, { taskQueue, workflowId: uuid4(), args: ['test', 'foo', 'bar'] });
+      const result = env.client.workflow.execute(existing, {
+        taskQueue,
+        workflowId: uuid4(),
+        args: ['test', 'foo', 'bar'],
+      });
       t.is((await result).handler, 'existing');
       t.deepEqual((await result).args, ['test', 'foo', 'bar']);
     });
-  });
-}
+  } finally {
+    await env.teardown();
+  }
+});
--- a/packages/test/src/test-default-workflow.ts
+++ b/packages/test/src/test-default-workflow.ts
@@ -1,0 +1,39 @@
+/**
+ * Test usage of a default workflow handler
+ */
+import test from 'ava';
+import { v4 as uuid4 } from 'uuid';
+import { WorkflowClient } from '@temporalio/client';
+import { RUN_INTEGRATION_TESTS, Worker } from './helpers';
+import { existing } from './workflows/default-workflow-function';
+
+if (RUN_INTEGRATION_TESTS) {
+  test('Default workflow handler is used if requested workflow does not exists', async (t) => {
+    const taskQueue = `${t.title}-${uuid4()}`;
+    const worker = await Worker.create({
+      taskQueue,
+      workflowsPath: require.resolve('./workflows/default-workflow-function'),
+    });
+    const client = new WorkflowClient();
+    await worker.runUntil(async () => {
+      const result = client.execute('non-existing', { taskQueue, workflowId: uuid4(), args: ['test', 'foo', 'bar'] });
+      t.is((await result).handler, 'default');
+      t.is((await result).workflowType, 'non-existing');
+      t.deepEqual((await result).args, ['test', 'foo', 'bar']);
+    });
+  });
+
+  test('Default workflow handler is not used if requested workflow does exists', async (t) => {
+    const taskQueue = `${t.title}-${uuid4()}`;
+    const worker = await Worker.create({
+      taskQueue,
+      workflowsPath: require.resolve('./workflows/default-workflow-function'),
+    });
+    const client = new WorkflowClient();
+    await worker.runUntil(async () => {
+      const result = client.execute(existing, { taskQueue, workflowId: uuid4(), args: ['test', 'foo', 'bar'] });
+      t.is((await result).handler, 'existing');
+      t.deepEqual((await result).args, ['test', 'foo', 'bar']);
+    });
+  });
+}
--- a/packages/worker/src/workflow/vm-shared.ts
+++ b/packages/worker/src/workflow/vm-shared.ts
@@ -338,7 +338,7 @@ export abstract class BaseVMWorkflow implements Workflow {
     await new Promise(setImmediate);
     if (this.unhandledRejection) {
       return {
-        runId: activation.runId,
+        runId: this.activator.info.runId,
         failed: { failure: this.activator.errorToFailure(this.unhandledRejection) },
       };
     }
--- a/packages/worker/src/workflow/vm-shared.ts
+++ b/packages/worker/src/workflow/vm-shared.ts
@@ -11,10 +11,6 @@ import { partition } from '../utils';
 import { Workflow } from './interface';
 import { WorkflowBundleWithSourceMapAndFilename } from './workflow-worker-thread/input';
 
-export interface ActivationContext {
-  isReplaying: boolean;
-}
-
 // Not present in @types/node for some reason
 const { promiseHooks } = v8 as any;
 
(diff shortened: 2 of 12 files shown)


packages/workflow/src/interfaces.ts
```diff
@@ -380,3 +388,47 @@ export interface WorkflowCreateOptions {
...
+ *     handler: 'default',
+ *     workflowType,
+ *     args,
+ *   };
+ * };
+ *
+ * export default defaultWorkflow;
+ * ```
+ */
+export type DefaultWorkflowFunction = (workflowType: string, ...args: unknown[]) => WorkflowReturnType;
```
bergundy, line 434: Should we just send the args?
`workflowType` is obtainable with `workflowInfo().workflowType`

packages/workflow/src/interfaces.ts
```diff
@@ -380,3 +388,47 @@ export interface WorkflowCreateOptions {
...
+ * const defaultWorkflow: DefaultWorkflowFunction = async function (
+ *   workflowType: string,
+ *   ...args: unknown[]
+ * ): Promise<WorkflowTypeAndArgs> {
+ *   return {
+ *     handler: 'default',
+ *     workflowType,
+ *     args,
+ *   };
+ * };
```
bergundy, line 429: This return type is a bit confusing, I'd put a console log here instead.

```suggestion
 * ): Promise<void> {
 *   console.log(`Using default handler for `${workflowInfo().workflowType}`, { args });
 *   // Workflow logic comes here
 * };
```

packages/workflow/src/internals.ts
```diff
@@ -136,6 +137,11 @@ export class Activator implements ActivationHandler {
    */
   readonly signalHandlers = new Map<string, WorkflowSignalType>();
 
+  /**
+   * A signal handler that catches calls for non-registered signal names.
+   */
+  defaultSignalHandler: DefaultSignalHandler | undefined;
```
bergundy, line 143: nit:
```suggestion
  defaultSignalHandler?: DefaultSignalHandler;
```

packages/workflow/src/interfaces.ts
```diff
@@ -380,3 +388,47 @@ export interface WorkflowCreateOptions {
...
+  Args extends any[],
+  T extends SignalDefinition<Args> | QueryDefinition<Ret, Args>
+> = T extends SignalDefinition<infer A>
+  ? (...args: A) => void | Promise<void>
+  : T extends QueryDefinition<infer R, infer A>
+  ? (...args: A) => R
+  : never;
+
+/**
+ * ???
```
bergundy, line 406: Missed this?

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -338,7 +338,7 @@ export abstract class BaseVMWorkflow implements Workflow {
     await new Promise(setImmediate);
     if (this.unhandledRejection) {
       return {
-        runId: activation.runId,
+        runId: this.activator.info.runId,
```
bergundy, line 341: Curious why you changed this

packages/test/src/test-default-workflow.ts
```diff
@@ -0,0 +1,39 @@
...
+    const client = new WorkflowClient();
+    await worker.runUntil(async () => {
+      const result = client.execute('non-existing', { taskQueue, workflowId: uuid4(), args: ['test', 'foo', 'bar'] });
+      t.is((await result).handler, 'default');
+      t.is((await result).workflowType, 'non-existing');
+      t.deepEqual((await result).args, ['test', 'foo', 'bar']);
+    });
+  });
+
+  test('Default workflow handler is not used if requested workflow does exists', async (t) => {
```
bergundy, line 26:
```suggestion
  test('Default workflow handler is not used if requested workflow exists', async (t) => {
```

packages/test/src/test-default-workflow.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Test usage of a default workflow handler
+ */
+import test from 'ava';
+import { v4 as uuid4 } from 'uuid';
+import { WorkflowClient } from '@temporalio/client';
+import { RUN_INTEGRATION_TESTS, Worker } from './helpers';
+import { existing } from './workflows/default-workflow-function';
+
+if (RUN_INTEGRATION_TESTS) {
+  test('Default workflow handler is used if requested workflow does not exists', async (t) => {
```
bergundy, line 11:
```suggestion
  test('Default workflow handler is used if requested workflow does not exist', async (t) => {
```

packages/test/src/test-default-workflow.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Test usage of a default workflow handler
+ */
+import test from 'ava';
+import { v4 as uuid4 } from 'uuid';
+import { WorkflowClient } from '@temporalio/client';
+import { RUN_INTEGRATION_TESTS, Worker } from './helpers';
+import { existing } from './workflows/default-workflow-function';
+
+if (RUN_INTEGRATION_TESTS) {
```
bergundy, line 10: I prefer using the test environment for our newer tests.

See [this](https://github.com/temporalio/sdk-typescript/commit/b6669fc9eb6c8613ee3f673a80f33b087584c082#diff-426424443d6004f9a1ad57cffeb1dde07cf735727acd8861e2147d6a0af828e0) for reference.

bergundy: We'll get better coverage this way and it'll allow us to easily run our tests against different server implementations in the future.

bergundy: (better OS coverage)

packages/workflow/src/internals.ts
```diff
@@ -526,10 +532,13 @@ export class Activator implements ActivationHandler {
 
   public async signalWorkflowNextHandler({ signalName, args }: SignalInput): Promise<void> {
     const fn = this.signalHandlers.get(signalName);
-    if (fn === undefined) {
+    if (fn !== undefined) {
```
bergundy, line 535: Maybe check `!= null` just in case?

packages/workflow/src/workflow.ts
```diff
@@ -1168,20 +1157,48 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+  const activator = getActivator();
+  const bufferedSignals = activator.bufferedSignals;
+  while (bufferedSignals.length) {
+    if (activator.defaultSignalHandler) {
+      // We have a default signal handler, so all signals are dispatchable
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      activator.signalWorkflow(bufferedSignals.shift()!);
+    } else {
+      const foundIndex = bufferedSignals.findIndex(
+        (signal) => activator.signalHandlers.get(signal.signalName ?? '') !== undefined
```
bergundy, line 1193: You can use `has` here

packages/workflow/src/workflow.ts
```diff
@@ -1168,20 +1157,48 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+  const activator = getActivator();
+  activator.defaultSignalHandler = handler;
+  if (handler !== undefined) dispatchBufferedSignals();
+}
+
+function dispatchBufferedSignals() {
+  const activator = getActivator();
+  const bufferedSignals = activator.bufferedSignals;
+  while (bufferedSignals.length) {
+    if (activator.defaultSignalHandler) {
```
bergundy, line 1187: While I see why this is correct, it's a bit confusing that we check the default handler first.
Essentially we have 2 distinct functions in one, and I think it'd be clearer if we kept the logic in the signal setter.

packages/workflow/src/workflow.ts
```diff
@@ -1168,20 +1157,48 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
   const activator = getActivator();
   if (def.type === 'signal') {
     activator.signalHandlers.set(def.name, handler as any);
```
bergundy, line 1159: Let's change this to check if `handler != null` and unset the handler.

packages/workflow/src/workflow.ts
```diff
@@ -1168,20 +1157,48 @@ export function setHandler<Ret, Args extends any[], T extends SignalDefinition<A
...
+ * Signals are dispatched to the default signal handler in the order that they were accepted by the server.
+ *
+ * If this function is called multiple times for a given signal or query name the last handler will overwrite any previous calls.
+ *
+ * @param handler a function that will handle signals for non-registered signal names, or `undefined` to unset the handler.
+ */
+export function setDefaultSignalHandler(handler: DefaultSignalHandler | undefined): void {
+  const activator = getActivator();
+  activator.defaultSignalHandler = handler;
+  if (handler !== undefined) dispatchBufferedSignals();
```
bergundy, line 1180: JS people love their `null`s.

```suggestion
  if (handler != null) dispatchBufferedSignals();
```

bergundy (approved):
Overall LGTM.
Had some comments that should be easily addressable.

packages/workflow/src/interfaces.ts
mjameswh: Idk. Had some hesitation about this one.

On one side, having the `workflowType` arg helps make it clear that this function is not a regular workflow. Also, there is no equivalent way to get the current `signalName` in the default signal handler, so it gives some coherence between all "default handlers".

On the other side, if the user forget to add this `workflowType` argument, it get swallowed in by the following `...args` parameter without any compiler hint that he missed something.

TBH I have no strong opinion on either side.

packages/workflow/src/interfaces.ts
mjameswh: Told myself 20 times: "I must not forget to fix that one."

Guess I was too eager to finish yesterday. 😄

packages/worker/src/workflow/vm-shared.ts
mjameswh: In `test-workflows.ts`, we set `this.activator.info.runId` to the title of the test, but `activation.runId` to `test-runId`.

Then, if the activation completes without issue, the completion response use `this.activator.info.runId`, but if it fails, it uses `activation.runId`.

Now, in test-workflows.ts' `activate()`, we do:

```
  const { workflow, runId } = t.context;
  const completion = await workflow.activate(activation);
  t.deepEqual(completion.runId, runId);   // <------ HERE
  return completion;
}
```

There, we compare the `completion.runId` to the expected `runId`, which corresponds to `this.activator.info.runId`. The sum of all of this is that if the activation fails with an exception, the `t.deepEquals()` will report an incorrect `runId` before the test code itself gets the chance to even inspect the content of the activation response itself. That makes it more difficult to locate the true reason of the failure.

This can be resolved inside `test-workflows.ts` instead, but it seems to me that `this.activator.info.runId` is really the correct thing to use, rather `activation.runId`.

packages/test/src/test-default-workflow.ts
mjameswh: You're right. Will fix.

packages/workflow/src/workflow.ts
mjameswh: The difficulty is that handlers might get registered or unregistered from inside a signal handler. The dispatch function therefore must be reentrant-safe.

Imagine a situation where I receive signals (A, A, B, B, C, C), with following handlers:

```
setHandler(signalC, () => {
   setHandler(signalC, undefined);
   setHandler(signalB, () => {
       setHandler(signalB, undefined);
       setDefaultSignalHandler(() => { ... })
   })
})
```

This is why I always reevaluate the whole list, from the beginning, and restart every time a single signal gets dispatched.

Note that the dispatch loop is now O(N^2), but we don't expect N to grow anything big.

A possible optimisation here would be to have a "signal handler changed" counter, and only restart the loop if that counter changed while dispatching a signal. That would bring the dispatch function down to O(N) in most cases, but I don't think that the potential performance gain in extremely rare cases justify the increased complexity that this solution involves.

cretz:
Didn't read review, but weren't there some concerns that by using `export default` which is a common thing in JS that we'd be over-encouraging the use of dynamic workflow handlers? Did we want to consider making it a bit more unwieldy and/or requiring maybe an explicit opt-in boolean at registration time saying "I am intentionally using dynamic workflow" and throw if default exported function is present otherwise?

bergundy:
Yes, there's some concern about `export default` being common in JS.
We'd considered using a different convention but this seems like the best option since we don't want to prevent other strings, we can't use symbols, and default is reserved in JS.
I think requiring opt-in will be safer and also a bit more annoying but I can see a point for preferring safety especially when adding this convention to a mature SDK.

packages/workflow/src/interfaces.ts
mjameswh: Lol! That was totally unintended. Copy-Paste mistake.

packages/workflow/src/workflow.ts
bergundy: Good point! I didn't account for registration of signal handlers inside signal handlers.
We'll need to check this to ensure ordering.
It's probably worth moving this code to the activator since it's relying on some of its internals.

packages/workflow/src/internals.ts
```diff
@@ -538,14 +547,8 @@ export class Activator implements ActivationHandler {
...
 
-    const fn = this.signalHandlers.get(signalName);
-    if (fn === undefined) {
-      let buffer = this.bufferedSignals.get(signalName);
-      if (buffer === undefined) {
-        buffer = [];
-        this.bufferedSignals.set(signalName, buffer);
-      }
-      buffer.push(activation);
+    if (this.signalHandlers.get(signalName) === undefined && this.defaultSignalHandler === undefined) {
```
bergundy, line 550: nit: can use `has`

PR #1047, opened 2023-02-06
fix(workflow): SDK Internal patch causes non-determinism if no longer required

mjameswh:
## What changed

- On receiving notification of a SDK Internal Patch, an identical command is now immediately generated.

## Why

- If workflow code was modified so that it no longer required that SDK internal patch, then core would complain of non-determinism due to the fact that history contained a patch event, but no corresponding patch command had been generated by the workflow.
- SDK internal patch can't be marked as "deprecated", because they may have side effect on future events.

### DIFF
--- a/packages/test/src/workflows/sdk-internal-patch-after.ts
+++ b/packages/test/src/workflows/sdk-internal-patch-after.ts
@@ -1,0 +1,15 @@
+/**
+ * "Before" version of a workflow used to reproduce an issue with SDK internal patch,
+ * where the user modifies his code so that the SDK internal patch no longer gets generated.
+ *
+ * @module
+ */
+import { condition, sleep } from '@temporalio/workflow';
+
+/**
+ * Patches the workflow from ./patch-and-condition-pre-patch, adds a patched statement inside a condition.
+ */
+export async function sdkInternalPatch(): Promise<void> {
+  await condition(() => false, 10);
+  await sleep(10);
+}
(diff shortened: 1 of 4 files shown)


Sushisource:
This strongly justifies my fear of us writing markers internally ourselves :laughing: :sweat_smile: 

We should eventually get rid of all of these in favor of the newer mechanism I'm currently working on.

mjameswh:
> This strongly justifies my fear of us writing markers internally ourselves 😆 😅
> 
> We should eventually get rid of all of these in favor of the newer mechanism I'm currently working on.

No debate here, I fully agree.

bergundy (commented):
I think this will make the situation worse and result in change of command order

packages/test/src/workflows/sdk-internal-patch-after.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @module
+ */
+import { condition, sleep } from '@temporalio/workflow';
+
+/**
+ * Patches the workflow from ./patch-and-condition-pre-patch, adds a patched statement inside a condition.
+ */
+export async function sdkInternalPatch(): Promise<void> {
+  await condition(() => false, 10);
```
bergundy, line 13: I think we just live with the fact that changing condition duration from 0 to anything else is a non deterministic change due to this internal patch unless we want to generate another patch that wraps both cases.

Let's wait a bit longer to have a standard way to do internal SDK patches.

mjameswh: I agree. Closing.

PR #1048, opened 2023-02-06
chore: Update sdk-core

mjameswh:
## What changed

- Renamed metric `workflow_task_queue_poll_failed` to `workflow_task_execution_failed`, because that's its correct name. It used to be that.

- Fix slot metrics appearing to be off-by-one because of reservation.


mjameswh:
Superseded by #1072.

Issue #1052, opened 2023-02-10
[Feature Request] Retain type information regarding signal/query `name`


josiah-roberts:
### Is your feature request related to a problem? Please describe.

Everything here also applies to `QueryDefinition`, but I'll use `SignalDefinition` in these examples.

Because `SignalDefinition<Args>` does not retain the type of its `name`, it can be difficult to create broader systems that keep track of the legal signals and queries that can be sent to a workflow. 

For example:
```typescript
const signalA = defineSignal<[string]>('a');
const signalB = defineSignal<[string]>('b');

export function myWorkflow() {
  ... use these signals
}

// This doesn't work: these are just SignalDefinition<[string]>
// it matches *any* SignalDefinition<[string]>, not just 'a' and 'b'
export type SupportedSignals = typeof signalA | typeof signalB;

// This also doesn't work: SignalDefinition<_Args>['name'] is typed as string
export type SupportedSignalNames = typeof signalA.name | typeof signalB.name;
```

As a result, there is no straightforward way to define within the type system what the legal queries/signals to send to a workflow are. We can use proximity ("this signal was exported by the same module as the workflow, it probably makes sense to send it"), but we can't create structures that have `tsc` enforce this for us.

### Describe the solution you'd like

This issue can be fixed by varying the type of `name` with the type of the passed in string.

`SignalDefinition` becomes:
```typescript
export interface SignalDefinition<Args extends any[] = [], Name extends string = string> {
  type: 'signal';
  name: Name;
}
```

This presents some challenges for the definition of `defineSignal` though. You can't have partially-inferred generics in typescript, so we are forced to specify the name on both a type and a value level
```typescript
export function defineSignal<Name extends string, Args extends any[] = []>(name: Name): SignalDefinition<Args, Name> {
  return {
    type: 'signal',
    name,
  };
}

// Annoying to have to specify 'some name' twice
const stringSignal = defineSignal<'some-name', [string]>('some-name');
```

A curried alternative removes the duplication, though with a slightly unfamiliar style. This would be my preference.
```typescript
function defineNamedSignal<Name extends string>(name: Name): <Args extends any[]>() => SignalDefinition<Args, Name> {
  return () => ({
    type: 'signal',
    name,
  })
}

defineNamedSignal('some-name')<[string]>();
```

Finally, we could also make this level of specificity opt-in, so that only consumers who care to reason about names on a type-level pay the cost for it:

```typescript
[...]

PR #1056, opened 2023-02-13
Stronger query and signal definitions

### LINKED ISSUE #1051: [Feature Request] Retain type information for signal/query `_Args`

Everything here also applies to `QueryDefinition`, but I'll use `SignalDefinition` in these examples.

Because `SignalDefinition<Args>` does not use `_Args` in the resulting type, and because TypeScript is structurally typed, certain absurdities can arise. For example, the fact that a `SignalDefinition<[string]>` is a `SignalDefinition<[number]>`. The following is legal TS:

```typescript
function sendNumberSignal(s: SignalDefinition<[number]>, n: number) {
  // workflowHandle.signal(s, n)
}

const stringSignal = defineSignal<[string]>('some name');
sendNumberSignal(stringSignal, 21);
```

This creates some real problems when trying to perform type algebra or do higher-order reasoning about signal and query types for helpers and utility functions.

This issue can be fixed by no longer discarding the `<_Args>` type parameter, and instead using a virtual [type brand](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d) to preserve the information within the type system.

This brand doesn't exist at run-time, and since the symbol is not exported it is not possible for consumers to access the "virtual" property.

`SignalDefinition` becomes:
```typescript
declare const argsBrand: unique symbol;
export interface SignalDefinition<Args extends any[] = []> {
  type: 'signal';
  [argsBrand]: Args;
}
```

And `defineSignal` becomes:
```typescript
export function defineSignal<Args extends any[] = [], Name extends string = string>(name: Name): SignalDefinition<Args, Name> {
  return {
    type: 'signal',
    name,
  } as SignalDefinition<Args, Name>;
}
```

Having made these changes, `SignalDefinition<[string]>` is no longer assignable to `SignalDefinition<[number]>`, and type-level logic over the arg types of symbols becomes possible.

josiah-roberts:
## What was changed

This PR makes two refinements to the `SignalDefinition` and `QueryDefinition` types:
1. `Args` and `Ret` types are branded into the resulting types, using virtual symbolic brands. This addresses #1051
2. The `name` field is allowed to be narrowed to sub-type of `string` via an optional generic which defaults to `string`. I've used the least-intrusive option available here, which leaves behavior completely unchanged for existing callers, at the expense of making strict names a bit more verbose. This addresses #1052 

## Why?

Because #1051 and #1052 are making me sad.

## Checklist

1. Closes #1051 and #1052
2. How was this tested:  
    - [x] `npm run test`
    - [x] Added type-level tests
3. We may want to update the [Workflow TS docs](https://github.com/temporalio/documentation/blob/40176f12b79a73cc23481a3effdd98ddc2289e5e/docs-src/typescript/workflows.md)

### DIFF
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -47,7 +47,7 @@ export interface QueryDefinition<Ret, Args extends any[] = [], Name extends stri
    * This field is not present at run-time.
    */
   [argsBrand]: Args;
-    /**
+  /**
    * Virtual type brand to maintain a distinction between {@link QueryDefinition} types with different return types.
    * This field is not present at run-time.
    */
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -19,7 +19,7 @@ declare const argsBrand: unique symbol;
 /**
  * An interface representing a Workflow signal definition, as returned from {@link defineSignal}
  *
- * @remarks `Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.  
+ * @remarks `Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
  * `Name` can optionally be specified with a string literal type to preserve type-level knowledge of the signal name.
  */
 export interface SignalDefinition<Args extends any[] = [], Name extends string = string> {
@@ -36,7 +36,7 @@ declare const retBrand: unique symbol;
 /**
  * An interface representing a Workflow query definition as returned from {@link defineQuery}
  *
- * @remarks `Args` and `Ret` can be used for parameter type inference in handler functions and *WorkflowHandle methods.  
+ * @remarks `Args` and `Ret` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
  * `Name` can optionally be specified with a string literal type to preserve type-level knowledge of the query name.
  */
 export interface QueryDefinition<Ret, Args extends any[] = [], Name extends string = string> {
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -15,26 +15,31 @@ export type WorkflowQueryType = (...args: any[]) => any;
  */
 export type Workflow = (...args: any[]) => WorkflowReturnType;
 
+declare const argsBrand: unique symbol;
 /**
  * An interface representing a Workflow signal definition, as returned from {@link defineSignal}
  *
  * @remarks `_Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
  */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export interface SignalDefinition<_Args extends any[] = []> {
+export interface SignalDefinition<Args extends any[] = [], Name extends string = string> {
   type: 'signal';
-  name: string;
+  name: Name;
+  [argsBrand]: Args;
 }
 
+declare const retBrand: unique symbol;
 /**
  * An interface representing a Workflow query definition as returned from {@link defineQuery}
  *
  * @remarks `_Args` and `_Ret` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
  */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export interface QueryDefinition<_Ret, _Args extends any[] = []> {
+export interface QueryDefinition<Ret, Args extends any[] = [], Name extends string = string> {
   type: 'query';
-  name: string;
+  name: Name;
+  [argsBrand]: Args;
+  [retBrand]: Ret;
 }
 
 /** Get the "unwrapped" return type (without Promise) of the execute handler from Workflow type `W` */
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -19,12 +19,16 @@ declare const argsBrand: unique symbol;
 /**
  * An interface representing a Workflow signal definition, as returned from {@link defineSignal}
  *
- * @remarks `Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
+ * @remarks `Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.  
+ * `Name` can optionally be specified with a string literal type to preserve type-level knowledge of the signal name.
  */
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export interface SignalDefinition<Args extends any[] = [], Name extends string = string> {
   type: 'signal';
   name: Name;
+  /**
+   * Virtual type brand to maintain a distinction between {@link SignalDefinition} types with different args.
+   * This field is not present at run-time.
+   */
   [argsBrand]: Args;
 }
 
@@ -32,13 +36,21 @@ declare const retBrand: unique symbol;
 /**
  * An interface representing a Workflow query definition as returned from {@link defineQuery}
  *
- * @remarks `Args` and `Ret` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
+ * @remarks `Args` and `Ret` can be used for parameter type inference in handler functions and *WorkflowHandle methods.  
+ * `Name` can optionally be specified with a string literal type to preserve type-level knowledge of the query name.
  */
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 export interface QueryDefinition<Ret, Args extends any[] = [], Name extends string = string> {
   type: 'query';
   name: Name;
+  /**
+   * Virtual type brand to maintain a distinction between {@link QueryDefinition} types with different args.
+   * This field is not present at run-time.
+   */
   [argsBrand]: Args;
+    /**
+   * Virtual type brand to maintain a distinction between {@link QueryDefinition} types with different return types.
+   * This field is not present at run-time.
+   */
   [retBrand]: Ret;
 }
 
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -19,7 +19,7 @@ declare const argsBrand: unique symbol;
 /**
  * An interface representing a Workflow signal definition, as returned from {@link defineSignal}
  *
- * @remarks `_Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
+ * @remarks `Args` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
  */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 export interface SignalDefinition<Args extends any[] = [], Name extends string = string> {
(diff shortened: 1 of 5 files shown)


josiah-roberts:
@bergundy I'm running into an integration test issue, and wonder if you have any suggestions on where I should look to understand it.

I have `npm run test` passing, and most of the integration tests as well with `RUN_INTEGRATION_TESTS=true`. However, all of the `schedules` tests are failing with `4 DEADLINE_EXCEEDED: context deadline exceeded`. 

Any thoughts on what could cause this?

E.g.
```
@temporalio/test:   schedules › Can create schedule with startWorkflow action (no arg)
@temporalio/test:   Rejected promise returned by test. Reason:
@temporalio/test:   ServiceError {
@temporalio/test:     cause: Error {
@temporalio/test:       code: 4,
@temporalio/test:       details: 'context deadline exceeded',
@temporalio/test:       metadata: Metadata {
@temporalio/test:         internalRepr: Map { … },
@temporalio/test:         options: {},
@temporalio/test:       },
@temporalio/test:       message: '4 DEADLINE_EXCEEDED: context deadline exceeded',
@temporalio/test:     },
@temporalio/test:     message: 'Failed to describe schedule',
@temporalio/test:   }
@temporalio/test:   › ScheduleClient.rethrowGrpcError (/home/josiah-roberts/sdk-typescript/packages/client/src/schedule-client.ts:510:13)
@temporalio/test:   › ScheduleClient._describeSchedule (/home/josiah-roberts/sdk-typescript/packages/client/src/schedule-client.ts:282:12)
@temporalio/test:   › Object.describe (/home/josiah-roberts/sdk-typescript/packages/client/src/schedule-client.ts:415:21)
@temporalio/test:   › <anonymous> (src/test-schedules.ts:170:33)
```

bergundy:
@josiah-roberts I think this test is known to be flaky. CC @mjameswh.

packages/common/src/interfaces.ts
```diff
@@ -15,26 +15,31 @@ export type WorkflowQueryType = (...args: any[]) => any;
...
+ * @remarks `Args` and `Ret` can be used for parameter type inference in handler functions and *WorkflowHandle methods.
  */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export interface QueryDefinition<_Ret, _Args extends any[] = []> {
+export interface QueryDefinition<Ret, Args extends any[] = [], Name extends string = string> {
   type: 'query';
-  name: string;
+  name: Name;
+  [argsBrand]: Args;
+  [retBrand]: Ret;
```
bergundy, line 42: We could use some comments explaining the purpose of the brands.

bergundy (approved):
Would be nice to have some tests demonstrating that definitions can't be assigned incompatibly.

The tests shouldn't run any assertions, just validate compilation.

Let me know if you need help with this.
Thanks!

packages/common/src/interfaces.ts
josiah-roberts: Added :raised_hands:

josiah-roberts:
> Would be nice to have some tests demonstrating that definitions can't be assigned incompatibly.
> 
> The tests shouldn't run any assertions, just validate compilation.
> 
> Let me know if you need help with this. Thanks!

Done! I've gone ahead and actually called `define*` and used `typeof`, because it gives more full coverage - it actually exposed a [bug](98752e666fdd4bb588bfefc3b026675c8d961dbc), so tests working as-inteded.

Wherever possible in these tests, I've used specific assertions instead of `@ts-expect-error`, because it's very easy for you to end up in a state where `@ts-expect-error` is seeing an error, but not the *right* error as code drifts around.

packages/common/src/interfaces.ts
```diff
@@ -15,26 +15,43 @@ export type WorkflowQueryType = (...args: any[]) => any;
...
-  name: string;
+  name: Name;
+  /**
+   * Virtual type brand to maintain a distinction between {@link SignalDefinition} types with different args.
+   * This field is not present at run-time.
+   */
+  [argsBrand]: Args;
 }
 
+declare const retBrand: unique symbol;
```
bergundy, line 35: should this be exported?

josiah-roberts: No, it's intentionally not exported - the idea here is that unlike patching `__brand: T` onto the type, `[retBrand]: T` is _inaccessable_: callers cannot attempt to do `myQueryDef[retBrand]`, expect to get `Ret`, and get `undefined`, because they cannot access the [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) needed to look up the key.

This is also why I've used `declare const retBrand: unique symbol` instead of `const retBrand = Symbol()`; the idea is that this key a) exists only in the type system, not at run time and b) you cannot even attempt to read it. It's just a type discriminator.

packages/common/src/interfaces.ts
```diff
@@ -15,26 +15,43 @@ export type WorkflowQueryType = (...args: any[]) => any;
...
+export interface QueryDefinition<Ret, Args extends any[] = [], Name extends string = string> {
   type: 'query';
-  name: string;
+  name: Name;
+  /**
+   * Virtual type brand to maintain a distinction between {@link QueryDefinition} types with different args.
+   * This field is not present at run-time.
+   */
+  [argsBrand]: Args;
+    /**
```
jparise, line 50:
```suggestion
  /**
```

packages/common/src/interfaces.ts
bergundy: That makes sense.

packages/workflow/src/workflow.ts
```diff
@@ -1122,11 +1122,11 @@ function conditionInner(fn: () => boolean): Promise<void> {
  * Definitions are used to register handler in the Workflow via {@link setHandler} and to signal Workflows using a {@link WorkflowHandle}, {@link ChildWorkflowHandle} or {@link ExternalWorkflowHandle}.
  * Definitions can be reused in multiple Workflows.
  */
-export function defineSignal<Args extends any[] = []>(name: string): SignalDefinition<Args> {
+export function defineSignal<Args extends any[] = [], Name extends string = string>(name: Name): SignalDefinition<Args, Name> {
```
bergundy, line 1125: I don't love this extra type arg since it cannot be inferred from the function's parameter and requires duplicate definition.

Some alternatives, by least to most favorite:

```ts
export function defineSignalUglyVersion<Args extends any[] = any[], Name extends string = string>(
  name: Name,
  args?: Args
): SignalDefinition<Args, Name> {
  return {} as any;
}

export function namedSignalDefiner<Name extends string>(name: Name): Definer<Name> {
  return {} as any;
}

defineSignal<[string]>('a');
defineSignalUglyVersion('a', {} as [string]);
defineSignalUglyVersion('a');

interface Definer<Name extends string> {
  define<Args extends any[]>(): SignalDefinition<Args, Name>;
}

const s = namedSignalDefiner('a').define<[string]>();

// Another version of this can rewrite the returned object keys to be `signal${K}` but that's probably overkill.
export type SignalsForDefs<T> = {
  [K in keyof T & string]: T[K] extends (...args: infer A) => void ? SignalDefinition<A, K> : never;
};

export function defineSignals<Defs>(): SignalsForDefs<Defs> {
  return {} as any;
}

interface MySignals {
  a(): void;
}

const signals = defineSignals<MySignals>();
```

WDYT?

bergundy: Other variations of these solutions are possible too.

@mjameswh your input is also welcome.

mjameswh: Combining/extending some of @bergundy ideas by using overloads on `defineSignal`:

```
interface SignalDefinitionWithUnspecifiedArgs<Name extends string> extends SignalDefinition<[], Name> {
  withArgs<Args extends any[]>(): SignalDefinition<Args, Name>;
}

// defineSignal('signalA')
//  - Return value extends SignalDefinition<[], Name> which maintain backward compatibility
//  - Return value allows refining arg types using .withArgs()
//
export function defineSignal<Name extends string>(name: Name): SignalDefinitionWithUnspecifiedArgs<Name>;

// defineSignal('signalB', mySignalHandlerFunction)
//  - deduce signal args from the handler function
//  - technically nice, but probably rarely applicable in practice, as handlers are generally not in scope at the moment `defineSignal` is called (ie. handlers are generally defined as inner functions inside the workflow function itself)
//  - Can be worked around by creating an anonymous noop function just for the purpose of `defineSignal`, but that is no better than withArgs<[ ... ]>()
//
export function defineSignal<Args extends any[], Name extends string>(
  name: Name,
  fn: (...args: Args) => void | Promise<void>
): SignalDefinition<Args, Name>;

// defineSignal<[number, string]>('signalC')
//   - backward compatible form (note that in this case, Name is string)
//
// defineSignal<[number, string], 'signalC'>('signalC')
//   - Explicit format, where both Args and Name are specified
//
export function defineSignal<Args extends any[], Name extends string = string>(
  name: Name
): SignalDefinition<Args, Name>;

// Implementation
export function defineSignal<Args extends any[] = [], Name extends string = string>(
  name: Name
): SignalDefinitionWithUnspecifiedArgs<Name> | SignalDefinition<Args, Name> {
  return {
    type: 'signal',
    name,
  } as SignalDefinitionWithUnspecifiedArgs<Args, Name>;
}

// Example usage
const signalFoo = defineSignal('foo').withArgs<number, string>();
```

mjameswh: As a side note, we have two features on our road map that might also require revisiting the way signal handlers are defined:

- Support for signal/query introspection (ie. allow for the web UI to present a list of signals and queries supported by a given workflow, including the number and type of arguments)
- Make it possible to register multiple signal handlers at once, so that buffered signals get dispatch in the right order.

If we are to actually change the way signals are defined, I think we should at least consider either the new API actually get us closer from these two goals (don't mean we should _implement_ those features immediately). It would be disappointing that we encourage users to change their code now, and then come up with a completely different recommandation in three months...

bergundy: In the meantime, I suggest we merge this since it's only improving the existing situation and consider a better API later on.

josiah-roberts: Sorry I haven't been able to get back to these considerations - great conversation here around a lot of ergonomics options. As @bergundy mentioned, I've intentionally written this PR to be zero-impact on existing callers (other than type safety), and that does mean some compromises.

> Support for signal/query introspection (ie. allow for the web UI to present a list of signals and queries supported by a given workflow, including the number and type of arguments)

This would be fantastic; my team is about to implement a workflow abstraction layer to do this ourselves, based on the changes in this PR. But first-class support would of course be better.

josiah-roberts:
@bergundy please let me know if there's any blockers on my end for this, and I'll prioritize

bergundy:
@josiah-roberts no blockers, we will merge as is.
Thanks!

PR #1067, opened 2023-03-01
feat(worker): `Worker.runReplayHistories` returns an `AsyncIterableIterator`

### LINKED ISSUE #1064: [Feature Request] Return well-formed error OR success if Worker.runReplayHistories receives no histories

Today, it looks like if I pass a `histories` AsyncIterable that yields no results, I get an error like the one below, rather than one with a clear error code that I could catch/handle. Given that the goal is to find determinism errors, and that there wouldn't be any determinism errors if there are no histories, I question whether this should throw an error at all.

```
Error
    at _super (path/to/node_modules/rxjs/src/internal/util/createErrorClass.ts:13:22)
    at new EmptyErrorImpl (path/to/node_modules/rxjs/src/internal/util/EmptyError.ts:26:3)
    at Object.complete (path/to/node_modules/rxjs/src/internal/lastValueFrom.ts:71:18)
    at ConsumerObserver.complete (path/to/node_modules/rxjs/src/internal/Subscriber.ts:185:25)
    at SafeSubscriber.Subscriber._complete (path/to/node_modules/rxjs/src/internal/Subscriber.ts:132:24)
    at SafeSubscriber.Subscriber.complete (path/to/node_modules/rxjs/src/internal/Subscriber.ts:106:12)
    at checkComplete (path/to/node_modules/rxjs/src/internal/operators/mergeInternals.ts:48:18)
    at path/to/node_modules/rxjs/src/internal/operators/mergeInternals.ts:140:7
    at OperatorSubscriber._this._complete (path/to/node_modules/rxjs/src/internal/operators/OperatorSubscriber.ts:92:13)
    at OperatorSubscriber.Subscriber.complete (path/to/node_modules/rxjs/src/internal/Subscriber.ts:106:12)
    at path/to/node_modules/rxjs/src/internal/observable/innerFrom.ts:131:14
    at step (path/to/node_modules/rxjs/dist/cjs/internal/observable/innerFrom.js:33:23)
    at Object.next (path/to/node_modules/rxjs/dist/cjs/internal/observable/innerFrom.js:14:53)
    at fulfilled (path/to/node_modules/rxjs/dist/cjs/internal/observable/innerFrom.js:5:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  message: 'no elements in sequence'
```

Suppose I get histories like this:
```typescript
const query = `WorkflowType="${workflowType}" and ExecutionStatus = 'Running'
    ORDER BY StartTime ASC`;
const workflowList = this.workflowClient.list({
    query,
    pageSize: 50,
});
const histories = workflowList.intoHistories({ concurrency: 10 });
```

and call the replayer like this:
```typescript
const replayResults = await Worker.runReplayHistories(
    {
      workflowBundle,
      sinks,
      failFast: false, // can toggle to true if replay is slow, but will need to catch error
    },
    histories
);
```

If the # of `histories` returned from the query above happens to be 0, I'd like for the replay to succeed.
If there are reasons why the replayer should fail, I'd like for there to be a specific error code that I could handle from the `await Worker.runReplayHistories()` call.

### Additional context

n/a

bergundy:
This is a breaking change for an experimental API.

Closes #1064

- Removed the `failFast` replay worker option
- Removed the `ReplayResults` type in favor of the singular `ReplayResult`

### DIFF
--- a/packages/worker/src/replay.ts
+++ b/packages/worker/src/replay.ts
@@ -33,17 +33,10 @@ export class ReplayError extends Error {
   }
 }
 
-export interface ReplayResults {
-  readonly hasErrors: boolean;
-  /** Maps run id to information about the replay failure */
-  readonly errors: ReplayError[];
-}
-
-/**
- * @internal
- */
-export interface ReplayRunOptions {
-  failFast?: boolean;
+export interface ReplayResult {
+  workflowId: string;
+  runId: string;
+  error?: ReplayError;
 }
 
 /**
@@ -51,7 +44,7 @@ export interface ReplayRunOptions {
  *
  * @experimental - this API is not considered stable
  */
-export type ReplayHistoriesIterable = AsyncIterable<HistoryAndWorkflowId> | Iterable<HistoryAndWorkflowId>;
+export type ReplayHistoriesIterable = AsyncIterableIterator<HistoryAndWorkflowId> | Iterable<HistoryAndWorkflowId>;
 
 /**
  * Handles known possible cases of replay eviction reasons.
@@ -60,7 +53,11 @@ export type ReplayHistoriesIterable = AsyncIterable<HistoryAndWorkflowId> | Iter
  *
  * @internal
  */
-export function handleReplayEviction(evictJob: RemoveFromCache, workflowId: string, runId: string): ReplayError | null {
+export function evictionReasonToReplayError(
+  evictJob: RemoveFromCache,
+  workflowId: string,
+  runId: string
+): ReplayError | undefined {
   switch (evictJob.reason) {
     case EvictionReason.NONDETERMINISM:
       return new ReplayError(
@@ -81,7 +78,7 @@ export function handleReplayEviction(evictJob: RemoveFromCache, workflowId: stri
     // LANG_REQUESTED is used internally by Core to support duplicate runIds during replay.
     case EvictionReason.LANG_REQUESTED:
     case EvictionReason.CACHE_FULL:
-      return null;
+      return undefined;
     case undefined:
     case null:
     case EvictionReason.UNSPECIFIED:
--- a/packages/worker/src/replay.ts
+++ b/packages/worker/src/replay.ts
@@ -1,5 +1,6 @@
 import { HistoryAndWorkflowId } from '@temporalio/client';
 import { coresdk } from '@temporalio/proto';
+import { DeterminismViolationError } from '@temporalio/workflow';
 
 export type EvictionReason = coresdk.workflow_activation.RemoveFromCache.EvictionReason;
 export const EvictionReason = coresdk.workflow_activation.RemoveFromCache.EvictionReason;
@@ -10,33 +11,15 @@ export type RemoveFromCache = coresdk.workflow_activation.IRemoveFromCache;
  */
 export class ReplayError extends Error {
   public readonly name = 'ReplayError';
-
-  constructor(
-    /**
-     * Workflow ID of the Workflow that failed to replay
-     */
-    public readonly workflowId: string,
-    /**
-     * Run ID of the Workflow that failed to replay
-     */
-    public runId: string,
-    /**
-     * Whether or not this error is caused by non-determinism
-     */
-    public readonly isNonDeterminism: boolean,
-    /**
-     * Why replay failed
-     */
-    message: string
-  ) {
-    super(message);
-  }
 }
 
+/**
+ * Result of a single workflow replay
+ */
 export interface ReplayResult {
-  workflowId: string;
-  runId: string;
-  error?: ReplayError;
+  readonly workflowId: string;
+  readonly runId: string;
+  readonly error?: ReplayError | DeterminismViolationError;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/worker/src/replay.ts
```diff
@@ -33,25 +33,18 @@ export class ReplayError extends Error {
...
+  error?: ReplayError;
 }
 
 /**
  * An iterable on worflow histories and their IDs, used for batch replaying.
  *
  * @experimental - this API is not considered stable
  */
-export type ReplayHistoriesIterable = AsyncIterable<HistoryAndWorkflowId> | Iterable<HistoryAndWorkflowId>;
+export type ReplayHistoriesIterable = AsyncIterableIterator<HistoryAndWorkflowId> | Iterable<HistoryAndWorkflowId>;
```
mjameswh, line 47: Why impose `AsyncIterableIterator` here? Also, it should be coherent with the non-sync type.

packages/worker/src/replay.ts
```diff
@@ -81,7 +78,7 @@ export function handleReplayEviction(evictJob: RemoveFromCache, workflowId: stri
     // LANG_REQUESTED is used internally by Core to support duplicate runIds during replay.
     case EvictionReason.LANG_REQUESTED:
     case EvictionReason.CACHE_FULL:
```
mjameswh, line 80: I'm surprised that we consider `cache_full` as a non-error. Can the replay worker legitimately ever send a `cache_full` event?

packages/worker/src/replay.ts
```diff
@@ -60,7 +53,11 @@ export type ReplayHistoriesIterable = AsyncIterable<HistoryAndWorkflowId> | Iter
  *
  * @internal
  */
-export function handleReplayEviction(evictJob: RemoveFromCache, workflowId: string, runId: string): ReplayError | null {
+export function evictionReasonToReplayError(
+  evictJob: RemoveFromCache,
+  workflowId: string,
```
mjameswh, line 58: Do you see any real reason to keep `workflowId` and `runId` in `ReplayError`? They are already in `ReplayResult`...

Also, in `ReplayError` (can't put a comment there), `workflowId` and `isNonDeterminism` are both `readonly`, but `runId` isn't. They should be coherent.

packages/worker/src/worker-options.ts
```diff
@@ -408,14 +408,6 @@ export interface ReplayWorkerOptions
    *  @default workflow name from given history
    */
   replayName?: string;
```
mjameswh, line 410: Is it really useful to expose this config? Also, default comment is wrong (actual value: `(options.replayName ?? 'fake_replay_queue') + '-' + this.replayWorkerCount`).

packages/worker/src/worker.ts
```diff
@@ -505,10 +506,13 @@ export class Worker {
   ): Promise<void> {
     const validated = this.validateHistory(history);
     try {
-      await this.runReplayHistories(
-        { ...options, failFast: true }, // Always failFast in single replay
-        [{ history: validated, workflowId: workflowId ?? 'fake' }]
-      );
+      for await (const result of this.runReplayHistories(options, [
```
mjameswh, line 509: I'd make some `first` or `single` function in iterator-utils.ts... That would make it clear that we are expecting only one result here.

packages/worker/src/worker.ts
```diff
@@ -531,25 +535,67 @@ export class Worker {
...
+          // We replay one workflow at a time so the workflow ID comes from the histories iterable.
+          const error = evictionReasonToReplayError(evictJob, workflowId, runId);
+          yield {
+            workflowId,
+            runId,
+            error,
+          };
+        }
+      } catch (err) {
+        innerError = err;
```
mjameswh, line 571: No sure about the intention here. Shouldn't we rethrow?

packages/worker/src/worker.ts
```diff
@@ -531,25 +535,67 @@ export class Worker {
...
+        worker.shutdown();
+      } catch {
+        // ignore in case worker was already shutdown
+      }
+      try {
+        await runPromise;
+      } catch (err) {
+        /* eslint-disable no-unsafe-finally */
+        if (err instanceof native.ShutdownError) {
+          return;
```
mjameswh, line 585: Seems to me that we shoud check and rethrow `innerError` in this block

packages/worker/src/replay.ts
bergundy: I wanted it to have a `next` awaitable so I could race it the run promise but ended up not using that.
I think it's fine to revert.

packages/worker/src/replay.ts
bergundy: I'll fix that, thanks for noticing.

packages/worker/src/worker-options.ts
bergundy: Core used to require this value, some users are setting it so I didn't want to remove.

packages/worker/src/worker.ts
bergundy: Yeah, you're right, thanks.

packages/worker/src/worker.ts
bergundy: Hmm.. sure

packages/worker/src/replay.ts
bergundy: On second thought let's keep the change in case we need to await `next`. I can see it come in handy when Core supports concurrent replays.

packages/worker/src/replay.ts
bergundy: `ReplayError` is used for replaying a single history but I think it's fine that it doesn't give back the workflow ID and run ID.

packages/worker/src/worker-options.ts
mjameswh: We may deprecate it though. Or otherwise, we should rewrite the associated doc so that users understand why/when/what they should provide.

PR #1068, opened 2023-03-02
fix: Fail workflow on WorkflowExecutionAlreadyStartedError

### LINKED ISSUE #586: [Bug] Parent workflow doesn't fail when child workflow encounters WorkflowExecutionAlreadyStartedError

Ensure workflows properly exit and don't stay in the Running state indefinitely.

### Describe the bug

If a workflow, `ParentWorkflow` attempts to execute a `ChildWorkflow`, but uses the same workflow ID as an already executing instance of `ChildWorkflow`, the 2nd instance will fail as expected with `WorkflowExecutionAlreadyStartedError`. However, the `ParentWorkflow` instance that attempted to spawn the 2nd child workflow remains in the `Running` state indefinitely and never exits. Even if the first instance of `ChildWorkflow` completes, and the workflow is restarted, it remains stuck in Running.

### Minimal Reproduction

```typescript
import { sleep, executeChild } from '@temporalio/workflow';

export async function parentWorkflow(): Promise<string> {
  return await executeChild('childWorkflow', { 
    workflowId: 'childWorkflowId', // <-- intentionally static
  }
}

export async function childWorkflow(): Promise<void> {
  sleep(60000);
}
```

Launch two workflows concurrently
```bash
tctl workflow start --workflow_type parentWorkflow
tctl workflow start --workflow_type parentWorkflow
```

### Environment/Versions

- OS and processor: M1 Mac running Monterray 12.2.1
- Temporal Version: Typescript SDK 0.19.2
- Are you using Docker or Kubernetes or building Temporal from source? Docker

bergundy:
Closes #586

### DIFF
--- a/packages/common/src/errors.ts
+++ b/packages/common/src/errors.ts
@@ -1,3 +1,5 @@
+import { TemporalFailure } from './failure';
+
 /**
  * Thrown from code that receives a value that is unexpected or that it's unable to handle.
  */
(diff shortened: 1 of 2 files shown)


mjameswh (approved):

mjameswh:
Might be worth a feature test to assert that this behaviour is the same across all SDKs.

bergundy:
It's not the same behavior, but I agree it's worth a feature test.

PR #1070, opened 2023-03-03
fix: Misc fixes related to older Node releases

### LINKED ISSUE #1063: [Bug] Temporal does not use standard AbortController/AbortSignal typings

I am trying to pass `Context.current().cancellationSignal` to [`execa`](https://github.com/sindresorhus/execa) so child processes can be automatically killed when a workflow is cancelled.

### Describe the bug

TypeScript is not happy:

```
Argument of type 'import("/.../node_modules/abort-controller/dist/abort-controller").AbortSignal' is not assignable to parameter of type 'AbortSignal'.
Type 'AbortSignal' is missing the following properties from type 'AbortSignal': reason, throwIfAborted
```

Execa expects a standard library `AbortSignal` and Temporal is using a third-party `AbortSignal`.

### Additional context

I saw that ceab74d9d608a0f808cd2a9f63012064800f5044 introduced the `abort-controller` dependency for compatibility with Node 14 as `AbortController` was introduced in Node 15. Does the SDK have any official support policy? There have been two LTS releases since Node 14 (Node 16 and Node 18), both of which support AbortController. Can we just drop the `abort-controller` dependency and use the standard library?

mjameswh:
## What changed

- Completetly remove support for Node versions <= 14.17. Lot of our dependencies were already requiring Node 14.18+ anyway.
- Load package 'abort-controller' as a polyfill rather than a complete substitution. This will ensure using native implementation of that class in Node 16+ (fix #1063)
- Added the "node:" prefix on all of our Node built-in modules imports, except for files that might get loaded inside the Workflow sandbox.

### DIFF
--- a/packages/test/src/activities/cancellable-fetch.ts
+++ b/packages/test/src/activities/cancellable-fetch.ts
@@ -1,8 +1,9 @@
 import fetch from 'node-fetch';
+import type { AbortSignal as FetchAbortSignal } from 'node-fetch/externals';
 import { Context } from '@temporalio/activity';
 
 export async function cancellableFetch(url: string): Promise<Uint8Array> {
-  const response = await fetch(url, { signal: Context.current().cancellationSignal });
+  const response = await fetch(url, { signal: Context.current().cancellationSignal as FetchAbortSignal });
   const contentLengthHeader = response.headers.get('Content-Length');
   if (contentLengthHeader === null) {
     throw new Error('expected Content-Length header to be set');
--- a/packages/worker/src/workflow/vm-shared.ts
+++ b/packages/worker/src/workflow/vm-shared.ts
@@ -1,5 +1,5 @@
-import v8 from 'v8';
-import vm from 'vm';
+import v8 from 'node:v8';
+import vm from 'node:vm';
 import { SourceMapConsumer } from 'source-map';
 import { cutoffStackTrace, IllegalStateError } from '@temporalio/common';
 import { coresdk } from '@temporalio/proto';
--- a/packages/worker/src/workflow/vm-shared.ts
+++ b/packages/worker/src/workflow/vm-shared.ts
@@ -278,8 +278,7 @@ export abstract class BaseVMWorkflow implements Workflow {
     protected context: vm.Context | undefined,
     protected activator: Activator,
     readonly workflowModule: WorkflowModule,
-    public readonly isolateExecutionTimeoutMs: number,
-    public readonly hasSeparateMicrotaskQueue: boolean
+    public readonly isolateExecutionTimeoutMs: number
   ) {}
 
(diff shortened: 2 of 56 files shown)


packages/test/src/activities/cancellable-fetch.ts
```diff
@@ -1,8 +1,9 @@
 import fetch from 'node-fetch';
+import type { AbortSignal as FetchAbortSignal } from 'node-fetch/externals';
 import { Context } from '@temporalio/activity';
 
 export async function cancellableFetch(url: string): Promise<Uint8Array> {
-  const response = await fetch(url, { signal: Context.current().cancellationSignal });
+  const response = await fetch(url, { signal: Context.current().cancellationSignal as FetchAbortSignal });
```
bergundy, line 6: Sucks that we have to do this

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -356,13 +355,6 @@ export abstract class BaseVMWorkflow implements Workflow {
    */
   protected async tryUnblockConditions(): Promise<void> {
```
bergundy, line 356: Doesn't need to be async anymore

bergundy (approved):
Had one minor comment.
Curious which tools have issues with the `node:` prefix, I would prefer to keep it if possible.

packages/test/src/activities/cancellable-fetch.ts
mjameswh: I agree, but node-fetch intentionally redefine the `AbortSignal` interface to also include DOM WG's methods, so passing a standard Node 18 `AbortSignal` won't compile. We could technically go their way and redefine the `AbortSignal` interface to include these extra methods, but then we risk creating even more incompatibilities. I think this cast is the least bad option.

packages/worker/src/workflow/vm-shared.ts
mjameswh: Good catch

mjameswh:
> Curious which tools have issues with the node: prefix

The only specific case I can trace back ATM was an outdated version of Jest, but I remember there have been some other cases (including Webpack under certain circonstances).

> I would prefer to keep it if possible

Why do you care? TBH I'm very slightly inclined to use the prefix, but at this point, we have not been using it consistently anyway.

bergundy:
If we do go for consistency, I'd change it to be the other way around but I'm not religious about it.
Agree that being consistent is better.

mjameswh:
>If we do go for consistency, I'd change it to be the other way around but I'm not religious about it. Agree that being consistent is better.

Fair enough. Added 'node:' prefix everywhere possible

bergundy:
You should modify the PR description.

PR #1072, opened 2023-03-07
feat(worker): Don't forcefully shutdown worker

bergundy:
BREAKING CHANGE: `WorkerOptions.shutdownGraceTime` no longer forcefully shuts the worker down. Set `WorkerOptions.shutdownForceTime` to force shutdown.

As before, in-flight activities are automatically cancelled on worker shutdown. Setting `shutdownGraceTime` will add a delay before cancelling the activities. Any activities that respond to this synthetic cancellation will fail their current attempt as opposted to ending up cancelled.

- Fixes an issue where `assert-to-failure-interceptors` did not propagate exceptions properly
- Upgrade to latest Core


bergundy:
The tests uncovered a bug in Core related to the SDK internal patches and queries, reproduced [here](https://github.com/temporalio/sdk-core/pull/504), I hope to get some time to fix this today.

mjameswh:
The shutdown refactor LGTM.

Just a note: we might have some other adjustments/tests to do wrt to update of sdk-core to HEAD. At first sight, I think the following PRs might require some work in TS code, though I need to think more about this.

- https://github.com/temporalio/sdk-core/pull/475
- https://github.com/temporalio/sdk-core/pull/482
- https://github.com/temporalio/sdk-core/pull/489
- https://github.com/temporalio/sdk-core/pull/495

Not blocking this PR however.

mjameswh (approved):

PR #1077, opened 2023-03-09
feat(testing): Use temporal CLI to power local test environment

### LINKED ISSUE #1055: Power local test environment with Temporal CLI

bergundy:
BREAKING CHANGE: the `TemporaliteConfig` interface was deleted but it likely was not directly referenced as only the deprecated `create` method accepted it.

Closes #1055

### DIFF
--- a/packages/testing/src/index.ts
+++ b/packages/testing/src/index.ts
@@ -29,13 +29,13 @@ import {
   EphemeralServer,
   EphemeralServerConfig,
   getEphemeralServerTarget,
-  TemporaliteConfig,
+  DevServerConfig,
   TimeSkippingServerConfig,
 } from '@temporalio/core-bridge';
 import { filterNullAndUndefined } from '@temporalio/common/lib/internal-non-workflow';
 import { Connection, TestService } from './connection';
 
-export { TimeSkippingServerConfig, TemporaliteConfig, EphemeralServerExecutable } from '@temporalio/core-bridge';
+export { TimeSkippingServerConfig, DevServerConfig, EphemeralServerExecutable } from '@temporalio/core-bridge';
 export { EphemeralServerConfig };
 
 export interface TimeSkippingWorkflowClientOptions extends WorkflowClientOptions {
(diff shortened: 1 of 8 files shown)


packages/testing/src/index.ts
```diff
@@ -250,16 +251,14 @@ export class TestWorkflowEnvironment {
    * {@link createTimeSkipping} does. {@link supportsTimeSkipping} will always return `false` for this environment.
    * {@link sleep} will sleep the actual amount of time and {@link currentTimeMs} will return the current time.
    *
-   * Internally, this uses [Temporalite](https://github.com/temporalio/temporalite). Which is a self-contained binary
-   * for Temporal using Sqlite persistence. This will download Temporalite to a temporary directory by default if it
+   * Internally, this uses [Temporal CLI](https://github.com/temporalio/cli). Which is a self-contained binary
```
mjameswh, line 254: I think this comment should be reworked.

Here is what I have in mind, but I don't think this is final:

```suggestion
   * This local environment will be powered by [Temporal CLI](https://github.com/temporalio/cli), which is a self-contained executable for Temporal. By default, Temporal's database will not be persisted to disk, and no UI will be started (FIXME Not sure about this one).

The CLI executable will be downloaded and cached to a temporary directory. See {@link LocalTestWorkflowEnvironmentOptions.server.executable.type} if you'd prefer to provide the CLI executable by yourself.
```

mjameswh: As of 0.5.0/1.20.0, is there any restrictions left in CLI compared to a Docker with ES server? If so, it would be useful to mention these, or if we want to make sure that documentation doesn't become stall, at least provide a link to a page that mention those differences.

mjameswh (approved):
LGTM

While we're at this, I'm wondering if we should not take the opportunity to remove the deprecated `create(opts?: TestWorkflowEnvironmentOptions)` function. I expect that most users of that function will anyway get a broken build there since they need to update the `type` field to `dev-server`...

PR #1082, opened 2023-03-17
Accept WorkflowOptions as const

bergundy:
Before this change, TypeScript would have difficulties inferring the type for workflow options provided via the various SDK APIs when defined out of context.

 For example, the following code would not compile:
```ts
const options = {
  workflowId: uuid4(),
  taskQueue: 'some-queue',
  searchAttributes: { someIntAttr: [1] },
  args: [3],
} as const;
const handle = await client.start(workflows.sleeper, options);
```

Note that args can only be inferred if the options are defined `as const` or the args are cast into a tuple (e.g. `args: [3] as [number]`).


rylandg:
Ty for this

PR #1086, opened 2023-03-30
docs: add install protocal buffers step to CONTRIBUTING.md doc

jhubbardsf:
## What was changed

Added a line to CONTRIBUTING.md mentioning to install protocal buffers after rust.

## Why?

While following the contributing guide I got an error that protobuf wasn't installed on my MacBook during `npm ci` step.


bergundy (approved):
Thanks!
I would mention alternative installation methods but this is great as is

jhubbardsf:
> Thanks! I would mention alternative installation methods but this is great as is

Yeah, I wasn't sure if more explicit directions per OS would be helpful (like `brew install protobuf` on MacOS). But I figured linking them to protocal buffers and telling them to install it would probably let them figure it out pretty fast.

PR #1096, opened 2023-04-12
Upgrade core

Sushisource:
## Why?
Allows SNI stuff to work properly by default when applying domain overrides in tls config

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/core-bridge/Cargo.lock
+++ b/packages/core-bridge/Cargo.lock
@@ -33,55 +33,56 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "0.7.20"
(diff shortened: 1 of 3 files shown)


bergundy (approved):

bergundy (changes requested):
On second thought, let's not expose originOverride until we see a need for it to keep the native and pure JS connection options as similar as possible.

Sushisource:
> On second thought, let's not expose originOverride until we see a need for it to keep the native and pure JS connection options as similar as possible.

That defeats the purpose of having it at all, which is if something like this happens again, the user has a path to resolving it without us cutting a release.

bergundy (approved):

PR #1104, opened 2023-04-19
fix: Fix multiple tests on Windows and/or Temporal CLI dev server

mjameswh:
## What changed

- Multiple tests were design to check for file path, but assumed forward-slash as directory separator. These tests now correctly deal with Window's backward-slash file separator.

- Temporal CLI's dev server does not support arrays of ints or dates as Search Attribute values. Modify existing tests to not use such arrays. Note: the array syntax is still supported with ES visiblity and in Cloud.

- Temporal CLI's dev server does not support the "IN" operator in workflow in list workflow query. Silently ignore that failure if it happens.

- Some tests are expecting some specific error messages, that differs on Windows or with the Temporal CLI's dev server. Modified these checks to include the missing variants.

- Some utility scripts used in CI tests were expecting a Unix-style filesystem (ie. "/tmp"). Modified these scripts so that paths can be user-specified.

- The `spawnNpx` function used by CI tests was executing `npx` under `cmd`. This fails in some situation if the parent shell is PowerShell. Executing it as `npx.cmd` instead seems to resolve the problem.

### DIFF
--- a/packages/test/src/helpers.ts
+++ b/packages/test/src/helpers.ts
@@ -51,7 +51,8 @@ export function cleanStackTrace(ostack: string): string {
     cleanedStack
       .replace(/:\d+:\d+/g, '')
       .replace(/^\s*/gms, '    at ')
-      .replace(/\[as fn\] /, '');
+      .replace(/\[as fn\] /, '')
+      .replace(/\\/g, '/');
 
   return normalizedStack ? `${firstLine}\n${normalizedStack}` : firstLine;
 }
--- a/packages/test/src/test-native-connection.ts
+++ b/packages/test/src/test-native-connection.ts
@@ -21,7 +21,7 @@ if (RUN_INTEGRATION_TESTS) {
   test('NativeConnection errors have detail', async (t) => {
     await t.throwsAsync(() => NativeConnection.connect({ address: 'localhost:1' }), {
       instanceOf: TransportError,
-      message: /.*Connection refused.*/,
+      message: /.*Connection[ ]?refused.*/i,
     });
   });
 
--- a/packages/testing/src/utils.ts
+++ b/packages/testing/src/utils.ts
@@ -15,7 +15,11 @@ export async function waitOnNamespace(
         execution: { workflowId: 'fake', runId },
       });
     } catch (err: any) {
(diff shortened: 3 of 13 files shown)


packages/test/src/helpers.ts
```diff
@@ -51,7 +51,8 @@ export function cleanStackTrace(ostack: string): string {
     cleanedStack
       .replace(/:\d+:\d+/g, '')
       .replace(/^\s*/gms, '    at ')
-      .replace(/\[as fn\] /, '');
+      .replace(/\[as fn\] /, '')
```
bergundy, line 54: 🤦

packages/test/src/test-native-connection.ts
```diff
@@ -21,7 +21,7 @@ if (RUN_INTEGRATION_TESTS) {
   test('NativeConnection errors have detail', async (t) => {
     await t.throwsAsync(() => NativeConnection.connect({ address: 'localhost:1' }), {
       instanceOf: TransportError,
-      message: /.*Connection refused.*/,
+      message: /.*Connection[ ]?refused.*/i,
```
bergundy, line 24: When would the space be optional? Do we see the string `ConnectionRefused`?

packages/testing/src/utils.ts
```diff
@@ -15,7 +15,11 @@ export async function waitOnNamespace(
         execution: { workflowId: 'fake', runId },
       });
     } catch (err: any) {
-      if (err.details.includes('workflow history not found') || err.details.includes(runId)) {
+      if (
+        err.details.includes('workflow history not found') ||
+        err.details.includes('Workflow executionsRow not found') ||
```
bergundy, line 20: sigh

bergundy (approved):

packages/test/src/test-native-connection.ts
mjameswh: If I'm not mistaken, that one was on Windows.

Really, all of those fixes are things I discovered while working on the CI refactor branch. I extracted these to a distinct PR only so that git history remains sane. You know... "Why did we made that space optionnal? ... Because of a CI refactor?!? ehhhhh...." :smiley:

PR #1106, opened 2023-04-21
fix(workflow): Fix internal patches NDE on replay

mjameswh:
## What changed

- Remove support for SDK internal patches.
- Modify `condition(fn, 0)` patch to use the regular `patched()` construct instead.

## Why

- Internal patches were previously causing NDE on replay
- Future internal patches will be based on WFT SDK metadata.


packages/test/src/test-internal-patches.ts
bergundy, line 1: Put this in `test-integration-workflows.ts` instead please

bergundy (commented):
I'll go over this next week, too late in the day to think this through.

packages/test/src/test-internal-patches.ts
mjameswh: I don't think we have a way to force replay in test-integration.ts, do we?

bergundy: You can create a worker per test in the file I mentioned, not the old integration tests

packages/test/src/test-integration-workflows.ts
```diff
@@ -204,3 +209,52 @@ test('Worker allows heartbeating activities after shutdown has been requested',
...
+  const worker = await createWorker({
+    // Disable workflow caching, to force replay after the condition's sleep
+    maxCachedWorkflows: 0,
+  });
+  await worker.runUntil(async () => {
+    try {
+      await executeWorkflow(conditionTimeout0Simple);
+      t.pass();
+    } catch (e) {
+      t.fail((e as Error).message);
```
bergundy, line 236: Why catch instead of letting the test throw an exception with all of the context?

packages/test/src/test-integration-workflows.ts
```diff
@@ -204,3 +209,52 @@ test('Worker allows heartbeating activities after shutdown has been requested',
...
+
+test('Internal patches does not cause non-determinism error on replay', async (t) => {
+  const { createWorker, executeWorkflow } = helpers(t);
+  const worker = await createWorker({
+    // Disable workflow caching, to force replay after the condition's sleep
+    maxCachedWorkflows: 0,
+  });
+  await worker.runUntil(async () => {
+    try {
+      await executeWorkflow(conditionTimeout0Simple);
```
bergundy, line 233: Should assert on the return value

packages/test/src/test-integration-workflows.ts
```diff
@@ -204,3 +209,52 @@ test('Worker allows heartbeating activities after shutdown has been requested',
...
+    t.is(timerStartedEvents?.[0].timerStartedEventAttributes?.timerId, '1');
+    t.is(tsToMs(timerStartedEvents?.[0].timerStartedEventAttributes?.startToFireTimeout), 10000);
+    t.is(timerStartedEvents?.[1].timerStartedEventAttributes?.timerId, '2');
+    t.is(tsToMs(timerStartedEvents?.[1].timerStartedEventAttributes?.startToFireTimeout), 1);
+
+    const markersEvents = history.events?.filter(({ eventType }) => eventType === EVENT_TYPE_MARKER_RECORDED);
+    t.is(markersEvents?.length, 1);
+    t.is(markersEvents?.[0].markerRecordedEventAttributes?.markerName, CHANGE_MARKER_NAME);
+  });
+});
```
bergundy, line 260: TBH I don't think we even need this test, the fact that the workflow returns false should be enough to cover the functionality.

packages/test/src/test-integration-workflows.ts
```diff
@@ -204,3 +209,52 @@ test('Worker allows heartbeating activities after shutdown has been requested',
...
 });
+
+export async function conditionTimeout0Simple(): Promise<boolean | undefined> {
+  let validationTimerFired = false;
+  // Don't lower this value, otherwise, in CI, there is risk that the timer fire
+  // while the WFT is being processed, which would result in an UnhandledCommand.
+  workflow
+    .sleep(10_000)
+    .then(() => (validationTimerFired = true))
+    .catch((e) => console.log(e));
```
bergundy, line 220: Why do we need this timer? Isn't it enough to check that condition returned false?

packages/workflow/src/internals.ts
```diff
@@ -617,22 +618,21 @@ export class Activator implements ActivationHandler {
...
 
   public checkInternalPatchAtLeast(minimumPatchNumber: number): boolean {
-    if (this.internalPatchNumber >= minimumPatchNumber) return true;
-    if (!this.info.unsafe.isReplaying) {
-      this.internalPatchNumber = minimumPatchNumber;
+    const usePatch = !this.info.unsafe.isReplaying || this.knownInternalPatchNumber >= minimumPatchNumber;
+    if (usePatch && this.sentInternalPatchNumber < minimumPatchNumber) {
       this.pushCommand({
-        setPatchMarker: { patchId: `__sdk_internal_patch_number:${LATEST_INTERNAL_PATCH_NUMBER}`, deprecated: false },
+        setPatchMarker: { patchId: `__sdk_internal_patch_number:${minimumPatchNumber}`, deprecated: false },
```
bergundy, line 631: Hmm.. this will cause us to emit a patch for every patch increment.
That's probably fine considering that we won't use this mechanism anymore.

I'd strongly consider scrapping all of this complexity and using a single hard-coded `__sdk_internal_patch_number:1`.

bergundy: Yeah, let's just do that and implement the next patch using the new dedicated mechanism.

packages/test/src/test-integration-workflows.ts
```diff
@@ -20,6 +22,9 @@ import { activityStartedSignal } from './workflows/definitions';
 import { signalSchedulingWorkflow } from './activities/helpers';
 import { ConnectionInjectorInterceptor } from './activities/interceptors';
 
+const { EVENT_TYPE_TIMER_STARTED, EVENT_TYPE_MARKER_RECORDED } = iface.temporal.api.enums.v1.EventType;
```
mjameswh, line 25: Can remove these `const`

packages/test/src/test-integration-workflows.ts
```diff
@@ -3,6 +3,8 @@ import { ExecutionContext, TestFn } from 'ava';
 import { firstValueFrom, Subject } from 'rxjs';
 import { WorkflowFailedError, WorkflowHandle, WorkflowStartOptions } from '@temporalio/client';
 import { TestWorkflowEnvironment, workflowInterceptorModules } from '@temporalio/testing';
+import * as iface from '@temporalio/proto';
```
mjameswh, line 6: Can remove those imports

bergundy (approved):

PR #1107, opened 2023-04-25
chore: Upgrade opentelemetry dependencies

bergundy:
Reopened after upgrading the rest of the dependencies and fixing the tests, couldn't push to the PR author's branch.
Thanks @maxgurewitz for the contribution.


packages/client/src/schedule-types.ts
```diff
@@ -764,13 +764,13 @@ export const MONTHS = [
 ] as const;
 
 /** @experimental */
-export type Month = typeof MONTHS[number];
+export type Month = (typeof MONTHS)[number];
```
mjameswh, line 767: I don't understand why we have these extra parenthesis added everywhere. have we bumped prettier's version along the way?

mjameswh (approved):

packages/client/src/schedule-types.ts
bergundy: yeah, I nuked package-lock and node_modules

PR #1109, opened 2023-04-26
fix: Fix `null` payload data in `historyFromJSON`

mjameswh:
## What changed and Why

- `fromProto3JSON` doesn't allow `null` values on 'bytes' fields. This turns out to be a problem for payloads. This fix this situation by recursively descending on events and fixing payloads that have null data field

### DIFF
--- a/packages/common/src/proto-utils.ts
+++ b/packages/common/src/proto-utils.ts
@@ -24,20 +24,23 @@ function fixEnumValue<O extends Record<string, any>>(obj: O, attr: keyof O, pref
 
 // fromProto3JSON doesn't allow null values on 'bytes' fields. This turns out to be a problem for payloads.
 // Recursively descend on objects and array, and fix in-place any payload that has a null data field
-function fixPayloads<T>(e: T): void {
-  if (Array.isArray(e)) {
-    e.forEach(fixPayloads);
-  } else if (e && typeof e === 'object') {
-    if ('metadata' in e && 'data' in e) {
-      if ((e as any).data === null) {
-        delete (e as any)['data'];
-      }
-    } else if (!(e instanceof String)) {
-      for (const v of Object.values(e)) {
-        fixPayloads(v);
-      }
-    }
+function fixPayloads<T>(e: T): T {
+  function isPayload(p: any): p is JSONPayload {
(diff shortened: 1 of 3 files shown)


packages/common/src/proto-utils.ts
```diff
@@ -22,12 +22,33 @@ function fixEnumValue<O extends Record<string, any>>(obj: O, attr: keyof O, pref
...
+}
+
 function fixHistoryEvent(e: Record<string, any>) {
   const type = Object.keys(e).find((k) => k.endsWith('EventAttributes'));
   if (!type) {
     throw new TypeError(`Missing attributes in history event: ${JSON.stringify(e)}`);
   }
 
+  // Fix payloads with null data
+  fixPayloads(e);
```
bergundy, line 50: I'd rather not mutate user input

packages/test/src/test-proto-utils.ts
```diff
@@ -22,3 +22,26 @@ test('cancel-fake-progress-replay', async (t) => {
...
+        "workflowExecutionCompletedEventAttributes": {
+          "result": { "payloads": [{ "metadata": { "encoding": "YmluYXJ5L251bGw=" }, "data": null }] },
+          "workflowTaskCompletedEventId": "15",
+          "newExecutionRunId": ""
+        }
+      }
+    ]
+  }`;
+
+  const histJSON = JSON.parse(sample);
```
bergundy, line 44: Why bother parsing?

packages/common/src/proto-utils.ts
```diff
@@ -22,12 +22,33 @@ function fixEnumValue<O extends Record<string, any>>(obj: O, attr: keyof O, pref
   );
 }
 
+// fromProto3JSON doesn't allow null values on 'bytes' fields. This turns out to be a problem for payloads.
+// Recursively descend on objects and array, and fix in-place any payload that has a null data field
+function fixPayloads<T>(e: T): void {
+  if (Array.isArray(e)) {
+    e.forEach(fixPayloads);
+  } else if (e && typeof e === 'object') {
+    if ('metadata' in e && 'data' in e) {
```
bergundy, line 31: Not the neatest approach but I'll take it

mjameswh:
Made some improvements.

bergundy (approved):

PR #1117, opened 2023-05-12
feat: Workflow logger improvements

bergundy:
- [Lazily initialize the default sinks logger](https://github.com/temporalio/sdk-typescript/commit/540896e03e9e0d1564c3d2198a1bc7c57d5b1795)
- [Export `log` from workflow](https://github.com/temporalio/sdk-typescript/commit/5a2789661bc4afa2c66056b4307d14f48146a0d4) - a logger that funnels messages to the default worker logger sink.

### Example:
```ts

export async function myWorkflow(): Promise<void> {
  workflow.log.info('hello from my workflow', { key: 'value' });
}
```

### DIFF
--- a/packages/worker/src/workflow/reusable-vm.ts
+++ b/packages/worker/src/workflow/reusable-vm.ts
@@ -3,6 +3,8 @@ import { AsyncLocalStorage } from 'node:async_hooks';
 import vm from 'node:vm';
 import * as internals from '@temporalio/workflow/lib/worker-interface';
 import { IllegalStateError } from '@temporalio/common';
+import { getTimeOfDay } from '@temporalio/core-bridge';
+import { timeOfDayToBigint } from '../logger';
 import { Workflow, WorkflowCreateOptions, WorkflowCreator } from './interface';
 import { WorkflowBundleWithSourceMapAndFilename } from './workflow-worker-thread/input';
 import { BaseVMWorkflow, globalHandlers, injectConsole, setUnhandledRejectionHandler } from './vm-shared';
@@ -160,7 +162,11 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
       }
     ) as any;
 
-    workflowModule.initRuntime({ ...options, sourceMap: this.workflowBundle.sourceMap });
+    workflowModule.initRuntime({
+      ...options,
+      sourceMap: this.workflowBundle.sourceMap,
+      getTimeOfDay: () => timeOfDayToBigint(getTimeOfDay()),
+    });
     const activator = bag.__TEMPORAL_ACTIVATOR__ as any;
 
     const newVM = new ReusableVMWorkflow(options.info, context, activator, workflowModule, isolateExecutionTimeoutMs);
--- a/packages/workflow/src/workflow.ts
+++ b/packages/workflow/src/workflow.ts
@@ -38,7 +38,7 @@ import {
   WorkflowInfo,
 } from './interfaces';
 import { LocalActivityDoBackoff, getActivator, maybeGetActivator } from './internals';
-import { Sinks } from './sinks';
+import { LoggerSinks, Sinks } from './sinks';
 import { untrackPromise } from './stack-helpers';
 import { ChildWorkflowHandle, ExternalWorkflowHandle } from './workflow-handle';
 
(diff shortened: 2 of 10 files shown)


packages/workflow/src/workflow.ts
```diff
@@ -1253,3 +1253,20 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+ *
+ * Note that since sinks are used to power this logger, any log attributes must be transferable via the
+ * {@link https://nodejs.org/api/worker_threads.html#worker_threads_port_postmessage_value_transferlist | postMessage}
+ * API.
+ *
+ * `defaultSinks` accepts a user logger and defaults to the `Runtime`'s logger.
+ *
+ * See the documentation for `WorkerOptions`, `defaultSinks`, and `Runtime` for more information.
+ */
+export const log = loggerSinks.defaultWorkerLogger;
```
mjameswh, line 1272: I'm thinking that maybe we should clone arguments (ie. `JSON.parse(JSON.stringify(args))`) on Workflow side (before they get captured by the sink plumbing). I got bitten once because an object that was being passed to a sink was being modified afterward (but in the same WFT), so the sink would receive the object in its modified state rather than the object in the state it was at the moment the sink was called in code...

mjameswh: For similar reason, wondering if we should maybe capture the _real_ time at the moment `log` was called...

mjameswh: For now, I suggest we simply mark this as experimental. Let's get this in as is, and we'll see later if either of my previous concern turns out to be justified.

packages/workflow/src/workflow.ts
```diff
@@ -1253,3 +1253,20 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
 
 export const stackTraceQuery = defineQuery<string>('__stack_trace');
 export const enhancedStackTraceQuery = defineQuery<EnhancedStackTrace>('__enhanced_stack_trace');
+
+const loggerSinks = proxySinks<LoggerSinks>();
+
+/**
+ * Default workflow logger.
+ * This logger is replay-aware and will omit log messages on workflow replay.
+ * The messages emitted by this logger are funnelled to the worker's `defaultSinks`, which are installed by default.
```
lorensr, line 1262:
```suggestion
 * The messages emitted by this logger are funnelled to the worker's {@link defaultSinks}, which are installed by default.
```

packages/workflow/src/workflow.ts
```diff
@@ -1253,3 +1253,20 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+ *
+ * Note that since sinks are used to power this logger, any log attributes must be transferable via the
+ * {@link https://nodejs.org/api/worker_threads.html#worker_threads_port_postmessage_value_transferlist | postMessage}
+ * API.
+ *
+ * `defaultSinks` accepts a user logger and defaults to the `Runtime`'s logger.
```
lorensr, line 1268:
```suggestion
 * `defaultSinks` accepts a user logger and defaults to the {@link RuntimeOptions.logger | `Runtime`'s logger}.
```

packages/workflow/src/workflow.ts
```diff
@@ -1253,3 +1253,20 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+ *
+ * See the documentation for `WorkerOptions`, `defaultSinks`, and `Runtime` for more information.
```
lorensr, line 1270:
```suggestion
```

prefer linking above, when the terms are used. didn't see a place for `{@link WorkerOptions.sinks}`

lorensr (approved):

packages/workflow/src/workflow.ts
bergundy: I'll try that and see if it gets linked properly, defaultSinks is in a different package.

packages/workflow/src/workflow.ts
bergundy: I wouldn't want to impose JSON serializability here, `postMessage` accepts a wider variety of types.

Sushisource: Why not just use this which is mentioned in the linked docs? https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

I think the point James makes seems pretty important

bergundy: I agree that the point is important, I’ll look into it

bergundy: All good points @mjameswh , thanks

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -3,6 +3,8 @@ import { AsyncLocalStorage } from 'node:async_hooks';
 import vm from 'node:vm';
 import * as internals from '@temporalio/workflow/lib/worker-interface';
 import { IllegalStateError } from '@temporalio/common';
+import { getTimeOfDay } from '@temporalio/core-bridge';
```
mjameswh, line 6: I know this is not new, but any reason why getting nano time through core was preferred over [process.hrtime](https://nodejs.org/api/process.html#processhrtimebigint)?

bergundy: We need to compare timestamps of logs emitted by Core and Lang when buffering Core logs so they can be emitted in order.

bergundy: From the hrtime docs:
> These times are relative to an arbitrary time in the past, and not related to the time of day and therefore not subject to clock drift. The primary use is for measuring performance between intervals:

We need time of day in nanosecond resolution.

bergundy:
I'm merging this and opening a new PR to call structuredClone on every sink call.

PR #1119, opened 2023-05-24
fix: Emit metrics from core client

bergundy:
Apparently the comment in the code that `init_worker` will attach metrics to the client was wrong.
Fixed.

Had to put an `Arc` around the runtime, I don't think that's avoidable.


bergundy:
Tested this manually.

Sushisource (approved):

PR #1120, opened 2023-05-24
fix(client): Fix incorrect schedule spec boundaries

### LINKED ISSUE #1105: [Bug] `CalendarSpec` components not encoded correctly

Creating a Temporal schedule using the calendar-based specifications of times.

### Describe the bug

When creating a Temporal schedule using the calendar-based specifications of times, there is an issue with the encoding of the input values.
According to the documentation the `dayOfMonth` should accept numbers from 1 to 31 but e.g. input value 21 is rejected with the following error message (https://typescript.temporal.io/api/interfaces/client.CalendarSpec#dayofmonth).

`Error: Invalid CalendarSpec component for field dayOfMonth: '21' of type 'number'`

### Minimal Reproduction

Use the code below in the following sample.
https://github.com/temporalio/samples-typescript/blob/main/schedules/src/start-schedule.ts

```typescript
// https://typescript.temporal.io/api/classes/client.ScheduleClient#create
const schedule = await client.schedule.create({
  action: {
    type: 'startWorkflow',
    workflowType: reminder,
    args: [''],
    taskQueue: 'schedules',
  },
  scheduleId: 'sample-schedule',
  policies: {
    catchupWindow: '1 day',
    overlap: ScheduleOverlapPolicy.ALLOW_ALL,
  },
  spec: {
    calendars: [
      {
        year: 2023,
        month: 'APRIL',
        dayOfMonth: 21, // 21st day of month
        hour: 12,
        minute: 0,
        second: 0,
      },
    ],
  },
});
```

### Environment/Versions

- OS and processor: Apple M2
- Temporal Version: `"@temporalio/client": "^1.7.2"`

### Additional context

Here is a diff that we use to fix the issue.

```diff
diff --git a/node_modules/@temporalio/client/src/schedule-helpers.ts b/node_modules/@temporalio/client/src/schedule-helpers.ts
index 14cddbe..6a5d05a 100644
--- a/node_modules/@temporalio/client/src/schedule-helpers.ts
+++ b/node_modules/@temporalio/client/src/schedule-helpers.ts
@@ -66,7 +66,7 @@ const [encodeMinute, decodeMinue] = makeCalendarSpecFieldCoders(
 
 const [encodeHour, decodeHour] = makeCalendarSpecFieldCoders(
   'hour',
-  (x: number) => (typeof x === 'number' && x >= 0 && x <= 59 ? x : undefined),
+  (x: number) => (typeof x === 'number' && x >= 0 && x <= 23 ? x : undefined),
   (x: number) => x,
   [{ start: 0, end: 0, step: 0 }], // default to 0
   [{ start: 0, end: 23, step: 1 }]
@@ -74,7 +74,7 @@ const [encodeHour, decodeHour] = makeCalendarSpecFieldCoders(
 
 const [encodeDayOfMonth, decodeDayOfMonth] = makeCalendarSpecFieldCoders(
   'dayOfMonth',
-  (x: number) => (typeof x === 'number' && x >= 0 && x <= 6 ? x : undefined),
+  (x: number) => (typeof x === 'number' && x >= 0 && x <= 31 ? x : undefined),
   (x: number) => x,
   [{ start: 1, end: 31, step: 1 }], // default to *
   [{ start: 1, end: 31, step: 1 }]

```

mjameswh:
## What changed

Fix incorrect schedule spec boundaries validation on hour and day-of-month fields

## Why

- Fixes #1105
- Thanks to @florianmorath for reporting

### DIFF
--- a/packages/client/src/schedule-client.ts
+++ b/packages/client/src/schedule-client.ts
@@ -497,6 +497,7 @@ export class ScheduleClient extends BaseClient {
   }
 
   protected rethrowGrpcError(err: unknown, scheduleId: string, fallbackMessage: string): never {
+    if (err instanceof TypeError) throw err;
     if (isServerErrorResponse(err)) {
       if (err.code === grpcStatus.NOT_FOUND) {
         throw new ScheduleNotFoundError(err.details ?? 'Schedule not found', scheduleId);
--- a/packages/client/src/schedule-client.ts
+++ b/packages/client/src/schedule-client.ts
@@ -260,6 +260,7 @@ export class ScheduleClient extends BaseClient {
       const res = await this.workflowService.createSchedule(req);
       return { conflictToken: res.conflictToken };
     } catch (err: any) {
+      if (err instanceof TypeError) throw err;
       if (err.code === grpcStatus.ALREADY_EXISTS) {
         throw new ScheduleAlreadyRunning('Schedule already exists and is running', opts.scheduleId);
       }
@@ -305,6 +306,7 @@ export class ScheduleClient extends BaseClient {
         requestId: uuid4(),
       });
     } catch (err: any) {
+      if (err instanceof TypeError) throw err;
       this.rethrowGrpcError(err, scheduleId, 'Failed to update schedule');
     }
   }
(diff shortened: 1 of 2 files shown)


packages/client/src/schedule-client.ts
```diff
@@ -497,6 +497,7 @@ export class ScheduleClient extends BaseClient {
   }
 
   protected rethrowGrpcError(err: unknown, scheduleId: string, fallbackMessage: string): never {
+    if (err instanceof TypeError) throw err;
```
bergundy, line 500: Hmm.. wondering if we can catch the type error at a different level.

bergundy (approved):

packages/client/src/schedule-client.ts
mjameswh: We could, but then we would need that extra check in every function that does encodings. Not a big thing, but that's why I put it there.

bergundy: Just find it a bit out of place here.
I’ll let you make the call.

Issue #441, opened 2022-01-03
[Bug] Retry policy intervals get rounded down to the second


bergundy:
Passing the following activity options:
```js
{
  retry: { initialInterval: "500ms" },
}
```

Produces the following result on the server:
```json
{
  "retryPolicy": {
    "nonRetryableErrorTypes": [],
    "initialInterval": {
      "duration": 0
    },
    "backoffCoefficient": 2,
    "maximumInterval": {
      "duration": 10
    },
    "maximumAttempts": 0
},
}
```

The SDK should probably round up the durations (first check what the other SDKs do here).

mjameswh:
@bergundy I think this has been [resolved](https://github.com/temporalio/temporal/pull/1681) on server. Can we close?

bergundy:
Not sure, can you please test?

bergundy:
Need to verify that this is resolved

mjameswh:
Validated in #1121.

PR #1121, opened 2023-05-24
chore(workflow): add test for activity retry policies

mjameswh:
## What changed

- Added an integration test to assert fix of #441

### DIFF
--- a/packages/test/src/test-integration-workflows.ts
+++ b/packages/test/src/test-integration-workflows.ts
@@ -14,7 +14,8 @@ import {
 } from '@temporalio/worker';
 import * as activity from '@temporalio/activity';
 import * as workflow from '@temporalio/workflow';
-import { CancelReason } from '@temporalio/worker/src/activity';
+import { CancelReason } from '@temporalio/worker/lib/activity';
+import { tsToMs } from '@temporalio/common/lib/time';
 import { test as anyTest, bundlerOptions, Worker } from './helpers';
 import { activityStartedSignal } from './workflows/definitions';
 import { signalSchedulingWorkflow } from './activities/helpers';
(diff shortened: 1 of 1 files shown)


packages/test/src/test-integration-workflows.ts
```diff
@@ -14,7 +14,8 @@ import {
 } from '@temporalio/worker';
 import * as activity from '@temporalio/activity';
 import * as workflow from '@temporalio/workflow';
-import { CancelReason } from '@temporalio/worker/src/activity';
+import { CancelReason } from '@temporalio/worker/lib/activity';
```
bergundy, line 17: We should export this from worker or activity if we don’t already

bergundy (approved):

packages/test/src/test-integration-workflows.ts
mjameswh: ATM, that would have to be exported from 'worker'...

But why? We don't use that type on anything public, except for `CancelledFailure.message`, but we can't easily change the type of the `message` property (it's in common and may also be used in the context of a Workflow, which implies different set of values).

bergundy: Still useful for users to compare and I'd like to have a property on activity context for the cancel reason.

mjameswh: Another thing... on deserializing a `CancelledFailure` object from an incoming failure (ie. failure-converter.ts), the incoming message may come from a different SDK, and could therefore contain any message (ie. not bound to `CancelReason`)

bergundy: We don't have any API that throws CancelledFailure directly.

bergundy: I don't think the message property should be typed, we use CancelledFailure all over the SDK.

PR #1126, opened 2023-05-26
fix: Better error messages, and more

### LINKED ISSUE #367: [Feature Request] Provide better errors when user passes a nameless function to infer workflow name
See related discussion here: https://temporalio.slack.com/archives/C01DKSMU94L/p1636493017132500

### LINKED ISSUE #506: [Feature Request] Get proto comments in API docs

`protobufjs`-generated classes like:

https://typescript.temporal.io/api/classes/proto.temporal.api.workflowservice.v1.RecordActivityTaskHeartbeatByIdRequest

don't explain what the properties are beyond their types. 

It would be nice if the proto comments showed up:

https://github.com/temporalio/api/blob/eff34dc3a970cdbf3a10a7be827edf1b1528b125/temporal/api/workflowservice/v1/request_response.proto#L376

### Additional context

User question looking at current docs: https://temporalio.slack.com/archives/C01DKSMU94L/p1646242515967829

### LINKED ISSUE #553: [Feature Request] Error when Workflow APIs are used outside of vm
We should provide better error messages to save users some headache. Currently they get like

```
WorkflowFailedError [WorkflowExecutionFailedError]: Workflow execution failed
  cause: [TemporalFailure: Activity execution failed] {
    cause: TypeError: exports.storage.getStore is not a function
```    
    
when using `condition` in an Activity

### LINKED ISSUE #1125: [Bug] tsc in a project created with `npx @temporalio/create` fails on Node 20

Trying to create a new project based on the `child-workflow` sample, following the directions at: https://docs.temporal.io/dev-guide/typescript/foundations#install-a-temporal-sdk

### Describe the bug

TypeScript compilation fails in an unmodified new project, based on the `child-workflow` sample, because the `es2023` library definitions are not included with the version of TypeScript used in the sample project:

```
❯ npm run start.watch
...
TSError: ⨯ Unable to compile TypeScript:
node_modules/@tsconfig/node20/tsconfig.json:6:13 - error TS6046: Argument for '--lib' option must be: 'es5', 'es6', 'es2015', 'es7', [...]

6     "lib": ["es2023"],
              ~~~~~~~~
...
[nodemon] app crashed - waiting for file changes before starting...
```

### Minimal Reproduction

1. `npx @temporalio/create@latest sample-tsc-fail-repro`
2. Choose the `child-workflows` sample and accept the defaults for the rest of the prompts
3. In a new terminal, `temporal server start-dev`
4. `cd sample-tsc-fail-repro`
5. `npm run start.watch`
6. Observe the crash above

### Environment/Versions

- OS and processor: M1 Mac, macOS 13.4
- Temporal Version: `temporal version 0.9.0 (server 1.20.2) (ui 2.15.0)`
- Are you using Docker or Kubernetes or building Temporal from source? No
- Node Version: `v20.2.0`

### Additional context

I doubt this is specific to the `child-workflows` sample; I suspect the others suffer from the same issue (though I haven't tried).

I notice the `es2023` lib reference comes from the `tsconfig.json` included in `@tsconfig/node20`. If I look at the sample itself, I see that its `package.json` uses `@tsconfig/node16` (which presumably does not have the `es2023` lib reference). So it seems `@temporalio/create` is tweaking the `package.json` to use `node20` (which is the version I have installed). And indeed I can see this in the code: https://github.com/temporalio/sdk-typescript/blob/1709d45c5e649fd6b233243b7a80c49be5e706bd/packages/create-project/src/helpers/install.ts#L33

Also, the sample's `package.json` pins TypeScript to a 4.x release, and I don't think `es2023` is defined until 5.0.

I think there might be two ways of fixing this:
1. Bump the TypeScript version (and related dependencies) to TypeScript 5.x in all the samples
2. Stop messing with the `@tsconfig/nodeXX` version when creating the new package

It might be a good idea to do both, in any case. Staying up to date (i.e. (1)) is always good, and my thinking re. (2) is that people may be developing on the latest version of Node (currently, v20), but may want to use an LTS release (currently, v18) when deploying. So it may not necessarily make sense to depend on the current Node version in a new project.

mjameswh:
## What changed

- Better error message when `executeWorkflow()` and similar are given a workflowFunction fucntion that is anonymous (fix #367)
- Better error message when Workflow-only APIs are called from non-Workflow execution context (fix #553)
- On `create`, do not use `@tsconfig/node20` because it is broken (fix #1125)
- Include proto comments in API files generated by protobufjs (fix #506)
- Minor refactor of schedule's error handling

### DIFF
--- a/packages/common/src/workflow-options.ts
+++ b/packages/common/src/workflow-options.ts
@@ -159,7 +159,7 @@ export function extractWorkflowType<T extends Workflow>(workflowTypeOrFunc: stri
   if (typeof workflowTypeOrFunc === 'string') return workflowTypeOrFunc as string;
   if (typeof workflowTypeOrFunc === 'function') {
     if (workflowTypeOrFunc?.name) return workflowTypeOrFunc.name;
-    throw new TypeError(`Invalid workflow type: the workflow function is anonymous`);
+    throw new TypeError('Invalid workflow type: the workflow function is anonymous');
   }
   throw new TypeError(
     `Invalid workflow type: expected either a string or a function, got '${typeof workflowTypeOrFunc}'`
--- a/packages/common/src/workflow-options.ts
+++ b/packages/common/src/workflow-options.ts
@@ -154,3 +154,14 @@ export function compileWorkflowOptions<T extends CommonWorkflowOptions>(options:
     workflowTaskTimeout: msOptionalToTs(workflowTaskTimeout),
   };
 }
+
+export function extractWorkflowType<T extends Workflow>(workflowTypeOrFunc: string | T): string {
+  if (typeof workflowTypeOrFunc === 'string') return workflowTypeOrFunc as string;
+  if (typeof workflowTypeOrFunc === 'function') {
+    if (workflowTypeOrFunc?.name) return workflowTypeOrFunc.name;
+    throw new TypeError(`Invalid workflow type: the workflow function is anonymous`);
+  }
+  throw new TypeError(
+    `Invalid workflow type: expected either a string or a function, got '${typeof workflowTypeOrFunc}'`
+  );
+}
--- a/packages/create-project/src/helpers/install.ts
+++ b/packages/create-project/src/helpers/install.ts
@@ -35,9 +35,8 @@ export async function updateNodeVersion({ root }: InstallArgs): Promise<void> {
   const versionAlreadyInPackageJson = 16;
   const minimumValidVersion = 14;
 
-  // For some reason, the @tsconfig/node20 package exists, but is currently unusable because it
-  // refers to a --lib value that isn't supported in latest release of TypeScript.
-  // FIXME: Update this once @tsconfig/node20 is fixed.
+  // The @tsconfig/node20 sets "--lib es2023", which require TypeScript 5.x.
+  // FIXME: Remove this once samples have been updated to TypeScript ^5.0.0.
   const maximumValidVersion = 18;
 
   const tsconfigNodeVersion = Math.max(minimumValidVersion, Math.min(currentNodeVersion, maximumValidVersion));
--- a/packages/create-project/src/helpers/install.ts
+++ b/packages/create-project/src/helpers/install.ts
@@ -35,8 +35,15 @@ export async function updateNodeVersion({ root }: InstallArgs): Promise<void> {
   const versionAlreadyInPackageJson = 16;
   const minimumValidVersion = 14;
 
-  if (currentNodeVersion >= minimumValidVersion && currentNodeVersion !== versionAlreadyInPackageJson) {
-    const packageName = `@tsconfig/node${currentNodeVersion}`;
+  // For some reason, the @tsconfig/node20 package exists, but is currently unusable because it
+  // refers to a --lib value that isn't supported in latest release of TypeScript.
+  // FIXME: Update this once @tsconfig/node20 is fixed.
+  const maximumValidVersion = 18;
+
+  const tsconfigNodeVersion = Math.max(minimumValidVersion, Math.min(currentNodeVersion, maximumValidVersion));
+
+  if (tsconfigNodeVersion !== versionAlreadyInPackageJson) {
+    const packageName = `@tsconfig/node${tsconfigNodeVersion}`;
 
     const packageExists = await isUrlOk(`https://registry.npmjs.org/${packageName}`);
     if (packageExists) {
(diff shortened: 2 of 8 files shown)


packages/proto/scripts/compile-proto.js
```diff
@@ -59,6 +60,16 @@ async function compileProtos(dtsOutputFile, ...args) {
...
   try {
     await promisify(pbjs.main)([...pbjsArgs, '--target', 'static-module', '--out', tempFile]);
+
+    // pbts internally calls jsdoc, which do strict validation of jsdoc tags.
+    // Unfortunately, some protobuf comment about cron syntax contains the
+    // "@every" shorthand at the begining of a line, making it appear as a
+    // (invalid) jsdoc tag. We fix this by rewriting all cron shorthand
+    // using markdown "code" syntax.
+    let tempFileContent = await readFile(tempFile, 'utf8');
+    tempFileContent = tempFileContent.replace(/(@(?:yearly|monthly|weekly|daily|hourly|every))/g, '`$1`');
```
Sushisource, line 70: lovely, lol

Sushisource (approved):

packages/common/src/workflow-options.ts
```diff
@@ -154,3 +154,14 @@ export function compileWorkflowOptions<T extends CommonWorkflowOptions>(options:
     workflowTaskTimeout: msOptionalToTs(workflowTaskTimeout),
   };
 }
+
+export function extractWorkflowType<T extends Workflow>(workflowTypeOrFunc: string | T): string {
+  if (typeof workflowTypeOrFunc === 'string') return workflowTypeOrFunc as string;
+  if (typeof workflowTypeOrFunc === 'function') {
+    if (workflowTypeOrFunc?.name) return workflowTypeOrFunc.name;
+    throw new TypeError(`Invalid workflow type: the workflow function is anonymous`);
```
bergundy, line 162:
```suggestion
    throw new TypeError('Invalid workflow type: the workflow function is anonymous');
```

packages/create-project/src/helpers/install.ts
```diff
@@ -35,8 +35,15 @@ export async function updateNodeVersion({ root }: InstallArgs): Promise<void> {
   const versionAlreadyInPackageJson = 16;
   const minimumValidVersion = 14;
 
-  if (currentNodeVersion >= minimumValidVersion && currentNodeVersion !== versionAlreadyInPackageJson) {
-    const packageName = `@tsconfig/node${currentNodeVersion}`;
+  // For some reason, the @tsconfig/node20 package exists, but is currently unusable because it
+  // refers to a --lib value that isn't supported in latest release of TypeScript.
+  // FIXME: Update this once @tsconfig/node20 is fixed.
```
bergundy, line 40: I'd go as far as linking the issue where there's a bit more information for future reference.

packages/test/scripts/compile-proto.js
```diff
@@ -28,7 +28,17 @@ function mtime(path) {
...
 
 async function compileProtos(outputFile, ...args) {
-  const pbjsArgs = [...args, '--wrap', 'commonjs', '--force-long', '--no-verify', '--out', outputFile, ...protoFiles];
+  const pbjsArgs = [
+    ...args,
+    '--wrap',
+    'commonjs',
+    '--force-long',
+    '--no-verify',
+    '--alt-comment',
```
bergundy, line 37: Didn't know about this one, thanks

packages/workflow/src/workflow.ts
```diff
@@ -1290,3 +1316,9 @@ export const log: LoggerSinks['defaultWorkerLogger'] = Object.fromEntries(
     ];
   })
 ) as any;
+
+function assertInWorkflowContext(message: string): void {
```
bergundy, line 1320: nit: this assertion is always called before `getActivator`.
You can return the activator from this function.

packages/workflow/src/workflow.ts
```diff
@@ -250,6 +252,7 @@ async function scheduleLocalActivityNextHandler({
  * @hidden
  */
 export function scheduleActivity<R>(activityType: string, args: any[], options: ActivityOptions): Promise<R> {
+  assertInWorkflowContext('Workflow.scheduleActivity(...) may only be used from a Workflow Execution');
```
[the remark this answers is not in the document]
mjameswh: Which function? `scheduleActivity`? I personally think it's a good thing that they are. There are situations where `proxyActivities()` is not a good fit, eg. when activity name is dynamic, or when users want to add their own wrappers around some activity calls... It's still possible to use `proxies` in those cases (eg. `proxy[activityName](...)`), but calling `scheduleActivity` directly makes things much more understandable.

packages/create-project/src/helpers/install.ts
mjameswh: This is not actually a bug from their perspective. `@tsconfig/node20` require TypeScript 5.x.

Updated comment accordingly.

bergundy: Let’s upgrade to ts 5 in samples

PR #1128, opened 2023-05-30
fix: Remove most instanceof checks from SDK

bergundy:
The number one reported issue stems from instanceof checks and SDK module level variables, which fail when there are multiple versions of the SDK packages installed.

- Remove the instanceof checks in favor of coded `is` methods on the critical classes.
- Install the activity context local storage globally.

Note that the `is` methods do no use the class name, which may be lost when using minifiers.

### DIFF
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -35,7 +35,40 @@ checkExtends<RetryState, temporal.api.enums.v1.RetryState>();
 
 export type WorkflowExecution = temporal.api.common.v1.IWorkflowExecution;
 
-const isTemporalFailure = Symbol.for('__temporal_is_temporal_failure__');
+const isLineageTrackingError = Symbol.for('__temporal_is_lineage_tracking_error__');
+
+/**
+ * Base error class used to track the class lineage of any of its subclasses.
+ *
+ * This error is never directly instantiated but is used throughout the SDK code to support an alternative form of
+ * instanceof checks that works when multiple versions of the temporalio package are present.
+ *
+ * See {@link LineageTrackingError.is} and the `is` method on each subclass.
+ */
+export class LineageTrackingError extends Error {
+  public readonly name: string = 'LineageTrackingError';
+  /**
+   * Marker to determine whether an error is an instance of LineageTrackingError.
+   */
+  protected readonly [isLineageTrackingError] = true;
+  /**
+   * Array that stores subclass names.
+   * @hidden
+   */
+  public readonly lineage = Array<string>();
+
+  constructor(message?: string) {
+    super(message ?? undefined);
+  }
+
+  /**
+   * Instanceof check that is works when multiple versions of @temporalio/common are installed.
+   */
+  static is(error: unknown): error is LineageTrackingError {
+    // Cast to any since isLineageTrackingError is intentionally left protected.
+    return error instanceof LineageTrackingError || (error instanceof Error && (error as any)[isLineageTrackingError]);
+  }
+}
 
 /**
  * Represents failures that can cross Workflow and Activity boundaries.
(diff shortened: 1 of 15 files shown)


bergundy:
> I have a few worries...
> 
> 1. The `(e instanceof Error)` that we have in most places may itself not work correctly in cases where an error is passed from one execution context to another one.

We don't pass these across execution contexts and if that were an issue it would have been so for `instanceof` checks too.
In any case, using `Symbol.for` strict equality works across contexts, I tested by running this in a workflow:

```ts
const sym = Symbol.for('foo');
const externalSym = globalThis.constructor.constructor('return Symbol.for("foo")')();
console.log(sym === externalSym);
```

> 2. `error.name` checks are fragile not only in case of minimization (which is arguably unlikely very unlikely in this case for a few reasons) but also in case where classes are extended, or simply by setting `error.name` to something else.

I typically wouldn't expect users to extend these error classes and set custom names but I agree we need a sound solution.
Added a `LineageTrackingError` base class that checks the entire lineage.
It's adds some annoying boilerplate to defining the errors but it works and covers all of the cases.

> 3. The symbol approach seems a little bit stronger and could easily be used everywhere, though I'm not sure if symbols survive travelling across execution contexts...

See above.

mjameswh:
@bergundy Will review more in depth tomorrow morning, but at first look, I think that should work just fine.

packages/common/src/failure.ts
```diff
@@ -35,14 +35,49 @@ checkExtends<RetryState, temporal.api.enums.v1.RetryState>();
...
+  constructor(message?: string) {
+    super(message ?? undefined);
+  }
+
+  /**
+   * Instanceof check that is works when multiple versions of @temporalio/common are installed.
+   */
+  static is(error: unknown): error is LineageTrackingError {
+    // Cast to any since isLineageTrackingError is intentionally left protected.
+    return error instanceof LineageTrackingError || (error instanceof Error && (error as any)[isLineageTrackingError]);
```
mjameswh, line 69: I think I would skip the `error instanceof Error && ` check. At that point, the presence of the symbol is enough.

packages/common/src/failure.ts
```diff
@@ -53,6 +88,16 @@ export class TemporalFailure extends Error {
 
   constructor(message?: string | undefined | null, public readonly cause?: Error) {
     super(message ?? undefined);
+    this.lineage.unshift('TemporalFailure');
+  }
+
+  /**
+   * Instanceof check that is works when multiple versions of @temporalio/common are installed.
+   */
+  static is(error: unknown): error is TemporalFailure {
```
mjameswh, line 97: Shouldn't we make these functions public and recommend that users also use them instead of `instanceof`? Otherwise they may be receiving errors from activity, workflow or worker and trying to handle these against a different version of common... For the same reason, I think we'll need to add `.is` checks on all error types, not only those that _we_ need to check.

packages/workflow/src/internals.ts
```diff
@@ -619,10 +619,10 @@ export class Activator implements ActivationHandler {
   async handleWorkflowFailure(error: unknown): Promise<void> {
     if (this.cancelled && isCancellation(error)) {
       this.pushCommand({ cancelWorkflowExecution: {} }, true);
-    } else if (error instanceof ContinueAsNew) {
+    } else if (ContinueAsNew.is(error)) {
       this.pushCommand({ continueAsNewWorkflowExecution: error.command }, true);
     } else {
-      if (!(error instanceof TemporalFailure)) {
+      if (!TemporalFailure.is(error)) {
```
mjameswh, line 625: @bergundy I added these two. Were they left apart intentionally?

mjameswh:
Would it not have been easier to define a different Symbol for every Error class? Then `.is` would simply check for the presence of that property. That would get rid of `LineageTrackingError` and avoid the `Array.includes[]` check.

packages/workflow/src/internals.ts
bergundy: I must have missed this.

bergundy:
> Would it not have been easier to define a different Symbol for every Error class? Then `.is` would simply check for the presence of that property. That would get rid of `LineageTrackingError` and avoid the `Array.includes[]` check.

Sure that works for me. I like the idea. Let me change.

packages/common/src/failure.ts
bergundy: They are public.
I agree we should recommend using them instead of `instanceof`.

bergundy:
> > Would it not have been easier to define a different Symbol for every Error class? Then `.is` would simply check for the presence of that property. That would get rid of `LineageTrackingError` and avoid the `Array.includes[]` check.
> 
> Sure that works for me. I like the idea. Let me change.

PTAL.

mjameswh (approved):
LGTM

PR #1129, opened 2023-05-31
Create CODEOWNERS

jackdawm:
## What was changed
Add CODEOWNERS so that Security can own the Semgrep rules files and paths.

## Why?
We are adding Semgrep for static analysis to this repository, and only the security team should be able to approve exclusions from the policy.

## Checklist

How was this tested:
We ran this scanner on internal repos with this CODEOWNERS file and it worked as expected.


Sushisource (approved):

mjameswh:
@jackdawm That implies the security team also need _at least_ write access on the repo. Do you need _more_ than write access?

PR #1132, opened 2023-06-05
feat(worker): Add support for concurrent pollers

### LINKED ISSUE #608: [Feature Request] Better balancing of workflows between worker threads
Currently the SDK worker uses round robin to dispatch workflows to the different worker threads in the pool.
Due to the fact that workflows can be evicted, worker threads might be in a fairly unbalanced state where a single thread takes on significantly more work than the others.

A simple refactoring of this dispatcher to choose the worker thread with least workflows would improve the overall balancing of work between the threads.

### LINKED ISSUE #838: [Bug] Default `WorkerOptions.maxCachedWorkflows` is too low
Since #802 the default was significantly reduced.
For example on a 64GB machine it went down from 12600 to 761 and on a 16GB machine (nightly CI) from 2853 to 261.

We could in the future calculate this dynamically based on real usage instead of having a hard maximum but for now we should at least make the default a bit higher (details yet TBD).

mjameswh:
## What changed

- Added `WorkerOptions` properties `maxConcurrentWorkflowTaskPolls` and `maxConcurrentActivityTaskPolls`, which allow controlling the number of pollers to be used to fetch Workflow/Activity tasks from the  Task Queue. Properly adjusting these values should allow better filling of the corresponding execution slots.
- Remove the `@experimental` tag on `reuseV8Context`
- Better auto-configuration of `maxCachedWorkflows`, including using a different formula when `reuseV8Context` is enabled (fix #838)
- Default value for `maxConcurrentWorkflowTaskExecutions` has been reduced to 40 (was previously 100), as higher values increase the risk of Workflow Task Timeouts unless other options are also tuned. This was not problem previously because the single poller was unlikely to fill in all execution slot anyway, so max would rarely reached).
- Load balance create workflows requests amongst WF threads (fix #608)
- Better documentation of some performance/scaling-related properties in `WorkerOptions`.
- Add more options to internal performance test runners, including specifying mTLS options (for testing against Temporal Cloud, pollers and WF thread pool size)
- Updated Core to 782282e7

### DIFF
--- a/packages/test/src/workflows/log-sink-tester.ts
+++ b/packages/test/src/workflows/log-sink-tester.ts
@@ -16,7 +16,7 @@ export async function logSinkTester(): Promise<void> {
       wf.workflowInfo().historyLength
     }`
   );
-  // We rely on this test to run with workflow cache disabled. This sleep() will
+  // We rely on this test to run with workflow cache disabled. This sleep()
   // therefore ends the current WFT, evicting the workflow from cache, and thus
   // causing replay of the first sink call.
   await wf.sleep(1);
--- a/packages/test/src/workflows/log-sink-tester.ts
+++ b/packages/test/src/workflows/log-sink-tester.ts
@@ -7,7 +7,6 @@
 
 import * as wf from '@temporalio/workflow';
 import { LoggerSinks } from './definitions';
-import { successString } from './success-string';
 
 const { logger } = wf.proxySinks<LoggerSinks>();
 
@@ -17,10 +16,10 @@ export async function logSinkTester(): Promise<void> {
       wf.workflowInfo().historyLength
     }`
   );
-  // We rely on the test to run with max cached workflows of 1.
-  // Executing this child will flush the current workflow from the cache
-  // causing replay or the first sink call.
-  await wf.executeChild(successString);
+  // We rely on this test to run with workflow cache disabled. This sleep() will
+  // therefore ends the current WFT, evicting the workflow from cache, and thus
+  // causing replay of the first sink call.
+  await wf.sleep(1);
   logger.info(
     `Workflow execution completed, replaying: ${wf.workflowInfo().unsafe.isReplaying}, hl: ${
       wf.workflowInfo().historyLength
--- a/packages/testing/src/utils.ts
+++ b/packages/testing/src/utils.ts
@@ -18,6 +18,8 @@ export async function waitOnNamespace(
       if (
         err.details.includes('workflow history not found') ||
         err.details.includes('Workflow executionsRow not found') ||
+        err.details.includes('operation GetCurrentExecution') ||
+        err.details.includes('operation GetWorkflowExecution encountered not found') ||
         err.details.includes(runId)
       ) {
         break;
--- a/packages/testing/src/utils.ts
+++ b/packages/testing/src/utils.ts
@@ -1,6 +1,8 @@
 import { Connection } from '@temporalio/client';
 import { msToTs } from '@temporalio/common/lib/time';
 
+// Starting with 1.20, we should no longer need to wait on namespace
+// TODO: Remove all of this once we are confident that this is no longer required
 export async function waitOnNamespace(
   connection: Connection,
   namespace: string,
--- a/packages/worker/src/runtime.ts
+++ b/packages/worker/src/runtime.ts
@@ -541,31 +541,51 @@ export class Runtime {
 
   protected checkHeapSizeLimit(): void {
     if (process.platform === 'linux') {
-      const cgroupMemoryConstraint = Number(
-        this.tryReadFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.max') ??
-          this.tryReadFileSync(/* cgroup v1 */ '/sys/fs/cgroup/memory/memory.limit_in_bytes')
-      );
+      // References:
+      // - https://facebookmicrosites.github.io/cgroup2/docs/memory-controller.html
+      // - https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt
+      const cgroupMemoryConstraint =
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.high') ??
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.max') ??
+        this.tryReadNumberFileSync(/* cgroup v1 */ '/sys/fs/cgroup/memory/memory.limit_in_bytes');
+      const cgroupMemoryReservation =
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.low') ??
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.min') ??
+        this.tryReadNumberFileSync(/* cgroup v1 */ '/sys/fs/cgroup/memory/soft_limit_in_bytes');
 
-      if (cgroupMemoryConstraint < os.totalmem() && cgroupMemoryConstraint < v8.getHeapStatistics().heap_size_limit) {
-        const totalMemInMb = toMB(os.totalmem(), 0);
-        const suggestedOldSpaceSizeInMb = toMB(os.totalmem() * 0.75, 0);
+      const applicableMemoryConstraint = cgroupMemoryReservation ?? cgroupMemoryConstraint;
+      if (
+        applicableMemoryConstraint &&
+        applicableMemoryConstraint < os.totalmem() &&
+        applicableMemoryConstraint < v8.getHeapStatistics().heap_size_limit
+      ) {
+        let dockerArgs = '';
+        if (cgroupMemoryConstraint) {
+          dockerArgs += `--memory=${toMB(cgroupMemoryConstraint, 0)}m `;
+        }
+        if (cgroupMemoryReservation) {
+          dockerArgs += `--memory-reservation=${toMB(cgroupMemoryReservation, 0)}m `;
+        }
+
+        const suggestedOldSpaceSizeInMb = toMB(applicableMemoryConstraint * 0.75, 0);
 
         this.logger.warn(
           `This program is running inside a containerized environment with a memory constraint ` +
-            `(eg. docker --memory ${totalMemInMb}m). Node itself does not consider this memory constraint ` +
+            `(eg. '${dockerArgs}' or similar). Node itself does not consider this memory constraint ` +
             `in how it manages its heap memory. There is consequently a high probability that ` +
             `the process will crash due to running out of memory. To increase reliability, we recommend ` +
             `adding '--max-old-space-size=${suggestedOldSpaceSizeInMb}' to your node arguments. ` +
-            `Refer to https://docs.temporal.io/application-development/worker-performance for more ` +
-            `advice on tuning your workers.`
+            `Refer to https://docs.temporal.io/dev-guide/typescript/foundations#run-a-worker-on-docker ` +
+            `for more advices on tuning your Workers.`
         );
       }
     }
   }
 
-  protected tryReadFileSync(file: string): string | undefined {
+  protected tryReadNumberFileSync(file: string): number | undefined {
     try {
-      return fs.readFileSync(file, { encoding: 'ascii' }) as string;
+      const val = Number(fs.readFileSync(file, { encoding: 'ascii' }));
+      return isNaN(val) ? undefined : val;
     } catch (e) {
       return undefined;
     }
--- a/packages/worker/src/runtime.ts
+++ b/packages/worker/src/runtime.ts
@@ -571,7 +571,7 @@ export class Runtime {
             `the process will crash due to running out of memory. To increase reliability, we recommend ` +
             `adding '--max-old-space-size=${suggestedOldSpaceSizeInMb}' to your node arguments. ` +
             `Refer to https://docs.temporal.io/dev-guide/typescript/foundations#run-a-worker-on-docker ` +
-            `for more advices on tuning your Workers.`
+            `for more advice on tuning your Workers.`
         );
       }
     }
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -206,7 +206,7 @@ export interface WorkerOptions {
    * `maxConcurrentWorkflowTaskExecutions` slots despite a backlog of Workflow
    * Tasks in the Task Queue (ie. due to network latency). Can't be higher than
    * `maxConcurrentWorkflowTaskExecutions`.
-   * @default min(5, maxConcurrentWorkflowTaskExecutions)
+   * @default min(2, maxConcurrentWorkflowTaskExecutions)
    */
   maxConcurrentWorkflowTaskPolls?: number;
 
@@ -216,7 +216,7 @@ export interface WorkerOptions {
    * `maxConcurrentActivityTaskExecutions` slots despite a backlog of Activity
    * Tasks in the Task Queue (ie. due to network latency). Can't be higher than
    * `maxConcurrentActivityTaskExecutions`.
-   * @default min(5, maxConcurrentActivityTaskExecutions)
+   * @default min(2, maxConcurrentActivityTaskExecutions)
    */
   maxConcurrentActivityTaskPolls?: number;
 
@@ -243,6 +243,17 @@ export interface WorkerOptions {
    * @default `max(maxHeapMemory / 1GiB - 1, 1) * 250`
    */
   maxCachedWorkflows?: number;
+
+  /**
+   * Controls the number of Worker threads the Worker should create.
+   *
+   * Threads are used to create {@link https://nodejs.org/api/vm.html | vm }s for the isolated Workflow environment.
+   *
+   * New Workflows are created on this pool in a round-robin fashion.
+   *
+   * @default 1 for reuseV8Context, otherwise 8. Ignored if `debugMode` is enabled.
+   */
+  workflowThreadPoolSize?: number;
 
   /**
    * Longest interval for throttling activity heartbeats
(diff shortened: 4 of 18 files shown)


Sushisource (approved):

packages/worker/src/worker-options.ts
```diff
@@ -195,10 +195,31 @@ export interface WorkerOptions {
   /**
    * Maximum number of Workflow tasks to execute concurrently.
    * Adjust this to improve Worker resource consumption.
+   * Can't be lower than 2 if `maxCachedWorkflows` is non-zero.
    * @default 100
```
bergundy, line 199: We should lower this to avoid a large backlog and possibly timeouts

packages/worker/src/worker-options.ts
```diff
@@ -487,14 +510,17 @@ export function appendDefaultInterceptors(
...
+
   return {
     namespace: namespace ?? 'default',
     identity: `${process.pid}@${os.hostname()}`,
     shutdownGraceTime: 0,
-    maxConcurrentActivityTaskExecutions: 100,
     maxConcurrentLocalActivityExecutions: 100,
     enableNonLocalActivities: true,
-    maxConcurrentWorkflowTaskExecutions: 100,
+    maxConcurrentWorkflowTaskPolls: Math.min(5, maxConcurrentWorkflowTaskExecutions),
```
bergundy, line 522: Users should be able to configure these

mjameswh: If you mean that `maxConcurrentWorkflowTaskPolls` should be user-configurable, they already can. This line sets the default value. `...rest` below will overwrite that default value with the user's value if present. After that, `compileWorkerOptions` will warn if the value is invalid (which should only be possible if value was provided by the user).

packages/worker/src/worker-options.ts
mjameswh: Now, regarding `maxConcurrentActivityTaskExecutions` and `maxConcurrentLocalActivityTaskExecutions`, I would lower both default values to something in the range of 20-40, but we should really encourage people to self configure these based on their own workload.

mjameswh:
@bergundy Some major changes from the previous pass. Please take another look..

@Sushisource Would you mind taking a second look at changes in the worker-options.ts file? Any runtime constraint I forgot to check?

Also, in tests against Temporal Cloud, I had to set both WF and Act pollers to ~60 each to get peak performances for a single thread Worker running on my laptop. That seems way too high to use as a default value, but performance with only 2 pollers (ie. the minimum) were very disappointing. I settled on 10 pollers WF/Act as the default value. Comments?

packages/worker/src/worker-options.ts
```diff
@@ -53,6 +53,10 @@ export function isPathBundleOption(bundleOpt: WorkflowBundleOption): bundleOpt i
 
 /**
  * Options to configure the {@link Worker}
+ *
+ * Some options can signifciantly affect Workers performance. Default settings are generally appropriate for
+ * day-to-day developments, but unlikely to be suitable for production use. We recommend that you explicitely set
+ * values every performance-related options on production deployment.
```
Sushisource, line 59:
```suggestion
 * Some options can significantly affect Worker's performance. Default settings are generally appropriate for
 * day-to-day development, but unlikely to be suitable for production use. We recommend that you explicitly set
 * values for every performance-related option on production deployment.
```

packages/worker/src/worker-options.ts
```diff
@@ -64,6 +68,12 @@ export interface WorkerOptions {
 
   /**
    * A human-readable string that can identify your worker
+   *
+   * Note that in most production environments, the `identity` value set by default may be unhelful for traceability
+   * purpose. It is highly recommended that you set this value to something that will allow you to efficiently identify
+   * that particular Worker container/process/logs in your infrastructure (eg. the task ID allocated to this container
+   * by your orchestrator).
```
Sushisource, line 75:
```suggestion
   * Note that in most production environments, the `identity` value set by default may be unhelpful for traceability
   * purposes. It is highly recommended that you set this value to something that will allow you to efficiently identify
   * that particular Worker container/process/logs in your infrastructure (ex: the task ID allocated to this container
   * by your orchestrator).
```

packages/worker/src/worker-options.ts
```diff
@@ -193,12 +203,72 @@ export interface WorkerOptions {
...
+   * `maxConcurrentWorkflowTaskExecutions`. Your millage may vary.
+   *
+   * Setting this value higher than needed will generally not have negative impact on this Worker's performance; your
+   * server may however impose a limit on the total number of concurrent Workflow Task pollers.
+   *
+   * As a general guidelines:
+   * - By default, set this value to half of `maxConcurrentWorkflowTaskExecutions`.
+   * - **Increase** if actual number of Workflow Tasks being processed concurrently is lower than
+   *   `maxConcurrentWorkflowTaskExecutions` despite a backlog of Workflow Tasks in the Task Queue.
+   * - Keep low on Task Queues that have very few concurrent Workflow Executions.
```
Sushisource, line 254:
```suggestion
   * In general, a Workflow Worker's performance is mostly network bound (due to communication latency with the
   * Temporal server). Polling multiple Workflow Tasks concurrently helps compensate for this latency, by ensuring that
   * the Worker is not starved waiting for the server to return new Workflow Tasks to execute.
   *
   * This setting is highly related with {@link WorkerOptions.maxConcurrentWorkflowTaskExecutions}. In various
   * performance tests, we generally got optimal performance by setting this value to about half of
   * `maxConcurrentWorkflowTaskExecutions`. Your millage may vary.
   *
   * Setting this value higher than needed will generally not have negative impact on this Worker's performance; your
   * server may however impose a limit on the total number of concurrent Workflow Task pollers.
   *
   * As a general guidelines:
   * - By default, set this value to half of `maxConcurrentWorkflowTaskExecutions`.
   * - **Increase** if actual number of Workflow Tasks being processed concurrently is lower than
   *   `maxConcurrentWorkflowTaskExecutions` despite a backlog of Workflow Tasks in the Task Queue.
   * - Keep this value low for Task Queues which have very few concurrent Workflow Executions.
```

Sushisource (approved):
I think the option defaults make sense :+1: 

Just some wording fixes in the docstrings.

mjameswh:
> I think the option defaults make sense 👍
> 
> Just some wording fixes in the docstrings.

Thanks a lot! I need to find some good English spell/grammar checker that knows how to deal with source code, but hopefully, in the mean time, there is peer reviews 😆

packages/test/src/workflows/log-sink-tester.ts
```diff
@@ -17,10 +16,10 @@ export async function logSinkTester(): Promise<void> {
       wf.workflowInfo().historyLength
     }`
   );
-  // We rely on the test to run with max cached workflows of 1.
-  // Executing this child will flush the current workflow from the cache
-  // causing replay or the first sink call.
-  await wf.executeChild(successString);
+  // We rely on this test to run with workflow cache disabled. This sleep() will
+  // therefore ends the current WFT, evicting the workflow from cache, and thus
```
bergundy, line 20:
```suggestion
  // We rely on this test to run with workflow cache disabled. This sleep()
  // therefore ends the current WFT, evicting the workflow from cache, and thus
```

packages/testing/src/utils.ts
```diff
@@ -18,6 +18,8 @@ export async function waitOnNamespace(
       if (
         err.details.includes('workflow history not found') ||
         err.details.includes('Workflow executionsRow not found') ||
+        err.details.includes('operation GetCurrentExecution') ||
+        err.details.includes('operation GetWorkflowExecution encountered not found') ||
```
bergundy, line 22: There's no need to wait on namespaces anymore since 1.20 but it's better to improve the existing helper than to assume we're working with a new enough server for the time being.

packages/worker/src/runtime.ts
```diff
@@ -536,31 +536,51 @@ export class Runtime {
...
+      const cgroupMemoryConstraint =
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.high') ??
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.max') ??
+        this.tryReadNumberFileSync(/* cgroup v1 */ '/sys/fs/cgroup/memory/memory.limit_in_bytes');
+      const cgroupMemoryReservation =
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.low') ??
+        this.tryReadNumberFileSync(/* cgroup v2 */ '/sys/fs/cgroup/memory.min') ??
+        this.tryReadNumberFileSync(/* cgroup v1 */ '/sys/fs/cgroup/memory/soft_limit_in_bytes');
+
+      const applicableMemoryConstraint = cgroupMemoryReservation ?? cgroupMemoryConstraint;
```
bergundy, line 551: nit: don't read the files for `cgroupMemoryReservation` if non of the memory constraint files were found.

packages/worker/src/runtime.ts
```diff
@@ -536,31 +536,51 @@ export class Runtime {
...
           `This program is running inside a containerized environment with a memory constraint ` +
-            `(eg. docker --memory ${totalMemInMb}m). Node itself does not consider this memory constraint ` +
+            `(eg. '${dockerArgs}' or similar). Node itself does not consider this memory constraint ` +
             `in how it manages its heap memory. There is consequently a high probability that ` +
             `the process will crash due to running out of memory. To increase reliability, we recommend ` +
             `adding '--max-old-space-size=${suggestedOldSpaceSizeInMb}' to your node arguments. ` +
-            `Refer to https://docs.temporal.io/application-development/worker-performance for more ` +
-            `advice on tuning your workers.`
+            `Refer to https://docs.temporal.io/dev-guide/typescript/foundations#run-a-worker-on-docker ` +
+            `for more advices on tuning your Workers.`
```
bergundy, line 574:
```suggestion
            `for more advice on tuning your Workers.`
```

packages/worker/src/worker-options.ts
```diff
@@ -193,12 +203,72 @@ export interface WorkerOptions {
...
+   *
+   * Increasing this number will have no impact if Workflow Task pollers can't fill available execution slots fast
+   * enough. Therefore, when adjusting this value, you may want to similarly adjust `maxConcurrentWorkflowTaskPolls`.
+   * See {@link WorkerOptions.maxConcurrentWorkflowTaskPolls} for more information.
+   *
+   * Also, setting this value too high might cause Workflow Task timeouts due to the fact that the Worker is not able
+   * to complete processing accepted Workflow Tasks fast enough. Increasing the number of Workflow threads
+   * (see {@link WorkerOptions.workflowThreadPoolSize}) may help in that case.
+   *
+   * As a general guidelines:
```
bergundy, line 220:
```suggestion
   * General guidelines:
```

packages/worker/src/worker-options.ts
```diff
@@ -193,12 +203,72 @@ export interface WorkerOptions {
...
+   * - High latency to Temporal Server => Increase this number
+   * - Very short Workflow Tasks (no lengthy Local Activities) => increase this number
+   * - Very long/heavy Workflow Histories => decrease this number
+   * - Low CPU usage despite backlog of Workflow Tasks => increase this number
+   * - High number of Workflow Task timeouts => decrease this number
+   *
+   * In our reference performance test against Temporal Cloud, running with a single Workflow thread and the Reuse V8
+   * Context option enabled, we reached peak performance with a `maxConcurrentWorkflowTaskExecutions` of `120`, and
+   * `maxConcurrentWorkflowTaskPolls` of `60`. Your millage may vary.
```
bergundy, line 229: You might want to mention the latency from where you tested and the type of load if relevant

packages/worker/src/worker-options.ts
```diff
@@ -193,12 +203,72 @@ export interface WorkerOptions {
...
+   *
+   * Can't be lower than 2 if `maxCachedWorkflows` is non-zero.
+   * @default 40
```
bergundy, line 232: Should be mentioned in release notes (I'd put it in the PR description too) that the default changed.

packages/worker/src/worker-options.ts
```diff
@@ -193,12 +203,72 @@ export interface WorkerOptions {
...
+   * In general, a Workflow Worker's performance is mostly network bound (due to communication latency with the
+   * Temporal server). Polling multiple Workflow Tasks concurrently helps compensate for this latency, by ensuring that
+   * the Worker is not starved waiting for the server to return new Workflow Tasks to execute.
+   *
+   * This setting is highly related with {@link WorkerOptions.maxConcurrentWorkflowTaskExecutions}. In various
+   * performance tests, we generally got optimal performance by setting this value to about half of
+   * `maxConcurrentWorkflowTaskExecutions`. Your millage may vary.
+   *
+   * Setting this value higher than needed will generally not have negative impact on this Worker's performance; your
+   * server may however impose a limit on the total number of concurrent Workflow Task pollers.
```
bergundy, line 248: But it may have a negative impact on the cluster's perf and may result in workflow tasks timing out if they're not processed in a timely fashion.

packages/worker/src/worker-options.ts
```diff
@@ -210,19 +280,37 @@ export interface WorkerOptions {
...
+  /**
+   * Controls the number of Worker threads the Worker should create.
+   *
+   * Threads are used to create {@link https://nodejs.org/api/vm.html | vm }s for the isolated Workflow environment.
+   *
+   * New Workflows are created on the thread having the least active Workflows.
+   *
+   * @default 1 for reuseV8Context, otherwise 8. Ignored if `debugMode` is enabled.
+   */
+  workflowThreadPoolSize?: number;
```
bergundy, line 312: Remind me again why we're exposing this?
Have you seen this have any effect on perf?

packages/worker/src/worker-options.ts
```diff
@@ -210,19 +280,37 @@ export interface WorkerOptions {
...
-   * Cached Workflows continue execution from their last stopping point.
-   * If the Worker is asked to run an uncached Workflow, it will need to replay the entire Workflow history.
-   * Use as a dial for trading memory for CPU time.
+   * Cached Workflows continue execution from their last stopping point. If the Worker is asked to run an uncached
+   * Workflow, it will need to fetch and replay the entire Workflow history.
+   *
+   * **When `reuseV8Context` is disabled** — Most users are able to fit at least 250 Workflows per GB of available
+   * memory. The major factors contributing to a Workflow's memory weight are the size of allocations made by the
+   * Workflow itself and by all loaded librairies (including the Node JS builtin context), as well as the size of the
+   * Workflow bundle (code and source map). For the SDK test Workflows, we managed to fit 750 Workflows per GB. Your
```
bergundy, line 289: And the size of inputs (stored in memory state). Note that there's a bug in Core where it does not discard payloads from already replayed events.

packages/worker/src/worker-options.ts
bergundy: Also, I'm thinking the default of 8 is too high without reuse context. Need to test that.

packages/worker/src/runtime.ts
mjameswh: No. I really want to get both so that we can build the equivalent docker args below. I've seen a few cases recently where users had different value on request vs limit, where node process would get OOMkilled only once the host machine would be under pressure. Most often, the devs that were trying to diagnose these problem were not aware of how the pods had been configured, which slowed down research. I want to make it easier to find out these things.

mjameswh: Initially, I was even considering printing out a warning whenever request and limit are different, but ended up not doing so since the risk is rather moot if heap is calculated based on _request_ (rather than limit).

packages/worker/src/worker-options.ts
mjameswh: As far as I could see in perf tests, task timeouts were linked to setting `maxConcurrentWorkflowTaskExecutions` too high, not to increasing `maxConcurrentWorkflowTaskPolls`. Core won't poll anyway if there is no execution slot available.

Still, I will change the phrasing to mention that too high might be negative on the server cluster.

packages/worker/src/worker-options.ts
mjameswh: > Remind me again why we're exposing this? Have you seen this have any effect on perf?

Oh... I thought I had documented it, but yes, I've seen positive effect in tests.

With `reuseV8Context`, increasing this to 2 was a must to avoid WF Task Timeouts when setting `maxConcurrentWorkflowTaskExecutions` to higher values, and assuming WF poller was also high enough to keep execution slots filled (ie. we keep the pipeline enough so that WFT handling is CPU bound rather than network bound). At lower value, increasing the number of thread would have no effect (neither negative nor positive), since WFT handling is network bound.

Another reason for exposing this is for cases where people are running several WF Workers from a single node process, with `reuseV8Context` disabled, on rather modest machines. In that case, running 8 threads per worker uselessly creates resource contention. Recently had to recommend one customer to reduce this value on some workers, even though it was technically not exposed.

bergundy: Still 8 is excessive AFAICT, let's lower to 2 by default for non reuse.

packages/worker/src/worker-options.ts
```diff
@@ -217,16 +217,18 @@ export interface WorkerOptions {
...
    * - Very long/heavy Workflow Histories => decrease this number
    * - Low CPU usage despite backlog of Workflow Tasks => increase this number
    * - High number of Workflow Task timeouts => decrease this number
    *
-   * In our reference performance test against Temporal Cloud, running with a single Workflow thread and the Reuse V8
-   * Context option enabled, we reached peak performance with a `maxConcurrentWorkflowTaskExecutions` of `120`, and
-   * `maxConcurrentWorkflowTaskPolls` of `60`. Your millage may vary.
+   * In some performance test against Temporal Cloud, running with a single Workflow thread and the Reuse V8 Context
+   * option enabled, we reached peak performance with a `maxConcurrentWorkflowTaskExecutions` of `120`, and
+   * `maxConcurrentWorkflowTaskPolls` of `60` (worker machine: Apple M2 Max; ping of 74 ms to Temporal Cloud;
```
bergundy, line 229: I'm wondering what the CPU load was and number of wf/s (not that we need to document that).

bergundy (approved):

bergundy:
Awesome work on this!

PR #1133, opened 2023-06-05
fix(workflow): Make Local Activily timeouts in ActivityInfo match those of non-Local Activities

mjameswh:
## What changed

- Make Local Activily timeouts in ActivityInfo match those of non-Local Activities.
- Restructure LA tests to use the newly introduced "workflow next to test" format, which makes it easier to define a greater number of test cases.
- Updated core to 31099415

## Why

- Because that matches how the Temporal server would fill details for a regular Activity,
- Previously, when if user only set a Local Activity's `scheduleToCloseTimeout`, then TS SDK would crash when receiving the LA task, because `startToCloseTimeout` was undefined (but was mandatory). See [this forum post](https://community.temporal.io/t/how-to-make-local-activities-with-typescript-sdk/8414) for one example of this.


packages/test/src/test-local-activities.ts
```diff
@@ -1,339 +1,619 @@
...
+  ): Promise<WorkflowHandle<T>>;
+}
 
-test.beforeEach(async (t) => {
-  const title = t.title.replace('beforeEach hook for ', '');
-  const taskQueue = `test-local-activities-${title}`;
-  t.context = {
-    ...t.context,
-    client: new WorkflowClient(),
+function helpers(t: ExecutionContext<Context>): Helpers {
```
bergundy, line 53: Let's make sure we share this code.

bergundy (approved):

packages/test/src/test-local-activities.ts
mjameswh: Definitely my intention. But that will be a different PR, after 1.8.0 release.

Issue #1066, opened 2023-02-28
[Feature Request] add string types for `ms` package


akamos:
### Is your feature request related to a problem? Please describe.

When for example using the `sleep` method, it's possible to describe the time as a string, because of the built-in `ms` package. It would be great to use the string types from `ms` instead of a simple `string` type, to have typechecking on the given time argument.

### Describe the solution you'd like

The solution would be to use the typing provided by `ms`: https://github.com/vercel/ms/blob/78ce59eab01c197b9133bf9752ae01fbc11f4976/src/index.ts#L44

mjameswh:
@bergundy This is what I have in mind for this.

It maintains backward compatibility, at the cost of complaining about deprecation rather than entirely refusing incorrect syntaxes. We could encourage people to progressively switch to `Duration` instead of `number | string` in their own code, and schedule removal of the deprecated `number | string` form in a few months.

```
type Duration = number | MsFormatedValue;

export function sleep(ms: Duration): Promise<void>;

/** @deprecated */
export function sleep(ms: string): Promise<void>;

export function sleep(ms: number | Duration | string): Promise<void> {
  // ...
}
```

bergundy:
Hmm.. the only thing I'm worried about here is users reading these values from external configurations so we'll have to figure out what we want to do there but I think your suggestion makes total sense @mjameswh.

mjameswh:
Fixed by #1136.

PR #1136, opened 2023-06-07
feat!: Protect against ms durations errors

### LINKED ISSUE #1046: [Bug] `sleep('1 month')` results in UnhandledCommand WFT failure

Start and cancel a workflow that just calls an activity that sleeps.

### Describe the bug

WorkflowTaskFailed with message `UnhandledCommand`

### Minimal Reproduction

https://github.com/lorensr/workflow-statuses/blob/3b1d476fca708a8e1e98a711c9caa94e3da064cc/src/client.ts#L29-L35

image

mjameswh:
## What changed

- Added type definitions for "ms-formated strings"
- Introduced a Duration type which is either a ms-formated string or a number of milliseconds
- Make ms-formated-string to number conversions throw on invalid time specifications

## Why
- Invalid time specifiers would previously cause misterious issues: including unhandled commands (ie. `sleep('1 month')`) or `condition` behaving as if no timeout was provided (ie. `condition(fn, '1 month')`)
- Fix #1046 and #1066

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -15,7 +15,7 @@
     "@opentelemetry/api": "^1.4.1",
     "@temporalio/proto": "file:../proto",
     "long": "^5.2.0",
-    "ms": "^2.1.3",
+    "ms": "^3.0.0-canary.1",
     "proto3-json-serializer": "^1.0.3",
     "protobufjs": "^7.0.0"
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -18206,9 +18206,17 @@
         "@opentelemetry/api": "^1.4.1",
         "@temporalio/proto": "file:../proto",
         "long": "^5.2.0",
-        "ms": "^2.1.3",
+        "ms": "^3.0.0-canary.1",
         "proto3-json-serializer": "^1.0.3",
         "protobufjs": "^7.0.0"
+      }
+    },
+    "packages/common/node_modules/ms": {
+      "version": "3.0.0-canary.1",
+      "resolved": "https://registry.npmjs.org/ms/-/ms-3.0.0-canary.1.tgz",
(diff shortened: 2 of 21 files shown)


packages/common/package.json
```diff
@@ -15,7 +15,7 @@
     "@opentelemetry/api": "^1.4.1",
     "@temporalio/proto": "file:../proto",
     "long": "^5.2.0",
-    "ms": "^2.1.3",
+    "ms": "^3.0.0-canary.1",
```
bergundy, line 18: Canary?

bergundy (approved):
I'm approving but I'm a bit wary of introducing this breaking change (I'd at least mark it as such).
I can imagine users loading their durations from configuration or passing strings around and they'd need to cast to `StringValue` for compilation to pass.

packages/common/package.json
mjameswh: Yeah... That's not an error.

'ms' has been on "3.0.0-canary.1" since Sept 2021, with >50K weekly downloads. Maintainer is saying they're waiting for a bit more adoption before marking it as stable...

mjameswh:
> I'm approving but I'm a bit wary of introducing this breaking change (I'd at least mark it as such). I can imagine users loading their durations from configuration or passing strings around and they'd need to cast to `StringValue` for compilation to pass.

Yeah. I originally thought of accepting raw strings with deprecated function overloads, but most usages of durations are in types, making this impossible to achieve in a practical way (ie. we would need to duplicate every type definition that contains Duration parameters, plus overload every functions that accept one of these types).

Definitely agree on marking this as breaking.

candra-xendit:
Hi @mjameswh , @bergundy 
may i know does typescript support this mix of `type` for import?
afaik it need to be separated in the import or include both `type`

option 1
- ```
import type { Timestamp } from './time';
import { Duration } from './time';```

option 2
- ```
import type { Timestamp, Duration } from './time';
```

**Comment by mjameswh** [comment]
> Hi @mjameswh , @bergundy may i know does typescript support this mix of `type` for import? afaik it need to be separated in the import or include both `type`

Sure, like this:

```
import { Duration, type Timestamp } from './time';
```

PR #1138, opened 2023-06-19
feat(activity): Add Activity Context logger

mjameswh:
## What changed

- Added `Context.logger`. It defaults to `Runtime.instance().logger`
- Interceptors may override the value of `Context.logger` to set a custom logger.
- `ActivityInboundLogInterceptor` is still installed by default. It may optionally be used to set to a custom logger. Enabling this interceptor provides enriched metadata from activity context on each log entry.

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -71,7 +71,8 @@
 
 import 'abort-controller/polyfill'; // eslint-disable-line import/no-unassigned-import
 import { AsyncLocalStorage } from 'node:async_hooks';
-import { Duration, msToNumber } from '@temporalio/common/lib/time';
+import { Logger, Duration } from '@temporalio/common';
+import { msToNumber } from '@temporalio/common/lib/time';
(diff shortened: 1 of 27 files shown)


packages/activity/src/index.ts
```diff
@@ -243,6 +244,18 @@ export class Context {
    * The heartbeat implementation, injected via the constructor.
    */
   protected readonly heartbeatFn: (details?: any) => void;
+  /**
+   * The logger for this Activity.
+   *
+   * This defaults to the `Runtime`'s Logger (see {@link Runtime.logger}). If the {@link ActivityInboundLogInterceptor}
+   * is installed (by default, it is; see {@link WorkerOptions.interceptors}), then various attributes from the current
```
bergundy, line 251: I think here would be a good place to mention that you can subclass the interceptor to customize the attributes that get emitted by it.

bergundy (approved):

PR #1143, opened 2023-06-22
Ensure payload converters keep Uint8Array type equality (#1142)

### LINKED ISSUE #1142: [Bug] Binary and protobuf payload converters break Uint8Array type equality
### Describe the bug

Payload converters run in the workflow context but sometimes return objects, e.g., Uint8Array, that use constructors defined in the node context. This makes comparing objects with 'instanceof' unreliable, breaking some third-party libraries.

Also, some less commonly used converters rely on the Buffer class that is not visible in the workflow context.  

### Minimal Reproduction

See the temporalio/features repo under features/data_converters/{json_protobuf, binary_protobuf, and binary}

### Environment/Versions

All

antlai-temporal:
Closes #1142

## What was changed

Inject in the workflow context two new globals: Buffer and URL

  - Buffer delegates unsafe allocation methods to safe ones.

Modify BinaryPayloadConverter and ProtobufBinaryPayloadConverter so that they wrap Uint8Array with an instance of Uint8Array from the workflow context.

## Why?

Preserve instanceof semantics by payload converters

## Checklist

1. Closes #1142 

3. How was this tested:

Modified integration test args-and-return to check instanceof and that Buffer is properly injected


packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -88,7 +90,8 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
         },
       }
     );
-    const globals = { AsyncLocalStorage, assert, __webpack_module_cache__ };
+    const SafeBuffer = { ...Buffer, allocUnsafeSlow: Buffer.alloc, allocUnsafe: Buffer.alloc };
```
bergundy, line 93: Does this remove the Buffer constructor?

bergundy (approved):
Just one comment about the buffer constructor, otherwise LGTM

packages/worker/src/workflow/reusable-vm.ts
antlai-temporal: Yes, Buffer constructor is supposed to be deprecated because is unsafe, do we need to keep it?

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -88,7 +90,8 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
...
+    const globals = { AsyncLocalStorage, assert, __webpack_module_cache__, Buffer: SafeBuffer, url };
```
mjameswh, line 94: Here, `url` is the whole package's export. That means that the URL class won't be accessible directly as a global (ie. that would be `global.url.URL`). You will need to add URL and URLSearchParams directly into globals.

mjameswh: I'm a bit bothered by `global` namespace pollution. I think we should wrap reexported classes and modules into a `global.__temporal_injected_modules` object, except for those that would usually be accessible in `global`.

packages/core-bridge/sdk-core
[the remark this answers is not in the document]
antlai-temporal: Done

packages/worker/src/workflow/reusable-vm.ts
antlai-temporal: Removed Buffer and URL changes from this PR

packages/worker/src/workflow/reusable-vm.ts
antlai-temporal: Removed Buffer and URL changes from this PR

PR #1146, opened 2023-06-22
Worker versioning

Sushisource:
Support for the new Worker Versioning feature

### DIFF
--- a/packages/client/src/build-id-types.ts
+++ b/packages/client/src/build-id-types.ts
@@ -1,7 +1,7 @@
 import { temporal } from '@temporalio/proto';
 
 /**
- * Operations that can be passed to {@link WorkflowClient.updateWorkerBuildIdCompatability}.
+ * Operations that can be passed to {@link WorkflowClient.updateWorkerBuildIdCompatibility}.
  *
  * @experimental
  */
@@ -13,7 +13,7 @@ export type BuildIdOperation =
   | MergeSets;
 
 /**
- * This operation adds a new Build Id into a new set, which will be used as the default set for
+ * Adds a new Build Id into a new set, which will be used as the default set for
  * the queue. This means all new workflows will start on this Build Id.
  *
  * @experimental
@@ -24,7 +24,7 @@ interface NewIdInNewDefaultSet {
 }
 
 /**
- * This operation adds a new Build Id into an existing compatible set. The newly added ID becomes
+ * Adds a new Build Id into an existing compatible set. The newly added ID becomes
  * the default for that compatible set, and thus new workflow tasks for workflows which have been
  * executing on workers in that set will now start on this new Build Id.
  *
@@ -39,11 +39,11 @@ interface NewCompatibleVersion {
   existingCompatibleBuildId: string;
   // If set to true, the targeted set will also be promoted to become the
   // overall default set for the queue.
-  makeSetDefault?: boolean;
+  promoteSet?: boolean;
 }
 
 /**
- * This operation promotes a set of compatible Build Ids to become the current
+ * Promotes a set of compatible Build Ids to become the current
  * default set for the task queue. Any Build Id in the set may be used to
  * target it.
  *
@@ -55,8 +55,8 @@ interface PromoteSetByBuildId {
 }
 
 /**
- * This operation promotes a Build Id within an existing set to become the
- * default ID for that set.
+ * Promotes a Build Id within an existing set to become the default ID for that
+ * set.
  *
  * @experimental
  */
@@ -66,9 +66,9 @@ interface PromoteBuildIdWithinSet {
 }
 
 /**
- * This operation merges two sets into one set, thus declaring all the Build Ids in both as
- * compatible with one another. The default of the primary set is maintained as the merged set's
- * overall default.
+ * Merges two sets into one set, thus declaring all the Build Ids in both as
+ * compatible with one another. The default of the primary set is maintained as
+ * the merged set's overall default.
  *
  * @experimental
  */
--- a/packages/client/src/build-id-types.ts
+++ b/packages/client/src/build-id-types.ts
@@ -19,7 +19,7 @@ export type BuildIdOperation =
  * @experimental
  */
 interface NewIdInNewDefaultSet {
-  operation: 'NEW_ID_IN_NEW_DEFAULT_SET';
+  operation: 'newIdInNewDefaultSet';
   buildId: string;
 }
 
@@ -31,7 +31,7 @@ interface NewIdInNewDefaultSet {
  * @experimental
  */
 interface NewCompatibleVersion {
-  operation: 'NEW_COMPATIBLE_VERSION';
+  operation: 'newCompatibleVersion';
   // buildId The Build Id to add to an existing compatible set.
   buildId: string;
   // A Build Id which must already be defined on the task queue, and is used to
@@ -50,7 +50,7 @@ interface NewCompatibleVersion {
  * @experimental
  */
 interface PromoteSetByBuildId {
-  operation: 'PROMOTE_SET_BY_BUILD_ID';
+  operation: 'promoteSetByBuildId';
   buildId: string;
 }
 
@@ -61,7 +61,7 @@ interface PromoteSetByBuildId {
  * @experimental
  */
 interface PromoteBuildIdWithinSet {
-  operation: 'PROMOTE_BUILD_ID_WITHIN_SET';
+  operation: 'promoteBuildIdWithinSet';
   buildId: string;
 }
 
@@ -73,7 +73,7 @@ interface PromoteBuildIdWithinSet {
  * @experimental
  */
 interface MergeSets {
-  operation: 'MERGE_SETS';
+  operation: 'mergeSets';
   // A Build Id which is used to find the primary set to be merged.
   primaryBuildId: string;
   // A Build Id which is used to find the secondary set to be merged.
--- a/packages/client/src/build-id-types.ts
+++ b/packages/client/src/build-id-types.ts
@@ -26,16 +26,16 @@ interface NewIdInNewDefaultSet {
 /**
  * This operation adds a new Build Id into an existing compatible set. The newly added ID becomes
  * the default for that compatible set, and thus new workflow tasks for workflows which have been
- * executing on workers in that set will now start on this new Build ID.
+ * executing on workers in that set will now start on this new Build Id.
  *
  * @experimental
  */
 interface NewCompatibleVersion {
   operation: 'newCompatibleVersion';
-  // buildId The Build Id to add to an existing compatible set.
+  // The Build Id to add to an existing compatible set.
   buildId: string;
   // A Build Id which must already be defined on the task queue, and is used to
-  // find the compatible set to add the new ID to.
+  // find the compatible set to add the new id to.
   existingCompatibleBuildId: string;
   // If set to true, the targeted set will also be promoted to become the
   // overall default set for the queue.
@@ -43,8 +43,8 @@ interface NewCompatibleVersion {
 }
 
 /**
- * This operation promotes a set of compatible Build IDs to become the current
- * default set for the task queue. Any Build ID in the set may be used to
+ * This operation promotes a set of compatible Build Ids to become the current
+ * default set for the task queue. Any Build Id in the set may be used to
  * target it.
  *
  * @experimental
@@ -55,7 +55,7 @@ interface PromoteSetByBuildId {
 }
 
 /**
- * This operation promotes a Build ID within an existing set to become the
+ * This operation promotes a Build Id within an existing set to become the
  * default ID for that set.
  *
  * @experimental
@@ -81,7 +81,7 @@ interface MergeSets {
 }
 
 /**
- * Represents the sets of compatible Build ID versions associated with some
+ * Represents the sets of compatible Build Id versions associated with some
  * Task Queue, as fetched by {@link WorkflowClient.getWorkerBuildIdCompatability}.
  *
  * @experimental
@@ -97,7 +97,7 @@ export class WorkerBuildIdVersionSets {
   }
 
   /**
-   * Returns the default set of compatible Build IDs for the task queue these sets are
+   * Returns the default set of compatible Build Ids for the task queue these sets are
    * associated with.
    */
   public defaultSet(): BuildIdVersionSet {
@@ -105,7 +105,7 @@ export class WorkerBuildIdVersionSets {
   }
 
   /**
-   * Returns the overall default Build ID for the task queue these sets are
+   * Returns the overall default Build Id for the task queue these sets are
    * associated with.
    */
   public defaultBuildId(): string {
@@ -114,7 +114,7 @@ export class WorkerBuildIdVersionSets {
 }
 
 /**
- * Represents one set of compatible Build IDs.
+ * Represents one set of compatible Build Ids.
  *
  * @experimental
  */
@@ -123,13 +123,13 @@ export class BuildIdVersionSet {
 
   constructor(set: temporal.api.taskqueue.v1.ICompatibleVersionSet) {
     if (set == null || set.buildIds == null || set.buildIds.length === 0) {
-      throw new Error('Compatible version sets must contain at least one Build ID');
+      throw new Error('Compatible version sets must contain at least one Build Id');
     }
     this.buildIds = set.buildIds;
   }
 
   /**
-   * Returns the default Build ID for this set
+   * Returns the default Build Id for this set
    */
   public default(): string {
     return this.buildIds[this.buildIds.length - 1];
--- a/packages/client/src/build-id-types.ts
+++ b/packages/client/src/build-id-types.ts
@@ -2,6 +2,8 @@ import { temporal } from '@temporalio/proto';
 
 /**
  * Operations that can be passed to {@link WorkflowClient.updateWorkerBuildIdCompatability}.
+ *
+ * @experimental
  */
 export type BuildIdOperation =
   | NewIdInNewDefaultSet
@@ -13,6 +15,8 @@ export type BuildIdOperation =
 /**
  * This operation adds a new Build Id into a new set, which will be used as the default set for
  * the queue. This means all new workflows will start on this Build Id.
+ *
+ * @experimental
  */
 interface NewIdInNewDefaultSet {
   operation: 'NEW_ID_IN_NEW_DEFAULT_SET';
@@ -22,8 +26,9 @@ interface NewIdInNewDefaultSet {
 /**
  * This operation adds a new Build Id into an existing compatible set. The newly added ID becomes
  * the default for that compatible set, and thus new workflow tasks for workflows which have been
- * executing on workers in that set will now start on this new Build Id.
+ * executing on workers in that set will now start on this new Build ID.
  *
+ * @experimental
  */
 interface NewCompatibleVersion {
   operation: 'NEW_COMPATIBLE_VERSION';
@@ -37,41 +42,82 @@ interface NewCompatibleVersion {
   makeSetDefault?: boolean;
 }
 
+/**
+ * This operation promotes a set of compatible Build IDs to become the current
+ * default set for the task queue. Any Build ID in the set may be used to
+ * target it.
+ *
+ * @experimental
+ */
 interface PromoteSetByBuildId {
   operation: 'PROMOTE_SET_BY_BUILD_ID';
   buildId: string;
 }
 
+/**
+ * This operation promotes a Build ID within an existing set to become the
+ * default ID for that set.
+ *
+ * @experimental
+ */
 interface PromoteBuildIdWithinSet {
   operation: 'PROMOTE_BUILD_ID_WITHIN_SET';
   buildId: string;
 }
 
+/**
+ * This operation merges two sets into one set, thus declaring all the Build Ids in both as
+ * compatible with one another. The default of the primary set is maintained as the merged set's
+ * overall default.
+ *
+ * @experimental
+ */
 interface MergeSets {
   operation: 'MERGE_SETS';
+  // A Build Id which is used to find the primary set to be merged.
   primaryBuildId: string;
+  // A Build Id which is used to find the secondary set to be merged.
   secondaryBuildId: string;
 }
 
+/**
+ * Represents the sets of compatible Build ID versions associated with some
+ * Task Queue, as fetched by {@link WorkflowClient.getWorkerBuildIdCompatability}.
+ *
+ * @experimental
+ */
 export class WorkerBuildIdVersionSets {
   public readonly versionSets: BuildIdVersionSet[];
 
   constructor(resp: temporal.api.workflowservice.v1.GetWorkerBuildIdCompatibilityResponse) {
     if (resp == null || resp.majorVersionSets == null || resp.majorVersionSets.length === 0) {
-      throw new Error('Must be constructed froma a compatability response with at least one version set');
+      throw new Error('Must be constructed from a compatability response with at least one version set');
     }
     this.versionSets = resp.majorVersionSets.map((set) => new BuildIdVersionSet(set));
   }
 
+  /**
+   * Returns the default set of compatible Build IDs for the task queue these sets are
+   * associated with.
+   */
   public defaultSet(): BuildIdVersionSet {
     return new BuildIdVersionSet(this.versionSets[this.versionSets.length - 1]);
   }
 
+  /**
+   * Returns the overall default Build ID for the task queue these sets are
+   * associated with.
+   */
   public defaultBuildId(): string {
     return this.defaultSet().default();
   }
 }
 
+/**
+ * Represents one set of compatible Build IDs.
+ *
+ * @experimental
+ */
 export class BuildIdVersionSet {
   public readonly buildIds: string[];
 
@@ -82,6 +128,9 @@ export class BuildIdVersionSet {
     this.buildIds = set.buildIds;
   }
 
+  /**
+   * Returns the default Build ID for this set
+   */
   public default(): string {
     return this.buildIds[this.buildIds.length - 1];
   }
--- a/packages/client/src/build-id-types.ts
+++ b/packages/client/src/build-id-types.ts
@@ -1,13 +1,13 @@
 import { temporal } from '@temporalio/proto';
 
 /**
- * Operations that can be passed to {@link WorkflowClient.updateWorkerBuildIdCompatibility}.
+ * Operations that can be passed to {@link TaskQueueClient.updateBuildIdCompatibility}.
  *
  * @experimental
  */
 export type BuildIdOperation =
-  | NewIdInNewDefaultSet
-  | NewCompatibleVersion
+  | AddNewIdInNewDefaultSet
+  | AddNewCompatibleVersion
   | PromoteSetByBuildId
   | PromoteBuildIdWithinSet
   | MergeSets;
@@ -18,8 +18,8 @@ export type BuildIdOperation =
  *
  * @experimental
  */
-interface NewIdInNewDefaultSet {
-  operation: 'newIdInNewDefaultSet';
+interface AddNewIdInNewDefaultSet {
+  operation: 'addNewIdInNewDefaultSet';
   buildId: string;
 }
 
@@ -30,8 +30,8 @@ interface NewIdInNewDefaultSet {
  *
  * @experimental
  */
-interface NewCompatibleVersion {
-  operation: 'newCompatibleVersion';
+interface AddNewCompatibleVersion {
+  operation: 'addNewCompatibleVersion';
   // The Build Id to add to an existing compatible set.
   buildId: string;
   // A Build Id which must already be defined on the task queue, and is used to
@@ -82,7 +82,7 @@ interface MergeSets {
 
 /**
  * Represents the sets of compatible Build Id versions associated with some
- * Task Queue, as fetched by {@link WorkflowClient.getWorkerBuildIdCompatability}.
+ * Task Queue, as fetched by {@link TaskQueueClient.getBuildIdCompatability}.
  *
  * @experimental
  */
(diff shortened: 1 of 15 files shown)


Sushisource:
Still need to add some docstrings & experimental tags

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+  | MergeSets;
+
+/**
+ * This operation adds a new Build Id into a new set, which will be used as the default set for
+ * the queue. This means all new workflows will start on this Build Id.
+ *
+ * @experimental
+ */
+interface NewIdInNewDefaultSet {
+  operation: 'NEW_ID_IN_NEW_DEFAULT_SET';
```
bergundy, line 22: Looks like for schedules we made the action enum camel case. Let's do that here too.

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+  buildId: string;
+}
+
+/**
+ * This operation adds a new Build Id into an existing compatible set. The newly added ID becomes
+ * the default for that compatible set, and thus new workflow tasks for workflows which have been
+ * executing on workers in that set will now start on this new Build ID.
```
bergundy, line 29: Let's consistently use `Id` please.
```suggestion
 * executing on workers in that set will now start on this new Build Id.
```

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @experimental
+ */
+interface NewCompatibleVersion {
+  operation: 'NEW_COMPATIBLE_VERSION';
+  // buildId The Build Id to add to an existing compatible set.
```
bergundy, line 35:
```suggestion
  // The Build Id to add to an existing compatible set.
```

bergundy: Or maybe `A build Id ...`

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+  buildId: string;
+  // A Build Id which must already be defined on the task queue, and is used to
+  // find the compatible set to add the new ID to.
```
bergundy, line 38:
```suggestion
  // find the compatible set to add the new Id to.
```

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+  existingCompatibleBuildId: string;
+  // If set to true, the targeted set will also be promoted to become the
+  // overall default set for the queue.
+  makeSetDefault?: boolean;
```
bergundy, line 42: Ambiguous
```suggestion
  promoteSet?: boolean;
```

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/**
+ * This operation promotes a set of compatible Build IDs to become the current
```
bergundy, line 46:
```suggestion
 * This operation promotes a set of compatible Build Ids to become the current
```

bergundy: We haven't been consistent in the reset of the codebase, looks like we use `Id` in schedules docs. And a mix of them in workflow options :(

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+ * default set for the task queue. Any Build ID in the set may be used to
```
bergundy, line 47:
```suggestion
 * default set for the task queue. Any Build Id in the set may be used to
```

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @experimental
+ */
+interface PromoteSetByBuildId {
+  operation: 'PROMOTE_SET_BY_BUILD_ID';
+  buildId: string;
+}
+
+/**
+ * This operation promotes a Build ID within an existing set to become the
```
bergundy, line 58: Do we have to start all of the doc strings with "This operation"? Seems redundant.

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+ * default ID for that set.
```
bergundy, line 59:
```suggestion
 * default Id for that set.
```

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+interface MergeSets {
+  operation: 'MERGE_SETS';
+  // A Build Id which is used to find the primary set to be merged.
+  primaryBuildId: string;
+  // A Build Id which is used to find the secondary set to be merged.
+  secondaryBuildId: string;
+}
+
+/**
+ * Represents the sets of compatible Build ID versions associated with some
```
bergundy, line 84:
```suggestion
 * Represents the sets of compatible Build Id versions associated with some
```

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,137 @@
...
+ * Task Queue, as fetched by {@link WorkflowClient.getWorkerBuildIdCompatability}.
+ *
+ * @experimental
+ */
+export class WorkerBuildIdVersionSets {
```
bergundy, line 89: We don't return classes for simple objects like this all over the code base.
Instead define two interfaces:

```
export interface TaskQueueMetadata {
  versionSets: VersionSet[];
  defaultSet: string;
}

export interface VersionSet {
  buildIds: string[];
  defaultBuildId: string;
}
```

I regret that we didn't make the API return all of the task queue metadata `TaskQueueMetadata` as we're going to add more information to it.

For now you can name it `WorkerVersioningInfo`

packages/client/src/workflow-client.ts
```diff
@@ -69,6 +69,9 @@ import {
   WithDefaults,
 } from './base-client';
 import { mapAsyncIterable } from './iterators-utils';
+import { BuildIdOperation, WorkerBuildIdVersionSets } from './build-id-types';
+import IUpdateWorkerBuildIdCompatibilityRequest = temporal.api.workflowservice.v1.IUpdateWorkerBuildIdCompatibilityRequest;
+import { assertNever } from '@temporalio/common/lib/type-helpers';
```
bergundy, line 74: lint

packages/client/src/workflow-client.ts
```diff
@@ -581,6 +591,69 @@ export class WorkflowClient extends BaseClient {
     }
   }
 
+  /**
+   * Used to add new Build IDs or otherwise update the relative compatibility of Build IDs as
```
bergundy, line 595:
```suggestion
   * Used to add new Build Ids or otherwise update the relative compatibility of Build Ids as
```

packages/client/src/workflow-client.ts
```diff
@@ -581,6 +591,69 @@ export class WorkflowClient extends BaseClient {
...
+   * defined on a specific task queue for the Worker Versioning feature. For more on this feature,
+   * see https://docs.temporal.io/workers#worker-versioning
+   *
+   * @param taskQueue The task queue to make changes to.
+   * @param operation The operation to be performed.
+   *
+   * @experimental
+   */
+  public async updateWorkerBuildIdCompatability(taskQueue: string, operation: BuildIdOperation): Promise<void> {
```
bergundy, line 604: Make a new TaskQueueClient and put these methods on it and attach it to the top level `Client` instead.

`client.taskQueue.updateBuildIdCompatibility`

bergundy, line 604: spelling: compatibility

packages/common/src/type-helpers.ts
```diff
@@ -61,3 +61,10 @@ export function errorCode(error: unknown): string | undefined {
 
   return undefined;
 }
+
+/**
+ * Asserts that some type is the never type
+ */
+export function assertNever(x: never): never {
```
bergundy, line 68: Make a specific helper for this use case and throw a TypeError.

packages/client/src/workflow-client.ts
```diff
@@ -487,7 +490,14 @@ export class WorkflowClient extends BaseClient {
       try {
         res = await this.workflowService.getWorkflowExecutionHistory(req);
       } catch (err) {
-        this.rethrowGrpcError(err, { workflowId, runId }, 'Failed to get Workflow execution history');
+        this.rethrowGrpcError(
```
bergundy, line 493: Why?

packages/client/src/workflow-client.ts
bergundy: The word worker will be redundant with a separate client.
Also need a high level API for reachability, put that on the new client.

packages/core-bridge/sdk-core
bergundy, line 1: Do we need to update core for this PR?

packages/worker/src/worker-options.ts
```diff
@@ -81,12 +81,26 @@ export interface WorkerOptions {
   /**
    * A string that should be unique to the exact worker code/binary being executed.
    *
-   * This is used to populate the `binaryChecksum` attribute in history events originated from this Worker.
```
bergundy, line 84: Please mention that this will be used for binaryChecksum in older servers.

packages/worker/src/worker-options.ts
```diff
@@ -81,12 +81,26 @@ export interface WorkerOptions {
...
    *
    * @default `@temporalio/worker` package name and version + checksum of workflow bundle's code
+   *
+   * @experimental
    */
   buildId?: string;
 
+  /**
+   * If set true, this worker opts into the worker versioning feature. This ensures it only receives
+   * workflow tasks for workflows which it claims to be compatible with.
```
bergundy, line 96: I'd add a sentence that this affects how we use the buildId option above.

packages/client/src/build-id-types.ts
Sushisource: Yeah this is just a copy paste mistake

packages/client/src/build-id-types.ts
Sushisource: You led me astray in the airport!

packages/client/src/build-id-types.ts
Sushisource: I'm not sure we should rename them as it's consistent with what I did in the other SDKs right now, but that's not the biggest deal. If you really prefer that we can rename

packages/common/src/type-helpers.ts
Sushisource: What do you mean specific helper? This will work for all cases where you want to do exhausive checks, pretty useful. The throw is actually irrelevant, since what matters happens at compile time, but I changed it anyway.

packages/client/src/workflow-client.ts
Sushisource: Just got autoformatted like that. Not sure how it wasn't before.

packages/core-bridge/sdk-core
Sushisource: It already has been

packages/client/src/workflow-client.ts
Sushisource: I'm fine adding the reachability client but we didn't in Go/Java since there just wasn't a whole lot to abstract over. It can probably be a little nicer in TS though.

packages/test/src/test-worker-versioning.ts
```diff
@@ -0,0 +1,181 @@
...
+      operation: 'mergeSets',
+      primaryBuildId: '2.0',
+      secondaryBuildId: '1.0',
+    });
+    resp = await conn.taskQueue.getWorkerBuildIdCompatability(taskQueue);
+    assert.equal(resp?.defaultBuildId(), '2.0');
+
+    const reachResp = await conn.taskQueue.getBuildIdReachability({ buildIds: ['2.0', '1.0', '1.1'] });
+    assert.deepEqual(reachResp.buildIdReachability.get('2.0')?.taskQueueReachability.get(taskQueue), ['NewWorkflows']);
+    assert.deepEqual(reachResp.buildIdReachability.get('1.1')?.taskQueueReachability.get(taskQueue), []);
```
bergundy, line 176: This will fail once we add a grace time of 5 minutes for the reachability.

packages/test/src/test-worker-versioning.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    await conn.taskQueue.updateWorkerBuildIdCompatibility(taskQueue, {
+      operation: 'mergeSets',
+      primaryBuildId: '2.0',
+      secondaryBuildId: '1.0',
+    });
+    resp = await conn.taskQueue.getWorkerBuildIdCompatability(taskQueue);
+    assert.equal(resp?.defaultBuildId(), '2.0');
+
+    const reachResp = await conn.taskQueue.getBuildIdReachability({ buildIds: ['2.0', '1.0', '1.1'] });
```
bergundy, line 174: This whole part should run in an "eventually" block (retry for a while) because of eventual consistency.

packages/test/src/test-worker-versioning.ts
```diff
@@ -0,0 +1,181 @@
...
+    assert.equal(resp?.defaultBuildId(), '1.1');
+
+    // Target nonexistent build ID
+    await t.throwsAsync(
+      conn.taskQueue.updateWorkerBuildIdCompatibility(taskQueue, {
+        operation: 'newCompatibleVersion',
+        buildId: '1.2',
+        existingCompatibleBuildId: 'amnotreal',
+      }),
+      { message: /amnotreal not found/ }
```
bergundy, line 142: What's the error class that's thrown here?

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,226 @@
...
+  /** The kind of reachability this request is concerned with. */
+  reachability?: ReachabilityType;
+}
+
+export interface ReachabilityResponse {
+  /** Maps Build Ids to their reachability information. */
+  buildIdReachability: Map<string, BuildIdReachability>;
+}
+
+type ReachabilityTypeResponse = ReachabilityType | 'NotFetched';
```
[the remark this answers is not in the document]
Sushisource: That rename doesn't make sense to me, it's not the response type - it's the type *in the response*. Keeping unspecified also seems sort of a loss since the whole point of the higher level API is to make things more clear, but unspecified is not clear.

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,226 @@
...
+   *  concerned with.
+   */
+  taskQueues?: string[];
+  /** The kind of reachability this request is concerned with. */
+  reachability?: ReachabilityType;
+}
+
+export interface ReachabilityResponse {
+  /** Maps Build Ids to their reachability information. */
+  buildIdReachability: Map<string, BuildIdReachability>;
```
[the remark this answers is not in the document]
Sushisource: Record works fine

packages/client/src/workflow-client.ts
Sushisource: Yeah but the CI formatter doesn't correct it. :shrug:

packages/common/src/type-helpers.ts
Sushisource: Any extension of the enum will render this as a compile error. The throw could only happen I believe if you were doing weird JS things and going around TS.

packages/client/src/build-id-types.ts
```diff
@@ -0,0 +1,148 @@
...
+   * Returns the default set of compatible Build Ids for the task queue these sets are
+   * associated with.
+   */
+  defaultSet(): BuildIdVersionSet;
+
+  /**
+   * Returns the overall default Build Id for the task queue these sets are
+   * associated with.
+   */
+  defaultBuildId(): string;
```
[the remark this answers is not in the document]
Sushisource: Cool, I forgot getters exist. IMO it's... kinda wrong, since it hides arbitrary cost operations behind what looks like pure memory access, but, it's nice for interface compatibility.

packages/core-bridge/sdk-core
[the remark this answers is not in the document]
Sushisource: It's not unrelated, it's needed.

packages/common/src/type-helpers.ts
bergundy: No, the server may be updated with new enum values and the SDK may not have those.

packages/test/src/test-worker-versioning.ts
Sushisource: I think it's just `Error`:

```
Error {
    code: 5,
    details: 'targeted compatible_version amnotreal not found',
    metadata: Metadata {
      internalRepr: Map {
        'content-type' => Array [ … ],
        'grpc-status-details-bin' => Array [ … ],
      },
      options: {},
    },
    message: '5 NOT_FOUND: targeted compatible_version amnotreal not found',
  }
  ```

File: packages/test/src/test-worker-versioning.ts (hunk 19 above)
Comment (Sushisource) [reply packages/test/src/test-worker-versioning.ts:176]: I can just assert that the unreachable ones are present and not bother with them being empty? We don't really need to test that server is sending the right thing back, just that we give empty when they aren't reachable.

Other option is just unit test with fake data.

File: packages/test/src/test-worker-versioning.ts (hunk 21 above)
Comment (bergundy) [reply packages/test/src/test-worker-versioning.ts:144]: You need to wrap the call in `rethrowGrpcError` like we do in the other clients.

Comment (Sushisource) [reply packages/test/src/test-worker-versioning.ts:144]: K. I added not found but it's not really possible to tie one specific build id to the error in all cases, but that seems fine since the error message itself will have that info and the main thing is matching on type.

File: packages/common/src/type-helpers.ts (hunk 15 above)
Comment (Sushisource) [reply packages/common/src/type-helpers.ts:73]: Sure, makes sense.

File: packages/client/src/build-id-types.ts [hunk 25]
```diff
@@ -0,0 +1,148 @@
...
+ * Represents one set of compatible Build Ids.
+ *
+ * @experimental
+ */
+export interface BuildIdVersionSet {
+  // All build IDs contained in the set.
+  readonly buildIds: string[];
+
+  // Returns the default Build Id for this set
+  default(): string;
```
bergundy, line 118: Same with the other interface, make it an attribute not a method.

packages/client/src/index.ts
```diff
@@ -38,3 +38,4 @@ export * from './workflow-client';
 export * from './workflow-options';
 export * from './schedule-types';
 export * from './schedule-client';
+export { WorkerBuildIdVersionSets, BuildIdVersionSet, BuildIdOperation } from './build-id-types';
```
bergundy, line 41:
```suggestion
export * from './task-queue-client';
export { WorkerBuildIdVersionSets, BuildIdVersionSet, BuildIdOperation } from './build-id-types';
```

packages/client/src/workflow-client.ts
bergundy: Can you please reset this file, there's no relevant change here.

packages/client/src/index.ts
Sushisource: We don't actually want to export everything though

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,271 @@
...
+        reachability: reachabilityTypeToProto(options.reachability),
+      });
+    } catch (e) {
+      this.rethrowGrpcError(e, 'Unexpected error fetching Build Id reachability');
+    }
+    return reachabilityResponseFromProto(resp);
+  }
+
+  protected rethrowGrpcError(err: unknown, fallbackMessage: string): never {
+    if (isServerErrorResponse(err)) {
```
bergundy, line 144: See how James changed this method in workflow-client:

```
    if (isGrpcServiceError(err)) {
      rethrowKnownErrorTypes(err);
```

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,271 @@
...
+/**
+ * There are different types of reachability:
+ *   - `NewWorkflows`: The Build Id might be used by new workflows
+ *   - `ExistingWorkflows` The Build Id might be used by open workflows and/or closed workflows.
+ *   - `OpenWorkflows` The Build Id might be used by open workflows
+ *   - `ClosedWorkflows` The Build Id might be used by closed workflows
+ */
+export type ReachabilityType = 'NewWorkflows' | 'ExistingWorkflows' | 'OpenWorkflows' | 'ClosedWorkflows';
+
+interface BaseReachabilityOptions {
```
bergundy, line 168: Missing export
```suggestion
export interface BaseReachabilityOptions {
```

packages/client/src/task-queue-client.ts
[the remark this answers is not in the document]
bergundy: Still need to export this

packages/client/src/task-queue-client.ts
Sushisource: This doesn't actually get used directly by users, only through the `ReachabilityOptions` type

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,271 @@
...
+          continue;
+        }
+        if (tqr.reachability == null) {
+          taskQueueReachability[tqr.taskQueue] = [];
+          continue;
+        }
+        taskQueueReachability[tqr.taskQueue] = tqr.reachability.map(reachabilityTypeFromProto);
+      }
+    }
+    buildIdReachability[bir.buildId ?? ''] = { taskQueueReachability };
```
bergundy, line 232: I mentioned this before, using empty strings for keys is weird, let's make a symbol for the unversioned build ID (and please export it).

```ts
export const UnversionedBuildId = Symbol.for('__temporal_unversionedBuildId')
```

This function will return 
```
 { buildIdReachabilty: Record<string | UnversionedBuildId, BuildIdReachability> }
```

And in the request you'll accept
```
buildIds: Array<string | UnversionedBuildId>
```

packages/common/src/type-helpers.ts
bergundy: I guess my only request would be to pass in the error message to this function so it doesn't just say `Unexpected object` which isn't very helpful especially with enums.

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+        break;
+      default:
+        assertNever(operation);
+    }
+    try {
+      await this.workflowService.updateWorkerBuildIdCompatibility(request);
+    } catch (e) {
+      if (isGrpcServiceError(e)) {
+        rethrowKnownErrorTypes(e);
+      }
```
bergundy, line 100: Put this in `rethrowGrpcError` to avoid repetition.

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+    let resp;
+    try {
+      resp = await this.workflowService.getWorkerBuildIdCompatibility({
+        taskQueue,
+        namespace: this.options.namespace,
+      });
+    } catch (e) {
+      if (isGrpcServiceError(e)) {
+        rethrowKnownErrorTypes(e);
+      }
```
bergundy, line 122: Same

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+      resp = await this.workflowService.getWorkerTaskReachability({
+        namespace: this.options.namespace,
+        taskQueues: options.taskQueues,
+        buildIds,
+        reachability: reachabilityTypeToProto(options.reachability),
+      });
+    } catch (e) {
+      if (isGrpcServiceError(e)) {
+        rethrowKnownErrorTypes(e);
+      }
```
bergundy, line 160: Same, avoid repetition.

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+ */
+export interface BaseReachabilityOptions {
+  /**
+   * A list of build ids to query the reachability of. Currently, at least one Build Id must be
+   * specified, but this restriction may be lifted in the future.
+   */
+  buildIds: (string | UnversionedBuildIdType)[];
+  /**
+   *  A list of task queues with Build Ids defined on them that the request is
+   *  concerned with.
```
bergundy, line 202:
```suggestion
   * A list of task queues with Build Ids defined on them that the request is
   * concerned with.
```

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+ * @experimental
+ */
+export type LoadedTaskQueueClientOptions = LoadedWithDefaults<TaskQueueClientOptions>;
+
+/**
+ * A stand-in for a Build Id for unversioned Workers
+ * @experimental
+ */
+export const UnversionedBuildId = Symbol.for('__temporal_unversionedBuildId');
+type UnversionedBuildIdType = typeof UnversionedBuildId;
```
bergundy, line 28:
```suggestion
export type UnversionedBuildIdType = typeof UnversionedBuildId;
```

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+  /** The kind of reachability this request is concerned with. */
+  reachability?: ReachabilityType;
+}
+
+export interface ReachabilityResponse {
+  /** Maps Build Ids to their reachability information. */
+  buildIdReachability: Record<string | UnversionedBuildIdType, BuildIdReachability>;
+}
+
+type ReachabilityTypeResponse = ReachabilityType | 'NotFetched';
```
bergundy, line 214:
```suggestion
export type ReachabilityTypeResponse = ReachabilityType | 'NotFetched';
```

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+    case 'NewWorkflows':
+      return temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_NEW_WORKFLOWS;
+    case 'ExistingWorkflows':
+      return temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_EXISTING_WORKFLOWS;
+    case 'OpenWorkflows':
+      return temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_OPEN_WORKFLOWS;
+    case 'ClosedWorkflows':
+      return temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_CLOSED_WORKFLOWS;
+    default:
+      assertNever(type);
```
bergundy, line 238:
```suggestion
      assertNever(type, 'Unknown reachability type');
```

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+    case temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_NEW_WORKFLOWS:
+      return 'NewWorkflows';
+    case temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_EXISTING_WORKFLOWS:
+      return 'ExistingWorkflows';
+    case temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_OPEN_WORKFLOWS:
+      return 'OpenWorkflows';
+    case temporal.api.enums.v1.TaskReachability.TASK_REACHABILITY_CLOSED_WORKFLOWS:
+      return 'ClosedWorkflows';
+    default:
+      return assertNever(rtype);
```
bergundy, line 278:
```suggestion
      return assertNever(rtype, 'Unknown reachability type');
```

packages/client/src/task-queue-client.ts
```diff
@@ -0,0 +1,297 @@
...
+ * It could be because:
+ * - Id passed is incorrect
+ * - Build Id has been scavenged by the server.
+ *
+ * @experimental
+ */
+export class BuildIdNotFoundError extends Error {
+  public readonly name: string = 'BuildIdNotFoundError';
+
+  constructor(message: string) {
```
bergundy, line 294: If you're not tired from me discovering things every round, you could specify which build id is not found in this exception, could have a `buildId: string` attribute on this class.

packages/client/src/workflow-client.ts
bergundy, line 1: nit: No need to change this file, please reset.

packages/test/src/integration-tests.ts
bergundy, line 1: Reset changes to this file please.

packages/test/src/test-worker-versioning-unit.ts
bergundy, line 1: nit: merge into the other test file.

packages/test/src/test-worker-versioning.ts
bergundy: Add the type to the assertion.

packages/test/src/test-worker-versioning.ts
```diff
@@ -0,0 +1,190 @@
...
+      operation: 'mergeSets',
+      primaryBuildId: '2.0',
+      secondaryBuildId: '1.0',
+    });
+    resp = await conn.taskQueue.getBuildIdCompatability(taskQueue);
+    assert.equal(resp?.defaultBuildId, '2.0');
+
+    await asyncRetry(
+      async () => {
+        const reachResp = await conn.taskQueue.getBuildIdReachability({ buildIds: ['2.0', '1.0', '1.1'] });
```
bergundy, line 180: Can you please add a test for reachability of the unversioned build id?

packages/client/src/task-queue-client.ts
Sushisource: You can't actually do that, because in the merge operation it could be either of two and the error isn't readily parseable. Mentioned here: https://github.com/temporalio/sdk-typescript/pull/1146#discussion_r1244088033

packages/test/src/test-worker-versioning-unit.ts
Sushisource: I couldn't see any obvious way to have different `test` instances in the same file

packages/test/src/integration-tests.ts
Sushisource: Why? This will work with old versions

packages/test/src/integration-tests.ts
```diff
@@ -624,14 +634,28 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
...
       CustomDoubleField: [3.14],
-      BuildIds: ['unversioned', `unversioned:${t.context.worker.options.buildId}`],
     });
+    let checksum: any;
+    if (BinaryChecksums != null) {
+      t.true(BinaryChecksums.length === 1);
+      checksum = BinaryChecksums[0];
+    } else {
+      t.true(BuildIds!.length === 2);
+      t.deepEqual(BuildIds![0], 'unversioned');
```
bergundy, line 651: nit: just deepEquals on the BuildIds.

bergundy (approved):

PR #1147, opened 2023-06-23
feat(client): Throw more specific errors from Client APIs

### LINKED ISSUE #976: [Bug] `WorkflowNotFoundError` used when namespace not found
### Describe the bug

image

### Options

In order of not breaking to most breaking:

1. Add "namespace does not exist" to the list of reasons:

image

https://typescript.temporal.io/api/classes/workflow.WorkflowNotFoundError

2. Add `NamespaceNotFoundError`
3. Remove `WorkflowNotFoundError` and just throw `ServiceError`

### Meta

1. Are there other possible reasons for `grpcStatus.NOT_FOUND`?
2. For all the other gRPC status codes we check for throwing a specific error, are there cases for those codes we haven't considered? Can search for `grpcStatus.*` and `status.*`

### LINKED ISSUE #1145: [Bug] Client.schedule.create is giving "Schedule not found" error if namespace is incorrect
### Describe the bug

Using a schedule client configured with a namespace that doesn't exist, trying to create a new schedule results in a "Schedule not found" exception, rather than something like "Namespace not found".

Very similar errors probably exists elsewhere on ScheduleClient and WorkflowClient.

mjameswh:
## What changed

- All Client APIs now throw `NamespaceNotFoundError` when appropriate, rather than `WorkflowNotFound`, `ScheduleNotFound` or `ActivityNotFound` (fix #1145, fix #976)
- Errors that happen while preparing API requests from a `Client`, and particularly errors happening while converting input objects to payloads, now results in throwing an appropriate exception, rather than an opaque ServiceError.

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -19,7 +19,6 @@ import {
   WorkflowNotFoundError,
   WorkflowResultType,
   extractWorkflowType,
-  NamespaceNotFoundError,
 } from '@temporalio/common';
 import { composeInterceptors } from '@temporalio/common/lib/interceptors';
 import { History } from '@temporalio/common/lib/proto-utils';
@@ -32,7 +31,7 @@ import {
   filterNullAndUndefined,
 } from '@temporalio/common/lib/internal-non-workflow';
 import { temporal } from '@temporalio/proto';
-import { isServerErrorResponse, ServiceError, WorkflowContinuedAsNewError, WorkflowFailedError } from './errors';
+import { ServiceError, WorkflowContinuedAsNewError, WorkflowFailedError, isGrpcServiceError } from './errors';
 import {
   WorkflowCancelInput,
   WorkflowClientInterceptor,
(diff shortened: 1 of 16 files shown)


packages/client/src/workflow-client.ts
```diff
@@ -581,14 +582,19 @@ export class WorkflowClient extends BaseClient {
...
-  protected rethrowGrpcError(err: unknown, workflowExecution: WorkflowExecution, fallbackMessage: string): never {
+  protected rethrowGrpcError(err: unknown, fallbackMessage: string, workflowExecution?: WorkflowExecution): never {
     if (isServerErrorResponse(err)) {
       if (err.code === grpcStatus.NOT_FOUND) {
-        throw new WorkflowNotFoundError(
-          err.details ?? 'Workflow not found',
-          workflowExecution.workflowId,
-          workflowExecution.runId
-        );
+        const matcher = err.message.match(/^5 NOT_FOUND: Namespace (.*?) is not found./);
```
bergundy, line 588: Don’t parse the error message, use the gRPC response header as I suggested in the corresponding issue to get the structured error.

bergundy (approved):

PR #1148, opened 2023-06-23
chore: Update CHANGELOG.md file in prep for 1.8.0 and make last minute API changes

mjameswh:
## What changed

- Updated CHANGELOG.md file in prep for 1.8.0.
- [Update core to get local activity start-to-close fix](https://github.com/temporalio/sdk-typescript/pull/1148/commits/e920da3263e7f2248738d533ae27b6c41eb6d8ad)
- [Fix flake in test-isolation](https://github.com/temporalio/sdk-typescript/pull/1148/commits/7153d4dc4449ed59d2859a81dd9cc8ca66b304c9)
- [Rename activity context logger to log, moar changelog edits](https://github.com/temporalio/sdk-typescript/pull/1148/commits/8379ca42ab60a09d251242393d1515cb142bb740)
- [Add versioning changelog and minor API changes](https://github.com/temporalio/sdk-typescript/pull/1148/commits/7999b73536d168e4e5f8caed82768afe2a7693a7)

### DIFF
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -83,7 +83,7 @@ export interface WorkerOptions {
    *
    * This is used to uniquely identify the worker's code for a handful of purposes, including the
    * worker versioning feature if you have opted into that with
-   * {@link WorkerOptions.useWorkerVersioning}. It will also populate the `binaryChecksum` field
+   * {@link WorkerOptions.useVersioning}. It will also populate the `binaryChecksum` field
    * on older servers.
    *
    * @default `@temporalio/worker` package name and version + checksum of workflow bundle's code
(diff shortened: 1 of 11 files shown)


packages/worker/src/worker-options.ts
```diff
@@ -101,7 +101,7 @@ export interface WorkerOptions {
    *
    * @experimental
    */
-  useWorkerVersioning?: boolean;
```
Sushisource, line 104: This would make it inconsistent with the other SDKs. I get not wanting `worker` inside worker options, but, the redundancy isn't exactly a huge deal and this is clear in tying the feature name to the option name.

Sushisource (approved):

packages/worker/src/worker-options.ts
bergundy: Maybe we should change the other SDKs? It feels needlessly redundant to me in the context of worker options.

PR #1150, opened 2023-06-27
fix!: Make local activity not found error non retryable

bergundy:
While there may be a case for making this failure retryable, it seems like the common user expectation would be that a not found local activity would fail with a non retryable failure as those must be registered along with the calling workflow.


lorensr:
What do we do for not found workflows and activities? Seems like it might
be one of the kinds of “oops I need to redeploy to fix” bug that people
would appreciate the WE to stay open for

Sushisource (approved):
IMO we can't really make this retryable even if we wanted to because it's gonna spin forever WFT heartbeating and that'll look real ugly so, yeah, this makes sense to me.

bergundy:
Replaced with #1152

PR #1152, opened 2023-06-28
fix: Fail workflow task if local activity not registered with worker

bergundy:
Previously the worker would needlessly retry these local activities.
With this change, the worker fails the workflow task allowing a user to deploy a fix.

Replaces https://github.com/temporalio/sdk-typescript/pull/1150 with better behavior.
Matches Java behavior https://github.com/temporalio/sdk-java/pull/1628.

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -23,6 +23,7 @@ import { delay, filter, first, ignoreElements, last, map, mergeMap, takeUntil, t
 import type { RawSourceMap } from 'source-map';
 import { Info as ActivityInfo } from '@temporalio/activity';
 import {
+  ActivityFunction,
   DataConverter,
   decompileRetryPolicy,
   defaultPayloadConverter,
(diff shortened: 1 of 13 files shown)


packages/worker/src/worker.ts
```diff
@@ -1217,6 +1218,7 @@ export class Worker {
                   try {
                     const decodedActivation = await this.workflowCodecRunner.decodeActivation(activation);
                     const unencodedCompletion = await state.workflow.activate(decodedActivation);
+                    this.validateCompletion(unencodedCompletion);
```
cretz, line 1221: Would it be clearer/possible to validate this where the local activity is invoked? You I suppose this is preferred over passing a collection of activity names into the sandbox?

bergundy: I can't pass in activity names if we want to support dynamic activities.

bergundy: That would have been my preference though.

bergundy: To make this error catchable in the workflow.

cretz: > I can't pass in activity names if we want to support dynamic activities.

If there is a dynamic activity on the worker, instead of activity names, you tell the workflow that scheduling a local activity will never fail (i.e. that "all" names are valid).

Sushisource (approved):

PR #1156, opened 2023-06-30
Add versioning intent to commands

Sushisource:
https://github.com/temporalio/sdk-typescript/pull/1146 accidentally left out `VersioningIntent` for commands. This adds that.

### DIFF
--- a/packages/common/src/activity-options.ts
+++ b/packages/common/src/activity-options.ts
@@ -2,6 +2,7 @@ import type { coresdk } from '@temporalio/proto';
 import { RetryPolicy } from './retry-policy';
 import { checkExtends } from './type-helpers';
 import { Duration } from './time';
+import { VersioningIntent } from './versioning-intent';
 
 // Avoid importing the proto implementation to reduce workflow bundle size
 // Copied from coresdk.workflow_commands.ActivityCancellationType
@@ -100,6 +101,14 @@ export interface ActivityOptions {
    * @default true
    */
   allowEagerDispatch?: boolean;
+
+  /**
+   * When using the Worker Versioning feature, specifies whether this Activity should run on a
+   * worker with a compatible Build Id or not. See {@link VersioningIntent}.
+   *
+   * @experimental
+   */
+  versioningIntent?: VersioningIntent;
 }
 
 /**
--- a/packages/test/src/test-workflows.ts
+++ b/packages/test/src/test-workflows.ts
@@ -22,7 +22,6 @@ import { parseWorkflowCode } from '@temporalio/worker/lib/worker';
 import * as activityFunctions from './activities';
 import { cleanStackTrace, REUSE_V8_CONTEXT, u8 } from './helpers';
 import { ProcessedSignal } from './workflows';
-import VersioningIntent = coresdk.common.VersioningIntent;
 
 export interface Context {
   workflow: VMWorkflow | ReusableVMWorkflow;
@@ -758,7 +757,7 @@ test('cancelWorkflow', async (t) => {
           startToCloseTimeout: msToTs('10m'),
           taskQueue: 'test',
           doNotEagerlyExecute: false,
-          versioningIntent: VersioningIntent.UNSPECIFIED,
+          versioningIntent: coresdk.common.VersioningIntent.UNSPECIFIED,
         }),
       ])
     );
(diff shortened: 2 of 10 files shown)


bergundy (commented):
I’m debating if we need this enum at all in Typescript because we only have 3 options, which can be represented with an optional boolean.
I don’t think we will have another intent later on.

Sushisource:
I don't think it matters much. This is probably a bit more obvious

packages/common/src/activity-options.ts
```diff
@@ -100,6 +101,14 @@ export interface ActivityOptions {
...
    */
   allowEagerDispatch?: boolean;
+
+  /**
+   * When using the Worker Versioning feature, specifies whether this Activity should run on a
+   * worker with a compatible Build Id or not. See {@link VersioningIntent}.
+   *
+   * @experimental
+   */
+  versioningIntent?: VersioningIntent;
```
bergundy, line 111: This is a bit messy but for _most_ of the SDK we used enums for this type of option: https://github.com/temporalio/sdk-typescript/blob/ab58f964afe2c0ffb22107b1802a656e35cdc7b6/packages/common/src/activity-options.ts#L9

bergundy: I would do it the way you did it here if I had to do it all over again

Sushisource: So do you want it to be an enum? I dunno that being consistent like that is terribly important, if this is slightly nicer.

mjameswh: I'm generally in favor of consistency, yet I really don't mind going against the trend on this precise question. TypeScript enums unfortunately have quite a few quirks, so I'd be happy to walk away from them from now on. Plus we already have a few places where we used unions of strings for things that could have been enums instead.

My only concern here is that you made those strings Titlecase. In most places where we already used unions of strings, we used ALL_UPPERCASE syntax, with a very few places using camelCase. I think I'd prefer staying with ALL_UPPERCASE, as it makes it more obvious when looking at TS code that it represents some constant keyword, where other syntaxes tends to look like just plain strings.

> I would do it the way you did it here if I had to do it all over again

Not a priority, but we could easily consider accepting both strings and enums in all of these places, and deprecate these enums, so people slowly move away from them. It might be a little bit harder to get away from cases where SDK gives back such enums, but actually, I don't think think there are that much cases.

packages/test/src/test-workflows.ts
```diff
@@ -22,6 +22,7 @@ import { parseWorkflowCode } from '@temporalio/worker/lib/worker';
 import * as activityFunctions from './activities';
 import { cleanStackTrace, REUSE_V8_CONTEXT, u8 } from './helpers';
 import { ProcessedSignal } from './workflows';
+import VersioningIntent = coresdk.common.VersioningIntent;
```
mjameswh, line 25: TIL about this `import ... = something other than require` syntax. Can't even find any documentation about it. Nice to know.

mjameswh (approved):
Just one nit thing.

packages/common/src/activity-options.ts
bergundy: I like the suggestion of transitioning to string unions everywhere. I think our string union types are all camel case though. @mjameswh can you come up with a guideline?

packages/test/src/test-workflows.ts
bergundy: I think this is an old way of importing, I would rather avoid this style for consistency.

bergundy:
@mjameswh please merge once we decide on the casing for string unions.
And after that I would like to cut a 1.8.1 release.

packages/common/src/activity-options.ts
mjameswh: @bergundy I just did some search on the code.

Except for string union types introduced by the Versioning feature (which is still experimental and probably almost unused in the wild yet), all publicly exported string unions are ALL_UPPERCASE.

The only places where we are not doing so are types that are union of other types/interfaces with a discriminant field (for example `EphemeralServerExecutable.type` can be either `'cached-download'` or `'existing-path'; similar for `HeartbeatInput`.

Existing enums were also very consistently using ALL_UPPERCASE.

So I'd maintain that convention.

packages/test/src/test-workflows.ts
Sushisource: Didn't even realize. I think the IDE did it automatically

Sushisource:
Thanks @mjameswh !

packages/common/src/activity-options.ts
bergundy: Schedule action is camel case..

mjameswh: > Schedule action is camel case..

Yeah. No convention holds at the moment for discriminants in union of types/interfaces. I did not fixed these in Versioning either, as I was running out of time to cut a release today.

I will continue analyze on this Monday.

PR #1158, opened 2023-07-04
feat(workflow): Logger improvements

bergundy:
- [Make log attributes optional for defaultWorkerLogger sink](https://github.com/temporalio/sdk-typescript/commit/8d3a3f1770a0962b9f92367f31795d6ce9f66994)
- [Move getting log context from interceptor to sink](https://github.com/temporalio/sdk-typescript/commit/4984534d584b366602998b5f1cda992877d1577a) in order to attach log attributes to all `workflow.log` calls, not just ones made by the `WorkflowInboundLogInterceptor`

The approach taken in this PR trades off the power of modifying the default log attributes from workflow context for simplicity and future extensibility where we may expose workflow task interceptors on the main Node.js thread that can be used to augment the default attributes for logs emitted in `worker.ts` (e.g. https://github.com/temporalio/sdk-typescript/blob/dbcc9e8ef5832296ed739c2bf9c846761b46b47e/packages/worker/src/worker.ts#L1201).

### DIFF
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -13,6 +13,7 @@ import { Runtime } from './runtime';
 import { InjectedSinks } from './sinks';
 import { MiB } from './utils';
 import { defaultWorflowInterceptorModules, WorkflowBundleWithSourceMap } from './workflow/bundler';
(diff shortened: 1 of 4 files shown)


bergundy:
Decided to go with #1159

PR #1162, opened 2023-07-06
fix: Remove instanceof Error checks from `is` methods

bergundy:
This doesn't work when errors are passed from different v8 contexts.
Only known to affect ShutdownError with jest (see https://github.com/jestjs/jest/issues/2549).

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -117,7 +117,7 @@ export class CompleteAsyncError extends Error {
    * Instanceof check that works when multiple versions of @temporalio/activity are installed.
    */
   static is(error: unknown): error is CompleteAsyncError {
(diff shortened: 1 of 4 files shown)


bergundy:
@mjameswh I'm debating at least adding a `typeof err === 'object'` for some basic sanity.
WDYT?

mjameswh:
> I'm debating at least adding a typeof err === 'object' for some basic sanity.

Well... If you go that way, then you can remove the optional chain operator after that.

But TBH I don't think that's useful. As far as I can think of, the only case where a `typeof` check before the optional chain operator would make any difference is if the left part is an unresolvable reference (e.g. a variable that doesn't exist), which is impossible in the context of these methods, as `error` is a function argument.

mjameswh (approved):
LGTM

PR #1166, opened 2023-07-09
Use custom symbol-based impl of `instanceof`

mjameswh:
## Context

#1128 introduced a new strategy, based on presence of a symbol, for checking `instanceof` on Error classes defined by SDK. However this turned out to work incorrectly when running in Jest, because of the remaining `instanceof Error` check. #1162 tried to fix that one, but missed the particular case of converting errors thrown from Core.

## What changed

- Introduce a new decorator function `SymbolBasedInstanceOfError` that adds a custom `instanceof` handler to error classes. It does so by defining a `[Symbol.hasInstance]` function on the object constructor. That custom `instanceof` handler look for the presence of a specific symbol.
- Replace the recently introduced `.is(...)` function by the `.[Symbol.hasInstance]()` function, which makes it automatically used by the `instanceof` operator, in place of the default algorithm. That means that user code automatically benefit from this, without change to their easing code.
- Introduced a new internal helper function `isError`, which uses a duck-typing strategy. Fixed all remaining cases of `instanceof Error` by calling `isError` instead (except for the `create-project`, scripts, and tests, all of which are not meant to be used in contexts where that would matter).
- Similarly strengthened some other type checking helpers
- Apply this new `instanceof` mechanism on all Error and Failure objects. That certainly include some errors that are not meant to be exposed to users or to travel across the Workflow sandbox boundary; however, considering how Jest may also play us tricks, trying to pin point exactly which one are could be subject to instanceof issues seems like an unreasonable approach.
- Also replaced all `instanceof Array` by `isArray()`.

## Limitations

- We are not adding custom `instanceof` handlers for JavaScript's builtin objects (Object, Array, Map, Uint8Array, Date, Promise, etc.). I believe this would technically be possible, but it could likely cause other/unpredictable issues. We should instead use `isXxx` kind of helpers for these things.

## Future work

- We'll need to define isXxx helpers for all built-in objects on which we need to do `instanceof` checks (eg. Uint8Array, Date, Promise, etc). This will be covered in a different PR (don't want to delay 1.8.2 for this)

### DIFF
--- a/packages/client/src/schedule-client.ts
+++ b/packages/client/src/schedule-client.ts
@@ -9,7 +9,7 @@ import {
 } from '@temporalio/common/lib/internal-non-workflow';
 import { temporal } from '@temporalio/proto';
 import { optionalDateToTs, optionalTsToDate, optionalTsToMs, tsToDate } from '@temporalio/common/lib/time';
-import { symbolBasedInstanceOf } from '@temporalio/common/lib/type-helpers';
+import { SymbolBasedInstanceOfError } from '@temporalio/common/lib/type-helpers';
 import { CreateScheduleInput, CreateScheduleOutput, ScheduleClientInterceptor } from './interceptors';
 import { WorkflowService } from './types';
 import { isGrpcServiceError, ServiceError } from './errors';
@@ -527,10 +527,8 @@ export class ScheduleClient extends BaseClient {
  *
  * @experimental
  */
-@symbolBasedInstanceOf('ScheduleAlreadyRunning')
+@SymbolBasedInstanceOfError('ScheduleAlreadyRunning')
 export class ScheduleAlreadyRunning extends Error {
-  public readonly name: string = 'ScheduleAlreadyRunning';
-
   constructor(message: string, public readonly scheduleId: string) {
     super(message);
   }
@@ -544,10 +542,8 @@ export class ScheduleAlreadyRunning extends Error {
  *
  * @experimental
  */
-@symbolBasedInstanceOf('ScheduleNotFoundError')
+@SymbolBasedInstanceOfError('ScheduleNotFoundError')
 export class ScheduleNotFoundError extends Error {
-  public readonly name: string = 'ScheduleNotFoundError';
-
   constructor(message: string, public readonly scheduleId: string) {
     super(message);
   }
--- a/packages/client/src/schedule-client.ts
+++ b/packages/client/src/schedule-client.ts
@@ -527,7 +527,7 @@ export class ScheduleClient extends BaseClient {
  *
  * @experimental
  */
-@symbolBasedInstanceOf('ScheduleAlreadyExistsError')
+@symbolBasedInstanceOf('ScheduleAlreadyRunning')
 export class ScheduleAlreadyRunning extends Error {
   public readonly name: string = 'ScheduleAlreadyRunning';
 
--- a/packages/client/src/schedule-client.ts
+++ b/packages/client/src/schedule-client.ts
@@ -9,6 +9,7 @@ import {
 } from '@temporalio/common/lib/internal-non-workflow';
 import { temporal } from '@temporalio/proto';
 import { optionalDateToTs, optionalTsToDate, optionalTsToMs, tsToDate } from '@temporalio/common/lib/time';
+import { symbolBasedInstanceOf } from '@temporalio/common/lib/type-helpers';
 import { CreateScheduleInput, CreateScheduleOutput, ScheduleClientInterceptor } from './interceptors';
 import { WorkflowService } from './types';
 import { isGrpcServiceError, ServiceError } from './errors';
@@ -526,6 +527,7 @@ export class ScheduleClient extends BaseClient {
  *
  * @experimental
  */
+@symbolBasedInstanceOf('ScheduleAlreadyExistsError')
 export class ScheduleAlreadyRunning extends Error {
   public readonly name: string = 'ScheduleAlreadyRunning';
 
@@ -542,6 +544,7 @@ export class ScheduleAlreadyRunning extends Error {
  *
  * @experimental
  */
+@symbolBasedInstanceOf('ScheduleNotFoundError')
 export class ScheduleNotFoundError extends Error {
   public readonly name: string = 'ScheduleNotFoundError';
 
--- a/packages/common/src/type-helpers.ts
+++ b/packages/common/src/type-helpers.ts
@@ -87,12 +87,17 @@ export function assertNever(msg: string, x: never): never {
   throw new TypeError(msg + ': ' + x);
 }
 
-export type Class = {
-  new (...args: any[]): any;
-  prototype: object;
+export type Class<E extends Error> = {
+  new (...args: any[]): E;
+  prototype: E;
 };
 
 /**
+ * A decorator to be used on error classes. It adds the 'name' property AND provides a custom
+ * 'instanceof' handler that works correctly across execution contexts.
+ *
+ * ### Details ###
+ *
  * According to the EcmaScript's spec, the default behavior of JavaScript's `x instanceof Y` operator is to walk up the
  * prototype chain of object 'x', checking if any constructor in that hierarchy is _exactly the same object_ as the
  * constructor function 'Y'.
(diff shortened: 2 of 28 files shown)


packages/common/src/type-helpers.ts
```diff
@@ -73,3 +86,45 @@ export function errorCode(error: unknown): string | undefined {
...
+ *
+ * In practice, this tends to cause issues when crossing the workflow-sandboxing boundary (since Node's vm module
+ * really creates new execution realms), as well as when running tests using Jest (see https://github.com/jestjs/jest/issues/2549
+ * for some details on that one).
+ *
+ * This function injects a custom 'instanceof' handler into the prototype of 'clazz', which is both cross-realm safe and
+ * cross-copies-of-the-same-lib safe. It works by adding a special symbol property to the prototype of 'clazz', and then
+ * checking for the presence of that symbol.
+ */
+export function makeCrossRealmSafeInstanceOf(clazz: { prototype: any }, markerName: string): void {
```
bergundy, line 109: What happens if a user has both an older SDK version and 1.8.2?
Will instanceof checks still work or do we rely on the symbol?
Do we hit the `else` clause in line 118?

bergundy: Also, this could have been a class decorator.

packages/create-project/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/create",
-  "version": "1.8.0",
+  "version": "1.8.1",
```
bergundy, line 3: Why is this change showing up here?

packages/test/src/test-type-helpers.ts
```diff
@@ -0,0 +1,143 @@
+import vm from 'vm';
```
bergundy, line 1: Can you test that if you define the same class twice in the same context with the same symbol, instanceof works?

packages/common/src/type-helpers.ts
```diff
@@ -73,3 +86,45 @@ export function errorCode(error: unknown): string | undefined {
...
+      } else {
+        // 'this' must be a _subclass_ of clazz that doesn't redefined [Symbol.hasInstance], so that it inherited
+        // from clazz's [Symbol.hasInstance]. If we don't handle this particular situation, then
+        // `x instanceof SubclassOfParent` would return true for any instance of 'Parent', which is clearly wrong.
+        //
+        // Ideally, it'd be preferable to avoid this case entirely, by making sure that all subclasses of 'clazz'
+        // redefine [Symbol.hasInstance], but we can't enforce that. We therefore fallback to the default instanceof
+        // behavior (which is NOT cross-realm safe).
+        return this.prototype.isPrototypeOf(error); // eslint-disable-line no-prototype-builtins
+      }
```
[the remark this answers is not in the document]
mjameswh: If that was only about our own code, I'd agree.

However, if users themselves subclass one of our class (some type of specific `ApplicationFailure` for example, though we strongly discourage doing that), then I do not expect them to make it cross realm safe. The helper function is not even exposed publicly, and I don't think we should expect users to understand low details of these issues. In that case, making `instanceof` fails systematically would likely cause more damage than it would resolve.

An alternative I have been considering, however, would be to use a simple constructor’s name comparision in those cases. This would save users from a low probability of false negative, but introduces an even lower probability of false positive. In the end, I think however it might turns out being more predictable for users.

packages/common/src/type-helpers.ts
mjameswh: > What happens if a user has both an older SDK version and 1.8.2?
> Will instanceof checks still work or do we rely on the symbol?
> Do we hit the else clause in line 118?

Let say we have `theObject instanceof TheClass`. The `instanceof` handler is determined solely by the right hand side argument (ie. `TheClass`).

If `TheClass`'s module is <1.8.2, then the default `instanceof` algorithm will be used, no matter the version of the module from which `theObject` was instantiated.
- If `theObject` was instantiated using exactly `TheClass` constructor function object (ie. the module has the same file location and the same execution context), then `instanceof` will work as expected.
- In all other cases, the `instanceof` check is guaranteed to return false.
- That is the behavior we know from pre-1.8.0 releases.

If `TheClass`'s module is 1.8.2+, then, the symbol based check will be used, no matter the version of the module with which `theObject` was instantiated. Consequently:
- If `theObject` was instantiated using a pre-1.8.0 module, then the symbol based check will obviously fail; anyway `instanceof` would also fail in that situation, so no regression here.

  AFAICT the only possible regression that could results from this are the cases where we were previously checking an error's `name` rather than using the `instanceof` operator (for example, the worker shutdown throwable). It would be very easy to modify `makeCrossRealmSafeInstanceOf` to also allow name based comparisions, but I'm afraid that would only make things less predictable, as it would work when `TheClass`'s come from 1.8.2+ and `theObject` from any version, but wouldn't work when `theObject` is 1.8.2+, but `TheClass` is older.
- If `theObject` was instantiated using 1.8.0 or 1.8.1, then I expect that the symbol based `instanceof` check should work on classes that were marked with the symbol in 1.8.0; it would however fail on classes that were not marked in 1.8.0 but are now marked in 1.8.2.

  Once again, these would fail on the default `instanceof` algorithm anyway, so would not be a regression. Note however that the opposite (`TheClass` is 1.8.0, and `theObject` is 1.8.2+) wouldn't work. Therefore, mixing 1.8.0/1 with 1.8.2+ modules should definitively be avoided.
- If `theObject` was instantiated using 1.8.2+, then the symbol based `instanceof` handler should work correctly, even if `TheClass` and `theObject` are from different package versions.

> Do we hit the else clause in line 118?

In practice, the `else` clause should only ever be reached if a user extends one of our own class, on which we registered the symbol-based `instanceof`, then use that custom class as the right hand side argument of `instanceof`.

For example, lets imagine that the user does this:

```
 class MyCustomSubclassOfTemporalFailure extends TemporalFailure {
    // ...
 }
 const theObject = new MyCustomSubclassOfTemporalFailure();

 const isInstanceOf = (theObject instanceof MyCustomSubclassOfTemporalFailure);
```

Here, `MyCustomSubclassOfTemporalFailure`'s inherits the `@@hasInstance` function from `TemporalFailure`. Therefore, `instanceof` will implicitly call `TemporalFailure`'s `hasInstance`'s function, as `MyCustomSubclassOfTemporalFailure[@@hasInstance](theObject)`. That means that `this` will be equal to `MyCustomSubclassOfTemporalFailure`, but `clazz` will be `TemporalFailure`. We therefore fall into the `else` clause, and revert to the default behavior of `instanceof`, which works only in same-context cases.

mjameswh: > Also, this could have been a class decorator.

Yeah, I was also considering that option, but was delaying because I want to make sure that it is safe to use in Node.js 14.18 to 20+ with the current version of TS. Digging into that right now.

packages/common/src/type-helpers.ts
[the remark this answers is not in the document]
antlai-temporal: It makes sense... And for our code, is there are a way we could make it more robust, e.g., throwing in debug mode during testing, so that we can make sure we never miss a subclass...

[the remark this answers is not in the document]
mjameswh: I'm running out of time, so I'll leave this the way it is for now, but I was indeed thinking about introducing some "paranoid validation" to be enabled during tests to assert that we are not missing any context-related issues. That would probably not be in this `else` case however (or _not only_ in there), as it would be very likely that we miss some other objects. Might try hook [around calls to the sandbox](https://github.com/mjameswh/sdk-typescript/blob/aa737e22937c3b258391a47c8a848404c422957b/packages/worker/src/workflow/vm.ts#L47-L61).

packages/client/src/schedule-client.ts
```diff
@@ -526,6 +527,7 @@ export class ScheduleClient extends BaseClient {
  *
  * @experimental
  */
+@symbolBasedInstanceOf('ScheduleAlreadyExistsError')
```
mjameswh, line 530: Wrong name here

bergundy (approved):

airhorns:
It'd be really great to get this released! Just spent a lot of time debugging this very same issue from a bunch of errant thrown `ShutdownError`s from our jest suites and would love to make it go away 💖

jgnieuwhof:
FYI - we worked around this by using the following package:

https://www.npmjs.com/package/jest-environment-node-single-context

mjameswh:
@airhorns Indeed. That's coming very soon.

Issue #1169, opened 2023-07-13
[Bug] ProtobufJSONPayloadConverter needs Buffer to encode/decode binary protobuf fields


antlai-temporal:
### Describe the bug
The workflow context does not have access to the node.js Buffer class, but the 
`proto3-json-serializer` library, used by the ProtobufJSONPayloadConverter, needs it to
convert binary fields to strings with base64 encoding.

The result is that workflows cannot encode/decode protobufs that contain binary data by using this converter. 

### Minimal Reproduction

 See features repo (`features/data_converter/json_protobuf/feature.ts`)

### Environment/Versions
All

PR #1170, opened 2023-07-13
Fix ProtobufJSONPayloadConverter to not require Buffer in the workflo…

antlai-temporal:
## What was changed

This is a fix to #1169 

The ProtobufJsonPayloadConverter ensures that Buffer is globally available, and if it is not, it just adds it when needed, but leaving the global space untouched after that. 

It also needs to ensure that no Buffer objects are leaked, replacing them with Uint8Array objects. 

## Why?

Protobufs containing binary fields cannot be properly encoded/decoded with this converter without the Buffer class, and this class is not available by default in the workflow context.

## Checklist

1. Closes 
#1169 

2. How was this tested:

Added a system test

### DIFF
--- a/packages/common/src/converter/protobuf-payload-converters.ts
+++ b/packages/common/src/converter/protobuf-payload-converters.ts
@@ -121,17 +121,57 @@ export class ProtobufJsonPayloadConverter extends ProtobufPayloadConverter {
       return undefined;
     }
 
-    const jsonValue = protoJsonSerializer.toProto3JSON(value);
+    const hasBufferChanged = setBufferInGlobal();
+    try {
+      const jsonValue = protoJsonSerializer.toProto3JSON(value);
 
-    return this.constructPayload({
-      messageTypeName: getNamespacedTypeName(value.$type),
-      message: encode(JSON.stringify(jsonValue)),
-    });
+      return this.constructPayload({
+        messageTypeName: getNamespacedTypeName(value.$type),
+        message: encode(JSON.stringify(jsonValue)),
+      });
+    } finally {
+      resetBufferInGlobal(hasBufferChanged);
+    }
   }
 
   public fromPayload<T>(content: Payload): T {
-    const { messageType, data } = this.validatePayload(content);
-    return protoJsonSerializer.fromProto3JSON(messageType, JSON.parse(decode(data))) as unknown as T;
+    const hasBufferChanged = setBufferInGlobal();
+    try {
+      const { messageType, data } = this.validatePayload(content);
+      const res = protoJsonSerializer.fromProto3JSON(messageType, JSON.parse(decode(data))) as unknown as T;
+      replaceBuffers(res);
+      return res;
+    } finally {
+      resetBufferInGlobal(hasBufferChanged);
+    }
+  }
+}
+
+function replaceBuffers<X>(obj: X) {
+  if (obj != null && typeof obj === 'object') {
+    for (const [key, value] of Object.entries(obj)) {
+      if (Buffer.isBuffer(value)) {
+        type T = keyof typeof obj;
+        // Need to copy. `Buffer` manages a pool slab, internally reused when Buffer objects are GC.
+        obj[key as T] = new Uint8Array(value) as any;
+      } else {
+        replaceBuffers(value);
+      }
+    }
+  }
+}
+
+function setBufferInGlobal(): boolean {
+  if (typeof globalThis.Buffer === 'undefined') {
+    globalThis.Buffer = globalThis.constructor.constructor('return globalThis.Buffer')();
+    return true;
+  }
+  return false;
+}
+
+function resetBufferInGlobal(hasChanged: boolean): void {
+  if (hasChanged) {
+    delete (globalThis as any).Buffer;
   }
 }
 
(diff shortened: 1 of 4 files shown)


packages/common/src/converter/protobuf-payload-converters.ts
```diff
@@ -121,17 +121,57 @@ export class ProtobufJsonPayloadConverter extends ProtobufPayloadConverter {
...
+        obj[key as T] = new Uint8Array(value) as any;
+      } else {
+        replaceBuffers(value);
+      }
+    }
+  }
+}
+
+function setBufferInGlobal(): boolean {
+  if (typeof globalThis.Buffer === 'undefined') {
```
lorensr, line 165: A. Thanks for fixing this bug of mine 😄
B. Did you consider adding Buffer to global in [overrideGlobals](https://github.com/temporalio/sdk-typescript/blob/0a792088e13930b9dd4f0fcbebf24fc0b6cc6b74/packages/workflow/src/worker-interface.ts#L24) and leave it there?

antlai-temporal: We had a long discussion on whether to add Buffer to globals or not, and for now the decision is not to do it because it can make it more difficult to port workers to non-nodejs environments...

lorensr: I see, thanks!

packages/common/src/converter/protobuf-payload-converters.ts
```diff
@@ -121,17 +121,57 @@ export class ProtobufJsonPayloadConverter extends ProtobufPayloadConverter {
...
+      const res = protoJsonSerializer.fromProto3JSON(messageType, JSON.parse(decode(data))) as unknown as T;
+      replaceBuffers(res);
+      return res;
+    } finally {
+      resetBufferInGlobal(hasBufferChanged);
+    }
+  }
+}
+
+function replaceBuffers<X>(obj: X) {
```
bergundy, line 150: It's unfortunate that we have to do this. And BTW, we will need to iterate over arrays too, not just objects.

I looked at the source code for the proto json library and it seems like they only use the Buffer class for base64 conversion.
Could you please submit an issue to request that https://github.com/googleapis/proto3-json-serializer-nodejs be made browser compatible and not depend on `Buffer`? We could suggest that they use https://www.npmjs.com/package/base64-js for base64 conversion, which is used by the Buffer polyfill package.
This would be the ideal solution IMHO, we could of course also do the work and submit a PR to the proto3-json-serializer library.

In the meantime, let's keep this unfortunate work around.

antlai-temporal: This should already iterate over arrays. Their type is 'object', entries() gives you the array index, and you can index with a string an array... But performance for large arrays will be horrible, so let me fix that...

As we discussed, the protojson repo looks like it is in maintenance mode, so I have not much hope, but it is worth a try!

I think that long term we will have to support a set of nodejs-agnostic polyfills for user code, and rely less on third-party dependencies for the core, e.g., fork proto3-json...

antlai-temporal: I just filed an issue googleapis/proto3-json-serializer-nodejs#74 about direct browser support, will see...

packages/common/src/converter/protobuf-payload-converters.ts
```diff
@@ -121,17 +121,66 @@ export class ProtobufJsonPayloadConverter extends ProtobufPayloadConverter {
...
   }
 
   public fromPayload<T>(content: Payload): T {
-    const { messageType, data } = this.validatePayload(content);
-    return protoJsonSerializer.fromProto3JSON(messageType, JSON.parse(decode(data))) as unknown as T;
+    const hasBufferChanged = setBufferInGlobal();
+    try {
+      const { messageType, data } = this.validatePayload(content);
+      const res = protoJsonSerializer.fromProto3JSON(messageType, JSON.parse(decode(data))) as unknown as T;
+      replaceBuffers(res);
```
mjameswh, line 142: Is there any risk here that `res` itself might be an instance of Buffer? Looks like `replaceBuffers` never check the root object itself.

packages/common/src/converter/protobuf-payload-converters.ts
```diff
@@ -121,17 +121,66 @@ export class ProtobufJsonPayloadConverter extends ProtobufPayloadConverter {
...
+      for (const [key, value] of Object.entries(obj)) {
+        replaceBuffersImpl(value, key, obj);
+      }
+    }
+  }
+}
+
+function setBufferInGlobal(): boolean {
+  if (typeof globalThis.Buffer === 'undefined') {
+    globalThis.Buffer = globalThis.constructor.constructor('return globalThis.Buffer')();
```
mjameswh, line 175: Not sure what's the exact cost of these eval-style calls, but we should probably avoid doing these _repeatedly_.

Please either cache the right hand side of this line to a module level const, or pass it from `overrideGlobals` under some internal object (eg. `globalThis.__temporal_outter_globals.Buffer` or something like that).

mjameswh (approved):
Just two minor questions/suggestions. LGTM overall.

packages/common/src/converter/protobuf-payload-converters.ts
antlai-temporal: Adding a top level blob without metadata is a really bad idea, but you are right, we cannot control what the app will do... Fixed it.

packages/common/src/converter/protobuf-payload-converters.ts
antlai-temporal: Done

PR #1171, opened 2023-07-13
Fix argument name in executeChild signature

dandavison:
## What was changed
An argument name in a function type signature.

## Why?
The semantics implied by the name were incorrect. 

## Checklist

2. How was this tested:
`npm run test`

4. Any docs updates needed?
I believe that the API reference docs at https://typescript.temporal.io/api/namespaces/workflow#executechild will update automatically although I haven't researched exactly how that happens.

I believe that this change should not have any backwards compatibility problems, although it is technically possible that something could be introspecting on the argument name.

### DIFF
--- a/packages/workflow/src/workflow.ts
+++ b/packages/workflow/src/workflow.ts
@@ -774,7 +774,7 @@ export async function executeChild<T extends Workflow>(
  * @return The result of the child Workflow.
  */
 export async function executeChild<T extends Workflow>(
-  workflowType: T,
+  workflowFunc: T,
   options: WithWorkflowArgs<T, ChildWorkflowOptions>
 ): Promise<WorkflowResultType<T>>;
 


packages/workflow/src/workflow.ts
```diff
@@ -774,7 +774,7 @@ export async function executeChild<T extends Workflow>(
  * @return The result of the child Workflow.
  */
 export async function executeChild<T extends Workflow>(
-  workflowType: T,
+  workflowFunc: T,
   options: WithWorkflowArgs<T, ChildWorkflowOptions>
 ): Promise<WorkflowResultType<T>>;
```
dandavison, line 779: cf the corresponding `startChild` signature above:

```typescript
/**
 * Start a child Workflow execution
 *
 * - Returns a client-side handle that implements a child Workflow interface.
 * - Deduces the Workflow type and signature from provided Workflow function.
 * - By default, a child will be scheduled on the same task queue as its parent.
 *
 * A child Workflow handle supports awaiting completion, signaling and cancellation via {@link CancellationScope}s.
 * In order to query the child, use a {@link WorkflowClient} from an Activity.
 */
export async function startChild<T extends Workflow>(
  workflowFunc: T,
  options: WithWorkflowArgs<T, ChildWorkflowOptions>
): Promise<ChildWorkflowHandle<T>>;
```

Sushisource (approved):

dandavison:
Thanks @Sushisource.

PR #1181, opened 2023-07-27
fix(workflow): Fix mutliple sinks oddities

mjameswh:
## What changed

- FIx incorrect values of `wokrflowInfo.historyLength` and `workflowInfo.unsafe.isReplaying` as reported in out-of-sandbox log messages and as argument of sink function implementations.
- Sink functions configured with `callDuringReplay = false` may no longer be invoked from a replay-only worker (ie. `Worker.runReplayHistories()`); it  could previously get called in some cases on the very last WFT.
- Sink functions are now called serially, rather than in parallel. This ensure correct ordering of sink calls, even when sink functions execute asynchronously. This may however increase total WFT execution time, with a very minor risk of causing WFT Timeout.

### DIFF
--- a/packages/test/src/test-sinks.ts
+++ b/packages/test/src/test-sinks.ts
@@ -146,13 +146,17 @@ if (RUN_INTEGRATION_TESTS) {
   });
 
   test('Sink functions are not called during replay if callDuringReplay is unset', async (t) => {
-    const recordedMessages = Array<string>();
+    const recordedMessages = Array<{ message: string; hl: number; isReplaying: boolean }>();
     const taskQueue = `${__filename}-${t.title}`;
     const sinks: InjectedSinks<workflows.LoggerSinks> = {
       logger: {
         info: {
-          async fn(_info, message) {
-            recordedMessages.push(message);
+          async fn(info, message) {
+            recordedMessages.push({
+              message,
+              hl: info.historyLength,
+              isReplaying: info.unsafe.isReplaying,
+            });
           },
         },
       },
@@ -178,19 +182,31 @@ if (RUN_INTEGRATION_TESTS) {
     ]);
 
     t.deepEqual(recordedMessages, [
-      'Workflow execution started, replaying: false, hl: 3',
-      'Workflow execution completed, replaying: false, hl: 8',
+      {
+        message: 'Workflow execution started, replaying: false, hl: 3',
+        hl: 3,
+        isReplaying: false,
+      },
+      {
+        message: 'Workflow execution completed, replaying: false, hl: 8',
+        hl: 8,
+        isReplaying: false,
+      },
     ]);
   });
 
   test('Sink functions are called during replay if callDuringReplay is set', async (t) => {
-    const recordedMessages = Array<string>();
+    const recordedMessages = Array<{ message: string; hl: number; isReplaying: boolean }>();
     const taskQueue = `${__filename}-${t.title}`;
     const sinks: InjectedSinks<workflows.LoggerSinks> = {
       logger: {
         info: {
-          async fn(_info, message) {
-            recordedMessages.push(message);
+          async fn(info, message) {
+            recordedMessages.push({
+              message,
+              hl: info.historyLength,
+              isReplaying: info.unsafe.isReplaying,
+            });
           },
           callDuringReplay: true,
         },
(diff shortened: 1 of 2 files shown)


packages/test/src/test-sinks.ts
```diff
@@ -146,13 +146,17 @@ if (RUN_INTEGRATION_TESTS) {
   });
 
   test('Sink functions are not called during replay if callDuringReplay is unset', async (t) => {
-    const recordedMessages = Array<string>();
+    const recordedMessages = Array<{ message: string; hl: number; isReplaying: boolean }>();
```
bergundy, line 149: Make it more readable

```suggestion
    const recordedMessages = Array<{ message: string; historyLength: number; isReplaying: boolean }>();
```

packages/worker/src/worker.ts
```diff
@@ -1245,8 +1251,7 @@ export class Worker {
                     // Fatal error means we cannot call into this workflow again unfortunately
                     if (!isFatalError) {
                       const externalCalls = await state.workflow.getAndResetSinkCalls();
-                      // TODO: state.info.searchAttributes are not updated here
```
bergundy, line 1248: This once only applied to search attributes (hence the TODO) but we've added more mutable properties to info since.
There should be a better way to ensure that `info` doesn't go stale.
Maybe we should get it from workflow context when extracting the sink calls, that'll give us search attributes too.

packages/worker/src/worker.ts
```diff
@@ -1297,30 +1302,33 @@ export class Worker {
...
    */
-  protected async processSinkCalls(externalCalls: SinkCall[], info: WorkflowInfo, isReplaying: boolean): Promise<void> {
+  protected async processSinkCalls(externalCalls: SinkCall[], workflowInfo: WorkflowInfo): Promise<void> {
+    // When processing workflows through runReplayHistories, Core may still send non-replay activations on the very
+    // last Workflow Task in some cases. Though Core is technically exact here, the fact that sinks marked with
+    // callDuringReplay = false may get called on a replay worker is definitely a surprising behavior. For that
+    // reason, we extend the isReplaying flag in this case to also include anything running under in a replay worker.
+    const isReplaying = workflowInfo.unsafe.isReplaying || this.isReplayWorker;
     const { sinks } = this.options;
-    await Promise.all(
```
bergundy, line 1302: Not convinced we should be calling these sequentially, if a user has many sink calls that do IO this will delay the completion of WFTs.
We should also consider failing activations on failed sink calls, users may be relying on sinks for reporting.

packages/worker/src/worker.ts
```diff
@@ -1297,30 +1302,33 @@ export class Worker {
    * This function does not throw, it will log in case of missing sinks
    * or failed sink function invocations.
    */
-  protected async processSinkCalls(externalCalls: SinkCall[], info: WorkflowInfo, isReplaying: boolean): Promise<void> {
+  protected async processSinkCalls(externalCalls: SinkCall[], workflowInfo: WorkflowInfo): Promise<void> {
+    // When processing workflows through runReplayHistories, Core may still send non-replay activations on the very
+    // last Workflow Task in some cases. Though Core is technically exact here, the fact that sinks marked with
+    // callDuringReplay = false may get called on a replay worker is definitely a surprising behavior. For that
+    // reason, we extend the isReplaying flag in this case to also include anything running under in a replay worker.
+    const isReplaying = workflowInfo.unsafe.isReplaying || this.isReplayWorker;
```
bergundy, line 1310: Let's leave this for now and think about a long term solution in Core.

packages/worker/src/worker.ts
```diff
@@ -1297,30 +1302,33 @@ export class Worker {
...
     const { sinks } = this.options;
-    await Promise.all(
-      externalCalls.map(async ({ ifaceName, fnName, args }) => {
-        const dep = sinks?.[ifaceName]?.[fnName];
-        if (dep === undefined) {
-          this.log.error('Workflow referenced an unregistered external sink', {
+    for (const { ifaceName, fnName, args } of externalCalls) {
```
bergundy, line 1312: I would to make serial processing configurable in the sink options (next to callDuringReplay) and default to concurrent.

packages/worker/src/worker.ts
mjameswh: Yeah, that's what I had in mind. My initial take at this was to add the `workflowInfo` object from the sandbox context into `SinkCall` objects.

Reverted that however, as this is not only about sink calls. We're also exposing the worker side `workflowInfo` object on log messages, and eventually, we would make it available to payload encoders.

I'm now considering sending `workflowInfo` on activation completion instead.

mjameswh:
Pushed some major changes. Open to discussion on that.

Still other things to fix however + need to add tests.

packages/test/src/test-sinks.ts
```diff
@@ -15,6 +21,21 @@ class DependencyError extends Error {
...
+  fn: (info: WorkflowInfo, message: string, attrs?: Record<string, unknown>) => Promise<void>,
+  opts?: Omit<InjectedSinkFunction<any>, 'fn'>
+): InjectedSinks<DefaultLoggerSinks> {
+  return {
+    defaultWorkerLogger: {
+      trace: { fn, ...(opts ?? {}) },
+      debug: { fn, ...(opts ?? {}) },
+      info: { fn, ...(opts ?? {}) },
+      warn: { fn, ...(opts ?? {}) },
+      error: { fn, ...(opts ?? {}) },
```
bergundy, line 34: Just `...opts` works, you can splat an undefined | null variable

packages/test/src/workflows/log-before-timing-out.ts
```diff
@@ -1,10 +1,9 @@
-import { proxySinks } from '@temporalio/workflow';
-import { LoggerSinks } from './definitions';
+import { LoggerSinks, proxySinks } from '@temporalio/workflow';
 
-const { logger } = proxySinks<LoggerSinks>();
+const { defaultWorkerLogger } = proxySinks<LoggerSinks>();
 
 export async function logAndTimeout(): Promise<void> {
-  logger.info('logging before getting stuck');
+  defaultWorkerLogger.info('logging before getting stuck');
```
bergundy, line 6: Why not use `workflow.log`? Don't want the extra context?

mjameswh: I did exactly that at some point, but ended reverting that back.

The thing that was bugging me was that log entries captured by the default logger included default lifecycle log messages, emmited from the interceptor. Added some logic to filter them out, but that only made the tests even more complex. Concluded that using a custom sink for that just the simplest approach for that specific case. Plus, having tests for _custom_ sinks is desirable.

packages/test/src/test-sinks.ts
mjameswh: No, really?!? I always thought the spread operator was not valid on null/undefined values. Turns out that's only true for array spread, not for object spread. TIL!

packages/test/src/test-workflows.ts
```diff
@@ -1549,7 +1551,11 @@ test('logAndTimeout', async (t) => {
     message: 'Script execution timed out after 200ms',
   });
   const calls = await workflow.getAndResetSinkCalls();
-  delete calls[0].args[1][LogTimestamp];
+  // Ignore LogTimestamp and workflowInfo for the purpose of this compare
```
bergundy, line 1554:
```suggestion
  // Ignore LogTimestamp and workflowInfo for the purpose of this comparison
```

packages/worker/src/worker.ts
```diff
@@ -1297,30 +1305,57 @@ export class Worker {
...
-            ifaceName,
-            fnName,
+
+    const filteredCalls = externalCalls
+      // Map sink call to the corresponding sink function definition
+      .map((call) => ({ call, sink: sinks?.[call.ifaceName]?.[call.fnName] }))
+      // Reject calls to undefined sink definitions
+      .filter(({ call: { ifaceName, fnName }, sink }) => {
+        if (sink !== undefined) return true;
+        this.log.error('Workflow referenced an unregistered external sink', {
```
bergundy, line 1317: Might want to give a hint here if the sink is the `defaultWorkerLogger` how to fix the situation.

packages/worker/src/worker.ts
```diff
@@ -1297,30 +1305,57 @@ export class Worker {
...
-          }
         }
-      })
-    );
+      },
+      sink: sink as InjectedSinkFunction<any>,
+    }));
+
+    const pendingPromises: Promise<void>[] = [];
+    for (const { sink, call } of mappedCalls) {
```
bergundy, line 1344: You still want different sink functions to run concurrently, I would group the sink calls by `ifaceName, fnName` and run each group concurrently.

packages/worker/src/worker.ts
```diff
@@ -1297,30 +1305,57 @@ export class Worker {
...
+      if (sink.callSerially) {
+        if (pendingPromises.length > 0) {
+          await Promise.allSettled(await pendingPromises);
```
bergundy, line 1347: Can just use `Promise.all`, we don't expect any of the promises to throw.

packages/worker/src/sinks.ts
```diff
@@ -17,18 +17,76 @@
...
+   * More formally, any pending executions of sink functions prior to this one will be awaited for
+   * completion before this one is executed, and completion of this sink function call will be
+   * awaited for completion before any subsequent sink function call is executed.
+   *
+   * Note that serializing sink function calls may increase total execution time, and therefore
+   * increase the risk of Workflow Task Timeout.
+   *
+   * Defaults to `false`.
+   */
+  callSerially?: boolean;
```
bergundy, line 89: Instead of a boolean make it a `concurrency: SinkConcurrency` enum, that we can (in the future) choose to run all sinks for an interface serially vs. sinks of the same function.

bergundy: Also would be nice to allow failing a WFT on sink failure, but that can also wait for another PR.

packages/worker/src/worker.ts
mjameswh: Grouping on both would lose relative ordering. Doesn't seem desirable. I'd be ok grouping in iface name instead. Can move the concurrency node to this level.

packages/worker/src/worker.ts
```diff
@@ -1297,27 +1304,36 @@ export class Worker {
...
-            fnName,
+      filteredCalls.map(async ({ call, sink }) => {
+        try {
+          await sink?.fn(call.workflowInfo, ...call.args);
+        } catch (error) {
+          this.log.error('External sink function threw an error', {
+            ifaceName: call.ifaceName,
+            fnName: call.fnName,
+            error,
+            workflowInfo: call.workflowInfo,
```
bergundy, line 1335: Are you passing around workflow info for each call because it can be mutated during an activation? (e.g. search attributes)

packages/workflow/src/workflow.ts
```diff
@@ -911,11 +911,19 @@ export function proxySinks<T extends Sinks>(): T {
...
                   ifaceName: ifaceName as string,
                   fnName: fnName as string,
                   // Only available from node 17.
                   args: (globalThis as any).structuredClone ? (globalThis as any).structuredClone(args) : args,
+                  // Clone the workflowInfo object so that any further mutations to it does not get reflected in sink
+                  workflowInfo: {
+                    ...info,
+                    // Make sure to clone any sub-property that may get mutated during the lifespan of an activation
+                    searchAttributes: { ...info.searchAttributes },
+                    memo: info.memo ? { ...info.memo } : undefined,
```
bergundy, line 925: I wonder if we can ensure statically that any future nested properties also get copied.
You might want to use `structuredClone` if available instead of manually cloning.

bergundy (approved):
Nice, left a couple of comments, feel free to address and merge.

packages/workflow/src/workflow.ts
mjameswh: That's certainly a possibility, though I'd only consider this if we polyfill `structuredClone` on Node 14 and 16. I really want to avoid leaving that kind of bugs.

PR #1203, opened 2023-08-16
fix: Remove accidental import of large protos into workflow library

bergundy:
The import was added between 1.8.0 and 1.8.1.

Reduces the SDK test workflow bundle from 2.55MB to 1.04MB.


bergundy:
Eventually we should add some assertion on bundle size in our tests.

mjameswh (approved):

Issue #695, opened 2022-06-13
[Feature Request] Add `WorkflowInfo.historySizeInBytes`


lorensr:
Blocked by: server adding history size

lorensr:
Waiting on: https://github.com/temporalio/sdk-core/issues/564

lorensr:
Ready ☺️

cretz:
Don't forget `isContinueAsNewSuggested`

PR #1223, opened 2023-08-24
feat(workflow): Add historySizeInBytes and continueAsNewSuggested to WorkflowInfo (#695)

antlai-temporal:
## What was changed

Makes visible to workflows the new info fields historySizeInBytes and continueAsNewSuggested

## Why?
Makes it easier for workflow code to decide when to continueAsNew() before history becomes  too large

## Checklist

1. Closes 
#695 
2. How was this tested:
A new system test and unit test added

3. Any docs updates needed?

jsdoc of the new fields


packages/workflow/src/interfaces.ts
```diff
@@ -70,6 +70,23 @@ export interface WorkflowInfo {
...
+   * Size of Workflow history in bytes until the current Workflow Task.
+   *
+   * This value changes during the lifetime of an Execution.
+   *
+   * You may safely use this information to decide when to {@link continueAsNew}.
+   */
+  historySizeBytes: number;
+
+  /**
+   * A hint provided by the last Workflow Task Started event recommending whether to
```
lorensr, line 83:
```suggestion
   * A hint provided by the current WorkflowTaskStarted event recommending whether to
```

packages/test/src/test-workflows.ts
```diff
@@ -353,6 +356,14 @@ test('successString', async (t) => {
   compareCompletion(t, req, makeSuccess([makeCompleteWorkflowExecution(defaultPayloadConverter.toPayload('success'))]));
 });
 
+test('continueAsNewSuggested', async (t) => {
```
bergundy, line 359: Can you get an integration test for this please instead of this unit test?
I'm considering removing (most of) these unit tests.

mjameswh: @bergundy I suggested a UT on that one because we can't make assumptions on when will the server reports `suggestContinueAsNew = true`. Sure, we could run for a long time and expect that, at some point, the flag will turn from false to true, but that seems more suitable for the feature test repo.

Sushisource: +1. Unit tests are very useful, specifically in situations like this which James gets at.

packages/workflow/src/worker-interface.ts
```diff
@@ -189,8 +189,18 @@ export function activate(activation: coresdk.workflow_activation.WorkflowActivat
       if (activation.historyLength == null) {
         throw new TypeError('Got activation with no historyLength');
       }
+      if (activation.historySizeBytes == null) {
```
mjameswh, line 192: @Sushisource What will we get at this point in `activation.historySizeBytes` and `activation.continueAsNewSuggested` if we process a WFT dating back from before these fields were added into server? Is it safe to assume the fact core will have filled these props with non-null values?

packages/client/src/types.ts
```diff
@@ -36,6 +36,7 @@ export interface WorkflowExecutionInfo {
   taskQueue: string;
   status: { code: proto.temporal.api.enums.v1.WorkflowExecutionStatus; name: WorkflowExecutionStatusName };
   historyLength: number;
+  historySizeBytes: number;
```
mjameswh, line 39: I believe this prop (and some others) will not be available in some situations, notably on info returned from visibility APIs (ie. list workflows) for workflows that haven't yet completed execution (there might be other cases, I'm not sure about the exact extent of that limitation, and I assume it may possibly vary depending on which visibility backend is used).

We should therefore make that prop optionnal, and it would be pertinent to add some comments to that effect.

Also, we should have at least one test validating this prop through workflow client.

bergundy: Good catch @mjameswh

packages/workflow/src/interfaces.ts
antlai-temporal: Much clear, thanks

packages/test/src/test-workflows.ts
antlai-temporal: We had a discussion about this. The system test is just a loop with Sleep(1) so that's easy... However, the current threshold is 4K activations, and against a cloud target that's probably 6-7 minutes just for that test. Also, this is a recent change, not sure if it has made it to the cli server yet. And then, the whole discussion on whether this is the right place to check that the server behaves... 

BTW, I don't think that `features` should become the system test repo either, I like the fact that is trying to be simple and focus on one "feature" at a time, rather than stressing  end-to-end with a combination of "features"...

bergundy: You could use a workflow that does a bunch of search attribute upserts.

bergundy: Integration tests are bound to test some of the server behavior. But without them you can't be sure that the feature actually works E2E.

packages/workflow/src/worker-interface.ts
antlai-temporal: I was tempted to remove the check, in sdk-core/core/src/worker/workflow/machines/workflow_machines.rs:250 is set to 0, and then it uses `unwrap_or_default()`, so it should not be null
 @Sushisource can you confirm this?

packages/client/src/types.ts
antlai-temporal: One thing that I found confusing is that this is a different "historySizeBytes" that is coming from WorkflowExecutionInfo and not from WorkflowTaskStarted. This is not an optional field, that it defaults to 0, why would it not be available?

packages/test/src/test-workflows.ts
antlai-temporal: So each upsert creates a different activation but we can batch many in one round-trip, is that what you mean?

I'm a big fan of E2E testing, the only issue is whether it belongs to this repo or not...

mjameswh: Fair enough, there's no use arguing this over and over. Let's just add an integration test. However, I would _also_ keep that unit test, because it has a clear, well defined semantic, which E2E test won't have on that specific property.

@antlai-temporal The current default thresholds for the server to suggest CAN is 4096 events or 4MB ([see here](https://github.com/temporalio/temporal/blob/548c1506bf831cf836a6232d86a118bc89477308/service/history/configs/config.go#L475C1-L478)). You can do multiple commands per WFT to grow history very quickly, eg.: if you start 512 timers concurrently of 1ms each, on the next WFT, history will have grown by ~1024 events (TimerStarted and TimerFired). History should reach the threshold very quickly. Now, don't assume that server will actually raise the flag at that exact value. Simply continue looping until either the flag turns true (success) or the history reaches some really unreasonable length (eg. > 40K events), at which point you can fail.

packages/client/src/types.ts
mjameswh: @antlai-temporal If you dont know about the Visibility API and Visibility Store yet, get a look [here](https://docs.temporal.io/clusters#visibility) and [here](https://docs.temporal.io/cluster-deployment-guide#visibility-store).

The visibility store is optimized for efficient listing/searching across a large number of workflows. For that reason, it is eventual consistent, and not all details of a workflow are stored there (in particular, details that would be too large, that would change too often, that would have little value in search/list use cases, and other reasons).

For example, look at the `History Length` and `History Size` columns in the following picture:

Screenshot 2023-08-25 at 3 54 26 PM

antlai-temporal: Thanks James, but even if the value is missing in the database, we are directly reading from a proto that has non-optional int64 fields, so they should default to zero, right? I guess that is the reason we did not see it with the historyLength. 
I was wondering why the GUI did not show zeros, maybe they don't use protobufs...

packages/test/src/test-workflows.ts
antlai-temporal: I added the system test that you described. 
I also added another one that checks historySizeBytes is visible in WorkflowExecutionInfo

packages/client/src/types.ts
mjameswh: Now, back to our original discussion... The problem with reporting history length and history size (and probably other fields too) as 0 is simply that these values are surprising for developers who are not aware of the fact that the visibility API only report these values on workflow executions that are no longer executing. We periodically get user questions on that subject.

A solution to this can be as simple as adding doc comments on fields in that interface (I think pretty much all of it can be copied from elsewhere), then add extra notes on these specific fields saying that these may not be available in some contexts. However, I don't like the idea of saying something similar to "this field may reports 0 when doing visibily api and if the workflow is still executing"; marking that field as optional seems like a more appropriate thing to do (and then, mention that "that information may not be available when ...".

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,54 @@ test('Condition 0 patch sets a timer', async (t) => {
...
+  await worker.runUntil(handle.result());
+  const historySizeBytes = (await handle.describe()).historySizeBytes;
+  t.assert(historySizeBytes > 100);
+});
+
+export async function suggestedCAN(): Promise<boolean> {
+  const MAX_EVENTS = 40_000;
+  const BATCH_SIZE = 100;
+  const EVENTS_PER_COMMAND = 2;
+  for (let i = 0; i < MAX_EVENTS / (BATCH_SIZE * EVENTS_PER_COMMAND); i++) {
```
mjameswh, line 247: You don't need to do the math. Just check `workflowInfo().historyLength < 40_000`. That will also get rid of the counter variable.

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,54 @@ test('Condition 0 patch sets a timer', async (t) => {
...
+    const batch: Promise<void>[] = [];
```
mjameswh, line 248: nit - I would personally go for this idiom:

```
await Promise.all(
  new Array(BATCH_SIZE).fill(undefined).map((_) => workflow.sleep(1))
)
```

packages/test/src/test-workflows.ts
```diff
@@ -111,6 +111,9 @@ async function createWorkflow(
       taskQueue: 'test',
       searchAttributes: {},
       historyLength: 3,
+      // Assuming 100 bytes per entry
```
mjameswh, line 114: That comment isn't helpful

packages/workflow/src/worker-interface.ts
Sushisource: Well since these come across to lang as protos, they're just gonna be zeroed out / false if there's no real value to read. Whether or not they are null depends on how the TS bridge converts Rust structures to TS. Nulls don't exist in Rust, of course, and these fields are not `Option`s, so I can't imagine why they'd ever get converted to null

mjameswh: Right... 🤔 Seems then like a few of the checks around here are totally useless. `historyLength`, `historySizeBytes` and `continueAsNewSuggested` can't possibly be null or undefined, as they are all primitives from in the protobuf definition, so they would anyway return their default value if missing. Same for `isReplaying` a few lines below.

packages/client/src/types.ts
antlai-temporal: The fact that we can explicitly set it to zero, and after de-serialization will ended up as missing is very confusing, I will do the change to explicitly set it to zero, and put a comment that zero means that the value may, or may not, have been set. 
I don't think we should talk about the visibility API in the SDK comments, the UI could show the missing fields as "pending", or add a comment pop-up. In any case, most users of the UI are not going to look at the comments in the code anyway...

packages/workflow/src/worker-interface.ts
antlai-temporal: Makes sense. The same for continueAsNewSuggested. As long as these values are going through core there will always have a default value. The only exception is unit tests that handcraft the activations... @mjameswh are you OK for me to remove them?

packages/client/src/types.ts
antlai-temporal: After reading @Sushisource I think I understand what is going on. Rust is *always* going to provide a default value, otherwise the program wouldn't compile since they are not Optional. The neon library just do the standard conversion in the bridge, so as long as they are coming from Core, they are always going to have a default value. That explains why the visibility API may see missing values but we don't...

@mjameswh can you confirm that the only path for this protos into the sdk-typescript is via core? If that's the case we should not make them optional, because they are not, and let other consumers make their own choices...

packages/test/src/test-integration-workflows.ts
antlai-temporal: Much cleaner, thanks

packages/test/src/test-integration-workflows.ts
antlai-temporal: Done

packages/test/src/test-sinks.ts
```diff
@@ -134,6 +137,8 @@ if (RUN_INTEGRATION_TESTS) {
       parent: undefined,
       searchAttributes: {},
       historyLength: 3,
+      historySizeBytes: expectedHistorySizeBytes,
```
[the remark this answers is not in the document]
antlai-temporal: Makes sense, done

packages/test/src/test-workflows.ts
antlai-temporal: removed

packages/client/src/types.ts
mjameswh: Oh, I see where you're coming from... It seems that you may be a bit confused about how things work in the `@temporalio/client` package vs how they work in `@temporal/worker` and `@temporalio/workflows`.

The rest of this PR is all about stuff happening on the Worker side and inside the Workflow context. However, in this specific GitHub comment, we are looking at file `packages/client/src/types.ts`. That is part of the _client_ API (eg. `new Client().workflow.start(...)`).

In TS, the client API isn't based on Core SDK (and just to make it clear: TS is the only SDK that uses different gRPC implementations for its client vs its worker). Instead, it uses a pure JS gRPC implementation. The main reason for this is that it allows using the TS client in contexts where native code would not be possible.

So absolutely no Rust implied here.

> I don't think we should talk about the visibility API in the SDK comments, the UI could show the missing fields as "pending", or add a comment pop-up. In any case, most users of the UI are not going to look at the comments in the code anyway...

Sorry, I should have gave you a better introduction regarding the visibility API.

I used a screenshot of our web UI only as a very quick demonstration of the fact that some of the properties returned by the visibility API are known to be unavailable in some cases. However, the visibility API is not only about our Web UI. It is an API meant to be used programmatically by Temporal app developers (and yes, people are indeed building useful and interesting stuff based on that API).

For example, one can do:

```
  // Find workflows executions that might have been affected by some prior bug
  for await (const workflowInfo of client.workflow.list({ query: 'WorkflowType="MySuperCoolWorkflow"' })) {
    if (workflowInfo.status === 'RUNNING' && workflowInfo.historyLength > 15_000) {
      console.log(`Need to fix workflow: ${workflowInfo.workflowId}/${workflowInfo.runId}`);
    }
  }
```

Now, that `list()` function returns an async iterator of `WorkflowExecutionInfo`, ie. the same interface as what would be returned by `workflowHandle.describe()`. Note that this interface is part of the `@temporalio/client` public API, and essentially meant to be used by users of the Client API. `@temporalio/worker` and others never use the `WorkflowExecutionInfo` interface, nor do they ever need to parse an instance of the `IWorkflowExecutionInfo` protobuf message.

So, a user iterating on `WorkflowExecutionInfo` objects returned by the `list()` function could very naturally expect that properties of that object contain actual values, especially if there is no doc and no language annotation indicating that that property might not match expectations. Take the snippet I gave previously. It would return absolutely no workflow. Does that seems easy to figure out for a regular Temporal developer?

packages/workflow/src/worker-interface.ts
mjameswh: > As long as these values are going through core there will always have a default value. 

To be exact, both Core _and_ Protobufjs are providing us guarantees here.

> are you OK for me to remove them?

Yeah. If we're sure they don't bring any actual value, then I'd prefer not having such checks. They led us in the wrong direction this time, and could very likely lead us in a wrong direction again in the future if we don't get rid of them.

For unit tests, we could simply reencode handcrafted activations objects into actual protobufjs objects. Something like this this (not sure what is the difference between `create` and `fromObject`):

```
  const act = coresdk.workflow_activation.WorkflowActivation.fromObject(activation);
  // ...
```

antlai-temporal: One catch with protobufjs is that they add the default values to the prototype and then Object#hasOwnProperty (and for of) does not work protobufjs/protobuf.js/#521 .  Is this going to be an issue for client/visibility API?

packages/client/src/types.ts
antlai-temporal: Thanks, this is a great explanation. I didn't know that the client API bypasses core, and also it was common for end users to write their own visibility api tools/scripts...This is much clearer now!

To summarize, for the client path protobufjs will give us a default value (zero) but it may be the wrong value for the visibility API because it is eventually consistent, and programmatic clients will break. Also, hasOwnProperty() on default values will not work, breaking JSON.stringify() and many other iterators...

So you are right, we need a comment explaining all this. 

Not sure what to do about declaring it optional in the SDK, because the problem is not that a value could be missing, protobufjs will take care of that, the problem is that the default value may be completely wrong... And we can't tell whether it was explicitly set or is a default because it is not declared with an "optional" field in the proto definition. And if we change the proto we will break most clients, and nobody else uses "optional" in protos APIs anyway...

On the other hand, declaring it optional in Typescript, and setting the default value explicitly, even if it is redundant, it could be much more clear for someone reading the code... Anyway, your choice...

bergundy: I feel like having unset fields is preferable than using zero values, and is consistent with the other SDK APIs and general best practices in JS/TS.
Please note that the TS interface for for `WorkflowExecutionInfo` defines every field as nullable:
```
                    /** WorkflowExecutionInfo historySizeBytes */
                    historySizeBytes?: (Long|null);
```

mjameswh: > To summarize, for the client path protobufjs will give us a default value (zero) but it may be the wrong value for the visibility API because it is eventually consistent

As it is now, a Workflow's `historyLength`, `historySizeBytes` and a few other attributes are only get sent to Elastic Search when Workflow Execution closes (see [here](https://github.com/temporalio/temporal/blob/8f11ba122fbbef537b07ca6b1823f143c9f36a0b/common/persistence/visibility/store/elasticsearch/visibility_store.go#L1012-L1020) and [here](https://github.com/temporalio/temporal/blob/8f11ba122fbbef537b07ca6b1823f143c9f36a0b/common/persistence/visibility/store/elasticsearch/visibility_store.go#L234-L238)). These values are not simply "lagging behind" as one would often expect for eventual consistency, they are simply not stored at all.

Another important thing: it is impossible for a workflow execution to actually have an history length of 0, or a size of 0. From the moment a workflow execution gets created, it already contains at least two events.

So lets take the `historySizeBytes` field as an example. There is only two possibilities:
- `protobufEncodedWorkflowExecutionInfo.historySizeBytes !== 0`: that value is valid (and presumably, it can no longer change, but let's not write this out, it case that might not be true for some other visibility engines);
- `protobufEncodedWorkflowExecutionInfo.historySizeBytes === 0`: that can only mean that the visibility store doesn't have that search attribute (presumably because the workflow execution has not completed yet, but actually, that's not even important). From a user's perspective, that is an undefined value.

> Also, hasOwnProperty() on default values will not work, breaking JSON.stringify() and many other iterators...

That's not an issue if they are using the high level client API (ie. `Client.workflow.list(...)`), as we don't return the protobuf objects to the user (well, we do expose the `raw`, as an escape hatch if they need something they can't get otherwise; but there's zero guarantee regarding that field, so no worry here).

packages/workflow/src/worker-interface.ts
mjameswh: I don't see any potential issue here. Activations are only meant for internal usage, and that change would only be for unit tests anyway.

packages/client/src/types.ts
mjameswh: @bergundy That's the interface of the protobuf `WorkflowExecutionInfo` class, but we also have an interface of the same name in `@temporalio/client`, which unfortunately define `historyLength` as a non-optional property.

Making it optional at this point would be a breaking change. My worry on that one is really about usages of `workflowHandle.describe()` (existing uses of `list()` are less of a concern here, since if they were using the `historyLength` field, then they were incorrect anyway, so forcing them to fix this is not a bad thing). Similarly, `historySizeBytes` shouldn't be optional on return of `workflowHandle.describe()`, though it should be on return of `list()`.

So I'd say the best thing would be to introduce a new `WorkflowExecutionSummary` interface, which extends `WorkflowExecutionInfo` but make some of its field optional, then make `list()` return an iterator of `WorkflowExecutionSummary` instead of `WorkflowExecutionInfo`. That would follow what was done in Schedules.

@bergundy Does that make sense to you?

bergundy: I think we can expect history length to always be present on `WorkflowExecutionInfo` so it's fine that it's not optional.
And I support this for any other field for which we have this level of certainty that it will always be present.

The fact that `list` returns an empty field is wrong so we'd need to at least set the default value there.

> So I'd say the best thing would be to introduce a new WorkflowExecutionSummary interface, which extends WorkflowExecutionInfo but make some of its field optional, then make list() return an iterator of WorkflowExecutionSummary instead of WorkflowExecutionInfo. That would follow what was done in Schedules.

This may be our best option, I just learned today that the parent execution isn't set on list result either, making this approach even more attractive.

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,49 @@ test('Condition 0 patch sets a timer', async (t) => {
...
+  await worker.runUntil(handle.result());
+  const historySizeBytes = (await handle.describe()).historySizeBytes;
+  t.assert(historySizeBytes > 100);
+});
+
+export async function suggestedCAN(): Promise<boolean> {
+  const MAX_EVENTS = 40_000;
+  const BATCH_SIZE = 100;
+  while (workflow.workflowInfo().historyLength < MAX_EVENTS) {
+    await Promise.all(new Array(BATCH_SIZE).fill(undefined).map((_) => workflow.sleep(1)));
```
bergundy, line 247: nit: I'd use an upsert search attributes instead, it's cheaper to run.

packages/client/src/types.ts
antlai-temporal: historySizeBytes behavior mimics historyLength, I was just using whatever it was decided for historyLength. @bergundy 
 Shall I also leave it as not optional? It probably makes sense just to be consistent...

And for adding WorkflowExecutionSummary, shall we do that with the PR that fixes the parent execution or here?

packages/test/src/test-integration-workflows.ts
antlai-temporal: It is actually pretty fast, and even against the cloud, it will succeed in just 20 round trips (the actual server settings  are 4K events, not 40K)

packages/workflow/src/interfaces.ts
```diff
@@ -70,6 +70,23 @@ export interface WorkflowInfo {
...
   historyLength: number;
 
+  /**
+   * Size of Workflow history in bytes until the current Workflow Task.
+   *
+   * This value changes during the lifetime of an Execution.
+   *
+   * You may safely use this information to decide when to {@link continueAsNew}.
+   */
+  historySizeBytes: number;
```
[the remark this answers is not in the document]
antlai-temporal: It is set by Rust (Core) to zero when the server does not provide it. I'll add a comment explaining this...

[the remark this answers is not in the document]
bergundy: I would prefer if we change 0 to unset to avoid this confusion.

packages/client/src/types.ts
bergundy: It depends on our confidence level for when this attribute will be non optional.

I think it was added a few releases ago, and if that's the case, let's keep it optional.

bergundy: > And for adding WorkflowExecutionSummary, shall we do that with the PR that fixes the parent execution or here?

I would wait on that for a followup PR. Don't want to unnecessarily block this one.

(apologize for the late comments)

packages/workflow/src/worker-interface.ts
antlai-temporal: Checks removed

packages/client/src/types.ts
antlai-temporal: Ok, historySizeBytes is now optional and has a comment about >1.20 server release needed. There is also a system test using describe()

packages/client/src/helpers.ts
```diff
@@ -59,6 +59,8 @@ export async function executionInfoFromRaw<T>(
     },
     // Safe to convert to number, max history length is 50k, which is much less than Number.MAX_SAFE_INTEGER
     historyLength: raw.historyLength!.toNumber(),
+    // Exact truncation for multi-petabyte histories
+    historySizeBytes: raw.historySizeBytes!.toNumber(),
```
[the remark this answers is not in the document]
antlai-temporal: Made optional

packages/workflow/src/interfaces.ts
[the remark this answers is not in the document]
antlai-temporal: Sure, I just changed it to optional in WorkflowInfo, and also convert zero to undefined.

packages/client/src/helpers.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line import/no-named-as-default
+import Long from 'long';
```
mjameswh, line 2: Please put the eslint directive comment on the same line as the `import`, like it's done in schedule-helpers.ts. Otherwise, these comments sometime get broken by tools that sort imports.

packages/test/src/integration-tests.ts
```diff
@@ -694,6 +695,9 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
       workflowType: 'returnWorkflowInfo',
       workflowId,
       historyLength: 3,
+      // historySizeBytes changes in every run, e.g., due to variable encoding of process id
+      historySizeBytes: result.historySizeBytes,
```
mjameswh, line 699: Same as in the other test, group fields that we intentionally ignore at the end

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,49 @@ test('Condition 0 patch sets a timer', async (t) => {
...
+  await workflow.sleep(1);
+  const after = workflow.workflowInfo().historySizeBytes;
+  return [before, after];
+}
+
+test('HistorySizeBytes grows with new WFT', async (t) => {
+  const { createWorker, executeWorkflow } = helpers(t);
+  const worker = await createWorker();
+  const [before, after] = await worker.runUntil(executeWorkflow(historySizeBytesGrows));
+  t.assert(after && before && after > before && before > 100);
```
mjameswh, line 230: Consistency: use `t.true()` instead of `t.assert()`. They are really doing the same anyway.

packages/workflow/src/interfaces.ts
```diff
@@ -66,10 +66,33 @@ export interface WorkflowInfo {
    *
    * This value changes during the lifetime of an Execution.
    *
+   * When this value is zero the history length is unknown.
```
mjameswh, line 69: Scratch that. It is not possible for `historyLength` to ever be unknown/zero in WorkflowInfo (that is only possible on client side, specifically in response to `list()`).

packages/client/src/helpers.ts
```diff
@@ -47,6 +49,9 @@ export async function executionInfoFromRaw<T>(
   dataConverter: LoadedDataConverter,
   rawDataToEmbed: T
 ): Promise<Replace<WorkflowExecutionInfo, { raw: T }>> {
+  const optionalLongToNumber = (x: Long | null | undefined): number | undefined =>
+    x === undefined || x === null ? undefined : x.toNumber();
```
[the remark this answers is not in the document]
antlai-temporal: Optional chaining really helps here. 
I don't think we should be using == instead of ===, many javascript programmers get confused by it, and it doesn't hurt to be a bit more explicit.

packages/client/src/helpers.ts
antlai-temporal: I don't need it anymore...

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,49 @@ test('Condition 0 patch sets a timer', async (t) => {
...
+  const worker = await createWorker();
+  const handle = await startWorkflow(historySizeBytesGrows);
+
+  await worker.runUntil(handle.result());
+  const historySizeBytes = (await handle.describe()).historySizeBytes;
+  t.assert(historySizeBytes && historySizeBytes > 100);
+});
+
+export async function suggestedCAN(): Promise<boolean> {
+  const MAX_EVENTS = 40_000;
```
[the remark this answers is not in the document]
antlai-temporal: Switched to camel. I think they are useful for documentation, making it more clear what they are for...

packages/test/src/test-integration-workflows.ts
antlai-temporal: done

packages/workflow/src/interfaces.ts
antlai-temporal: Done

packages/test/src/test-workflows.ts
```diff
@@ -353,6 +355,14 @@ test('successString', async (t) => {
   compareCompletion(t, req, makeSuccess([makeCompleteWorkflowExecution(defaultPayloadConverter.toPayload('success'))]));
 });
 
+test('continueAsNewSuggested', async (t) => {
```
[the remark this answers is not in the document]
antlai-temporal: I added a check that is initially `false`

packages/workflow/src/worker-interface.ts
```diff
@@ -186,11 +186,19 @@ export function activate(activation: coresdk.workflow_activation.WorkflowActivat
...
         activator.now = tsToMs(activation.timestamp);
       }
-      if (activation.historyLength == null) {
-        throw new TypeError('Got activation with no historyLength');
-      }
-      activator.info.unsafe.isReplaying = activation.isReplaying ?? false;
-      activator.info.historyLength = activation.historyLength;
+
+      // The Rust Core ensures that these activation fields are not null
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
mjameswh, line 191: Oh... That turns out quite ugly with all of these eslint directive comments 😞

Ok, lets put back the `?? false` on booleans, and try do something about the rest. Maybe this?

```ts
// Both Core and Protobuf guarantees that these fields won't be null
activator.info.unsafe.isReplaying = activation.isReplaying ?? false;
activator.info.historyLength = activation.historyLength as number;
activator.info.continueAsNewSuggested = activation.continueAsNewSuggested ?? false;
// historySizeBytes === 0 means WFT was generated by pre-1.20.0 server
activator.info.historySizeBytes = activation.historySizeBytes?.toNumber() || undefined;
```

packages/workflow/src/interfaces.ts
```diff
@@ -70,6 +70,23 @@ export interface WorkflowInfo {
...
+   */
+  historySizeBytes: number;
+
+  /**
+   * A hint provided by the current WorkflowTaskStarted event recommending whether to
+   * {@link continueAsNew}.
+   *
+   * This value changes during the lifetime of an Execution.
+   */
+  continueAsNewSuggested: boolean;
```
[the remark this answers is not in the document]
mjameswh: Should specifically mention that this is supported only on Temporal Server 1.20+, and yes, that it will always be false on older servers.

packages/workflow/src/worker-interface.ts
antlai-temporal: Yes, too many directives...
Changing compiler directives by run-time checks is not always a good idea, but I don't think this will affect performance much...

packages/workflow/src/interfaces.ts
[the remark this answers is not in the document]
antlai-temporal: Comment updated

packages/client/src/types.ts
```diff
@@ -36,6 +36,8 @@ export interface WorkflowExecutionInfo {
   taskQueue: string;
   status: { code: proto.temporal.api.enums.v1.WorkflowExecutionStatus; name: WorkflowExecutionStatusName };
   historyLength: number;
+  // Only available in server version 1.20 or later
+  historySizeBytes?: number;
```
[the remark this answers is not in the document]
cretz: We have history length and history size in all SDKs. I can understand how it is confusing because of how people use size in other places. But you could turn around and say "historyLength" is confusing because in some places length is meant to be the byte size (such as HTTP response headers and string length in Go). So we should have called it "historyEventCount".

But that's overthinking it.

packages/client/src/helpers.ts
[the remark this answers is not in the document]
bergundy: +1 for `x == null`, we've standardized on it in the SDK codebase.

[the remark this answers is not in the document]
antlai-temporal: OK, let's follow the standard

antlai-temporal:
I think the last issue left is to decide whether to rename to historySize(). @cretz  has a point that all the others SDKs are using it. I think that having both historyLength() and historySize() in the same interface is confusing, but not a big deal...

@bergundy Shall I go ahead and rename?

packages/client/src/types.ts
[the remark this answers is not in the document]
mjameswh: +1 to remove the `Bytes` suffix, for consistency (both in the TS SDK itself, and across SDKs)

packages/client/src/types.ts
```diff
@@ -36,6 +36,8 @@ export interface WorkflowExecutionInfo {
   taskQueue: string;
   status: { code: proto.temporal.api.enums.v1.WorkflowExecutionStatus; name: WorkflowExecutionStatusName };
   historyLength: number;
+  // Only available in server version 1.20 or later
```
mjameswh, line 39: We want that comment to be visible to users. `// ...` comments are not.

Use a `/** ... */` block comment, and copy what you have on the `historySize` property in WorkflowInfo.

packages/client/src/types.ts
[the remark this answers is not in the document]
antlai-temporal: Done

packages/client/src/types.ts
antlai-temporal: Done

antlai-temporal:
@mjameswh and @bergundy Renamed historySizeBytes to historySize, kept attribute in WorkflowExecutionInfo as optional, i.e., replacing zero by undefined, but attribute in WorkflowInfo is no longer optional, i.e.,, defaults to zero, as discussed...

packages/client/src/types.ts
```diff
@@ -36,6 +36,15 @@ export interface WorkflowExecutionInfo {
   taskQueue: string;
   status: { code: proto.temporal.api.enums.v1.WorkflowExecutionStatus; name: WorkflowExecutionStatusName };
   historyLength: number;
+  /**
+￼   * Size of Workflow history in bytes until the current Workflow Task.
```
mjameswh, line 40: `until the current Workflow Task` don't really make sense in this structure. There is no "current workflow task". Similar for `This value changes during the lifetime of an Execution`. I don't think it adds real value in this context.

Last thing: `historySize` is available in server versions >= 1.20 (ie. starting with 1.20, not greater than)

antlai-temporal: Updated comment

mjameswh (approved):
LGTM. Thanks @antlai-temporal!

PR #1225, opened 2023-08-25
feat(worker): Retain meta on log entries forwarded from core

### LINKED ISSUE #1195: Forwarded logs from the core do not include metadata
We're using the following configuration:
```
  Runtime.install({
    logger: new DefaultLogger("INFO", (logEntry): void => {
      // process log entry
    }),
    telemetryOptions: {
      logging: {
        forward: {},
      },
    },
  });
```

Unfortunately it seems like logs from the core do not include any metadata. The `logEntry.meta` field is `undefined`. Especially log entries like `Failing workflow task` need to include the workflow id and run id for them to be actionable.

mjameswh:
## What changed

- Retain meta on log entries forwarded from core (fix #1195)

### DIFF
--- a/packages/core-bridge/src/helpers.rs
+++ b/packages/core-bridge/src/helpers.rs
@@ -1,6 +1,6 @@
 use crate::errors::*;
 use neon::{prelude::*, types::buffer::TypedArray};
-use std::{fmt::Display, future::Future, sync::Arc, collections::HashMap};
+use std::{fmt::Display, future::Future, sync::Arc};
 
 /// Send a result to JS via callback using a [Channel]
 pub fn send_result<F, T>(channel: Arc<Channel>, callback: Root<JsFunction>, res_fn: F)
@@ -204,18 +204,12 @@ pub fn serde_value_to_js_value<'a>(cx: &mut impl Context<'a>, val: serde_json::V
             Ok(arr.upcast())
         }
         serde_json::Value::Object(map) => {
-            let obj: Handle<'a, JsObject> = cx.empty_object();
-            for (k, v) in map {
-                let k = cx.string(snake_to_camel(k));
-                let v=  serde_value_to_js_value(cx, v)?;
-                obj.set(cx, k, v)?;
-            }
-            Ok(obj.upcast())
+            hashmap_to_js_value(cx, map).map(|v| v.upcast())
         }
     }
 }
 
-pub fn hashmap_to_js_value<'a>(cx: &mut impl Context<'a>, map: HashMap<String, serde_json::Value>) -> JsResult<'a, JsObject> {
+pub fn hashmap_to_js_value<'a>(cx: &mut impl Context<'a>, map: impl IntoIterator<Item = (String, serde_json::Value)>) -> JsResult<'a, JsObject> {
     let obj: Handle<'a, JsObject> = cx.empty_object();
     for (k, v) in map {
         let k = cx.string(snake_to_camel(k));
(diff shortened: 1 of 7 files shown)


mjameswh:
@bergundy @Sushisource I'll take both of your input on this one. Pretty simple on the surface, but I have the feeling it could have some subtile consequences.

bergundy (commented):
Looks like there are some unhandled lint errors.

I'm a bit hesitant to merge this as is.
Core uses different log tags than TS, we should make sure that `workflowType`, `workflowId`, etc.. are converted to the TS standard.

Sushisource:
>  Core uses different log tags than TS, we should make sure that workflowType, workflowId, etc.. are converted to the TS standard.

This seems a little backwards to me - IMO all the Core-using languages should just use the same ones Core does so that we aren't doing conversion for every language. Core can even change if we don't like it right now, since nothing is depending on this, but we should do that there and then ensure every language is consistent and that way we don't have any obnoxious conversion layer

packages/worker/src/runtime.ts
```diff
@@ -265,31 +265,42 @@ export class Runtime {
...
+    const doPoll = async () => {
+      const logs = await poll(this.native);
+      for (const log of logs) {
+        const meta: Record<string | symbol, unknown> = {
+          [LogTimestamp]: timeOfDayToBigint(log.timestamp),
+        };
+        if (log.fields) {
+          for (const [key, value] of Object.entries(log.fields)) {
+            meta[key] = JSON.parse(value);
+          }
```
Sushisource, line 285: This is the only bit that is maybe mildly worrisome, doing quite a bit more parsing now. That's really got to happen though without heading down the flatbuffers road or something, so it's not like there's an alternative.

Otherwise this change looks totally fine to me pending what we decide about key names

bergundy:
> > Core uses different log tags than TS, we should make sure that workflowType, workflowId, etc.. are converted to the TS standard.
> 
> This seems a little backwards to me - IMO all the Core-using languages should just use the same ones Core does so that we aren't doing conversion for every language. Core can even change if we don't like it right now, since nothing is depending on this, but we should do that there and then ensure every language is consistent and that way we don't have any obnoxious conversion layer

That's a bit too late though now, users have already been using the TS convention for a while. If we were to do this from scratch, I'd agree with you.
Maybe we can do the conversion core-side (or originally emit attributes with camel case based on config) instead of in the bridge?

packages/worker/src/runtime.ts
bergundy: Agree, I'm a bit worried about the perf implications here.

mjameswh: I do share that worry, though I'm not sure if this is actually a thing. TBH that's pretty much _the_ thing I really wanted opinions about.

The perf impact pretty much comes down to the number and complexity of metadata that are produced by Core. Converting a few integers to string then back to numbers is definitely not optimal, yet not that bad. Same for a few short strings composed of almost ascii-7 characters. Converting large/complex JSON objects would definitely be something else. But at this point, I'm not sure what I should expect to find out in these metadata.

> it's not like there's an alternative

Well, sure there is. I ended up going for JSON just for simplicity, but I was initially thinking of recursively creating the JS object tree directly in Rust code ([here](https://github.com/temporalio/sdk-typescript/pull/1225/files#diff-198eab910c3297070c03356ffb4f223a429afb656d8bf06ee59132bea522223bR210-R215)). I still consider that as an option. Should not be a difficult thing.

bergundy: Yeah, if we have the rust types and it’s all primitives, I would do the conversion at the bridge

mjameswh:
I moved the work on fixing the RxJS issue on Worker shutdown to #1238. This PR will be left open for work on exporting log metadata.

packages/worker/src/runtime.ts
Sushisource: Ah, yeah that makes perfect sense. I think I'd go for that. Though _honestly_, it is not immediately obvious to me it'd actually be faster. JSON parsers are one of the most optimized things on the planet and I wouldn't be completely surprised to find that it actually ends up being faster than constructing the JS objects in the bridge and passing them back, since I'm not entirely sure what that entails.

We'd need to actually bench it. However, I don't think the actual impact is going to be super meaningful one way or the other. Not sure it's worth spending the time on that.

Sushisource:
> > > Core uses different log tags than TS, we should make sure that workflowType, workflowId, etc.. are converted to the TS standard.
> > 
> > 
> > This seems a little backwards to me - IMO all the Core-using languages should just use the same ones Core does so that we aren't doing conversion for every language. Core can even change if we don't like it right now, since nothing is depending on this, but we should do that there and then ensure every language is consistent and that way we don't have any obnoxious conversion layer
> 
> That's a bit too late though now, users have already been using the TS convention for a while. If we were to do this from scratch, I'd agree with you. Maybe we can do the conversion core-side (or originally emit attributes with camel case based on config) instead of in the bridge?

Yeah, like I said we can change core, but then any other languages that don't match should ideally just inherit that new convention. We could have toggles in core for snake or camel case or whatever, but if there's anything that's different beyond that, I vote we just change whatever language is doing that and bury the problem forever

packages/core-bridge/src/helpers.rs
```diff
@@ -187,3 +187,56 @@ where
...
+            }
+            Ok(arr.upcast())
+        }
+        serde_json::Value::Object(map) => {
+            let obj: Handle<'a, JsObject> = cx.empty_object();
+            for (k, v) in map {
+                let k = cx.string(snake_to_camel(k));
+                let v=  serde_value_to_js_value(cx, v)?;
+                obj.set(cx, k, v)?;
+            }
```
Sushisource, line 212: Looks like this can just use the below function?

Sushisource (approved):

packages/core-bridge/src/helpers.rs
```diff
@@ -187,3 +187,56 @@ where
...
+    for (k, v) in map {
+        let k = cx.string(snake_to_camel(k));
+        let v = serde_value_to_js_value(cx, v)?;
+        obj.set(cx, k, v)?;
+    }
+    Ok(obj)
+}
+
+fn snake_to_camel(s: String) -> String {
+    let mut result = String::new();
```
bergundy, line 229: nit: can probably optimize this but let's start with this and see if it becomes an issue (I predict it won't).

bergundy (approved):

packages/core-bridge/src/helpers.rs
mjameswh: Yeah, I doubt that will make a huge difference, but went for some obvious low hanging fruits, notably by not creating a new string unless there is actually a '_' somewhere, and by presizing the output string's buffer.Could also work at bytes level

PR #1235, opened 2023-09-08
chore: Remove experimental flags on Schedules, Failure Converters, Local Activities and Replay Histories

mjameswh:
## What was changed

The following APIs are no longer marked as experimental:

- Schedules
- Failure Converters
- Local Activities
- Replay Histories

For info, at this date, the following APIs are still marked as experimental:

- Versioning / TaskQueueClient (newly introduced)
- Client's gRPC retry options (not sure we want to commit to long time support on this, and for sure, not before we upgrade to grpc-js 1.8)
- All telemetry options (likely to change due to upcoming advanced metrics support)

This one is deemed to perpertual 'experimental'

- Workflow Internal Interceptors (this is for advanced use cases, and we don't want to commit to long term support)

### DIFF
--- a/packages/workflow/src/interceptors.ts
+++ b/packages/workflow/src/interceptors.ts
@@ -195,6 +195,8 @@ export interface DisposeInput {}
  * Interceptor for the internals of the Workflow runtime.
  *
  * Use to manipulate or trace Workflow activations.
+ *
+ * @experimental This API is for advanced use cases and may change in the future.
  */
 export interface WorkflowInternalsInterceptor {
   /**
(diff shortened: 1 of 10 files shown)


packages/workflow/src/interceptors.ts
```diff
@@ -195,8 +195,6 @@ export interface DisposeInput {}
  * Interceptor for the internals of the Workflow runtime.
  *
  * Use to manipulate or trace Workflow activations.
- *
- * @experimental
```
bergundy, line 199: Let's keep this one experimental, this is really for advanced use cases and we should reserve the right to modify our internal at will.

bergundy (approved):
Just one comment, otherwise LGTM

packages/workflow/src/interceptors.ts
mjameswh: Fair enough

PR #1236, opened 2023-09-11
changed the signal definition on ExternalWorkflowHandle

gabrielsantosblanchet:
changed the signal definition on ExternalWorkflowHandle so it matches the definition of BaseWorkflowHandle

## What was changed

the definition of signal was not the same in the interface ExternalWorkflowHandle and BaseWorkflowHandleClass

## Why?

Made sure the definition of Signal are the same because it causes problem where you have a typescript error when you call signal on an variable that was defined as ExternalWorkflowHandle | ChildWorkflowHandle.

## Checklist
ran test before: 222 passed  | 9 todo
build with no error
ran test after: 222 passed | 9 todo

2. How was this tested:
N/A

3. Any docs updates needed?
no docs updates should be needed

### DIFF
--- a/packages/workflow/src/workflow-handle.ts
+++ b/packages/workflow/src/workflow-handle.ts
@@ -17,7 +17,10 @@ export interface ExternalWorkflowHandle {
    * await handle.signal(incrementSignal, 3);
    * ```
    */
-  signal<Args extends any[] = []>(def: SignalDefinition<Args> | string, ...args: Args): Promise<void>;
+  signal<Args extends any[] = [], Name extends string = string>(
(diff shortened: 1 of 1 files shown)


josiah-roberts (approved):
Thanks, @gabrielsantosblanchet!

mjameswh:
Thanks you @gabrielsantosblanchet for this fix, and @josiah-roberts for feedback.

mjameswh:
@gabrielsantosblanchet Next time, please make sure to enable the "Allow edits and access to secrets by maintainers" option when opening a PR.

I cloned this PR to #1237, in order to add some basic test.

PR #1243, opened 2023-09-13
chore: Allow manually starting stress tests in CI

mjameswh:
## What was changed

- Make it possible to manually start CI stress tests

### DIFF
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -16,10 +16,10 @@ on:
         required: true
         type: string
       test-timeout-minutes:
-        description: Timeout for the test in minutes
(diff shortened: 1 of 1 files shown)


.github/workflows/stress.yml
```diff
@@ -1,6 +1,31 @@
...
+        options:
+          - ci-stress
+          - ci-nightly
+        required: true
+        type: string
+      test-timeout-minutes:
+        description: Timeout for the test in minutes
+        required: true
+        type: number
+        default: 20
```
bergundy, line 22: I think we should put the defaults for nightly and stress in the description.

bergundy (approved):

PR #1249, opened 2023-09-21
fix(activity): Change to arrow function notation so class context is not lost

SergeiMeza:
## What was changed

Added missing bindings on activity context class.

## Why?

Since we are restoring the `Context` class from `AsyncLocalStorage`. Class scope is missing and thus rebinding is necessary. Attempting to use `Context.current().heartbeat()` or `Context.current().sleep()` would result on `this` property is undefined error.

## Checklist

1. Closes 

2. How was this tested:

I tested it with the `activities-cancellation-heartbeating` SDK example.

3. Any docs updates needed?

Since it addresses a current bug, no doc changes are necessary.

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -285,7 +285,7 @@ export class Context {
    */
   public heartbeat = (details?: unknown): void => {
     this.heartbeatFn(details);
-  }
+  };
 
   /**
    * Gets the context of the current Activity.
@@ -311,5 +311,5 @@ export class Context {
       handle = setTimeout(resolve, msToNumber(ms));
     });
     return Promise.race([this.cancelled.finally(() => clearTimeout(handle)), timer]);
-  }
+  };
 }
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -297,6 +297,8 @@ export class Context {
     if (store === undefined) {
       throw new Error('Activity context not initialized');
     }
+    store.sleep = store.sleep.bind(store);
+    store.heartbeat = store.heartbeat.bind(store);
     return store;
   }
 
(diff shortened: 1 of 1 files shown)


packages/activity/src/index.ts
```diff
@@ -297,6 +297,8 @@ export class Context {
     if (store === undefined) {
       throw new Error('Activity context not initialized');
     }
+    store.sleep = store.sleep.bind(store);
+    store.heartbeat = store.heartbeat.bind(store);
```
bergundy, line 301: Instead of this you could define sleep as:

```ts
sleep: (ms: Duration): Promise<void> => {
}
```

Same goes for heartbeat.

bergundy (commented):
Thanks for the contribution.
I think this is fine and we should support destructuring the activity context.

packages/activity/src/index.ts
SergeiMeza: I guess they would be equivalent. Though, Current SDK and samples code style heavily prefers `function` methods over `array function` methods.

SergeiMeza: @bergundy Should I change it to arrow function? It's okay but in the future if a normal function using `this` is added, it will break if not bound as in the commit above.

SergeiMeza: @bergundy Done

SergeiMeza: @bergundy Done

SergeiMeza: I also test it using the sdk samples so it's working

SergeiMeza:
@bergundy It seems linter is misconfigured. It doesn't accept arrow functions inside class bodies.

bergundy:
We need to add a linter exception, @mjameswh or I can take care of that if you want. Otherwise this is good to merge.
Thanks again.

SergeiMeza:
Yes, please. I'll move on to further integrating temporal into our tech stack.

bergundy:
Looks like it wasn't a lint error, it was a formatting error. `npm run format` fixed it.
Merged, thanks @SergeiMeza.

giuliohome:
When is next release planned? I see this was merged on Sep 25 but as of today Nov 20 the samples are still broken. Either they should be fixed meanwhile or this should be published.

PR #1252, opened 2023-09-26
feat(activity): Export some of activity.Context members at module level

### LINKED ISSUE #1247: [Bug] Properties of `WorkflowInfo` and `Activity.Context` should not be mutable by user code
### Describe the bug

The objects returned by the `wf.workflowInfo()` and `act.Context.current()` functions can be mutated at will by user code. Though there might be arguably legitimate use cases for adding _custom_ properties into these objects, it appears undesirable, and even dangerous to allow mutating the value of existing properties.

The consequence of such mutations varies. As an extreme case, workflow code can actually flip the value of `wf.workflowInfo().unsafe.isReplaying`, which results in the `wf.patched()` API function to misbehave.

### Minimal Reproduction

```
export async function asyncActivityWorkflow(): Promise<string> {
  workflowInfo().attempt = 100;
  (workflowInfo() as any).something = 'something';
  
  workflowInfo().unsafe.isReplaying = true;
  if (patched('test')) {
    console.log(`This should be printed at least once`);
  }

  const answer = await completeSomethingAsync();

  console.log(`workflowInfo().attempt: ${workflowInfo().attempt}`);
  console.log(`workflowInfo().something: ${(workflowInfo() as any).something}`);

  return `The Peon says: ${answer}`;
}
```

produce the following output:

```
[asyncActivityWorkflow(activities-examples)] workflowInfo().attempt: 100
[asyncActivityWorkflow(activities-examples)] workflowInfo().something: something
```

mjameswh:
## What changed

- The `@temporalio/activity` module now exports `activityInfo()`, `log`, `sleep()`, `heartbeat()`, `cancelled` and `cancellationSignal`, avoiding the necessity of spreading the activity context in activity code to reach those.

- Marked most fields of `Context`, `ActivityInfo`, `WorkflowInfo` and others as readonly.

- Modified the way `WorkflowInfo` is updated internally by the SDK, so that the structure is copy-on-write; this should significantly reduce the number of time that object needs to be deep cloned (eg. it was previously cloned on every sink call).

## Why

- Better DX for activity developers as it is no longer required to explicitly traverse the Activity Context object to reach most activity APIs.
- Fixes #1247


packages/activity/src/index.ts
```diff
@@ -200,18 +200,34 @@ export interface Info {
...
-  public info: Info;
+  public readonly info: Info;
+
   /**
-   * Await this promise in an Activity to get notified of cancellation.
+   * A Promise that fails with a {@link CancelledFailure} when cancellation of this activity is requested. The promise
+   * is guaranteed to never successfully resolve. Await this promise in an Activity to get notified of cancellation.
    *
-   * This promise will never resolve—it will only be rejected with a {@link CancelledFailure}.
+   * Note that to get notified of cancellation, an activity must _also_ do {@link Context.heartbeat}.
```
bergundy, line 225:
```suggestion
   * Note that to get notified of cancellation, an activity must _also_ {@link Context.heartbeat}.
```

packages/activity/src/index.ts
```diff
@@ -222,13 +238,17 @@ export class Context {
    * {@link https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options child_process}
    * to abort a child process, as well as other built-in node modules and modules found on npm.
    *
+   * Note that to get notified of cancellation, an activity must _also_ do {@link Context.heartbeat}.
```
bergundy, line 241:
```suggestion
   * Note that to get notified of cancellation, an activity must _also_ {@link Context.heartbeat}.
```

packages/activity/src/index.ts
```diff
@@ -255,7 +275,11 @@ export class Context {
     heartbeat: (details?: any) => void,
     logger: Logger
   ) {
-    this.info = info;
+    // Note that we don't freeze the context object itself, because some users add properties to it (generally from
+    // interceptors). We even had samples promoting that practice. We should probably deprecate that at some point, but
+    // lets give us some time before we do.
+    this.info = deepFreeze(info);
```
bergundy, line 281: I think it's fine if the context remains mutable and we allow users to attach to it, I'd even not freeze the info object, it's fine if we just mark it readonly. I haven't seen issues with this in over 2 years that we've had this SDK.

packages/activity/src/index.ts
```diff
@@ -283,33 +307,104 @@ export class Context {
...
   };
 }
+
+/**
+ * The current Activity's context.
+ */
+export function activityInfo(): Info {
+  // For consistency with workflow.workflowInfo(), we want activityInfo() to be a function, rather than a const object.
+  return Context.current().info;
+}
```
bergundy, line 334: I wish `workflowInfo` was just called `info` but this is fine for consistency.

packages/activity/src/index.ts
```diff
@@ -283,33 +307,104 @@ export class Context {
...
+
+/**
+ * The logger for this Activity.
+ *
+ * This is a shortcut for `Context.current().log` (see {@link Context.log}).
+ */
+export const log: Logger = new Proxy({} as Logger, {
```
bergundy, line 341: This doesn't have to be a proxy, you could hardcode all of the methods instead.

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -225,7 +225,14 @@ export class VMWorkflowThreadProxy implements Workflow {
       throw new TypeError(`Got invalid response output from Workflow Worker thread ${output}`);
     }
     output.calls.forEach((call) => {
-      call.workflowInfo.unsafe.now = Date.now;
+      const info = call.workflowInfo;
```
bergundy, line 228: Why does this matter?

packages/workflow/src/internals.ts
```diff
@@ -296,7 +296,7 @@ export class Activator implements ActivationHandler {
     registeredActivityNames,
   }: WorkflowCreateOptionsInternal) {
     this.getTimeOfDay = getTimeOfDay;
-    this.info = info;
+    this.info = deepFreeze(info);
```
bergundy, line 299: This is taking it too far, just mark readonly and move on.

packages/activity/src/index.ts
mjameswh: Good point

packages/worker/src/workflow/threaded-vm.ts
mjameswh: Not sure what you mean. Are you asking why I care about rehydrating `unsafe.now`?

bergundy: I mean the cloning

bergundy (approved):

packages/worker/src/workflow/threaded-vm.ts
mjameswh: Oh, that... Just because `unsafe.now` is now readonly, so technically can't be modified in-place. But fair point, I'll cast the field to not readonly.

PR #1254, opened 2023-09-29
feat: Expose nonStickyToStickyPollRatio WorkerOption

bergundy:
## What changed

- `WorkerOptions` now expose a `nonStickyToStickyPollRatio` property, which control the proportion of Workflow Task Pollers assigned to polling from Regular Task Queue vs to polling from Sticky Tasks Queues.


bergundy:
Tested by putting a print statement to verify that the value is propagated.

packages/worker/src/worker-options.ts
```diff
@@ -256,6 +256,21 @@ export interface WorkerOptions {
...
+   * minimum for either poller is 1, so if `maxConcurrentWorkflowTaskPolls` is 1 and sticky queues are
+   * enabled, there will be 2 concurrent polls.
+   *
+   * ⚠️T This API is experimental and may be removed in the future if the poll scaling algorithm changes.
+   *
+   * @experimental
+   *
+   * @default 0.2
+   */
+  nonStickyToStickyPollRatio?: number;
```
mjameswh, line 272: This must also be added to [core-bridge's `WorkerOptions`](https://github.com/bergundy/sdk-typescript/blob/pr/bergundy/1254/packages/core-bridge/ts/index.ts#L264), since we're reading it from Rust side.

mjameswh: Have to admit though that, as it currently stands, this is somewhat maintenance overhead with little gain, as all props from the "outer" WorkerOptions are passed to bridge's createWorker anyway 🤔.

Could be pertinent to consider making a new object [here](https://github.com/bergundy/sdk-typescript/blob/pr/bergundy/1254/packages/worker/src/worker.ts#L221), matching exactly the bridge's expectation...

bergundy: I'll add it for the typing and obviously the right thing to do. FYI I tested and it's not required.

packages/worker/src/worker-options.ts
```diff
@@ -686,6 +705,7 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
     reuseV8Context: reuseV8Context ?? false,
     debugMode: debugMode ?? false,
     interceptors: appendDefaultInterceptors({}),
+    nonStickyToStickyPollRatio: nonStickyToStickyPollRatio ?? 0.2,
```
mjameswh, line 708: nit: Should probably add some boundary checks here, either here or in core. AFAICS, someone could set `nonStickyToStickyPollRatio: 20.0` and core would just happily start with 20 times the expected number of pollers...

mjameswh (approved):
LGTM

packages/worker/src/worker-options.ts
bergundy: This should be validated on the core side IMHO, less logic for every lang SDK.

bergundy:
I'll fix the trivial build error soon

mjameswh:
First time I see this time out in 'features' test: https://github.com/temporalio/sdk-typescript/actions/runs/6553208106/job/17798109271?pr=1254.

PR #1268, opened 2023-10-20
feat(worker): No longer require `defaultSink` and `appendDefaultInterceptors`

### LINKED ISSUE #1256: [Feature Request] Deprecate `appendDefaultInterceptors`
## Context

- `appendDefaultInterceptors()` is quite easy to miss => Bad developer experience.
- Since 1.8, prepending `defaultSink()` is no longer required => Better DX

## Description

- Make it so that Workflow life cycle logging no longer require calling `appendDefaultInterceptors()`, either by moving those log calls out of the interceptor into the activator, or by forcibly injecting that interceptor.
- In both case, `appendDefaultInterceptors()` would now be a noop, which can then be deprecated.
- Same for activity logger

mjameswh:
## What changed

- Deprecate `defaultSink()` and `LoggerSinks`. SDK's logger sink is now forcibly injected. Registering a custom logger sink named `defaultWorkerLogger` is still supported, but discouraged and deprecated.
- Deprecate `appendDefaultInterceptors()` and `defaultWorkflowInterceptorModules`. SDK's interceptors are now forcibly injected. Using `appendDefaultInterceptors()` to customize the activity logger is deprecated.
- Refactored sink code internally

## Why

- Fix #1256
- Better DX
- Extracted these changes from the metrics branch, to make it easier to review.


packages/worker/src/worker-options.ts
```diff
@@ -407,11 +408,11 @@ export interface WorkerOptions {
   /**
    * A mapping of interceptor type to a list of factories or module paths.
    *
-   * By default, {@link ActivityInboundLogInterceptor} and {@link WorkflowInboundLogInterceptor} are installed. If you
-   * wish to customize the interceptors while keeping the defaults, use {@link appendDefaultInterceptors}.
-   *
    * When using {@link workflowBundle}, these Workflow interceptors (`WorkerInterceptors.workflowModules`) are not used.
    * Instead, provide them via {@link BundleOptions.workflowInterceptorModules} when calling {@link bundleWorkflowCode}.
+   *
+   * The SDK itself may automatically append interceptors required to support its own features Explicitely calling
```
antlai-temporal, line 414: typo explicitly
Does the order matter, i.e., the interceptors added are run before or after the user provided ones?

packages/worker/src/worker-options.ts
```diff
@@ -407,11 +408,11 @@ export interface WorkerOptions {
...
+   * `appendDefaultInterceptors()` is no longer required.
```
antlai-temporal, line 415: What happens if they are also provided, do we run them twice , or just ignore them...

packages/worker/src/worker-options.ts
```diff
@@ -444,10 +445,9 @@ export interface WorkerOptions {
    * guarantees, please consider using local activities instead. For use cases that require
    * _exactly-once_ or _at-most-once_ execution guarantees, please consider using regular activities.
    *
-   * The SDK itself may register sinks functions required to support workflow features. At the moment, the only such
-   * sink is 'defaultWorkerLogger', which is used by the workflow context logger (ie. `workflow.log.info()` and
-   * friends); other sinks may be added in the future. You may override these default sinks by explicitely registering
-   * sinks with the same name.
+   * The SDK itself may register sinks functions required to support workflow features. The name,
+   * signature and semantic of these sinks functions is considered an internal detail and may change
+   * in the future without notice. Please do not use them directly nor overide them.
```
antlai-temporal, line 450: typo override

packages/worker/src/worker-options.ts
```diff
@@ -602,59 +599,33 @@ export interface ReplayWorkerOptions
...
+ *
+ * @deprecated Calling `appendDefaultInterceptors()` is no longer required. To configure a custom logger, set the
+ *             {@see Runtime.logger} property instead.
  */
 export function appendDefaultInterceptors(
   interceptors: WorkerInterceptors,
   logger?: Logger | undefined
 ): WorkerInterceptors {
+  if (!logger) return interceptors;
   return {
```
antlai-temporal, line 626: Do you need to clone interceptors to respect the original semantics?

packages/workflow/src/sinks.ts
```diff
@@ -44,14 +45,64 @@ export interface SinkCall {
 }
 
 /**
- * Sink interface for forwarding logs from the Workflow sandbox to the Worker
+ * Get a reference to Sinks for exporting data out of the Workflow.
+ *
```
antlai-temporal, line 49: What happens with logging for activities, does this affect how activities logging is implemented?

antlai-temporal (approved):
I have added a few minor issues. My main question is how logging will change for activities. I assume that local activities are the same as workflow logging, but what happens outside the sandbox?

packages/test/src/test-runtime.ts
```diff
@@ -118,13 +118,6 @@ if (RUN_INTEGRATION_TESTS) {
     }
   });
 
-  test.serial('Runtime.instance() throws meaningful error when passed invalid tracing.otel.url', (t) => {
```
bergundy, line 121: Maybe use metrics otel url for this test instead?

packages/worker/src/index.ts
```diff
@@ -40,9 +47,16 @@ export {
...
 } from './worker';
 export {
+  /**
+   * @deprecated Including `appendDefaultInterceptors()` in the worker options is no longer required.
+   */
   appendDefaultInterceptors,
   CompiledWorkerOptions,
-  defaultSinks,
+  /**
+   * @deprecated Including `defaultSink()` in the worker options is no longer required. To configure
```
bergundy, line 56:
```suggestion
   * @deprecated Including `defaultSinks()` in the worker options is no longer required. To configure
```

packages/worker/src/index.ts
```diff
@@ -56,13 +70,12 @@ export {
...
-} from './workflow/bundler';
+export { BundleOptions, bundleWorkflowCode, WorkflowBundleWithSourceMap } from './workflow/bundler';
+/**
+ * @deprecated Including `defaultWorkflowInterceptorModules` in BundlerOptions.workflowInterceptorModules is no longer required.
+ */
+export const defaultWorkflowInterceptorModules = [];
+/**
+ * @deprecated Including `defaultWorkflowInterceptorModules` in BundlerOptions.workflowInterceptorModules is no longer required.
+ */
+export const defaultWorflowInterceptorModules = [];
```
bergundy, line 81: I think it's fine to just remove this typo but 🤷

packages/worker/src/worker-options.ts
bergundy, line 414:
```suggestion
   * ⚠️ New in 1.9.0 ⚠️ 
   *
   * The SDK itself may automatically append interceptors required to support its own features. Explicitly calling
```

packages/worker/src/worker-options.ts
bergundy, line 415:
```suggestion
   * ⚠️ Before version 1.9.0, calling `appendDefaultInterceptors` was required to keep the SDK's default logging interceptors. This is no longer the case and the default interceptors have been integrated into the SDK internals.
```

packages/worker/src/worker-options.ts
```diff
@@ -704,9 +675,18 @@ export function addDefaultWorkerOptions(options: WorkerOptions): WorkerOptionsWi
...
+      activityInbound: [
+        ...(options.interceptors?.activityInbound ?? []),
+        (ctx) => new ActivityInboundLogInterceptor(ctx),
+      ],
+      workflowModules: options.interceptors?.workflowModules ?? [],
+    },
     nonStickyToStickyPollRatio: nonStickyToStickyPollRatio ?? 0.2,
-    sinks: { ...defaultSinks(), ...sinks },
+    sinks: {
+      ...initLoggerSink(),
```
bergundy, line 687: Wondering if it's worth initializing this with the runtime's sink here to avoid obtaining it later from the runtime instance.

packages/workflow/src/logs.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * NOTE: Specifying a custom logger through {@link defaultSink} or by manually registering a sink named
+ * `defaultWorkerLogger` has been deprecated. Please use {@link Runtime.logger} instead.
+ */
+export const log: WorkflowLogger = Object.fromEntries(
+  (['trace', 'debug', 'info', 'warn', 'error'] as Array<keyof WorkflowLogger>).map((level) => {
+    return [
+      level,
+      (message: string, attrs?: Record<string, unknown>) => {
+        const activator = assertInWorkflowContext('Workflow.log(...) may only be used from a Workflow Execution.');
```
bergundy, line 46:
```suggestion
        const activator = assertInWorkflowContext('Workflow.log(...) may only be used from workflow context.');
```

bergundy (approved):
LGTM.

You could also get rid of the otel dependency and the tracing code in the TS worker and the bridge and remove SDK tracing altogether.

packages/worker/src/worker-options.ts
mjameswh: No need to worry about that. Both list of interceptors will be spread-cloned in `addDefaultWorkerOptions`/`bundleWorkflowCode`.

packages/workflow/src/sinks.ts
mjameswh: No. Sinks is a Workflow only feature. They are required because workflow sandboxing makes it impossible for Workflow code to make direct calls to IO stuff (eg. logging and metrics probably involve newtwork requests and/or file system accesses). There is no such constraint for activity code, so we don't need similar indirections for activities.

packages/worker/src/worker-options.ts
mjameswh: Yes, ordering matters. Interceptors are called in order, from the first to the last, each one making the call to the next one, and the last one calling the original function. I'll add this bit of info to the doc.

mjameswh: Now, that leaves us with a minor problem: until now, `appendDefaultInterceptors()` was _appending_ our log interceptors. In this PR, I retained that ordering, but I think it is wrong. Built-in interceptors should be prepended, so that they capture what happens inside user's interceptors. For example, if the workflow function completed succesfully, but a custom interceptor wrapping it throws an exception, I'd say that WorkflowLogInterceptor should report that exception, rather than success.

I'm therefore changing this to _prepending_ internal interceptors instead. That might slightly change SDK's observable behaviors, which could techically be breaking in some rare cases, but I feel ok with that one. @bergundy WDYT?

bergundy: Yeah.. I think prepending is fine if we want to capture the durations and alterations made by user interceptors.
For the default log attributes though we should ensure that users can modify our defaults.

packages/worker/src/worker-options.ts
mjameswh: We would have run them twice. I modified code tonot prepend internal interceptors if they "seems" to already be there. But we can't catch _all_ cases that might results in logging wf/activity life cycle events twice.

packages/worker/src/worker-options.ts
mjameswh: Of course. That is what will happen by prepending internal interceptor.

mjameswh:
I completed removing tracing support, but that turns out to be quite a large change by itself, so moved that to a [distinct PR](https://github.com/temporalio/sdk-typescript/pull/1269).

packages/testing/src/index.ts
```diff
@@ -412,7 +412,8 @@ export class MockActivityEnvironment extends events.EventEmitter {
       promise,
       abortController.signal,
       heartbeatCallback,
-      Runtime.instance().logger
+      Runtime.instance().logger,
+      (attrs) => attrs
```
mjameswh, line 416: Should probably inject some basic log attributes here

mjameswh:
Will split this in three smaller PRs.

PR #1270, opened 2023-10-26
feat: Reduce the default Core log level to WARN and non Core Rust logs to ERROR

bergundy:
- Match Python and .NET SDKs.
- Reduce noise.


Sushisource (approved):
Oh wow I'm surprised we had `INFO` for all other Rust logs too

PR #1272, opened 2023-10-27
feat(client): Add `withAbortSignal` context manager

bergundy:
## What was changed

Clients and the Connection class get a `withAbortSignal` context manager that can be used to cancel all ongoing connections started within a function's scope.

```ts
const ctrl = new AbortController();
setTimeout(() => ctrl.abort(), 10_000);
// 👇 throws if incomplete by the timeout.
await conn.withAbortSignal(ctrl.signal, () => { /* make client or direct grpc calls */);
```

## Why?

Provide a way to abort client grpc calls.

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -69,6 +69,7 @@
  * @module
  */
 
+// Keep this around until we drop support for Node 14.
 import 'abort-controller/polyfill'; // eslint-disable-line import/no-unassigned-import
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { Logger, Duration, LogLevel, LogMetadata } from '@temporalio/common';
--- a/packages/client/src/base-client.ts
+++ b/packages/client/src/base-client.ts
@@ -71,13 +71,13 @@ export class BaseClient {
 
   /**
    * Set an {@link https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal | `AbortSignal`} that, when aborted,
-   * cancels any ongoing requests executed in `fn`'s scope.
+   * cancels any ongoing service requests executed in `fn`'s scope.
    *
    * @returns value returned from `fn`
    *
    * @see {@link Connection.withAbortSignal}
    */
-  async withAbortSignal<ReturnType>(abortSignal: AbortSignal, fn: () => Promise<ReturnType>): Promise<ReturnType> {
+  async withAbortSignal<R>(abortSignal: AbortSignal, fn: () => Promise<R>): Promise<R> {
     return await this.connection.withAbortSignal(abortSignal, fn);
   }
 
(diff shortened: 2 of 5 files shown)


packages/activity/src/index.ts
```diff
@@ -69,6 +69,7 @@
  * @module
  */
 
+// Keep this around until we drop support for Node 14.
```
Sushisource, line 72: Maybe link to an issue # so this is easily searchable

packages/client/src/base-client.ts
```diff
@@ -1,3 +1,5 @@
+// Keep this around until we drop support for Node 14.
+import 'abort-controller/polyfill'; // eslint-disable-line import/no-unassigned-import
```
mjameswh, line 2: I don't think that polyfill is actually making any difference. `AbortController` was added in 14.17, and we only support 14.18+. And based on that polyfill's [source code](https://github.com/mysticatea/abort-controller/blob/master/polyfill.js), it appears that the polyfill is a noop if that class already exists.

packages/client/src/base-client.ts
```diff
@@ -67,6 +69,18 @@ export class BaseClient {
     return await this.connection.withDeadline(deadline, fn);
   }
 
+  /**
+   * Set an {@link https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal | `AbortSignal`} that, when aborted,
+   * cancels any ongoing requests executed in `fn`'s scope.
```
dandavison, line 74: nit: in other docstrings in this file we refer to "service requests" -- could bring them all in line with each other.

packages/client/src/base-client.ts
```diff
@@ -67,6 +69,18 @@ export class BaseClient {
...
+   *
+   * @returns value returned from `fn`
+   *
+   * @see {@link Connection.withAbortSignal}
+   */
+  async withAbortSignal<ReturnType>(abortSignal: AbortSignal, fn: () => Promise<ReturnType>): Promise<ReturnType> {
```
dandavison, line 80: Are you thinking that there's an argument for entire words over single letter type parameters in places where the type will be displayed to users?

packages/client/src/connection.ts
```diff
@@ -397,7 +404,27 @@ export class Connection {
...
+   *
+   * @returns value returned from `fn`
+   *
+   * @example
+   *
+   * ```ts
+   * const ctrl = new AbortController();
+   * setTimeout(() => ctrl.abort(), 10_000);
+   * // 👇 throws if incomplete by the timeout.
+   * await conn.withAbortSignal(ctrl.signal, () => client.start(options));
```
dandavison, line 422: `start()` is often something that returns immediately, perhaps choose a different name for the example e.g. `execute`?

PR #1278, opened 2023-11-01
docs: Specify default `VersioningIntent`

lorensr:
Wrapped comment based on prettierrc's printwidth: 120

### DIFF
--- a/packages/common/src/activity-options.ts
+++ b/packages/common/src/activity-options.ts
@@ -106,6 +106,8 @@ export interface ActivityOptions {
    * When using the Worker Versioning feature, specifies whether this Activity should run on a
    * worker with a compatible Build Id or not. See {@link VersioningIntent}.
    *
+   * @defaultValue 'COMPATIBLE'
+   *
    * @experimental
    */
   versioningIntent?: VersioningIntent;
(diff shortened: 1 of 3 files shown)


packages/common/src/activity-options.ts
```diff
@@ -106,6 +106,8 @@ export interface ActivityOptions {
    * When using the Worker Versioning feature, specifies whether this Activity should run on a
    * worker with a compatible Build Id or not. See {@link VersioningIntent}.
    *
+   * @defaultValue 'COMPATIBLE'
```
bergundy, line 109: I think it’s just `@default`

lorensr: tsdoc, right? i don't see @default, only: https://tsdoc.org/pages/tags/defaultvalue/

mjameswh: Interesting... Yeah, it indeed look like `@defaultValue` is the right one. `@default` appears in [generated doc files](https://typescript.temporal.io/api/interfaces/client.BackoffOptions#factor), but that's because it is configured as a reserved discretionary block tag. The output (at least in VS Code) is different.

mjameswh: @lorensr For this PR, please use `@default` anyway, as this is our current standard.

We can and should consider changing our standard, but that should be a repo-wide find and replace, and I would like to see the generated html doc before we commit to `@defaultValue`.

bergundy: `@default` is the jsdoc standard, tooling seems to support both. Haven’t checked how vs code displays them though

mjameswh (approved):
LGTM. Thanks Loren!

PR #1279, opened 2023-11-03
fix(proto): protobuf exports may be incorrect if minimized

mjameswh:
## What changed

- In a few places, we reflectively visit and transform our protobuf definitions. To determine the type of node being visited (Root, Namespace or Type), we would previously check the object's `constructor.name`. However, that name could get mangled if the package was bundled and minimized, which would generally result in error messages such as:

  ```
  <app-dir>/node_modules/@temporalio/client/src/types.ts:62
  export const { WorkflowService } = proto.temporal.api.workflowservice.v1;
                                                    ^
  TypeError: Cannot read properties of undefined (reading 'api')
  ```

  We now check `constructor.className` instead, which is a convenient property added by the protobufjs library, specifically for the purpose of reflection.

## Note

- We provide no guarantee regarding the inclusion of `@temporalio/*` packages as part of bundled and/or minimized applications. We do not test such usage at this time. We strongly recommend configuring your bundler to treat these packages as external.

### DIFF
--- a/packages/common/src/converter/protobuf-payload-converters.ts
+++ b/packages/common/src/converter/protobuf-payload-converters.ts
@@ -193,7 +193,7 @@ function isProtobufType(type: unknown): type is Type {
   return (
     isRecord(type) &&
     // constructor.name may get mangled by minifiers; thanksfuly protobufjs also sets a className property
-    (type.constructor.name === 'Type' || (type.constructor as any).className === 'Type') &&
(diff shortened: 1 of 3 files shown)


bergundy (approved):
Let's always use `className` but otherwise LGTM.
Thanks!

iamnafets:
Wow this bug driving me absolutely crazy, thanks for the PR.

PR #1281, opened 2023-11-08
fix: make it possible to omit port from connection address

kaiba42:
## What was changed
Allows users to specify that they would like to omit the port entirely from the address used in the Connection constructor.
Users can now specify an address of the form `temporal.svc.mydomain.com:null` which will translate to the address `temporal.svc.mydomain.com`.

## Why?
Currently it is impossible to specify an address for use in the Connection constructor that omits the port entirely. As a user, if I try to specify the address `temporal.svc.mydomain.com`, it will translate to the address `temporal.svc.mydomain.com:7233` with the default port of 7233 set.

This can break network connectivity for load balancer & ingress configurations that rely on raw host names without ports.
> Note: The Temporal Python SDK does not have this behavior, and allows users to specify hosts without ports.

## Checklist
None

1. How was this tested:
I changed the code on a local copy, and saw that connectivity from my temporal client to my temporal server was restored. 

2. Any docs updates needed?

I'm not sure where to update documentation, but somewhere it should be documented that users can set the port to the string `"null"` to omit setting the port in the address for Connections.

### DIFF
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -149,7 +149,7 @@ function normalizeGRPCConfig(options?: ConnectionOptions): ConnectionOptions {
     // eslint-disable-next-line prefer-const
     let [host, port] = rest.address.split(':', 2);
     port = port || '7233';
-    rest.address = `${host}:${port}`;
+    rest.address = port === "null" ? host : `${host}:${port}`;
   }
   const tls = normalizeTlsConfig(tlsFromConfig);
   if (tls) {


mjameswh:
This doesn't make sense to me, and it doesn't work if I test that locally. The `${host}:${port}` syntax is the only way to provide a port number to the gRPC for NodeJS library. There is no distinct `port` option, and the place you modified is the only one that fills in a default port of 7233. gRPC's default port is 443, not 7233.

Can you please give a bit more specific details on how the current behavior is breaking? Could it be that you are trying to connect through a generic HTTP or gRPC proxy that listens on port 443 rather than 7233? If so, you may simply write your server address as `${host}:443`. If there's a concern regarding the mTLS certificate and/or SNI name, then it should be possible to fix this by setting the `tls.serverNameOverride` connection property.

bergundy:
Temporal's default port is `7233` and what the SDK defaults to.
If you to use a different port you should set it explicitly.
I don't think we should accept `host:null` which is not a valid address.

kaiba42:
hey apologies for the delay in response. Happy to give more clarity.

I'm running a kubernetes cluster using emissary-ingress. I use emissary-ingress to route requests to my temporal server, which I have created a `Mapping` for at `[temporal.svc.mydomain.com](http://temporal.svc.mydomain.com/)`. The kubernetes resource yaml for this `Mapping` is:
```yaml
apiVersion: [getambassador.io/v2](http://getambassador.io/v2)
kind: Mapping
metadata:
  creationTimestamp: "2023-10-24T08:25:38Z"
  generation: 1
  labels:
    [cue.contrib.flux.io/name](http://cue.contrib.flux.io/name): temporal
    [cue.contrib.flux.io/namespace](http://cue.contrib.flux.io/namespace): flux-system
  name: temporal
  namespace: temporal
  resourceVersion: "2111"
spec:
  ambassador_id:
  - --apiVersion-v3alpha1-only--default
  bypass_auth: true
  grpc: true
  host: [temporal.svc.mydomain.com](http://temporal.svc.mydomain.com/)
  prefix: /temporal.api.workflowservice.v1.WorkflowService/
  rewrite: /temporal.api.workflowservice.v1.WorkflowService/
  service: temporal-frontend.temporal:7233
  timeout_ms: 180000
```
which routes incoming requests to this service, created by the temporal helm chart:
```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    meta.helm.sh/release-name: temporal
    meta.helm.sh/release-namespace: temporal
  creationTimestamp: "2023-10-24T08:25:43Z"
  labels:
    [app.kubernetes.io/component](http://app.kubernetes.io/component): frontend
    [app.kubernetes.io/instance](http://app.kubernetes.io/instance): temporal
    [app.kubernetes.io/managed-by](http://app.kubernetes.io/managed-by): Helm
    [app.kubernetes.io/name](http://app.kubernetes.io/name): temporal
    [app.kubernetes.io/part-of](http://app.kubernetes.io/part-of): temporal
    [app.kubernetes.io/version](http://app.kubernetes.io/version): 1.21.5
    helm.sh/chart: temporal-0.27.0
    [helm.toolkit.fluxcd.io/name](http://helm.toolkit.fluxcd.io/name): temporal
    [helm.toolkit.fluxcd.io/namespace](http://helm.toolkit.fluxcd.io/namespace): temporal
  name: temporal-frontend
  namespace: temporal
  resourceVersion: "2374"
spec:
  clusterIP: 10.43.000.000
  clusterIPs:
  - 10.43.000.000
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: grpc-rpc
    port: 7233
    protocol: TCP
    targetPort: rpc
  selector:
    [app.kubernetes.io/component](http://app.kubernetes.io/component): frontend
    [app.kubernetes.io/instance](http://app.kubernetes.io/instance): temporal
    [app.kubernetes.io/name](http://app.kubernetes.io/name): temporal
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```
Here are the error logs I'm getting with the SDK as is.
Running this code:
```javascript
const address = "[temporal.svc.mydomain.com](http://temporal.svc.mydomain.com/)"
const connection = await Connection.connect({ address });
const client = new TemporalClient({ connection });
});
```
I get:
```bash
/project/node_modules/@grpc/grpc-js/src/client.ts:176
        callback(new Error('Failed to connect before the deadline'));
                 ^
Error: Failed to connect before the deadline
    at checkState (/project/node_modules/@grpc/grpc-js/src/client.ts:176:18)
    at Timeout._onTimeout (/project/node_modules/@grpc/grpc-js/src/channel.ts:745:9)
    at listOnTimeout (node:internal/timers:569:17)
    at processTimers (node:internal/timers:512:7)
error Command failed with exit code 1.
```

In contrast, when I run the Python SDK with the same configuration, I connect to the Temporal server and can run workflows successfully.

The code change in this PR definitely feels hacky, but I wanted to make sure that it is not a breaking change. It's effectively an escape hatch that let's me mimic the behavior of the Python SDK.

If you see an issue with my ingress configuration that would remedy this issue, I'm open to changing my ingress config, but my first instinct was to make the Typescript and Python SDKs give the same behavior with the same config.

kaiba42:
@mjameswh what was the error you saw when running this locally that failed?

PR #1283, opened 2023-11-08
chore: Deprecate `defaultSinks`

mjameswh:
## What changed

- Deprecate `defaultSink()` and `LoggerSinks`. SDK's logger sink is now forcibly injected. Registering a custom logger sink named `defaultWorkerLogger` is still supported, but discouraged and deprecated.
- Refactored sink code internally.

## Why

  - Promoting clear usage patterns regarding Workflow logging that works out-of-the-box;
  - Discourage the use of insufficiently documented features (eg. possibility of redirecting workflow logs by registering a custom `defaultWorkerLogger` sink);
  - Remove the requirement that users should call `defaultSinks()` if they were to add custom sinks, which was error prone and a recurring source of support question.

### DIFF
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -9,6 +9,7 @@ import { ActivityInboundLogInterceptor } from './activity-log-interceptor';
 import { NativeConnection } from './connection';
 import { WorkerInterceptors } from './interceptors';
 import { Logger } from './logger';
+import { initLoggerSink } from './workflow/logger';
 import { Runtime } from './runtime';
 import { InjectedSinks } from './sinks';
 import { MiB } from './utils';
@@ -444,10 +445,9 @@ export interface WorkerOptions {
    * guarantees, please consider using local activities instead. For use cases that require
    * _exactly-once_ or _at-most-once_ execution guarantees, please consider using regular activities.
    *
-   * The SDK itself may register sinks functions required to support workflow features. At the moment, the only such
-   * sink is 'defaultWorkerLogger', which is used by the workflow context logger (ie. `workflow.log.info()` and
-   * friends); other sinks may be added in the future. You may override these default sinks by explicitely registering
-   * sinks with the same name.
+   * The SDK itself may register sinks functions required to support workflow features. The name,
+   * signature and semantic of these sinks functions is considered an internal detail and may change
+   * in the future without notice. Please do not use them directly nor override them.
    */
   sinks?: InjectedSinks<any>;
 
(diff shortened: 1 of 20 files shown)


packages/worker/src/worker-options.ts
```diff
@@ -444,10 +445,9 @@ export interface WorkerOptions {
...
    *
-   * The SDK itself may register sinks functions required to support workflow features. At the moment, the only such
-   * sink is 'defaultWorkerLogger', which is used by the workflow context logger (ie. `workflow.log.info()` and
-   * friends); other sinks may be added in the future. You may override these default sinks by explicitely registering
-   * sinks with the same name.
+   * The SDK itself may register sinks functions required to support workflow features. The name,
+   * signature and semantic of these sinks functions is considered an internal detail and may change
+   * in the future without notice. Please do not use them directly nor override them.
    */
   sinks?: InjectedSinks<any>;
```
Sushisource, line 452: Should we perhaps put the internal sinks in a different, unexposed field so we don't need this disclaimer in the docstring?

packages/workflow/src/sinks.ts
```diff
@@ -44,14 +45,64 @@ export interface SinkCall {
...
+            get(_, fnName) {
+              return (...args: any[]) => {
+                const activator = assertInWorkflowContext(
+                  'Proxied sinks functions may only be used from a Workflow Execution.'
+                );
+                activator.sinkCalls.push({
+                  ifaceName: ifaceName as string,
+                  fnName: fnName as string,
+                  // Sink function don't get called immediately. Make a clone of sink's args, so that further mutations
+                  // to these objects don't corrupt the args that the sink function will receive. Only available from node 17.
```
Sushisource, line 94:
```suggestion
                  // Sink function doesn't get called immediately. Make a clone of the sink's args, so that further mutations
                  // to these objects don't corrupt the args that the sink function will receive. Only available from node 17.
```

Sushisource (approved):
LGTM. Comments aren't blocking.

packages/worker/src/worker-options.ts
mjameswh: No. The reason behind that disclaimer is that previously and until now, user overriding our default sinks (specifically `defaultWorkerLogger`) was allowed, documented, and even encouraged. So this is really a change of direction. Therefore, I want to preserve backward compatibility for code that was using that approach, but strongly discourage that practice going forward.

New internal sinks that we'll add from now on will use names that clearly indicate they are meant for internal usage only (eg. `__temporal_metrics` or similar), and their interface will not be exposed publicly, making it very unlikely that someone would consider overriding these as a good idea.

Sushisource: Works for me

packages/worker/src/worker-options.ts
```diff
@@ -444,10 +445,9 @@ export interface WorkerOptions {
    * guarantees, please consider using local activities instead. For use cases that require
    * _exactly-once_ or _at-most-once_ execution guarantees, please consider using regular activities.
    *
-   * The SDK itself may register sinks functions required to support workflow features. At the moment, the only such
-   * sink is 'defaultWorkerLogger', which is used by the workflow context logger (ie. `workflow.log.info()` and
-   * friends); other sinks may be added in the future. You may override these default sinks by explicitely registering
-   * sinks with the same name.
+   * The SDK itself may register sinks functions required to support workflow features. The name,
+   * signature and semantic of these sinks functions is considered an internal detail and may change
+   * in the future without notice. Please do not use them directly nor override them.
```
bergundy, line 450: I'd just say that the names can't start with `__temporal` here and change the built-in logger sink to use that.
If a user provides a `defaultWorkflowLogger` we'll still accept it and translate it to `__temporal_runtime_logger` or something.

bergundy (approved):
Overall LGTM, thanks for clearing up this mess.

PR #1284, opened 2023-11-08
chore: Deprecate `ActivityInboundLogInterceptor`

mjameswh:
## What changed

- Attributes from the current Activity context are now automatically included as metadata on every log entries by the Activity Worker, and some key events of the Activity's lifecycle will automatically be logged (at 'DEBUG' level for most messages; 'WARN' for failures). Logged Activity attributes can be customized by registering an `ActivityOutboundCallsInterceptor` that intercepts the `getLogAttributes()` method.

- Deprecated `ActivityInboundLogInterceptor`; registering such an interceptor with a custom logger will still work, but will prevent the use of the newly introduced `getLogAttributes()` interceptor. The same is true about modifying the context logger directly (eg. `context.log = myCustomLogger`).

- Deprecated `ActivityInboundCallsInterceptorFactory`. Activity Interceptors should now be instantiated using `ActivityInterceptorsFactory`.

- `MockActivityEnvironment` now accept a list of activity interceptors, and more closely match an actual activity execution.

## Why

  - Promoting clear usage patterns regarding Activity logging that works out-of-the-box;
  - Discourage the use of insufficiently documented features (eg. possibility of redirecting activity logs by explicitly setting `context.log = myCustomLogger`, or by registering an `ActivityInboundLogInterceptor` with a custom logger as argument);
  - Remove the requirement that users should call `appendDefaultInterceptors()` if they were to register custom interceptors, which was error prone and a recurring source of support question.


packages/common/src/converter/failure-converter.ts
```diff
@@ -34,10 +48,8 @@ export function cutoffStackTrace(stack?: string): string {
   const lines = (stack ?? '').split(/\r?\n/);
   const acc = Array<string>();
   lineLoop: for (const line of lines) {
```
bergundy, line 50: No need for this label any more.

packages/common/src/converter/failure-converter.ts
```diff
@@ -15,17 +15,31 @@ import {
...
   /** Workflow run anything in context */
-  /\s+at Script\.runInContext \((?:node:vm|vm\.js):\d+:\d+\)/,
-];
+  /\s+at Script\.runInContext \((?:node:vm|vm\.js):\d+:\d+\)/
+);
+
+/**
+ * Any stack trace frames that match any of those wil be ignored
+ */
+const IGNORED_STACK_FRAMES_PATTERNS = combineRegExp(
```
bergundy, line 37:
```suggestion
 * Any stack trace frames that match any of those will be dropped
 */
const DROPPED_STACK_FRAMES_PATTERNS = combineRegExp(
```

packages/test/src/test-activity-log-interceptor.ts
```diff
@@ -26,150 +22,166 @@ test.before(() => {
...
+    }
+  );
+  await env.run(async () => {
+    activity.log.debug('log message from activity');
+  });
+  const logs = (env.context as MyTestActivityContext).logs;
+  const activityLogEntry = logs.find((entry) => entry.message === 'log message from activity');
+  t.not(activityLogEntry, undefined);
+  t.is(activityLogEntry?.level, 'DEBUG');
+  t.is(activityLogEntry?.meta?.['custom'], 'attribute');
```
bergundy, line 186:
```suggestion
  t.is(activityLogEntry?.meta?.custom, 'attribute');
```

packages/testing/src/index.ts
```diff
@@ -387,6 +407,8 @@ export const defaultActivityInfo: activity.Info = {
   currentAttemptScheduledTimestampMs: 1,
 };
 
+const mockActivityContextStorage: AsyncLocalStorage<() => Promise<unknown>> = new AsyncLocalStorage();
```
bergundy, line 410: I like the idea of code reuse here but IMHO it's not worth the complication.
If all you want is to inject interceptors, I'd add that functionality here.

packages/worker/src/activity-log-interceptor.ts
```diff
@@ -1,50 +1,42 @@
...
+import { activityLogAttributes } from './activity';
+import { Runtime } from './runtime';
 
 /**
- * Returns a map of attributes to be set on log messages for a given Activity
+ * This interceptor was previously used to log Activity execution starts and their completions. It is now deprecated
+ * and behaves as a noop in most cases. It is only kept arround to avoid breaking code out there that was previously
+ * refering to it.
+ *
+ * @deprecated `ActivityInboundLogInterceptor` is deprecated. Activity life cycle events are now automatically logged
```
bergundy, line 12:
```suggestion
 * @deprecated `ActivityInboundLogInterceptor` is deprecated. Activity lifecycle events are now automatically logged
```

packages/worker/src/activity-log-interceptor.ts
```diff
@@ -1,50 +1,42 @@
...
+ *             by the SDK. To customize activity log attributes, register a custom `ActivityOutboundCallsInterceptor`
```
bergundy, line 13:
```suggestion
 *             by the SDK. To customize activity log attributes, register a custom {@link ActivityOutboundCallsInterceptor}
```

packages/worker/src/activity-log-interceptor.ts
```diff
@@ -59,27 +51,7 @@ export class ActivityInboundLogInterceptor implements ActivityInboundCallsInterc
...
-        this.ctx.log.debug('Activity completed', { durationMs });
-      } else if ((error instanceof CancelledFailure || isAbortError(error)) && this.ctx.cancellationSignal.aborted) {
-        this.ctx.log.debug('Activity completed as cancelled', { durationMs });
-      } else if (error instanceof CompleteAsyncError) {
-        this.ctx.log.debug('Activity will complete asynchronously', { durationMs });
-      } else {
-        this.ctx.log.warn('Activity failed', { error, durationMs });
-      }
-    }
+    // Logging of activity's life cycle events is now handled in `worker/src/activity.ts`
```
bergundy, line 54:
```suggestion
    // Logging of activity'slife cycle events is now handled in `worker/src/activity.ts`
```

packages/worker/src/interceptors.ts
```diff
@@ -44,6 +66,10 @@ export interface WorkerInterceptors {
    * List of factory functions returning {@link ActivityInboundCallsInterceptor}s
    */
   activityInbound?: ActivityInboundCallsInterceptorFactory[];
+  /**
+   * List of factory functions returning {@link ActivityOutboundCallsInterceptor}s
+   */
+  activityOutbound?: ActivityOutboundCallsInterceptorFactory[];
```
bergundy, line 72: Let's change this to:

```
activityInbound?: ActivityInboundCallsInterceptorFactory[]; // <-- deprecate this
activity?: ActivityInterceptorsFactory[]; <-- same as the workflow counterpart
```

bergundy: It'll help you reference the outbound from the inbound in case they need to interact (and reuse the same implementation for both).

packages/testing/src/index.ts
mjameswh: It's also to get proper lifecycle event logging, catching of cancellation error, correct stack trace truncation...

mjameswh: It should however be possible to avoid this `AyncLocalStorage` by passing the activity function to `Activity.runNoEncoding()`, and from there to `Activity.execute()`... Let me see if that turns out better.

packages/worker/src/interceptors.ts
mjameswh: Done

packages/interceptors-opentelemetry/src/worker/index.ts
```diff
@@ -17,6 +23,10 @@ export interface InterceptorOptions {
  * provided in the Activity input headers.
  */
 export class OpenTelemetryActivityInboundInterceptor implements ActivityInboundCallsInterceptor {
+  static asFactory(options?: InterceptorOptions): ActivityInterceptorsFactory {
+    return (ctx) => ({ inbound: [new OpenTelemetryActivityInboundInterceptor(ctx, options)] });
+  }
```
bergundy, line 28: I'd rather not expose this.

packages/test/src/activities/interceptors.ts
```diff
@@ -1,9 +1,23 @@
...
-import { ActivityExecuteInput, ActivityInboundCallsInterceptor, Next } from '@temporalio/worker';
+import {
+  ActivityExecuteInput,
+  ActivityInboundCallsInterceptor,
+  ActivityInterceptorsFactory,
+  Next,
+} from '@temporalio/worker';
 
 export class ConnectionInjectorInterceptor implements ActivityInboundCallsInterceptor {
+  public static asFactory(
```
bergundy, line 12: Don't love that we have this here but I don't mind since this is internal.

```suggestion
  public static createFactory(
```

packages/test/src/workflows/log-sink-tester.ts
```diff
@@ -34,3 +34,10 @@ export async function logSinkTester(): Promise<void> {
     }`
   );
 }
+
+// eslint-disable-next-line deprecation/deprecation
+const { defaultWorkerLogger } = wf.proxySinks<wf.LoggerSinks>();
+
+export async function useDepreatedLoggerSinkWorkflow(): Promise<void> {
+  defaultWorkerLogger.info(`Log message from workflow`, { workflowId: wf.workflowInfo().workflowId });
```
bergundy, line 42: nit:
```suggestion
  defaultWorkerLogger.info('Log message from workflow', { workflowId: wf.workflowInfo().workflowId });
```

bergundy: We should have a lint rule that prevents this for consistency (not sure if it affects performance even if it's irrelevant in this test).

packages/testing/src/index.ts
```diff
@@ -396,33 +401,31 @@ export const defaultActivityInfo: activity.Info = {
 export class MockActivityEnvironment extends events.EventEmitter {
   public cancel: (reason?: any) => void = () => undefined;
```
bergundy, line 402: Can we document that the `context` is shared for any activity that is run in this instance and that once `cancel` is called, any further activity runs in this env will be automatically cancelled?

packages/worker/src/interceptors.ts
```diff
@@ -31,19 +31,55 @@ export interface ActivityInboundCallsInterceptor {
...
  * Structure for passing in Worker interceptors via {@link WorkerOptions}
  */
 export interface WorkerInterceptors {
   /**
-   * List of factory functions returning {@link ActivityInboundCallsInterceptor}s
+   * List of factory functions that instanciate {@link ActivityInboundCallsInterceptor}s and
+   * {@link ActivityOutboundCallsInterceptor}s.
    */
-  activityInbound?: ActivityInboundCallsInterceptorFactory[];
+  activity?: ActivityInterceptorsFactory[];
```
bergundy, line 74: I think just a single factory is fine since each factory already returns an array of interceptors but I can see that this is consistent with workflows so I'm okay leaving it for the extra flexibility.

bergundy (approved):

packages/interceptors-opentelemetry/src/worker/index.ts
mjameswh: Oh, really? TBH I feel quite the opposite...

I want to avoid requiring users to know exactly which kinds of interceptors and sinks need to be instantiated for SDK provided optional features. That evolves badly, as one day, they may need only an activity inbound interceptor, but then, at some point in the future, they would also require an activity outbound interceptor. And even though `ConnectionInjectorInterceptor` is not a public feature, I think applying the same pattern internally make sense.

As it currently stand, I think that gives a reasonably clean API.

      activities,
      taskQueue: 'test-otel',
      interceptors: {
        workflowModules: [require.resolve('./workflows/otel-interceptors')],
        activity: [
          OpenTelemetryActivityInboundInterceptor.asFactory(),
          ConnectionInjectorInterceptor.asFactory(connection),
        ],
      },
      sinks,
    });

mjameswh: Now, that's essentially the same concern that led to the introduction of `WorkflowCoverage.augmentWorkerOptions` and friends, which is arguably an even more complete approach. The only thing I don't really like about it is the fact that it wraps around the whole `WorkerOptions`, which is kinda the objective of course, but has the unwanted side effect of creating some "distance" between the `Worker.create` and the actual options. But I can live with that.

Using that pattern instead of the `asFactory()` one I introduced previously, we could get something like this:

```
    const worker = await Worker.create(
      OpenTelemetryContextPropagator.augmentWorkerOptions(
        ConnectionInjector.augmentWorkerOptions({
          activities,
          taskQueue: 'test-otel',
        })
      )
    );
```

Or, what I consider to be the best of all solutions, something similar to this:

```
      activities,
      taskQueue: 'test-otel',
      extensions: [
        OpenTelemetryContextPropagator,
        ConnectionInjector,
      ]
    })
```

mjameswh: All of that to say that I'm open to reconsider this in the interest of improving DX. In the immediate, though, I'll move on with what I already proposed in this PR. We can revisit that next week.

packages/test/src/workflows/log-sink-tester.ts
mjameswh: >We should have a lint rule that prevents this for consistency (not sure if it affects performance even if it's irrelevant in this test).

FYI it has absolutely no performance impact. In V8, template strings get segmented at parse time. If the template string contains no expression segment, the parser emit a `StringLiteral` AST node rather than a `TemplateLiteral` node.

But yes, I agree with the consistency concern. Will add a eslint rule for that in a different PR, to avoid polluting this one with tons of insignificant quote changes.

packages/worker/src/interceptors.ts
mjameswh: Again, I think the opposite would be easier to work with, ie. accept an array of factory, but have each factory return at most one inbound and one outbound, but I went for consistency.

packages/interceptors-opentelemetry/src/worker/index.ts
mjameswh: Reverted `asFactory` for the moment. After thinking about this more, I really want to come up an approach that is not limited to activity interceptors, more in the spirit of the "extension" concept I mentioned. This will be done at a later point. In the mean time, it wouldn't be desirable to introduce a new approach (ie. `asFactory`) that would then be deprecated soon after.

packages/worker/src/interceptors.ts
mjameswh: Fixed.

PR #1286, opened 2023-11-09
docs: Change serverNameOverride to not be testing only

kollektiv:
## What was changed
Change serverNameOverride to not be testing only

## Why?
Using serverNameOverride is a legitimate use case in production cases

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/common/src/internal-non-workflow/tls-config.ts
+++ b/packages/common/src/internal-non-workflow/tls-config.ts
@@ -3,7 +3,6 @@ export interface TLSConfig {
   /**
    * Overrides the target name used for SSL host name checking.
    * If this attribute is not specified, the name used for SSL host name checking will be the host from {@link ServerOptions.url}.
-   * This _should_ be used for testing only.
    */
   serverNameOverride?: string;
   /**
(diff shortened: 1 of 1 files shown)


mjameswh:
I agree this is not only about testing, yet I think a word of caution could be pertinent as that option may have security impact. Please let me check first what we do in other SDKs…

packages/common/src/internal-non-workflow/tls-config.ts
```diff
@@ -3,7 +3,6 @@ export interface TLSConfig {
   /**
    * Overrides the target name used for SSL host name checking.
    * If this attribute is not specified, the name used for SSL host name checking will be the host from {@link ServerOptions.url}.
-   * This _should_ be used for testing only.
```
bergundy, line 6: Please mention when this is valid to use apart from testing.

mjameswh:
> I agree this is not only about testing, yet I think a word of caution could be pertinent as that option may have security impact. Please let me check first what we do in other SDKs…

AFAICS I can't find any explanation on equivalent options in other SDKs. Doesn't mean we shouldn't, only that we don't have a precedent. I think giving a rough idea of when this might be useful, including a few examples use cases.

kollektiv:
@bergundy @mjameswh Added an example into the docs. Wdyt?

bergundy (approved):

PR #1290, opened 2023-11-13
chore: Deprecate `WorkflowLogInterceptor`

mjameswh:
## What changed

- Attributes from the current Workflow are now automatically included as metadata on every log entries by the Workflow Worker, and some key events of the Workflow's lifecycle will automatically be logged (at 'DEBUG' level for most messages; 'WARN' for failures).

- Deprecated `WorkflowLogInterceptor`; it is now a no-op.

- Deprecated `appendDefaultInterceptors`. It is now a no-op in all cases, except when a custom logger is supplied, in which case it adds a `ActivityInboundLogInterceptor` to register that custom logger on Activity context.

## Why

  - Promoting clear usage patterns regarding Workflow logging that works out-of-the-box;
  - Remove the requirement that users should call `appendDefaultInterceptors()` if they were to register custom interceptors, which was error prone and a recurring source of support question.


packages/workflow/src/logs.ts
```diff
@@ -64,10 +67,52 @@ export const log: WorkflowLogger = Object.fromEntries(
...
+        } else if (error instanceof ContinueAsNew) {
+          log.debug('Workflow continued as new');
+          throw error;
+        }
+      }
+      log.warn('Workflow failed', { error });
+      throw error;
+    }
+  );
+  // Avoid showing this interceptor in stack trace query
```
bergundy, line 101:
```suggestion
  // Avoid showing this promise in stack trace query
```

bergundy (approved):

PR #1296, opened 2023-11-17
Delete .github/workflows/semgrep.yml

jackdawm:
## What was changed
Delete the Semgrep workflow file in this repo.

## Why?
Now that the Semgrep CI workflow is shared and runs on every repo, this per-repo file is no longer needed.

### How was this tested:

Saw the org-level action show up in the Actions tab in other repos and it runs successfully.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,9 +3,3 @@
 # @temporalio/sdk will be requested for review when
 # someone opens a pull request.
 * @temporalio/sdk
-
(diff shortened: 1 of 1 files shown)


krousey (approved):

mjameswh:
@jackdawm Can we also remove entries in [the CODEOWNER file](https://github.com/temporalio/sdk-typescript/blob/main/.github/CODEOWNERS#L7-L11)?

mjameswh (approved):

PR #1277, opened 2023-10-31
Workflow Update

dandavison:
Fixes https://github.com/temporalio/sdk-typescript/issues/1160

- Implement support for Workflow Update

### API

To see examples using the new API, see
- https://github.com/temporalio/features/pull/362
- `packages/test/src/test-integration-update-*`

Briefly:

```typescript
// An Update is defined similarly to Query and Signal. The second type parameter (input) is optional
const myUpdate = wf.defineUpdate<string, [string]>('myUpdate');

// A handler is a sync or async function; the optional validator is always sync.
wf.setHandler(myUpdate, handler, { validator });

// `await wfHandle.executeUpdate` to start an update and block until it completes or fails.
// The signature resembles `WorkflowClient.start` in that the arguments are supplied
// in the optional options object
// (as opposed to the variadic signatures of wfHandle.query and wfHandle.signal)
// An error is thrown if the RPC times out (we do not poll for this call)
// or if the validator rejects, or the handler throws ApplicationFailure.
const updateResult = await handle.executeUpdate(myUpdate, { args: ['a'] });

// `startUpdate` to block until Accepted and obtain a handle to the Update. Signature the same as `executeUpdate`.
const updateHandle = await handle.startUpdate(myUpdate, { args: ['a'] });
// Blocking call to fetch result, with polling.
// Validation failures or ApplicationFailure in the handler would throw here.
const updateResult = await updateHandle.result();
```

### How was this tested
   - features tests https://github.com/temporalio/features/pull/362
   - `packages/test/src/integration-tests/test-update*`

### Checklist

1. **Closes** https://github.com/temporalio/sdk-typescript/issues/1160

2. **Any docs updates needed?**
   Yes

3. **Not done in this PR:**
   - [ ] Test that non-ApplicationFailure error in handler fails the task
   - [ ] Tests involving continue-as-new (e.g. providing coverage for `firstExecutionRunId`)
   - [x] Support late registration
   - [ ] Fail task if workflow commands created during validation
   - [ ] Support setting a default update handler

PR #1299, opened 2023-11-21
Refactor integration tests

dandavison:
This is pure refactoring of the integration tests, done in preparation for implementation of Workflow Update in https://github.com/temporalio/sdk-typescript/pull/1277.


bergundy (approved):

bergundy:
Please fix the lint warnings before merging.

packages/test/src/integration-tests-old.ts
mjameswh, line 1: Can you please add a comment early in this file that explains how these tests are "old" (ie. they are still valid and up-to-date, but need to be converted to the new testing approach).

packages/test/src/integration-tests/helpers.ts
```diff
@@ -0,0 +1,105 @@
...
+  DefaultLogger,
+  LogLevel,
+  Runtime,
+  WorkerOptions,
+  WorkflowBundle,
+  bundleWorkflowCode,
+} from '@temporalio/worker';
+import * as workflow from '@temporalio/workflow';
+import { ConnectionInjectorInterceptor } from '../activities/interceptors';
+import { Worker, test as anyTest, bundlerOptions } from '../helpers';
```
mjameswh, line 19: Would it be difficult/disruptive to merge the content of this file into `../helpers.ts`?

mjameswh (approved):
I'm just a bit uncomfortable with file names (`-old` and `helpers`). I gave some details inline.

Not blocking however. Feel free to merge as is if you prefer. These can be addressed at a later point.

dandavison:
Thanks for reviewing. The file name in this PR I am most uncomfortable with is `test-integration-workflows.ts => integration-tests/test-misc.ts`! Please rename that if you can think of a better name. But I think the `-old` suffix is helpful. My objectives with this PR were to eliminate duplication of helper boilerplate, make it easy to discern which are our new-style integration tests, and which our old, while allowing the new-style integration tests to be spread over multiple files (e.g. for the coming update tests, which are in a few different files due to the use of difference interceptors). We don't have to keep the new-style ones in a directory; we could have `test-integration-*.ts` and `test-integration-old-*.ts`.

packages/test/src/integration-tests/helpers.ts
dandavison: That sounds good, I'm happy for any further renaming/reorganizing tweaks to be done here; I just wanted to start getting things in shape while adding new tests for Update.

Issue #1262, opened 2023-10-18
[Feature Request] Expose `StartWorkflowOptions.workflowStartDelay`


mjameswh:
### Describe the solution you'd like

- `workflowStartDelay` was introduced recently in server. Need to expose it in `StartWorkflowOptions.workflowStartDelay`

cretz:
I recommend calling this just `startDelay`

PR #1300, opened 2023-11-21
feat(client): Add startDelay to workflow (signal) start options

antlai-temporal:
## What was changed
 The Temporal server introduced the ability to delay the start of a workflow using the attribute `startDelay` in version 1.21. 

This PR enables it for the TypeScript SDK.  

There are also some format changes to Rust files in `core-bridge` after using `npm run format`. They were missing in previous commits because npm run lint.check does not validate Rust files with `cargo fmt`.

1. Closes 
#1262 
2. How was this tested:
  Two system tests added

### DIFF
--- a/packages/client/src/workflow-options.ts
+++ b/packages/client/src/workflow-options.ts
@@ -1,10 +1,7 @@
-import {
-  CommonWorkflowOptions,
-  SignalDefinition,
-  WithCompiledWorkflowOptions,
-  WithWorkflowArgs,
-  Workflow,
-} from '@temporalio/common';
+import { CommonWorkflowOptions, SignalDefinition, WithWorkflowArgs, Workflow } from '@temporalio/common';
+import { Duration, msOptionalToTs } from '@temporalio/common/lib/time';
+import { Replace } from '@temporalio/common/lib/type-helpers';
+import { google } from '@temporalio/proto';
 
 export * from '@temporalio/common/lib/workflow-options';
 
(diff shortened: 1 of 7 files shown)


packages/test/src/test-integration-workflows.ts
```diff
@@ -282,3 +283,34 @@ test('Activity initialInterval is not getting rounded', async (t) => {
...
+
+test('Start of workflow is delayed', async (t) => {
+  const { startWorkflow } = helpers(t);
+  // This workflow never runs
+  const handle = await startWorkflow(runTestActivity, {
+    startDelay: '5678s',
+  });
+  const { events } = await handle.fetchHistory();
+  const workflowExecutionStartedEvent = events?.find((ev) => ev.workflowExecutionStartedEventAttributes);
+  console.log(workflowExecutionStartedEvent);
```
dandavison, line 295: I think you meant to delete the `console.log` calls here and below.

packages/test/src/test-integration-workflows.ts
```diff
@@ -282,3 +283,34 @@ test('Activity initialInterval is not getting rounded', async (t) => {
...
+  const startDelay = workflowExecutionStartedEvent?.workflowExecutionStartedEventAttributes?.firstWorkflowTaskBackoff;
```
dandavison, line 296: nit: you could use `!` instead of `?` here and below since the assertion on the next line clearly expects it not to be `undefined`.

packages/test/src/test-integration-workflows.ts
```diff
@@ -282,3 +283,34 @@ test('Activity initialInterval is not getting rounded', async (t) => {
...
+    signal: workflows.interruptSignal,
+    signalArgs: ['Never called'],
+  });
+
+  const { events } = await handle.fetchHistory();
+  const workflowExecutionStartedEvent = events?.find((ev) => ev.workflowExecutionStartedEventAttributes);
+  console.log(workflowExecutionStartedEvent);
+  const startDelay = workflowExecutionStartedEvent?.workflowExecutionStartedEventAttributes?.firstWorkflowTaskBackoff;
+  t.is(tsToMs(startDelay), 4678000);
+});
```
dandavison, line 316: I see that the test failures in CI also occur locally.

dandavison (changes requested):
Just some minor comments, let me know when it's ready for another review.

packages/client/src/workflow-options.ts
```diff
@@ -37,6 +34,36 @@ export interface WorkflowOptions extends CommonWorkflowOptions {
    * @default true
    */
   followRuns?: boolean;
+
+  /**
+   * Amount of time to wait before starting the workflow.
+   * This is ignored within a Schedule, as expected by {@link ScheduleClient.create}.
```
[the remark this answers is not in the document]
mjameswh: You can remove that second sentence entirely. `ScheduleOptionsStartWorkflowAction` already pick only pertinent props from `WorkflowStartOptions`. Just make sure that this prop is effectively absent from `ScheduleOptionsStartWorkflowAction` (should be, but in case...)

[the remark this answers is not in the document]
antlai-temporal: Yes, it was just that the programmer maybe confused if they set it, and they assume the schedule will also be delayed... 
Anyway, maybe more confusing than helpful, deleting...

packages/test/src/test-integration-workflows.ts
antlai-temporal: Oops... Thanks

packages/test/src/test-integration-workflows.ts
antlai-temporal: I'm not sure that is true, `tsToMs` already takes care of the case when is `undefined`, and if the test failed, it will be `undefined`, and we only know that at runtime...

mjameswh (approved):
Looks good, thansk!

PR #1305, opened 2023-11-27
Refactor rxjs activation processing

dandavison:
This PR refactors an important function that has high complexity and is also long (250 lines), and contains several important sub-components that would benefit from being given their own names and abstracted out as separate functions. (Although I wasn't familiar with rxjs, I'm not critizing its use here -- I believe we are getting a lot of leverage out of it, and that it is aiding rather than hindering comprehension.)

The starting point was
```typescript
workflowOperator(): OperatorFunction<WorkflowActivation, Uint8Array>
```
Its job is to create an rxjs operator to perform the worker's processing of all activations (from all workflows).

We break `workflowOperator` up as follows:

_(Choosing good names is important if this change is to be useful. I'd love to hear any alternative suggestions regarding the names.)_

First, we move most of `workflowOperator` into the main top-level polling operator, so that it contains a very short high-level plan of the computation. I.e. it takes in the stream of _all_ activations, groups them according to the workflow execution they belong to, and hands off processsing of the substreams:

```typescript
  /**
   * Poll for Workflow activations, handle them, and report completions.
   */
  protected workflow$(): Observable<void> {
    // ...
    return this.workflowPoll$().pipe(
      closeableGroupBy((activation) => activation.runId),
      mergeMap(this.handleWorkflowActivations.bind(this)),
      mergeMap(async (completion) => {
        await this.nativeWorker.completeWorkflowActivation(completion.buffer.slice(completion.byteOffset));
      }),
     // ...
    );
  }
```

Next `handleWorkflowActivations` is responsible for all activations from a single workflow execution:

```typescript
  /**
   * Process activations from the same workflow execution to an observable of completions.
   */
  protected handleWorkflowActivations(
    activations$: CloseableGroupedObservable<string, coresdk.workflow_activation.WorkflowActivation>
  ): Observable<{
    completion?: Uint8Array | undefined;
    close: boolean;
  }>
...
    return merge(activations$$, evictions$).pipe(
      // ...
      mergeMapWithState(this.handleActivation.bind(this), undefined),
...
// (34 lines)  
```

Then `handleActivation` is responsible for handling an individual activation. It contains a lot of non-trivial error handling logic, and passes off the actual work to a helper function:

```typescript
  /**
   * Process a single activation to a completion.
   */
  protected async handleActivation(
    workflow: WorkflowWithLogAttributes | undefined,
    { activation, synthetic }: { activation: coresdk.workflow_activation.WorkflowActivation; synthetic: boolean }
  ): Promise<{
    state: WorkflowWithLogAttributes | undefined;
    output: { completion?: Uint8Array; close: boolean };
  }>
...  
          workflow = await this.createWorkflow(activation, maybeStartWorkflow.startWorkflow || undefined);
...
// (97 lines)
```

Finally, `createWorkflow` is the helper function used by `handleActivation`.

```typescript
  protected async createWorkflowWithLogAttributes(
    activation: coresdk.workflow_activation.WorkflowActivation,
    startWorkflow?: coresdk.workflow_activation.IStartWorkflow
  ): Promise<WorkflowWithLogAttributes>
// (98 lines)  
```

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -60,7 +60,7 @@ import {
   ReplayResult,
 } from './replay';
 import { History, Runtime } from './runtime';
-import { closeableGroupBy, mapWithState, mergeMapWithState } from './rxutils';
+import { CloseableGroupedObservable, closeableGroupBy, mapWithState, mergeMapWithState } from './rxutils';
 import { byteArrayToBuffer, convertToParentWorkflowType } from './utils';
 import {
   addDefaultWorkerOptions,
@@ -110,6 +110,7 @@ type Promisify<T> = T extends (...args: any[]) => void
   : never;
 
 type PatchJob = NonNullableObject<Pick<coresdk.workflow_activation.IWorkflowActivationJob, 'notifyHasPatch'>>;
+type StartWorkflowJob = NonNullableObject<Pick<coresdk.workflow_activation.IWorkflowActivationJob, 'startWorkflow'>>;
 
 type WorkflowActivation = coresdk.workflow_activation.WorkflowActivation;
 
@@ -175,6 +176,11 @@ export interface NativeReplayHandle {
 export interface WorkerConstructor {
   create(connection: NativeConnection, options: CompiledWorkerOptionsWithBuildId): Promise<NativeWorkerLike>;
   createReplay(options: CompiledWorkerOptionsWithBuildId): Promise<NativeReplayHandle>;
+}
+
+interface WorkflowWithLogAttributes {
+  workflow: Workflow;
+  logAttributes: Record<string, unknown>;
 }
 
 function isOptionsWithBuildId<T extends CompiledWorkerOptions>(options: T): options is T & { buildId: string } {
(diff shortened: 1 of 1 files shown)


packages/worker/src/worker.ts
```diff
@@ -993,260 +998,261 @@ export class Worker {
...
+              runId: activation.runId,
+              successful: {},
+            }).finish();
+        return { state: undefined, output: { close, completion } };
+      }
+
+      if (state === undefined) {
+        // Find a workflow start job in the activation jobs list
+        const maybeStartWorkflow = activation.jobs.find((j) => j.startWorkflow);
+        if (maybeStartWorkflow !== undefined) {
```
mjameswh, line 1098: Add the following to the top of this file (next to the `PatchJob` type):

```
type StartWorkflowJob = NonNullableObject<Pick<coresdk.workflow_activation.IWorkflowActivationJob, 'startWorkflow'>>;
```

then modify this line to:

```
        const maybeStartWorkflow = activation.jobs.find((j): j is StartWorkflowJob => j.startWorkflow != null);
```

By doing this, when you get you will know that `maybeStartWorkflow.startWorkflow` is defined.

packages/worker/src/worker.ts
```diff
@@ -993,260 +998,261 @@ export class Worker {
...
+          state = await this.getWorkflowState(activation, maybeStartWorkflow.startWorkflow || undefined);
+        } else {
```
mjameswh, line 1100: Please rename that function. It is not about getting a workflow's state, it is about instantiating the workflow execution context (choose the word you want: context/sandbox/isolate). Also, at this point, `maybeStartWorkflow.startWorkflow` can't be undefined (it was checked the line before).

dandavison: Thanks. I've changed the function name to `createWorkflowWithLogAttributes()` and I've renamed the local variable `state` to `workflow`.

mjameswh:
Thanks for the initiative. That logic is indeed quite difficult to get into. And more importantly, the way it is designed makes it very difficult to validate as a programmer.

TBH I'm not personally convinced that rxjs is helping enough to justify the code complexity it introduces. We have recently refactored other parts of the SDK to rewrite rxjs based algorithms using more traditional approaches (mainly pure loops, try/catch, promises), and generator functions or EventEmitter for the few cases where we needed multiplexing or merging event streams. Up to this point, this has resulted in a bit more code, yet significantly easier to understand and debug. But we don't need to go there immediately.

I agree that splitting the existing code can help a lot. I think however that there might be better split points. See inlined comments for details

packages/worker/src/worker.ts
```diff
@@ -993,260 +998,248 @@ export class Worker {
...
-              group$.close();
-              this.numCachedWorkflowsSubject.next(this.numCachedWorkflowsSubject.value - 1);
-            }
-          }),
-          takeWhile(({ close }) => !close, true /* inclusive */)
-        );
+  protected handleWorkflowActivations(
+    activations$: CloseableGroupedObservable<string, coresdk.workflow_activation.WorkflowActivation>
+  ): Observable<Uint8Array> {
+    const evictions$ = this.workflowPollerStateSubject.pipe(
```
mjameswh, line 1006: @bergundy We still have a pending discussion regarding synthetic evictions. WDYT?

packages/worker/src/worker.ts
```diff
@@ -1457,7 +1463,7 @@ export class Worker {
       return EMPTY;
     }
     return this.workflowPoll$().pipe(
-      this.workflowOperator(),
+      this.activationsOperator(),
       mergeMap(async (completion) => {
```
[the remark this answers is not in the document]
dandavison: > Or actually, I think I would inline part of activationOperator here

This is a nice improvement, thanks! It makes the top-level logic clearer like you say (and also deletes 15 lines).

packages/worker/src/worker.ts
mjameswh: @dandavison Supposing that we keep synthetic evictions, please rename to `syntheticEvictionActivations$` or similar. Also, add a comment above that explains that this pipe injects a synthetic eviction activation when the wf worker transition to no longer polling. I know there's a comment below, but I feel its not doing a job at helping reader understand the reason of this whole chain.

dandavison: Renamed and added comment.

packages/worker/src/worker.ts
dandavison: Thanks! Type guards are one of several things I've learned about Typescript from working on this codebase.  Maybe double check that https://github.com/temporalio/sdk-typescript/commit/3c74d2a044f5642dbb1bdedfa098a1cf4c2eaaed does what you intended.

Sushisource:
Thanks for doing this Dan, I've always found this section a bit confusing as well.

> TBH I'm not personally convinced that rxjs is helping enough to justify the code complexity it introduces.

I generally agree. I find it has quite surprising behavior every time I end up needing to dig into it. That said I also agree it's not worth going in and changing everything just to change it.

I don't have any detailed comment on the review, but, just chiming in I support the idea.

mjameswh:
Not approving yet, I want to get a second look at all of this tomorrow or Thursday, with a fresh mind.

bergundy:
Thanks @dandavison I tend to agree that rxjs is overkill for the simple worker logic. I had a plan to get rid of it a while back but it was never a priority.
Looks like you have enough eyes on this, deferring to @mjameswh for review.

bergundy:
I'd also support closing this and removing rxjs entirely if you feel you have time for that.

dandavison:
I don't have an opinion yet on whether we should use rxjs, and I think I'd be happy to defer to the original authors of the sdk on that question. I think though that the restructuring here is a helpful first step towards any subsequent replacement of rxjs, so I vote that we do merge this restructuring.

mjameswh:
Rewriting all of this to non-rxjs wouldn't be exceedingly difficult nor time-consuming, but I don't think this should be our first step.

I mean, the number one issue with that file IMO is low cohesion. It contains very specialized logics for too many different things. For that reason, I think we can significantly cut down on tech debt, with minimal effort, by simply splitting this into smaller, more tractable units. Introducing functions in correct places is a good step in that direction. Extracting a `WorkflowWorker` and an `ActivityWorker` could also help. And possibly a `WorkflowExecutionSupervisor` (whatever the name). Technically, our rxjs pipeline already adheres pretty much to that structure, it just makes it very hard to see and think about it.

Once we have isolated those units, then we can consider rewriting to non-rxjs, one unit at a time.

packages/worker/src/worker.ts
mjameswh: Just `createWorkflow()`. The `WithLogAttributes` is a pretty insignificant by-product.

dandavison: Done!

packages/worker/src/worker.ts
```diff
@@ -993,260 +999,246 @@ export class Worker {
...
+    const workflowCreator = this.workflowCreator!;
+    if (
+      !(
+        startWorkflow.workflowId != null &&
+        startWorkflow.workflowType != null &&
+        startWorkflow.randomnessSeed != null &&
+        startWorkflow.firstExecutionRunId != null &&
+        startWorkflow.attempt != null &&
+        startWorkflow.startTime != null
+      )
```
bergundy, line 1156: Can invert this whole condition.

packages/worker/src/worker.ts
```diff
@@ -993,260 +999,246 @@ export class Worker {
...
+    const workflowInfo: WorkflowInfo = {
+      workflowId,
+      runId: activation.runId,
+      workflowType,
+      searchAttributes:
+        (mapFromPayloads(searchAttributePayloadConverter, searchAttributes?.indexedFields) as SearchAttributes) || {},
+      memo: await decodeMapFromPayloads(this.options.loadedDataConverter, memo?.fields),
+      parent: convertToParentWorkflowType(parentWorkflowInfo),
+      lastResult: await decodeFromPayloadsAtIndex(this.options.loadedDataConverter, 0, lastCompletionResult?.payloads),
+      lastFailure: await decodeOptionalFailureToOptionalError(this.options.loadedDataConverter, continuedFailure),
```
bergundy, line 1193: nit: if you're refactoring this, you can use `Promise.all` to run the codecs concurrently.

mjameswh: Very good point, though that might also be addressed in a distinct PR. There are a few other places with the same issue. Should be easy to fix all of these in a single PR.

dandavison: Good point and I also vote let's do it in a follow-on PR. Although it's not a big thing, this PR is already touching a lot of critical and somewhat complex implementation, involving a library that we are slightly uncomfortable with in this context. So, in the event that we do discover some unexpected behavior, I would like us to have as few lines of investigation as possible.

packages/worker/src/worker.ts
dandavison: Also here, I vote we make deMorgan happy in a follow-on. In the PR so far I basically haven't changed anything other than the restructuring.

bergundy (approved):
Overall LGTM.

Thanks!

dandavison:
I've rebased this PR on main and resolved the merge conflict.

For the record: this PR conficted with https://github.com/temporalio/sdk-typescript/pull/1310. which made various changes to packages/worker/src/worker.ts but of those, the only change that altered the AST was changing the default value of `compiledOptions.reuseV8Context` to `true`.

PR #1307, opened 2023-12-02
Strengthen signal and query handler types to catch empty-argument signatures

dandavison:
These should both be type errors, but currently they are not:
```ts
  const signal = defineSignal<[string]>('a');
  wf.setHandler(signal, () => {});

  const query = defineQuery<string, [string]>('a');
  wf.setHandler(query, (): string => '');
```

I think that was unintentional, and this PR fixes it (but it is a breaking change).

### DIFF
--- a/packages/test/src/test-interface-type-safety.ts
+++ b/packages/test/src/test-interface-type-safety.ts
@@ -76,7 +76,7 @@ test('Signal handler type safety', (t) => {
   // @ts-expect-error signal handler must take string argument
   wf.setHandler(signal, (arg: number) => {});
 
-  // FIXME: @ts-expect-error signal handler must take string argument
+  // @ts-expect-error signal handler must take string argument
   wf.setHandler(signal, () => {});
 
   // @ts-expect-error signal handler must return void
@@ -93,7 +93,7 @@ test('Query handler type safety', (t) => {
   // @ts-expect-error query handler argument type must match
   wf.setHandler(query, (arg: number): string => '');
 
-  // FIXME: @ts-expect-error query handler argument type must match
+  // @ts-expect-error query handler argument type must match
   wf.setHandler(query, (): string => '');
 
   // @ts-expect-error query handler return type must match
--- a/packages/test/src/test-interface-type-safety.ts
+++ b/packages/test/src/test-interface-type-safety.ts
@@ -71,16 +71,16 @@ test('Can call signal on any WorkflowHandle', async (t) => {
 test('Signal handler type safety', (t) => {
   const signal = defineSignal<[string]>('a');
 
-  wf.setHandler(signal, (arg: string): void => {});
+  wf.setHandler(signal, (_arg: string): void => {});
 
   // @ts-expect-error signal handler must take string argument
-  wf.setHandler(signal, (arg: number) => {});
+  wf.setHandler(signal, (_arg: number) => {});
 
   // @ts-expect-error signal handler must take string argument
   wf.setHandler(signal, () => {});
 
   // @ts-expect-error signal handler must return void
-  wf.setHandler(signal, (arg: string): string => '');
+  wf.setHandler(signal, (_arg: string): string => '');
 
   t.pass();
 });
@@ -88,19 +88,19 @@ test('Signal handler type safety', (t) => {
 test('Query handler type safety', (t) => {
   const query = defineQuery<string, [string]>('a');
 
-  wf.setHandler(query, (arg: string): string => '');
+  wf.setHandler(query, (_arg: string): string => '');
 
   // @ts-expect-error query handler argument type must match
-  wf.setHandler(query, (arg: number): string => '');
+  wf.setHandler(query, (_arg: number): string => '');
 
   // @ts-expect-error query handler argument type must match
   wf.setHandler(query, (): string => '');
 
   // @ts-expect-error query handler return type must match
-  wf.setHandler(query, (arg: string): void => {});
+  wf.setHandler(query, (_arg: string): void => {});
 
   // @ts-expect-error query handler return type must match
-  wf.setHandler(query, (arg: string): number => 7);
+  wf.setHandler(query, (_arg: string): number => 7);
 
   t.pass();
 });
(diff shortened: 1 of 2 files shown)


packages/test/src/test-interface-type-safety.ts
```diff
@@ -66,3 +67,40 @@ test('Can call signal on any WorkflowHandle', async (t) => {
...
+test('Signal handler type safety', (t) => {
+  const signal = defineSignal<[string]>('a');
+
+  wf.setHandler(signal, (arg: string): void => {});
+
+  // @ts-expect-error signal handler must take string argument
+  wf.setHandler(signal, (arg: number) => {});
+
+  // @ts-expect-error signal handler must take string argument
+  wf.setHandler(signal, () => {});
```
dandavison, line 80: This one fails before this PR; the others pass.

packages/test/src/test-interface-type-safety.ts
```diff
@@ -66,3 +67,40 @@ test('Can call signal on any WorkflowHandle', async (t) => {
...
+test('Query handler type safety', (t) => {
+  const query = defineQuery<string, [string]>('a');
+
+  wf.setHandler(query, (arg: string): string => '');
+
+  // @ts-expect-error query handler argument type must match
+  wf.setHandler(query, (arg: number): string => '');
+
+  // @ts-expect-error query handler argument type must match
+  wf.setHandler(query, (): string => '');
```
dandavison, line 97: This one fails before this PR; the others pass.

packages/workflow/src/interfaces.ts
```diff
@@ -438,10 +438,10 @@ export type Handler<
...
   Args extends any[],
   T extends SignalDefinition<Args> | QueryDefinition<Ret, Args>
-> = T extends SignalDefinition<infer A>
-  ? (...args: A) => void | Promise<void>
-  : T extends QueryDefinition<infer R, infer A>
-  ? (...args: A) => R
+> = T extends SignalDefinition<Args>
+  ? (...args: Args) => void | Promise<void>
+  : T extends QueryDefinition<Ret, Args>
+  ? (...args: Args) => Ret
```
dandavison, line 444: `infer` was unnecessary here, since the type parameters in the true branch of the conditional are not narrowed w.r.t. to the input type parameters. And in fact, `infer` was harmful, because instead of using information about the non-emptiness of `Args` at the point where the `Handler` type was used, it was essentially pattern-matching against the `SignalDefinition` / `QueryDefinition` types and thus incorrectly permitting empty `Args`.

packages/test/src/test-interface-type-safety.ts
```diff
@@ -66,3 +67,40 @@ test('Can call signal on any WorkflowHandle', async (t) => {
 
   t.pass();
 });
+
+test('Signal handler type safety', (t) => {
+  const signal = defineSignal<[string]>('a');
+
+  wf.setHandler(signal, (arg: string): void => {});
```
mjameswh, line 74: Please prefix intentionnally unused variables with `_`.

packages/test/src/test-interface-type-safety.ts
mjameswh, line 80: Also validate `wf.setHandler(signal, async (arg: string): Promise<void> => {})`.

packages/test/src/test-interface-type-safety.ts
```diff
@@ -66,3 +67,40 @@ test('Can call signal on any WorkflowHandle', async (t) => {
...
+  wf.setHandler(query, (arg: number): string => '');
+
+  // @ts-expect-error query handler argument type must match
+  wf.setHandler(query, (): string => '');
+
+  // @ts-expect-error query handler return type must match
+  wf.setHandler(query, (arg: string): void => {});
+
+  // @ts-expect-error query handler return type must match
+  wf.setHandler(query, (arg: string): number => 7);
```
mjameswh, line 103: Also assert that `wf.setHandler(query, async (arg: string): Promise<string> => '')` is not accepted.

packages/test/src/test-interface-type-safety.ts
```diff
@@ -66,3 +67,40 @@ test('Can call signal on any WorkflowHandle', async (t) => {
...
+  // @ts-expect-error signal handler must take string argument
+  wf.setHandler(signal, () => {});
+
+  // @ts-expect-error signal handler must return void
+  wf.setHandler(signal, (_arg: string): string => '');
+
+  t.pass();
+});
+
+test('Query handler type safety', (t) => {
```
mjameswh, line 88: Should have a similar test for update handler type safety, no?

packages/test/src/test-interface-type-safety.ts
```diff
@@ -1,6 +1,7 @@
 import test from 'ava';
 import { defineSignal, defineQuery, ExternalWorkflowHandle, ChildWorkflowHandle, Workflow } from '@temporalio/workflow';
 import { WorkflowHandle } from '@temporalio/client';
+import * as wf from '@temporalio/workflow';
```
mjameswh, line 4: I'd expect you to get `IllegalStateException`s while running your new tests, since `setHandle()` assert that there is an active workflow execution context. You may do this instead:

```
const wf = { setHandler: () => {} } as unknown as typeof import('@temporalio/workflow');
```

mjameswh: Or move code that use `wf.setHandler()` inside an inner function that never gets called (like we did on `Can call signal on any WorkflowHandle`).

mjameswh:
Thanks! We should definitely try to have more of those tests. This is not the first time we notice failure to detect some form of argument incoherence.

I opened a [ticket here](https://github.com/temporalio/sdk-typescript/issues/1308) to make an inventory of which APIs don't yet have proper tests. Once you are done with this PR, would you mind updating that PR to reflect the current state of our tests (i.e. tick all items that are already covered)?

bergundy:
I’m not sure I understand the issue. The current behavior seems like the right behavior to me. We shouldn’t force users to accept all arguments.

dandavison:
> We shouldn’t force users to accept all arguments.

Ah, right. We want the rules regarding types of our signal/query/update handlers to be like Typescript's rules for function type compatibility.

```typescript
const signal = wf.defineSignal<[string]>('signal');
wf.setHandler(signal, () => {});

type H = (s: string) => void;
const h: H = () => {};
```

It's what you'd expect in Typescript, and in our context allows an argument to be added to a signal/update/query definition without having to update all handler definitions immediately (or ever). Closing; we can add more type-level tests in another PR.

PR #1310, opened 2023-12-06
feat(worker): Enable reuseV8Context by default

mjameswh:
## What was changed

- `WorkerOptions.reuseV8Context` is now enabled by default. It may be opted out by explicitly setting that option to false.

## Why?

- The Reuse V8 Context execution model brings significant performance improvements.
- The Reuse V8 Context execution model has been very stable since it was first introduced, with no issue being reported despite having been enabled in production by most of our largest TS SDK users for months.

### DIFF
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -487,9 +487,10 @@ export interface WorkerOptions {
    * From running basic stress tests we've observed 2/3 reduction in memory usage and 1/3 to 1/2 in CPU usage with this
    * feature turned on.
    *
-   * Note that we plan to turn this option on by default starting with 1.9.0.
+   * WARNING: There is currently no known reason to turn this off. The non-`reuseV8Context` execution model will be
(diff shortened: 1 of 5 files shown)


packages/worker/src/worker-options.ts
```diff
@@ -487,9 +487,10 @@ export interface WorkerOptions {
    * From running basic stress tests we've observed 2/3 reduction in memory usage and 1/3 to 1/2 in CPU usage with this
    * feature turned on.
    *
-   * Note that we plan to turn this option on by default starting with 1.9.0.
+   * WARNING: There is currently no known reason to turn this off. The non-`reuseV8Context` execution model will be
+   * completely removed in 1.10.0. Please update your worker options to use `reuseV8Context: true` as soon as possible.
```
Sushisource, line 491: Wait, doesn't this imply they want to _remove_ it from their worker options now?

Sushisource (approved):

PR #1311, opened 2023-12-06
[HOLD] feat(worker): Remove non-reuseV8Context execution model

mjameswh:
## What was changed

- Completely remove support for the legacy "V8 Context per Workflow" execution model (a.k.a. non-`reuseV8Context` mode).

## Why?

- The Reuse V8 Context execution model brings significant performance improvements.
- The Reuse V8 Context execution model has been very stable since it was first introduced, with no issue being reported despite having been enabled in production by most of our largest TS SDK users for months.
- Maintaining both non-`reuseV8Context` and `reuseV8Context` execution models add complexity to the code base, making it more difficult to introduce further optimizations to the workflow worker execution model.


mjameswh:
Closing, as there's little value in maintaining this until we are ready to actually merge it. This will not be very hard to redo from scratch.

PR #1312, opened 2023-12-07
getUpdateHandle

### LINKED ISSUE #1306: Add getUpdateHandle
Add `workflowHandle.getUpdateHandle(...)` method to obtain an `UpdateHandle` given update and workflow execution identifiers.

See also https://github.com/temporalio/sdk-python/issues/424

dandavison:
Add `getUpdateHandle` API. Fixes #1306.


packages/client/src/workflow-client.ts
```diff
@@ -351,7 +356,7 @@ export interface WorkflowUpdateHandle<Ret> {
   /**
    * The ID of the Run of the Workflow being targeted by this Update request.
    */
-  workflowRunId: string;
+  workflowRunId?: string;
```
dandavison, line 359: It was a mistake to have made this non-optional in #1277

packages/client/src/workflow-client.ts
```diff
@@ -164,6 +164,11 @@ export interface WorkflowHandle<T extends Workflow = Workflow> extends BaseWorkf
     options?: WorkflowUpdateOptions & { args?: Args }
   ): Promise<WorkflowUpdateHandle<Ret>>;
 
+  /**
+   * Get a handle to an Update.
+   */
+  getUpdateHandle<Ret>(updateId: string, workflowId: string, workflowRunId?: string): WorkflowUpdateHandle<Ret>;
```
mjameswh, line 170: `firstExecutionRunId`? Should probably add in an optional options object, like we did on `getHandle()`.

dandavison: Good call, I've added the options object. Shall we leave `firstExecutionRunId` for a subsequent PR? I don't have any test coverage for it yet.

mjameswh (approved):
LGTM

dandavison:
Thanks @mjameswh!

packages/client/src/workflow-client.ts
mjameswh: Yeah, leave it out for now. I'm not totally sure that this is pertinent, as the update handle represents an update that was sent to a specific workflow run... 🤔 I'd have to think a bit more about this, but don't have time for this ATM.

Issue #1081, opened 2023-03-16
[Proposal] Built-in registered handlers query


josiah-roberts:
[Link to thread](https://temporalio.slack.com/archives/C01DKSMU94L/p1678908893942829) with @bergundy

### Is your feature request related to a problem? Please describe.

There's a common need to keep track of what queries and signals are legal to send to a workflow _at this moment_. My team is creating a layer for this internally, but it would be ideal to solve it on a SDK level.

### Describe the solution you'd like

I'd like to see a built-in query which returns the queries and signals being currently listened to. This could take two forms:

**Separate**  
* A `QueryDefinition<Array<string>, []>` named `queries`
* A `QueryDefinition<Array<string>, []>` named `signals`

**Together**
* A `QueryDefinition<Array<{ name: string, type: 'query' | 'signal' }>, []>` named `handlers`

In either case, we can't preserve the _type information_ about these queries or signals, but we can at least maintain the knowledge of whether there is an active `setHandler(name, () => {...})` for a given name. This allows us to drive UX experiences that only allow **_currently legal_** operations, which is a powerful technique for code-as-source-of-truth on our workflows.

### Additional context

Here's a spike implementation for a `trackHandler` shim over `setHandler` that we're working on internally. Workflows need to be wrapped with `trackHandlers` in the form `export const myWorkflow = (args: Args) => trackHandlers(myWorkflowImpl, args);`. _If this was built-in, such wrapping would not be needed._

```typescript
const trackedHandlerStorage = new AsyncLocalStorage<Set<string>>();
const getTrackedHandlers = () => {
  const handlers = trackedHandlerStorage.getStore();
  if (!handlers) {
    throw new IllegalStateError('Not in a tracked context!');
  }
  return handlers;
};

export const trackHandler: typeof setHandler = (def, handler) => {
  const handlers = getTrackedHandlers();
  handlers.add(def.name);
  setHandler(def, ((...args) => {
    trackedHandlerStorage.enterWith(handlers);
    if (handler) {
      return handler(...(args as Parameters<typeof handler>));
    }
    return undefined;
  }) as typeof handler);
};

export const clearHandler = (def: Parameters<typeof setHandler>[0]) => {
  const handlers = getTrackedHandlers();
  handlers.delete(def.name);
  setHandler(def, undefined);
};

export const clearHandlers = (defs: Array<Parameters<typeof setHandler>[0]>) => {
  for (const def of defs) {
    clearHandler(def);
  }
};

[...]

PR #1319, opened 2023-12-15
feat(worker): Add a workflow metadata query

antlai-temporal:
## What was changed

Typescript SDK implementation of https://github.com/temporalio/features/issues/51 providing a built-in query to return
metadata of registered query, signal, and update handlers.

## Why?

See https://github.com/temporalio/features/issues/51 

## Checklist

1. Closes 
#1081 
2. How was this tested:

Added two integration tests.


packages/workflow/src/internals.ts
```diff
@@ -157,38 +160,76 @@ export class Activator implements ActivationHandler {
...
+        },
+        description: 'Returns a stack trace annotated with source information.',
+      },
+    ],
+    [
+      '__temporal_workflow_metadata',
+      {
+        handler: (): temporal.api.sdk.v1.IWorkflowMetadata => {
+          const workflowType = this.info.workflowType;
+          const description = this.info?.memo?.__temporal_workflow_description;
```
bergundy, line 208: Hmm... I wasn't part of the design but why put this in the memo? And if we're putting this in the memo, why also return it here?

bergundy (approved):
LGTM, had one clarifying question.

packages/workflow/src/internals.ts
bergundy: I think we should question if we want to put this in memo since we now have the standard query (although I see some advantages to it such as not needing a worker up to get the description).

If we want to use both memo and query, it might be better to have a dedicated method to set the description for a workflow.

packages/workflow/src/internals.ts
```diff
@@ -125,12 +125,15 @@ export class Activator implements ActivationHandler {
   /**
    * Mapping of update name to handler and validator
    */
-  readonly updateHandlers = new Map<string, { handler: WorkflowUpdateType; validator?: WorkflowUpdateValidatorType }>();
+  readonly updateHandlers = new Map<
+    string,
+    { handler: WorkflowUpdateType; validator?: WorkflowUpdateValidatorType; description?: string }
```
mjameswh, line 130: We prefer avoiding adhoc types. Seems like that one slipped through the Workflow Update PR.

Please make this a named type instead; same for `signalHandlers` below.

packages/workflow/src/workflow.ts
```diff
@@ -1143,15 +1144,17 @@ export function defineQuery<Ret, Args extends any[] = [], Name extends string =
  *
  * @param def an {@link UpdateDefinition}, {@link SignalDefinition}, or {@link QueryDefinition} as returned by {@link defineUpdate}, {@link defineSignal}, or {@link defineQuery} respectively.
  * @param handler a compatible handler function for the given definition or `undefined` to unset the handler.
+ * @param options an optional `description` of the handler and an optional update `validator` function.
  */
 export function setHandler<Ret, Args extends any[], T extends SignalDefinition<Args> | QueryDefinition<Ret, Args>>(
   def: T,
-  handler: Handler<Ret, Args, T> | undefined
+  handler: Handler<Ret, Args, T> | undefined,
+  options?: { description?: string }
```
mjameswh, line 1152: Please avoid adhoc types. Same at L1157 and L1249.

packages/workflow/src/internals.ts
mjameswh: I'm a bit reluctant with configuring this through a memo. At present, given that we don't support `upsertMemo`, that means the description would have to be provided by the code scheduling the workflow (i.e. client or parent workflow), which isn't desirable. Or, if we have `upsertMemo`, description would have to be set after starting the workflow, through a command, making this a billable feature.

packages/test/src/integration-tests-old.ts
```diff
@@ -398,7 +398,7 @@ export function runIntegrationTests(codec?: PayloadCodec): void {
     await t.throwsAsync(workflow.query('not found'), {
       instanceOf: QueryNotRegisteredError,
       message:
-        'Workflow did not register a handler for not found. Registered queries: [__stack_trace __enhanced_stack_trace isBlocked]',
+        'Workflow did not register a handler for not found. Registered queries: [__stack_trace __enhanced_stack_trace __temporal_workflow_metadata isBlocked]',
```
mjameswh, line 401: That's getting messy. I think we should consider not listing "internals" queries.

packages/workflow/src/internals.ts
antlai-temporal: I don't really want to change WorkflowExecutionInfo protos for something that is going to be rarely used, and its functionality is redundant to memo. And it has to work across continueAsNew, or multiple workers, so I cannot see how to do that without a proto change.

I'm inclined to just not set it at all, and if someone needs extra workflow info they can use memo + lookup with the  workflow id. Alternatively, we could do an explicit setter as @bergundy mentioned, after upsertMemo is implemented in TS. 

About the cost of upsertMemo, I don't think the WF description  is needed to use the query method, today the UI just uses the handlers names, and, in any case, the query costs are going to dominate anyway...

Any thoughts @bergundy @mjameswh ?

packages/workflow/src/workflow.ts
antlai-temporal: Done

packages/test/src/integration-tests-old.ts
antlai-temporal: It could be useful for knowing what built-in queries, like `__temporal_workflow_metadata`, are supported by the WF. 
Also the UI today may expect some handlers present when parsing the error message...

cretz, line 401: I know it sounds trivial, but can we update the comment at https://github.com/temporalio/api/blob/59df52af8d5ffdd4c2a9748e9711d436c0392aa9/temporal/api/sdk/v1/workflow_metadata.proto#L34 to be `__temporal_workflow_metadata` instead of `__temporal_getWorkflowMetadata`?

packages/workflow/src/internals.ts
cretz: Memo is not a good place to put this IMO. Memo takes up visibility store room and is only for information that needs to be returned on list/describe.

> I don't really want to change WorkflowExecutionInfo protos for something that is going to be rarely used, and its functionality is redundant to memo.

I don't think this information is related to existing protos and such. This is just a query return value. If there is a TS-SDK-specific problem of being able to set metadata for a workflow, ok, but this issue should be _just_ about returning this information from query and nothing more IMO.

antlai-temporal: @cretz I was following your proto schema that included a workflow description. What kind of workflow description you don't want to appear in list/describe? Do we need this at all?
I though that in all SDKs you need to modify protos if you want some workflow info to propagate across continueAsNew and worker's multiple execution resume. Is there a standard way across SDKs to do that without modifying protos?

cretz: > I was following your proto schema that included a workflow description
Which proto schema? If you mean https://github.com/temporalio/api/pull/336, that was meant purely as a query result and those messages aren't used otherwise (yet).

> What kind of workflow description you don't want to appear in list/describe? Do we need this at all?

I wasn't expecting list/describe to be affected by this at all. Those are server-side calls. Maybe one day, but for now this is just querying a running workflow for information about itself at this time (sorry if there was confusion on the feature repo issue, it got a bit scatted).

>  I though that in all SDKs you need to modify protos if you want some workflow info to propagate across continueAsNew and worker's multiple execution resume. Is there a standard way across SDKs to do that without modifying protos?

This is fixed code-based metadata, it is not runtime data necessarily. You should just answer the query with the information the workflow code knows about itself and nothing more.

packages/test/src/integration-tests-old.ts
antlai-temporal: https://github.com/temporalio/api/pull/342

mjameswh: UI doesn't and shouldn't _expect_ specific handlers to be present in the list returned on query not found error message. Using that error message to report existing queries is hack, not normalized across SDKs, and going forward, UI should call `__temporal_workflow_metadata` directly anyway.

packages/workflow/src/internals.ts
antlai-temporal: >This is fixed code-based metadata, it is not runtime data necessarily. You should just answer the query with the >information the workflow code knows about itself and nothing more.

OK, that changes things, I thought it could change per workflow instance and not only per workflow type.

I don't know a clean way to decorate a function in a way that works both for TS and JS. We could patch the function object to add a static method/getter, or define a static map mapping function names to descriptions, but they both look like a hack... 

We could also just return the function name in the description for now...

cretz: > I thought it could change per workflow instance and not only per workflow type.

It can change per workflow instance. People can manually register signal handlers for instance. But for the vast majority it won't change per instance. As for workflow description, I don't know of a good way for TS SDK to provide this kind of metadata and I'd be ok leaving it unset until decided.

> I don't know a clean way to decorate a function in a way that works both for TS and JS. We could patch the function object to add a static method/getter, or define a static map mapping function names to descriptions, but they both look like a hack...

Same. Users could call `setDescription` inside the workflow at runtime or somehow at registration time, but in the meantime I'm ok leaving this unset for TS SDK while we think about it.

> We could also just return the function name in the description for now...

I vote leave it unset

packages/test/src/integration-tests-old.ts
antlai-temporal: I agree with you on the UI, but unfortunately, depending on how the UI parses the query today, we may break it if we return an empty set. It is also unfortunate that `isBlocked` does not have the `__` internal prefix. Once they switch, we can do the cleanup with them...

cretz: :+1: Fixing the query-failed string to not include all queries anymore should be a separate task once we feel comfortable enough modern UIs are no longer using it (i.e. fixed version has been out there a while).

packages/workflow/src/interfaces.ts
```diff
@@ -456,3 +456,18 @@ export type DefaultSignalHandler = (signalName: string, ...args: unknown[]) => v
...
+
+/**
+ * A validator and description of an update handler.
+ */
+export type UpdateHandlerOptions<Args extends any[]> = { validator?: UpdateValidator<Args>; description?: string };
+
+/**
+ * Options for a generic handler.
+ */
+export type AnyHandlerOptions<Args extends any[]> = UpdateHandlerOptions<Args>;
```
Sushisource, line 473: Wouldn't it make more sense for `AnyHandlerOptions` to be `{ description?: string }`, with `SignalOrQueryHandlerOptions` being an alias to that (for now, possibly forever), and `UpdateHandlerOptions` being an extension of `AnyHandlerOptions` to include `validator?` ?

packages/test/src/integration-tests-old.ts
mjameswh: My point was not about query-failed no longer returning any queries, but only about not returning SDK _built-in_ queries. For example, if using an old UI, they should not see `__stack_trace`, `__enhanced_stack_trace` and `__temporal_workflow_metadata` in the query drop down. Returning user queries is still fine, for now.

packages/workflow/src/internals.ts
```diff
@@ -157,38 +156,74 @@ export class Activator implements ActivationHandler {
   /**
    * Mapping of query name to handler
    */
-  public readonly queryHandlers = new Map<string, WorkflowQueryType>([
+  public readonly queryHandlers = new Map<string, { handler: WorkflowQueryType; description?: string }>([
```
mjameswh, line 159: Use a named type

packages/workflow/src/interfaces.ts
mjameswh: I'm not even sure it is really pertinent to make a base common interface. Yes, turns out that Query, Signal and Update all accept a `description`, but do we ever care about that common denominator?

mjameswh: @antlai-temporal Let's make this instead:

```
export type QueryHandlerOptions = { description?: string };

export type SignalHandlerOptions = { description?: string };

export type UpdateHandlerOptions<Args extends any[]> = { validator?: UpdateValidator<Args>; description?: string };
```

... and no `AnyHandlerOptions`. Just use a union of those types where appropriate.

packages/workflow/src/workflow.ts
```diff
@@ -1143,15 +1146,17 @@ export function defineQuery<Ret, Args extends any[] = [], Name extends string =
  *
  * @param def an {@link UpdateDefinition}, {@link SignalDefinition}, or {@link QueryDefinition} as returned by {@link defineUpdate}, {@link defineSignal}, or {@link defineQuery} respectively.
  * @param handler a compatible handler function for the given definition or `undefined` to unset the handler.
+ * @param options an optional `description` of the handler and an optional update `validator` function.
  */
 export function setHandler<Ret, Args extends any[], T extends SignalDefinition<Args> | QueryDefinition<Ret, Args>>(
```
mjameswh, line 1151: Explode that overload into a `setHandler` for signal definition, and a distinct one for query definition.

packages/workflow/src/interfaces.ts
Sushisource: Yeah, I'm fine with that too.

packages/test/src/integration-tests-old.ts
antlai-temporal: If needed that's easy to filter by the UI using the `__` prefix. I like that we can quickly know what internal queries are supported for a given workflow...

packages/workflow/src/workflow.ts
antlai-temporal: Done

packages/workflow/src/internals.ts
antlai-temporal: Good catch, thanks!

antlai-temporal:
> I think we should prevent overriding built-in query handlers...

Yes,  I have opened an issue for that https://github.com/temporalio/sdk-typescript/issues/1343 I don't want to delay this one anymore...

PR #1320, opened 2023-12-16
Fix getUpdateHandle API

dandavison:
This PR fixes a mistake in the recently-introduced update `workflowHandle.getUpdateHandle(...)` API https://github.com/temporalio/sdk-typescript/pull/1312.  It should not take `workflowId` as an argument since the appropriate `workflowId` is the one on the `workflowHandle`.

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -167,7 +167,7 @@ export interface WorkflowHandle<T extends Workflow = Workflow> extends BaseWorkf
   /**
    * Get a handle to an Update of this Workflow.
    */
-  getUpdateHandle<Ret>(updateId: string, options?: GetWorkflowUpdateHandleOptions): WorkflowUpdateHandle<Ret>;
+  getUpdateHandle<Ret>(updateId: string): WorkflowUpdateHandle<Ret>;
 
   /**
    * Query a running or completed Workflow.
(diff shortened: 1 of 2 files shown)


packages/client/src/workflow-client.ts
```diff
@@ -165,13 +165,9 @@ export interface WorkflowHandle<T extends Workflow = Workflow> extends BaseWorkf
...
   /**
-   * Get a handle to an Update.
+   * Get a handle to an Update of this Workflow.
    */
-  getUpdateHandle<Ret>(
-    updateId: string,
-    workflowId: string,
-    options?: GetWorkflowUpdateHandleOptions
-  ): WorkflowUpdateHandle<Ret>;
+  getUpdateHandle<Ret>(updateId: string, options?: GetWorkflowUpdateHandleOptions): WorkflowUpdateHandle<Ret>;
```
bergundy, line 170: `runId` is already on the `WorkflowHandle`, let's not use `GetWorkflowUpdateHandleOptions`.

```suggestion
  getUpdateHandle<Ret>(updateId: string): WorkflowUpdateHandle<Ret>;
```

dandavison: I've removed the `GetWorkflowUpdateHandleOptions`. I think that there might be scenarios in which a user will want to construct an update handle with a run Id other than that in the workflow handle (e.g. the workflow handle lacks a run Id, but they know the run ID they want to target). As things stand in this PR they could do that by constructing a workflow handle manually with the desired run ID (as demonstrated in the unit tests added in this PR). But I agree that we should consider this issue offline and not make `getUpdateHandle` more complicated unless it is absolutely clear that it's needed.

bergundy (approved):

PR #1324, opened 2023-12-20
Fix metric meter initialization

Sushisource:
Prometheus initialization would explode because it needs tokio to exist, but was being initialized before tokio was.

### DIFF
--- a/packages/test/src/test-prometheus.ts
+++ b/packages/test/src/test-prometheus.ts
@@ -1,11 +1,11 @@
 import test from 'ava';
 import { v4 as uuid4 } from 'uuid';
+import fetch from 'node-fetch';
 import { WorkflowClient } from '@temporalio/client';
 import { DefaultLogger, NativeConnection, Runtime } from '@temporalio/worker';
 import * as activities from './activities';
 import { RUN_INTEGRATION_TESTS, Worker } from './helpers';
 import * as workflows from './workflows';
-import fetch from 'node-fetch';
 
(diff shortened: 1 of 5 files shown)


packages/test/src/test-prometheus.ts
```diff
@@ -0,0 +1,47 @@
...
+
+if (RUN_INTEGRATION_TESTS) {
+  test.serial('Prometheus metrics work', async (t) => {
+    const logger = new DefaultLogger('DEBUG');
+    Runtime.install({
+      logger,
+      telemetryOptions: {
+        metrics: {
+          prometheus: {
+            bindAddress: '0.0.0.0:9090',
```
mjameswh, line 18: Can we use a random port for this, and bind to 127.0.0.1 instead?

```
async function getPortFree() {
    return new Promise( res => {
        const srv = net.createServer();
        srv.listen(0, () => {
            const port = srv.address().port
            srv.close((err) => res(port))
        });
    })
}
```

mjameswh:
Not elegant, but that should do the work until we revisit `CoreRuntime::new`. Thanks!

mjameswh (approved):
Please make the prom test not bind on `0.0.0.0:9090`. Thanks!

packages/test/src/test-prometheus.ts
Sushisource: I knew I'd need to change that and forgot. Thanks for the reminder.

PR #1326, opened 2023-12-22
fix(worker): Disable Eager Activities if Worker has no activities

mjameswh:
## What changed and why

- Force `WorkerOptions.enableNonLocalActivities` to false if no activities are registered on a Worker, which automatically implies that Core will not request Eager Activity Dispatch for activities scheduled by workflows running on that Worker.

  Until now, Core would request Eager Activity Dispatch for activities scheduled on the same task queue as the Workflow, even when the Worker was being run without any activities registered (splitting Workflow and Activity execution to distinct Workers is a diagnostic strategy we often suggest). In those cases, the activity execution would time out on the first attempt, adding delay before the activity finally get executed.

- Enable several server features when running dev server for internal integration tests (applicable to tests using the new test helper).

- Fixed an issue in bridge that was preventing passing `extraArgs` to dev server.

- In `test-npm-init` CI tests, copy the current NPM version specifier from the SDK repo to the created sample repo. This is required in case packages get published to Verdaccio using a pre-release specifier (e.g. `1.9.0-rc.0`) as NPM will not consider a pre-release version to be an acceptable substitute for syntaxes such as `^1.8.0`, thus causing failure to install dependencies during sample testing.

### DIFF
--- a/packages/test/src/test-integration-workflows.ts
+++ b/packages/test/src/test-integration-workflows.ts
@@ -223,45 +223,49 @@ test('Start of workflow with signal is delayed', async (t) => {
   t.is(tsToMs(startDelay), 4678000);
 });
 
-export async function executeEagerActivity(): Promise<string> {
-  // If eager activity dispatch is working, then the task will always be dispatched to the workflow
-  // worker. Otherwise, chances are 50%-50% for either workers. We execute the activity 20 times to
-  // make sure that the workflow worker is really getting the task thanks to eager activity
-  // dispatch, and not out of pure luck.
-  for (let i = 0; i < 20; i++) {
-    const res = await workflow
+export async function executeEagerActivity(): Promise<string[]> {
+  const results: string[] = [];
+
+  const scheduleActivity = () =>
+    workflow
       .proxyActivities({ scheduleToCloseTimeout: '5s', allowEagerDispatch: true })
-      .testActivity()
-      .catch(() => 'failed');
-    if (res === 'failed') return res;
+      .then((res: string) => results.push(res));
+
+  for (let i = 0; i < 10; i++) {
+    // Schedule 3 activities at a time (`MAX_EAGER_ACTIVITY_RESERVATIONS_PER_WORKFLOW_TASK`)
+    await Promise.all([scheduleActivity(), scheduleActivity(), scheduleActivity()]);
   }
-  return 'success';
+  return results;
 }
 
-test('Worker request Eager Activity Dispatch if possible', async (t) => {
+test('Worker requests Eager Activity Dispatch if possible', async (t) => {
   const { createWorker, startWorkflow } = helpers(t);
+
+  // If eager activity dispatch is working, then the task will always be dispatched to the workflow
+  // worker. Otherwise, chances are 50%-50% for either workers. The test workflow schedule the
+  // activity 30 times to make sure that the workflow worker is really getting the task thanks to
+  // eager activity dispatch, and not out of pure luck.
 
   const activityWorker = await createWorker({
     activities: {
-      testActivity: () => 'failed',
+      testActivity: () => 'activity-only-worker',
     },
+    // Override the default workflow bundle, to make this an activity-only worker
     workflowBundle: undefined,
-    maxConcurrentActivityTaskPolls: 100,
   });
   const workflowWorker = await createWorker({
     activities: {
-      testActivity: () => 'success',
+      testActivity: () => 'workflow-and-activity-worker',
     },
-    maxConcurrentActivityTaskPolls: 2,
   });
   const handle = await startWorkflow(executeEagerActivity);
-  const result = await activityWorker.runUntil(workflowWorker.runUntil(handle.result()));
+  const results = await activityWorker.runUntil(workflowWorker.runUntil(handle.result()));
   const { events } = await handle.fetchHistory();
 
-  t.is(result, 'success');
+  t.true(results.every((res) => res === 'workflow-and-activity-worker'));
   t.false(events?.some?.((ev) => ev.activityTaskTimedOutEventAttributes));
   const activityTaskStarted = events?.filter?.((ev) => ev.activityTaskStartedEventAttributes);
-  t.is(activityTaskStarted?.length, 20);
+  t.is(activityTaskStarted?.length, 30);
   t.true(activityTaskStarted?.every((ev) => ev.activityTaskStartedEventAttributes?.attempt === 1));
 });
 
@@ -272,13 +276,20 @@ export async function dontExecuteEagerActivity(): Promise<string> {
     .catch(() => 'failed')) as string;
 }
 
-test('Worker dont request Eager Activity Dispatch if no activity registered', async (t) => {
+test("Worker doesn't request Eager Activity Dispatch if no activities are registered", async (t) => {
   const { createWorker, startWorkflow } = helpers(t);
+
+  // If the activity was eagerly dispatched to the Workflow worker even though it is a Workflow-only
+  // worker, then the activity execution will timeout (because tasks are not being polled) or
+  // otherwise fail (because no activity is registered under that name). Therefore, if the history
+  // shows only one attempt for that activity and no timeout, that can only mean that the activity
+  // was not eagerly dispatched.
 
   const activityWorker = await createWorker({
     activities: {
       testActivity: () => 'success',
     },
+    // Override the default workflow bundle, to make this an activity-only worker
     workflowBundle: undefined,
   });
   const workflowWorker = await createWorker({
--- a/packages/test/src/test-integration-workflows.ts
+++ b/packages/test/src/test-integration-workflows.ts
@@ -222,3 +222,75 @@ test('Start of workflow with signal is delayed', async (t) => {
   const startDelay = workflowExecutionStartedEvent?.workflowExecutionStartedEventAttributes?.firstWorkflowTaskBackoff;
   t.is(tsToMs(startDelay), 4678000);
 });
+
+export async function executeEagerActivity(): Promise<string> {
+  // If eager activity dispatch is working, then the task will always be dispatched to the workflow
+  // worker. Otherwise, chances are 50%-50% for either workers. We execute the activity 20 times to
+  // make sure that the workflow worker is really getting the task thanks to eager activity
+  // dispatch, and not out of pure luck.
+  for (let i = 0; i < 20; i++) {
+    const res = await workflow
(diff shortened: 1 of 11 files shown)


packages/test/src/test-integration-workflows.ts
```diff
@@ -222,3 +222,75 @@ test('Start of workflow with signal is delayed', async (t) => {
...
+    const res = await workflow
+      .proxyActivities({ scheduleToCloseTimeout: '5s', allowEagerDispatch: true })
+      .testActivity()
+      .catch(() => 'failed');
+    if (res === 'failed') return res;
+  }
+  return 'success';
+}
+
+test('Worker request Eager Activity Dispatch if possible', async (t) => {
```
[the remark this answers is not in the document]
bergundy: This doesn't test that the activity was eagerly dispatched.

I would run the two workers, start 3 activities (eager limit) and check the workflow worker handled the tasks.

packages/test/src/test-integration-workflows.ts
```diff
@@ -222,3 +222,75 @@ test('Start of workflow with signal is delayed', async (t) => {
...
+});
+
+export async function dontExecuteEagerActivity(): Promise<string> {
+  return (await workflow
+    .proxyActivities({ scheduleToCloseTimeout: '5s', allowEagerDispatch: true })
+    .testActivity()
+    .catch(() => 'failed')) as string;
+}
+
+test('Worker dont request Eager Activity Dispatch if no activity registered', async (t) => {
```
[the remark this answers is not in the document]
bergundy: Please assert that the activity doesn't have the requestEagerExecution flag set.

packages/worker/src/worker.ts
```diff
@@ -464,11 +464,7 @@ export class Worker {
     workflowBundle: WorkflowBundleWithSourceMapAndFilename,
     compiledOptions: CompiledWorkerOptions
   ): Promise<WorkflowCreator> {
-    const registeredActivityNames = new Set(
-      Object.entries(compiledOptions.activities ?? {})
-        .filter(([_, v]) => typeof v === 'function')
-        .map(([k]) => k)
-    );
+    const registeredActivityNames = new Set(compiledOptions.activities.keys());
```
bergundy, line 467:
```suggestion
    const registeredActivityNames = [...new Set(compiledOptions.activities.keys())];
```

packages/worker/src/worker-options.ts
```diff
@@ -752,6 +754,8 @@ export function compileWorkerOptions(opts: WorkerOptionsWithDefaults): CompiledW
     opts.maxConcurrentWorkflowTaskExecutions = 2;
   }
 
+  const activities = new Map(Object.entries(opts.activities ?? {}).filter(([_, v]) => typeof v === 'function'));
```
bergundy, line 757: This doesn't work with `Proxy` and non-enumerable properties, let's not break compatibility.
I wouldn't modify the activities object, but you can disable non local activities.

packages/test/src/test-integration-workflows.ts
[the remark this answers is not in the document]
mjameswh: > Please assert that the activity doesn't have the requestEagerExecution flag set.

Can't do that directly. `requestEagerExecution` only exists on the `ScheduleActivityTaskCommand` message that Core sends back to the server. The flag isn't present on anything after that point, either in WF History or on the Activity Task. Therefore we can only test that it got eagerly dispatched by the side effects of it.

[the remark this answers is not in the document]
mjameswh: In this case, if the activity was eagerly dispatched to the Workflow worker even though it is a Workflow-only worker, then the activity execution will timeout (because tasks are not being polled) or otherwise fail (because no activity is registered under that name). Therefore, if the history shows only one attempt for that activity and no timeout, that can only mean that the activity was not eagerly dispatched; at that point, we can't confirm exactly why it wasn't (e.g. could be that the feature is disabled on the server), which is why I also added the other test, that confirms that Eager Activity Execution works properly in other cases.

packages/test/src/test-integration-workflows.ts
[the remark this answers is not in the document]
mjameswh: > This doesn't test that the activity was eagerly dispatched.

This test asserts that 20 activities executed one after the other all get dispatched to the workflow+activity worker, and none to the activity-worker. That is statistically very unlikely to happen unless this is due to eager activity dispatch. As a proof of that, if I disable Eager Execution

> I would run the two workers, start 3 activities (eager limit) and check the workflow worker handled the tasks.

Just to be sure, are you simply suggesting starting three activities at once (ie. `MAX_EAGER_ACTIVITY_RESERVATIONS_PER_WORKFLOW_TASK`), rather than one, or are you suggesting some other changes here?

packages/worker/src/worker.ts
mjameswh: Uhh? Why? The Creators just bellow expect Sets, not Arrays.

packages/test/src/test-integration-workflows.ts
```diff
@@ -222,3 +222,75 @@ test('Start of workflow with signal is delayed', async (t) => {
...
+export async function executeEagerActivity(): Promise<string> {
+  // If eager activity dispatch is working, then the task will always be dispatched to the workflow
+  // worker. Otherwise, chances are 50%-50% for either workers. We execute the activity 20 times to
+  // make sure that the workflow worker is really getting the task thanks to eager activity
+  // dispatch, and not out of pure luck.
+  for (let i = 0; i < 20; i++) {
+    const res = await workflow
+      .proxyActivities({ scheduleToCloseTimeout: '5s', allowEagerDispatch: true })
+      .testActivity()
+      .catch(() => 'failed');
```
bergundy, line 235: Why do you need this catch block?

packages/test/src/test-integration-workflows.ts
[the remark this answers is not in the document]
bergundy: You're right, this is fine.
You could just schedule 3 at once, it's probably just as effective.

packages/test/src/test-integration-workflows.ts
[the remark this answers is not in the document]
bergundy: Yeah, this is fine, I missed this in the first pass.

packages/worker/src/worker.ts
bergundy: Sorry, I should've read this more closely, when I skimmed the diff it looked like you were converting this set to an array in line 865.

Ignore me.

packages/test/src/test-integration-workflows.ts
mjameswh: Removed

packages/worker/src/worker-options.ts
mjameswh: Fair point, though, to be exact, the behavior with `Proxy` and non-enumerable properties was already ill-defined and incorrect, so I'm not that much worried about breaking compatibility.

For example, starting a worker with `activities` defined like this would make it a workflow-only worker. Same if `activities` only contain non-enumerable properties.

    activities: new Proxy({}, {
      get(target, p, receiver) {
        return (activities as any)[p];
      },
    }),

bergundy: We have this condition that prevents starting an activity worker:

```
 if (this.options.activities === undefined || Object.keys(this.options.activities).length === 0)
```

You can still use `Proxy` for dynamic activities if you proxy a non empty object:

```ts
activities: new Proxy(activities, {
  get(target, p) {
    const f = (activities as any)[p];
    if (f == null) { return default }
    return f;
  },
}),
```

bergundy: But your point holds for if the activities are all non enumerable.

packages/test/src/test-integration-workflows.ts
```diff
@@ -222,3 +222,86 @@ test('Start of workflow with signal is delayed', async (t) => {
...
+    activities: {
+      testActivity: () => 'activity-only-worker',
+    },
+    // Override the default workflow bundle, to make this an activity-only worker
+    workflowBundle: undefined,
+  });
+  const workflowWorker = await createWorker({
+    activities: {
+      testActivity: () => 'workflow-and-activity-worker',
+    },
```
bergundy, line 259: This is much clearer, thanks!

packages/test/src/test-integration-workflows.ts
```diff
@@ -222,3 +222,86 @@ test('Start of workflow with signal is delayed', async (t) => {
...
+  });
+  const handle = await startWorkflow(executeEagerActivity);
+  const results = await activityWorker.runUntil(workflowWorker.runUntil(handle.result()));
+  const { events } = await handle.fetchHistory();
+
+  t.true(results.every((res) => res === 'workflow-and-activity-worker'));
+  t.false(events?.some?.((ev) => ev.activityTaskTimedOutEventAttributes));
+  const activityTaskStarted = events?.filter?.((ev) => ev.activityTaskStartedEventAttributes);
+  t.is(activityTaskStarted?.length, 30);
+  t.true(activityTaskStarted?.every((ev) => ev.activityTaskStartedEventAttributes?.attempt === 1));
```
bergundy, line 269: This doesn't matter

packages/test/src/test-integration-workflows.ts
```diff
@@ -222,3 +222,86 @@ test('Start of workflow with signal is delayed', async (t) => {
...
+  const workflowWorker = await createWorker({
+    activities: {
+      testActivity: () => 'workflow-and-activity-worker',
+    },
+  });
+  const handle = await startWorkflow(executeEagerActivity);
+  const results = await activityWorker.runUntil(workflowWorker.runUntil(handle.result()));
+  const { events } = await handle.fetchHistory();
+
+  t.true(results.every((res) => res === 'workflow-and-activity-worker'));
```
bergundy, line 265: can test this in workflow code.

packages/test/src/test-integration-workflows.ts
```diff
@@ -222,3 +222,86 @@ test('Start of workflow with signal is delayed', async (t) => {
...
+
+export async function executeEagerActivity(): Promise<string[]> {
+  const results: string[] = [];
+
+  const scheduleActivity = () =>
+    workflow
+      .proxyActivities({ scheduleToCloseTimeout: '5s', allowEagerDispatch: true })
+      .then((res: string) => results.push(res));
+
+  for (let i = 0; i < 10; i++) {
```
bergundy, line 234: Probably overkill to do 10 iterations but 🤷

bergundy (approved):
I would just like to bring back proxy support but other than that LGTM.

PR #1340, opened 2024-01-12
docs(release): Update publishing procedure

mjameswh:
## What was changed

- Updated publishing procedure

## Why?

- They are outdated, and it is desirable that everyone in the team knows what are our current processes

## Notes

- Work is in progress to automate most of this to GHA workflows.
  These instructions reflect the current, _de facto_ process, not the desired long term vision of how we should be releasing this project.


mjameswh (commented):
Sample repo now uses PNPM

scripts/publish.sh
```diff
@@ -0,0 +1,152 @@
...
+gh pr create
+git switch main
+git pull
+
+
+
+echo -e 'Please do the following:'
+echo -e ' 1. Open the \e]8;https://github.com/temporalio/sdk-typescript/releases/new?tag=v'"$version"'\e\\GitHub New Release page\e]8;;\e\\ and select the 'v"$version"' tag.'
+echo -e ' 2. In the Release Title field, enter '"$version"''
+echo -e ' 3. Paste the release notes inkto the description field'
```
chris-olszewski:
```suggestion
echo -e ' 3. Paste the release notes into the description field'
```

scripts/publish.sh
```diff
@@ -0,0 +1,152 @@
...
+else
+    echo -e ' 4. Make sure that the "Set as a pre-release" checkbox is checked'
+    echo -e '    and that the "Set as the latest release" checkbox is unchecked'
+fi
+echo -e ' 5. Press the "Save draft" button, then ask someone else to review'
+echo -e ' 6. Press the "Publish Release" buton to complete the release process'
+
+echo
+echo -e 'Press ENTER once this is completed.'
+read enterKey
```
chris-olszewski, line 89: I think we're missing actually publishing the packages to `npm`

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+
+pnpm -r exec npm version "${version}"
+git switch -c "release-v${version}"
+git add packages/*/package.json scripts/package.json
+git commit -m "Release v${version}"
+git push origin "release-v${version}"
+
+gh repo set-default temporalio/sdk-typescript
+gh pr create
+# Get PR approved and merged
```
chris-olszewski, line 66: We need to add a pause point here so the user has time to get the PR merged. Right now it will immediately checkout main before the PR has a chance to merge.

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/zsh
+set -euo pipefail
+
+local workdir=$( mktemp -d -t sdk-typescript-release )
+trap 'cd / && rm -rf "$workdir"' EXIT
```
chris-olszewski, line 5: This is a little aggressive, we might want to just suggest to end user to clean up. If a `pnpm publish` fails, we lose all progress and have to start from 0

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+read enterKey
+
+local count=$( find artifacts -type f -name "corebridge-native-*.zip" | wc -l )
+if [ $count -ne 5 ]; then
+    echo "The 'artifacts' directory does not contain exactly 5 files named 'corebridge-native-*.zip'"
+    echo "Aborting"
+    exit 1
+fi
+
+git clone --branch $source_branch --depth 1 --shallow-submodules --recurse-submodules https://github.com/temporalio/sdk-typescript.git
```
chris-olszewski, line 35: With GitHub removing password auth for HTTPS Git ops using SSH is a easier IMO than creating and using a PAT. 
```suggestion
git clone --branch $source_branch --depth 1 --shallow-submodules --recurse-submodules org-56493103@github.com:temporalio/sdk-typescript.git
```

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+export version="1.15.0"
+
+
+# Manually download all native artifacts from the latest build
+
+mkdir -p artifacts package/releases
+open artifacts
+
+echo -e 'Please do the following:'
+echo -e ' 1. Open the \e]8;;https://github.com/temporalio/sdk-typescript/actions/workflows/ci.yml?query=branch%3A'"$source_branch"'\e\\GHA status page\e]8;;\e\\ for the "Continuous Integration" workflow, on branch main.'
```
THardy98, line 20: Download artifacts from the "Build package" workflow, not "Continuous Integration".

"Build package" creates release artifacts that are significantly smaller than "Continuous Integration" debug packages.

scripts/publish.sh
THardy98, line 66: Also, for future reference - we are generally accepting of merging this PR even if CI fails given it's just a version bump

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+        --head "typescript-${version}"
+  )
+
+  # Update the Omes repo
+  (
+    git clone --depth 1 --shallow-submodules --recurse-submodules https://github.com/temporalio/omes.git
+    cd omes
+
+    # Note this only update _regular_ dependencies. There's no dev/peer dependencies at the moment.
+    npm install $(jq -r '.dependencies | keys[] | select(startswith("@temporalio/")) | . + "@^" + env.version' package.json)
```
THardy98, line 120: The only step needed is to bump `TYPESCRIPT_SDK_VERSION` in `versions.env` at the root dir to `${version}`

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+    git clone --depth 1 --shallow-submodules --recurse-submodules https://github.com/temporalio/samples-typescript.git
+    cd samples-typescript
+
+    npm i
+
+    # Update all samples
+    zx .scripts/upgrade-versions.mjs "^${version}"
+
+    # Update the package.json file
+    npm i
```
THardy98, line 143: use `pnpm install` to update the lockfile

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+echo
+echo 'Press ENTER to go on with publishing, or Ctrl+C to abort'
+
+read enterKey
+
+echo 'Bumping version numbers...'
+
+pnpm -r exec npm version "${version}"
+git switch -c "release-v${version}"
+git add packages/*/package.json scripts/package.json
```
THardy98, line 60: Make sure to `git add` packages that got bumped in `contrib` as well

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+git push origin "release-v${version}"
+
+gh repo set-default temporalio/sdk-typescript
+gh pr create
+# Get PR approved and merged
+git switch main
+git pull
+
+pnpm login
+pnpm -r publish
```
chris-olszewski, line 71: If backporting, then you must set `--tag` explicitly otherwise it will default to `latest`: https://pnpm.io/cli/publish#--tag-tag

mjameswh: NPM does not allow publishing without a tag.

So, in the case of backports, we'll tag releases with something like `latest-1.16`.

scripts/publish.sh
```diff
@@ -0,0 +1,155 @@
...
+ls -l packages/core-bridge/releases/*/*
+
+echo
+echo 'Press ENTER to go on with publishing, or Ctrl+C to abort'
+
+read enterKey
+
+echo 'Bumping version numbers...'
+
+pnpm -r exec npm version "${version}"
```
THardy98, line 58: `pnpm -r exec npm version --allow-same-version "${version}"`

useful when you've failed part way

scripts/publish.sh
THardy98: edit: now, bump in `mise.toml` and run `mise run sync-sdk:typescript`

PR #1344, opened 2024-01-23
Update webpack error doc link in worker/bundler.ts

lorisdev:
## What was changed
Outdated link to docs in the bundler script when there is a Webpack error

## Why?
To link users to the right documentation page.


bergundy (approved):
Thanks!

mjameswh:
Force merging: the failure on features-tests is unrelated

PR #1357, opened 2024-02-11
feat(worker): Set default webpack build configuration target to node

bernatixer:
## What was changed
This PR sets the default worker webpack build configuration target to `node`.

## Why?
Webpack default target is `web`, as described in [their docs](https://webpack.js.org/concepts/targets/), this brings some problems when running some packages i.e. [Prisma](https://github.com/prisma/prisma) that are unable to run with a web environment.

## Checklist

1. Closes: None, but happy to create if needed!

2. How was this tested:
While starting testing Temporal with our current codebase we found that we were unable to build workers given webpack default target is web, which is incompatible with Prisma way of executing.

3. Any docs updates needed?
As far as I saw, the current docs don't state the default config for webpack, but a link to the default configs could be added for possible future cases, to allow developers to dig into the problems faster.

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -238,6 +238,7 @@ exports.importInterceptors = function importInterceptors() {
       entry: [entry],
       mode: 'development',
       devtool: 'inline-source-map',
+      target: 'node',
       output: {
         path: distDir,
         filename: 'workflow-bundle-[fullhash].js',


bergundy:
This is by design.

While web isn't the best target it's preferable to node.

Workflows run in a deterministic sandboxed environment and cannot access node APIs, which may be non-deterministic.
You should only import packages that need to run in the node runtime in your activities and import only the activity types into your workflows.

Generally, any package that supports both web and node can be imported into the workflow sandbox.

PR #1363, opened 2024-02-19
feat(workflow): add durationMs to workflow logs

endihunter:
## What was changed
Added durationMs to "Workflow completion/failure" logs 

## Why?
It is useful to track the workflow duration

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/workflow/src/logs.ts
+++ b/packages/workflow/src/logs.ts
@@ -3,7 +3,7 @@ import { untrackPromise } from './stack-helpers';
 import { type Sink, type Sinks, proxySinks } from './sinks';
 import { isCancellation } from './errors';
 import { WorkflowInfo, ContinueAsNew } from './interfaces';
-import { assertInWorkflowContext } from './global-attributes';
+import { assertInWorkflowContext, getActivator } from './global-attributes';
 
 export interface WorkflowLogger extends Sink {
   trace(message: string, attrs?: Record<string, unknown>): void;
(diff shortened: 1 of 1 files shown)


packages/workflow/src/logs.ts
```diff
@@ -77,26 +77,34 @@ export const log: WorkflowLogger = Object.fromEntries(
 
 export function executeWithLifecycleLogging(fn: () => Promise<unknown>): Promise<unknown> {
   log.debug('Workflow started');
+  const activator = assertInWorkflowContext('Workflow.executeWithLifecycleLogging(...) may only be used from workflow context.');
```
mjameswh, line 80: Nit: Just use `getActivator()` here instead of `assertInWorkflowContext`. `executeWithLifecycleLogging` is not exposed publicly, so essentially no way that a user would call it improperly.

mjameswh:
Thanks! That indeed seems pertinent information.

Can you please fix the indentation (it must be 2 spaces, not 4)? Running `npm run format` from the root of the project should fix this.

packages/workflow/src/logs.ts
endihunter: @mjameswh done

endihunter:
@mjameswh indentation fixed, thanks!

bergundy:
This seems like it will confuse people because the workflow start time isn’t the real start time, it’s when the workflow has been loaded into the worker’s cache.

mjameswh:
> This seems like it will confuse people because the workflow start time isn’t the real start time, it’s when the workflow has been loaded into the worker’s cache.

@bergundy is right. `activator.getTimeOfDay()` returns the real wall clock time according to the Worker machine. That means that capturing `activator.getTimeOfDay()` _just before_ calling the workflow function will give the time that Workflow Execution entered the Worker's cache, which is totally insignificant in regard to the time at which that workflow started. `workflowInfo().runStartTime` should probably be used instead.

Reliably determining the completion time of the workflow execution also poses challenges. In most cases, capturing `workflowInfo().unsafe.now()` (or `Date.now()` from outside the workflow sandbox) after returning from the workflow function should provide a reasonable _estimate_ of the workflow completion time, within a few ms of the time at which the server will actually record the workflow completion.

However, there is no guarantee that the Worker's clock and Server's clock are in sync, so subtracting one from the other might therefore result in incorrect values. We must also consider that in degraded execution environments, there might be _seconds_ between the moment that the workflow function returns and the moment where the completion effectively get recorded to history by the server.

Now, that doesn't mean that determining the duration of a workflow from Worker's side doesn't make sense. We all do that regularly while looking at worker logs, and rarely get bit by that.

But I'm now thinking that it would not make sense for the SDK to expose this value by default if it is not reliable. For that reason, I'd be inclined to reject this PR.

Anyway, users that desire this metric, and understand its possible shortcomings, may easily implement that feature themselves using the following Workflow Interceptor:

```
export class WorkflowExecutionDurationInterceptor implements WorkflowInboundCallsInterceptor, WorkflowOutboundCallsInterceptor {
  private workflowCompleted = false;

  getLogAttributes(
    input: GetLogAttributesInput,
    next: Next<WorkflowOutboundCallsInterceptor, 'getLogAttributes'>
  ): Record<string, unknown> {
    if (!this.workflowCompleted) return next(input);
    return {
      ...next(input),
      durationMs: workflowInfo().unsafe.now() - workflowInfo().runStartTime.getTime(),
    };
  }

  execute(input: WorkflowExecuteInput, next: Next<WorkflowInboundCallsInterceptor, 'execute'>): Promise<unknown> {
    try {
      return next(input);
    } finally {
      this.workflowCompleted = true;
    }
  }
}
```

PR #1374, opened 2024-03-15
Fix comment typo on how default maxConcurrentActivityTaskPolls is calculated

taonic:
## What was changed

Fix a typo in comment.

## Why?

So it's consistent with the actual code.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


bergundy (commented):
Thanks.

I’m wondering if the defaults are right though. I think Go has a much lower default.

@mjameswh

bergundy (approved):

mjameswh:
>I’m wondering if the defaults are right though. I think Go has a much lower default.

There are very few cases where 10 pollers will be "too much" (to the point of causing side effects), but there are lots of cases where less than 10 would not be enough to keep the activity task execution slots filled…

Or put another way, if we set the default lower than 10, then most customers should increase that value in their prod settings, where now, most people just ignore it.

mjameswh (approved):

PR #1383, opened 2024-03-26
Update Core & pass through new metrics options

Sushisource:
## What was changed
Added new metrics options

Also autoformatted a bunch of stuff. Feel free to exclude that commit.

## Why?
Expose Core features

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1380

2. How was this tested:
Added tests. Unfortunately testing the specifics of the OTel metrics is not really reasonable, but, the option is passed through in the same way as Prom, which is.

3. Any docs updates needed?


mjameswh (approved):
LGTM

PR #1385, opened 2024-03-27
Api Key Support

Sushisource:
## What was changed
Add api-key helper methods for both kinds of clients

## Why?
Parity

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1367

2. How was this tested:
Upgraded tests

3. Any docs updates needed?

### DIFF
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -101,6 +101,14 @@ export interface ConnectionOptions {
   metadata?: Metadata;
 
   /**
+   * API key for Temporal. This becomes the "Authorization" HTTP header with "Bearer " prepended.
+   * This is only set if RPC metadata doesn't already have an "authorization" key.
+   *
+   * In order to dynamically set metadata, use {@link Connection.withApiKey}
+   */
+  apiKey?: string;
+
+  /**
    * Milliseconds to wait until establishing a connection with the server.
    *
    * Used either when connecting eagerly with {@link Connection.connect} or
@@ -113,7 +121,7 @@ export interface ConnectionOptions {
 }
 
 export type ConnectionOptionsWithDefaults = Required<
-  Omit<ConnectionOptions, 'tls' | 'connectTimeout' | 'callCredentials'>
+  Omit<ConnectionOptions, 'tls' | 'connectTimeout' | 'callCredentials' | 'apiKey'>
 > & {
   connectTimeoutMs: number;
 };
@@ -142,9 +150,16 @@ function addDefaults(options: ConnectionOptions): ConnectionOptionsWithDefaults 
  * - Convert {@link ConnectionOptions.tls} to {@link grpc.ChannelCredentials}
  * - Add the grpc.ssl_target_name_override GRPC {@link ConnectionOptions.channelArgs | channel arg}
  * - Add default port to address if port not specified
+ * - Set `Authorization` header based on API key (and not already set explicitly)
  */
 function normalizeGRPCConfig(options?: ConnectionOptions): ConnectionOptions {
   const { tls: tlsFromConfig, credentials, callCredentials, ...rest } = options || {};
+  if (rest.apiKey && !rest.metadata?.['Authorization']) {
+    if (!rest.metadata) {
+      rest.metadata = {};
+    }
+    rest.metadata['Authorization'] = `Bearer ${rest.apiKey}`;
+  }
   if (rest.address) {
     // eslint-disable-next-line prefer-const
     let [host, port] = rest.address.split(':', 2);
@@ -451,7 +466,39 @@ export class Connection {
    */
   async withMetadata<ReturnType>(metadata: Metadata, fn: () => Promise<ReturnType>): Promise<ReturnType> {
     const cc = this.callContextStorage.getStore();
-    return await this.callContextStorage.run({ ...cc, metadata: { ...cc?.metadata, ...metadata } }, fn);
+    return await this.callContextStorage.run(
+      {
+        ...cc,
+        metadata: { ...cc?.metadata, ...metadata },
+      },
+      fn
+    );
+  }
+
+  /**
+   * Set the apiKey for any service requests executed in `fn`'s scope.
+   *
+   * The provided api key will only be set if the current metadata does not already have an "authorization" key.
+   *
+   * @returns value returned from `fn`
+   *
+   * @example
+   *
+   * ```ts
+   * const workflowHandle = await conn.withApiKey('secret', () =>
+   *   conn.withMetadata({ otherKey: 'set' }, () => client.start(options)))
+   * );
+   * ```
+   */
+  async withApiKey<ReturnType>(apiKey: string, fn: () => Promise<ReturnType>): Promise<ReturnType> {
+    const cc = this.callContextStorage.getStore();
+    return await this.callContextStorage.run(
+      {
+        ...cc,
+        metadata: { Authorization: `Bearer ${apiKey}`, ...cc?.metadata },
+      },
+      fn
+    );
   }
 
   /**
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -95,6 +95,7 @@ export interface ConnectionOptions {
 
   /**
    * Optional mapping of gRPC metadata (HTTP headers) to send with each request to the server.
+   * Setting the `Authorization` header is mutually exclusive with the {@link apiKey} option.
    *
    * In order to dynamically set metadata, use {@link Connection.withMetadata}
    */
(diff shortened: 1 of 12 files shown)


packages/client/src/connection.ts
```diff
@@ -451,7 +466,39 @@ export class Connection {
...
+   *   conn.withMetadata({ otherKey: 'set' }, () => client.start(options)))
+   * );
+   * ```
+   */
+  async withApiKey<ReturnType>(apiKey: string, fn: () => Promise<ReturnType>): Promise<ReturnType> {
+    const cc = this.callContextStorage.getStore();
+    return await this.callContextStorage.run(
+      {
+        ...cc,
+        metadata: { Authorization: `Bearer ${apiKey}`, ...cc?.metadata },
```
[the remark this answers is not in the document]
Sushisource: It works that way, though? You use this method to change it for any particular call.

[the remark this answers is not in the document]
cretz: Hrmm, I might have expected that we'd mutate something the user can read back. But now I see unfortunately metadata wasn't even done that way. So there is no way for the user to retrieve the set API key or metadata? If I had a client and I wanted to remove a piece of metadata, how? Or if I wanted to check that I had an API key set, how would I? That's what I meant be "allows people to get and update API key independently of metadata", but it appears there is no "get" part in TS anyways?

Also, it is a bit unfortunate that for metadata (and now api key) we didn't make worker native connection and TS connection more similar in the ways of getting/setting these values.

But I guess these are general TypeScript SDK issues and not specific to this property, but if/when we ever offer the ability to get the API key or metadata, we will probably need to split these for easy use (get-then-update metadata is a common use case).

packages/client/src/connection.ts
```diff
@@ -142,9 +150,16 @@ function addDefaults(options: ConnectionOptions): ConnectionOptionsWithDefaults
  * - Convert {@link ConnectionOptions.tls} to {@link grpc.ChannelCredentials}
  * - Add the grpc.ssl_target_name_override GRPC {@link ConnectionOptions.channelArgs | channel arg}
  * - Add default port to address if port not specified
+ * - Set `Authorization` header based on API key (and not already set explicitly)
  */
 function normalizeGRPCConfig(options?: ConnectionOptions): ConnectionOptions {
   const { tls: tlsFromConfig, credentials, callCredentials, ...rest } = options || {};
+  if (rest.apiKey && !rest.metadata?.['Authorization']) {
```
mjameswh, line 157: I'd prefer we deal with this incoheencies with a TypeError, just like we do for the tls+credentials case below.

packages/client/src/connection.ts
```diff
@@ -142,9 +150,16 @@ function addDefaults(options: ConnectionOptions): ConnectionOptionsWithDefaults
...
+    if (!rest.metadata) {
+      rest.metadata = {};
+    }
```
mjameswh, line 160:
```
rest.metadata ??= {};
```

packages/client/src/connection.ts
```diff
@@ -142,9 +150,16 @@ function addDefaults(options: ConnectionOptions): ConnectionOptionsWithDefaults
  * - Convert {@link ConnectionOptions.tls} to {@link grpc.ChannelCredentials}
  * - Add the grpc.ssl_target_name_override GRPC {@link ConnectionOptions.channelArgs | channel arg}
  * - Add default port to address if port not specified
+ * - Set `Authorization` header based on API key (and not already set explicitly)
  */
 function normalizeGRPCConfig(options?: ConnectionOptions): ConnectionOptions {
   const { tls: tlsFromConfig, credentials, callCredentials, ...rest } = options || {};
```
mjameswh, line 156:
```suggestion
  const { tls: tlsFromConfig, credentials, callCredentials, metadata = {}, ...rest } = options || {};
```

... and then, replace `rest.metadata` everywhere below. No more need to check that metadata isn't empty.

packages/client/src/connection.ts
mjameswh: We should also forbid the credentials + apiKey use case, as `credentials` will override `metadata['Authorization']`

packages/client/src/connection.ts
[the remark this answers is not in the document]
mjameswh: Important! This must be:

```suggestion
        metadata: { ...cc?.metadata, Authorization: `Bearer ${apiKey}` },
```

In object spread, later entries override earlier ones, so the original version would not work.

packages/test/src/test-native-connection-headers.ts
```diff
@@ -47,9 +51,14 @@ test('NativeConnection passes headers provided in options', async (t) => {
       >,
       callback: grpc.sendUnaryData<temporal.api.workflowservice.v1.IPollActivityTaskQueueResponse>
     ) {
+      console.log(call.metadata);
```
mjameswh, line 54: Leftover?

Sushisource: Yes, thanks

packages/client/src/connection.ts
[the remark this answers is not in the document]
Sushisource: @mjameswh That was actually intentional, as the docstrings and the behavior I copied from the Python PR is that any `Authorization` header set in the metadata takes precedence.

packages/client/src/connection.ts
```diff
@@ -142,9 +150,22 @@ function addDefaults(options: ConnectionOptions): ConnectionOptionsWithDefaults
...
+  if (rest.apiKey) {
+    if (rest.metadata?.['Authorization']) {
+      throw new TypeError(
+        'Both `apiKey` option and `Authorization` header were provided. Only one makes sense to use at a time.'
+      );
+    }
+    if (credentials !== undefined) {
+      throw new TypeError(
+        'Both `apiKey` and `credentials` ConnectionOptions were provided. Only one makes sense to use at a time.'
```
bergundy, line 165: nit: I don't believe we typically put periods in our error messages.

packages/client/src/connection.ts
```diff
@@ -100,6 +100,14 @@ export interface ConnectionOptions {
    */
   metadata?: Metadata;
 
+  /**
+   * API key for Temporal. This becomes the "Authorization" HTTP header with "Bearer " prepended.
+   * This is only set if RPC metadata doesn't already have an "authorization" key.
+   *
+   * In order to dynamically set metadata, use {@link Connection.withApiKey} or {@link Connection.setApiKey}
+   */
+  apiKey?: string;
```
bergundy, line 109: If the setter accepts a function, you should accept a function here too:

```suggestion
  apiKey?: string | (() => string);
```

packages/client/src/connection.ts
```diff
@@ -100,6 +100,14 @@ export interface ConnectionOptions {
    */
   metadata?: Metadata;
 
+  /**
+   * API key for Temporal. This becomes the "Authorization" HTTP header with "Bearer " prepended.
+   * This is only set if RPC metadata doesn't already have an "authorization" key.
```
mjameswh, line 105: > This is only set if RPC metadata doesn't already have an "authorization" key.

This is no longer true. `apiKey` and `authorization` header are mutually exclusive. Also add a mention for this on `metadata`'s doc.

packages/client/src/connection.ts
```diff
@@ -100,6 +100,14 @@ export interface ConnectionOptions {
...
+   *
+   * In order to dynamically set metadata, use {@link Connection.withApiKey} or {@link Connection.setApiKey}
```
mjameswh, line 107:
```suggestion
   * In order to replace the API key at a later time, use {@link Connection.withApiKey} or {@link Connection.setApiKey}
```

packages/client/src/connection.ts
mjameswh: Yeah, forgot to mention that you would have to add `metadata` in object spread later on. That's still more JS-idiomatic.

packages/worker/src/connection-options.ts
```diff
@@ -28,11 +28,18 @@ export interface NativeConnectionOptions {
    * Set statically at connection time, can be replaced later using {@link NativeConnection.setMetadata}.
    */
   metadata?: Record<string, string>;
+
+  /**
+   * API key for Temporal. This becomes the "Authorization" HTTP header with "Bearer " prepended.
+   * This is only set if RPC metadata doesn't already have an "authorization" key.
```
mjameswh, line 34: > This is only set if RPC metadata doesn't already have an "authorization" key.

On creation of a `NativeConntection`, we should check for apikey+authorization metadata, and throw just like in client connection.

packages/client/src/connection.ts
Sushisource: Sure, I considered it but thought they'd probably just set it after construction if they need it dynamic, but, why not.

packages/client/src/connection.ts
Sushisource: Changed, but I have no idea why this matters lol.

packages/client/src/connection.ts
```diff
@@ -451,7 +497,52 @@ export class Connection {
...
+      },
+      fn
+    );
+  }
+
+  /**
+   * Set the {@link ConnectionOptions.apiKey} for all subsequent requests. A static string or a
+   * callback function may be provided.
+   */
+  setApiKey(apiKey: string | (() => string)): void {
```
bergundy, line 537: I know we discussed on slack that we want this URL but if we also have a callback on creation can we maybe punt on adding this API until we get user feedback that the callback approach isn't good enough?

It's easy to add APIs, hard to remove them.

bergundy (approved):
Overall LGTM.

Should we also mark these APIs as experimental? What did we do in other SDKs?

packages/client/src/connection.ts
Sushisource: Sorry, URL?

PR #1391, opened 2024-04-09
Add `logSource` and `taskQueue` metadata on log messages

mjameswh:
## What was changed

- Add `taskQueue` metadata attribute on all log messages emitted inside the context of a worker. It is there to make it easier to correlate logs when there are multiple workers running in a same process while diagnosing low level issues (e.g. while running tests).

- Add a `logSource` property in metadata. That property indicates where that log message was emitted from; possible values at this time are:
  - `workflow`: Messages emitted from workflow code (i.e. _user_ code, through the `workflow.log` object);
  - `activity`: Messages emitted from activity code (i.e. _user_ code, through the `activity.log` object);
  - `worker`: messages emitted by the worker itself, including workflow and activity life cycle events;
  - `core`: messages coming from Core SDK.
  
  That extra property will make it easier for users to implement fine-grained filtering of messages (eg. show DEBUG level messages for user-code, but only WARN for messages from the worker). Note that this is only about annotating log messages appropriately to make that possible. We don't plan to actually add support for that kind of filtering in the SDK itself. This can easily be implemented by users themselves in their custom runtime logger.

- `MockActivityEnvironment` no longer defaults the context logger to `Runtime.logger`, as it is not desirable for `Runtime` to get instantiated only for that purpose. Instead, it defaults to sending logs to the console. `MockActivityEnvironment` now accept a logger option, which makes it easier to capture log messages emitted while testing an activity.

- Fixed a Core-bridge build failure on Windows caused by Node's recent fix for CVE-2024-27980.


mjameswh:
Ideas of names for the "subsystem" attribute:

- `streamName`
- `logStreamName`
- `sourceCategory`
- `originContext`
- `logSource`
- `sourceComponent`
- `loggerName`

packages/core-bridge/src/helpers.rs
```diff
@@ -226,23 +226,64 @@ pub fn hashmap_to_js_value<'a>(
...
             let mut result = String::with_capacity(input.len());
+            snake_to_camel_mut(&input, &mut result);
+            result
+        }
+    }
+}
+
+// Convert from Core's log "targets" (actually full module name, like "temporal_sdk_core::worker::workflow"),
+// to the format used in our JavaScript side logger (i.e. "core/worker/workflow").
+pub fn rust_package_to_js_style(input: &str) -> String {
```
Sushisource, line 239: This seems slightly overkill to me, but you've done it so no reason to get rid of it.

packages/worker/src/worker-options.ts
```diff
@@ -625,6 +625,13 @@ export function defaultSinks(logger?: Logger): InjectedSinks<LoggerSinks> {
   // code that is still calling defaultSinks() expects return type to match the deprecated LoggerSinks API. Silently
   // cast just to mask type checking issues, even though we know this is wrong. Users shouldn't call functions directly
   // on the returned object anyway.
+
+  // If no logger was provided, the legacy behavior was to _lazyly_ set the sink's logger to the Runtime's logger.
```
Sushisource, line 629:
```suggestion
  // If no logger was provided, the legacy behavior was to _lazily_ set the sink's logger to the Runtime's logger.
```

Sushisource (approved):

packages/core-bridge/src/helpers.rs
bergundy: +1, just slap a `core` tag on every log record and be done with it.
Leave the Rust style tags verbatim so they're consistent when logging to console or being forwarded to lang.

packages/common/src/logger.ts
```diff
@@ -13,3 +13,83 @@ export interface Logger {
...
+   * - Sink processing issues;
+   * - Workflow Worker's lifecycle events.
+   */
+  workflowWorker = 'worker/workflow',
+
+  /**
+   * Log Source value for messages emited by the Workflow Bundling process,
+   * when being executed implicitly by the Workflow Worker.
+   */
+  workflowBundler = 'worker/bundler',
```
bergundy, line 71: I'd just call all of this `worker` so users can easily filter out worker generated logs.
There are tradeoffs between the different approaches so I'm fine with keeping this as well.

packages/core-bridge/src/helpers.rs
mjameswh: I'd strongly prefer consistency of the TS SDK with itself than TS SDK with other SDKs. We already took the decision of converting attribute names to TS style in the past, for that reason, and I don't think we should diverge here.

mjameswh: However, I wouldn't mind making TS SDK itself adopt the rust syntax for that specific purpose. Does that make sense to you?

mjameswh:
Flattened log sources: `workflow`, `activity`, `worker`, `core`.

PR #1401, opened 2024-04-23
Add `sdkComponent` and `taskQueue` metadata on log messages

mjameswh:
## What was changed

- Add `taskQueue` metadata attribute on all log messages emitted inside the context of a worker. It is there to make it easier to correlate logs when there are multiple workers running in a same process while diagnosing low level issues (e.g. while running tests).

- Add a `sdkComponent` property in metadata. That property indicates where that log message was emitted from; possible values at this time are:
  - `workflow`: Messages emitted from workflow code (i.e. _user_ code, through the `workflow.log` object);
  - `activity`: Messages emitted from activity code (i.e. _user_ code, through the `activity.log` object);
  - `worker`: messages emitted by the worker itself, including workflow and activity life cycle events;
  - `core`: messages coming from Core SDK.
  
  That extra property will make it easier for users to implement fine-grained filtering of messages (eg. show DEBUG level messages for user-code, but only WARN for messages from the worker). Note that this is only about annotating log messages appropriately to make that possible. We don't plan to actually add support for that kind of filtering in the SDK itself. This can easily be implemented by users themselves in their custom runtime logger.

- `MockActivityEnvironment` no longer defaults the context logger to `Runtime.logger`, as it is not desirable for `Runtime` to get instantiated only for that purpose. Instead, it defaults to sending logs to the console. `MockActivityEnvironment` now accept a logger option, which makes it easier to capture log messages emitted while testing an activity.

## Note

- This change was previously tracked in #1391, but that got merged to `main` before it was completed, and was consequently reverted in PR #1399. Given that a lot has changed in between, it makes more sense to start the PR over, rather than to reopen the original PR.


packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+ * example be used to implement fine-grained filtering of log messages.
+ *
+ * Note that there is no guarantee that this list will remain stable in the
+ * future; values may be added or removed, and messages that are currently
+ * emitted with some `logSource` value may use a different value in the future.
+ */
+export enum LogSource {
+  /**
+   * Log Source value for messages emited from Workflow code, using the {@link Workflow
+   * context logger|workflow.log}. The SDK itself never publishes messages with this source.
```
bergundy, line 29: Can you verify this link is rendered properly?

packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+   */
+  workflow = 'workflow',
+
+  /**
+   * Log Source value for messages emited from an activity, using the {@link activity
+   * context logger|Context.log}. The SDK itself never publishes messages with this source.
```
bergundy, line 35: Same...

packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+   * - Worker's, Activity's, and Workflow's lifecycle events;
+   * - Workflow Activation and Activity Task processing events;
+   * - Workflow Bundling messages;
+   * - Sink processing issues;
+   * - Issues in interracting with the Core SDK library.
+   */
+  worker = 'worker',
+
+  /**
+   * Log Source value for all messages emited by the Core SDK library.
```
bergundy, line 53:
```suggestion
   * Log Source value for all messages emitted by the Rust Core SDK library.
```

packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+  /**
+   * Log Source value for messages emited from a Temporal Worker instance.
+   *
+   * This notably includes:
+   * - Issues with worker or runtime configuration, or the JS execution environment;
+   * - Worker's, Activity's, and Workflow's lifecycle events;
+   * - Workflow Activation and Activity Task processing events;
+   * - Workflow Bundling messages;
+   * - Sink processing issues;
+   * - Issues in interracting with the Core SDK library.
```
bergundy, line 48:
```suggestion
   * - Issues when interacting with the Core SDK library.
```

packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+  activity = 'activity',
+
+  /**
+   * Log Source value for messages emited from a Temporal Worker instance.
+   *
+   * This notably includes:
+   * - Issues with worker or runtime configuration, or the JS execution environment;
+   * - Worker's, Activity's, and Workflow's lifecycle events;
+   * - Workflow Activation and Activity Task processing events;
+   * - Workflow Bundling messages;
```
bergundy, line 46:
```suggestion
   * - Workflow bundling messages;
```

packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+   * Log Source value for messages emited from an activity, using the {@link activity
+   * context logger|Context.log}. The SDK itself never publishes messages with this source.
+   */
+  activity = 'activity',
+
+  /**
+   * Log Source value for messages emited from a Temporal Worker instance.
+   *
+   * This notably includes:
+   * - Issues with worker or runtime configuration, or the JS execution environment;
```
bergundy, line 43: You're inconsistent in capitalization.
If you're capitalizing "Workflow", also capitalize "Worker" or don't for both.

packages/test/src/mock-native-worker.ts
```diff
@@ -163,11 +159,12 @@ export class Worker extends RealWorker {
   }
 
   public constructor(workflowCreator: WorkflowCreator, opts: CompiledWorkerOptions) {
-    // Worker.create() accesses Runtime.instance(), which has some side effects that would not happen (or that would
-    // happen too late) when creating a MockWorker. Force the singleton to be created now, if it doesn't already exist.
-    Runtime.instance();
+    const logger = withMetadata(Runtime.instance().logger, {
+      logSource: LogSource.worker,
+      taskQueue: opts.taskQueue ?? 'default',
```
bergundy, line 164: When would this be empty? Replay worker? Either document why this is here or just omit the task queue completely.

packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+ * Possible values of the `logSource` meta attributes on log messages. This
+ * attribute indicates which subsystem emitted the log message; this may for
+ * example be used to implement fine-grained filtering of log messages.
+ *
+ * Note that there is no guarantee that this list will remain stable in the
+ * future; values may be added or removed, and messages that are currently
+ * emitted with some `logSource` value may use a different value in the future.
+ */
+export enum LogSource {
+  /**
```
antlai-temporal, line 27: Do you also need a "client" source value?

packages/test/src/test-activity-log-interceptor.ts
```diff
@@ -116,7 +121,7 @@ test('Activity Worker logs when activity is cancelled with signal', async (t) =>
   t.not(entry, undefined);
   const { durationMs, ...rest } = entry?.meta ?? {};
   t.true(Number.isInteger(durationMs));
-  t.deepEqual(rest, activityLogAttributes(defaultActivityInfo));
+  t.deepEqual(rest, { ...activityLogAttributes(defaultActivityInfo), logSource: LogSource.worker });
```
bergundy, line 124: I wonder if we need to say this is a `logSource`, it's already an attribute on a `log message`.

Alternatives: `component`, `sdkComponent`, `module`. `source` on its own makes me think of a source code / location.

packages/test/src/test-mockactivityenv.ts
```diff
@@ -1,6 +1,16 @@
 import test from 'ava';
 import { MockActivityEnvironment } from '@temporalio/testing';
-import { CancelledFailure, Context } from '@temporalio/activity';
+import * as activity from '@temporalio/activity';
+import { Runtime } from '@temporalio/worker';
+
+test("MockActivityEnvironment doesn't implicitly instanciate Runtime", async (t) => {
```
bergundy, line 6:
```suggestion
test("MockActivityEnvironment doesn't implicitly instantiate Runtime", async (t) => {
```

packages/test/src/test-activity-log-interceptor.ts
bergundy: Even `subsystem` works for me.

packages/worker/src/activity.ts
```diff
@@ -38,14 +38,18 @@ export class Activity {
...
 
   constructor(
     public readonly info: Info,
     public readonly fn: ActivityFunction<any[], any> | undefined,
     public readonly dataConverter: LoadedDataConverter,
     public readonly heartbeatCallback: Context['heartbeat'],
+    workerLogger: Logger,
     interceptors: ActivityInterceptorsFactory[]
   ) {
+    this.logger = withMetadata(workerLogger, () => this.getLogAttributes());
```
bergundy, line 52: Put a comment explaining this is expected to be logged as a "worker" or, even better (IMHO), use the activity context logger here too.

packages/worker/src/logger.ts
```diff
@@ -87,3 +87,68 @@ export class DefaultLogger implements Logger {
...
+
+  warn(message: string, meta?: LogMetadata): void {
+    this.parentLogger.warn(message, this.resolveMetadata(meta));
+  }
+
+  error(message: string, meta?: LogMetadata): void {
+    this.parentLogger.error(message, this.resolveMetadata(meta));
+  }
+
+  private resolveMetadata(meta?: LogMetadata): LogMetadata {
```
bergundy, line 134: Why not do this eagerly instead of having to evaluate on every call?

packages/worker/src/activity.ts
mjameswh: >even better (IMHO), use the activity context logger here too.

This is exactly what we want to avoid. Users want to be able to distinguish messages emitted by the SDK vs their own code. But will add comment to make this more obvious.

packages/worker/src/logger.ts
mjameswh: We do eagerly when that's possible (most of the time), but there are case where attributes on a sublogger may change after the sublogger's creation, in particular the case of workflow's and activity's context.

packages/common/src/logger.ts
mjameswh: No, the client side API don't use a logger.

packages/common/src/logger.ts
mjameswh: Fixed.

packages/test/src/mock-native-worker.ts
mjameswh: Can't happen in this palce. Removed.

packages/test/src/test-activity-log-interceptor.ts
mjameswh: I want to avoid single words, especially generic ones (source, module, component) as it is more likely to conflict with user's own meta.

mjameswh: `sdkComponent` sounds good to me.

packages/worker/src/activity.ts
mjameswh: Renamed the property to `this.workerLogger`, and added more detailed comment.

packages/common/src/logger.ts
antlai-temporal: Not for this PR, but do we need one? How do we log client side errors?

packages/test/src/mock-native-worker.ts
```diff
@@ -192,6 +193,6 @@ export function isolateFreeWorker(options: WorkerOptions = defaultOptions): Work
         /* Nothing to destroy */
       },
     },
-    compileWorkerOptions(addDefaultWorkerOptions(options))
+    compileWorkerOptions(options, logger)
   );
```
antlai-temporal, line 197: I saw that addDefaultWorkerOptions(options,logger) is now called inside compileWorkerOptions. Is compileWorkerOptions internal only? Can we avoid adding defaults or it does not makes sense? Do we need to document this?

packages/worker/src/logger.ts
```diff
@@ -87,3 +87,68 @@ export class DefaultLogger implements Logger {
...
+
+class LoggerWithMetadata implements Logger {
+  private parentLogger: Logger;
+  private metaChain: (LogMetadata | (() => LogMetadata))[];
+
+  constructor(parent: Logger, meta: LogMetadata | (() => LogMetadata)) {
+    // Flatten recusive LoggerWithMetadata instances
+    if (parent instanceof LoggerWithMetadata) {
+      this.parentLogger = parent.parentLogger;
+      this.metaChain = LoggerWithMetadata.appendToChain(parent.metaChain, meta);
```
antlai-temporal, line 103: Can `this.parentLogger` be of type `LoggerWithMetadata`, i.e., more than one level deep... If not make a comment...

packages/worker/src/worker.ts
```diff
@@ -1334,7 +1344,7 @@ export class Worker {
       // The only way for this observable to be closed is by state changing to DRAINED meaning that all in-flight activities have been resolved and thus there should not be any heartbeats to send.
       this.takeUntilState('DRAINED'),
       tap({
-        complete: () => this.log.debug('Heartbeats complete'),
+        complete: () => this.logger.debug('Heartbeats complete'),
       }),
```
antlai-temporal, line 1348: Do you need logAttributes here?

packages/worker/src/worker.ts
```diff
@@ -1479,7 +1489,7 @@ export class Worker {
   protected workflow$(): Observable<void> {
     // This Worker did not register any workflows, return early
     if (this.workflowCreator === undefined) {
-      this.log.warn('No workflows registered, not polling for workflow tasks');
+      this.logger.warn('No workflows registered, not polling for workflow tasks');
       this.workflowPollerStateSubject.next('SHUTDOWN');
```
antlai-temporal, line 1493: logAttributes? There are a few others in this file with no attributes.

packages/workflow/src/logs.ts
```diff
@@ -67,6 +75,7 @@ export const log: WorkflowLogger = Object.fromEntries(
         return loggerSink[level](message, {
           // Inject the call time in nanosecond resolution as expected by the worker logger.
           [LogTimestamp]: activator.getTimeOfDay(),
+          sdkComponent: SdkComponent.workflow,
```
antlai-temporal, line 78: Do we need timeOfDayToBigint?

packages/worker/src/runtime.ts
```diff
@@ -283,7 +283,7 @@ export class Runtime {
       for (const log of logs) {
         const meta: LogMetadata = {
           [LogTimestamp]: timeOfDayToBigint(log.timestamp),
-          subsystem: log.target,
+          sdkComponent: SdkComponent.core,
```
antlai-temporal, line 286: Is the timestamp to Bigint something the bridge could do? It looks ugly to add the deprecated call with the "internal only" comment in the file above...

antlai-temporal (approved):

packages/worker/src/activity.ts
bergundy: It's debatable because these logs have all of the metadata that the user gets to control.
In my mind I always saw these logs that were previously in the interceptors associated with an activity or workflow.

bergundy (approved):
Apart from the debate about the workflow and activity lifecycle events being emitted as sdkComponent worker, it all LGTM.
I'm not blocking the PR on that though, leaving it up to you.
I'm wondering what others might say if you want to get more opinions.

packages/common/src/logger.ts
mjameswh: > Not for this PR, but do we need one? How do we log client side errors?

I don't think there's a real need to unify client side logging at the moment. Errors on client side are simply thrown back to user code.

Now, the `@grpc/grpc-js` library may technically produce log messages, and it accepts a logger. But I haven't seen anyone reach out for these messages, except when intentionally enabling gRPC's verbose/debug trace messages to diagnose a specific issue, which is not a use case we should be bothering.

Also, if we were at some point to start capturing client side logging, that couldn't go through the `Runtime` class, as we want the `@temporalio/client` package to be usable in some execution contexts where the `@temporalio/worker` package can't run. Sure, we can work this out, i.e. extract unified logger to the `common` package, but is it worth it?

If anyone really want to capture grpc's log messages, they can simply themselves calls that library's `setLogger` themselves.

packages/test/src/mock-native-worker.ts
mjameswh: These are not public APIs.

>  Can we avoid adding defaults or it does not makes sense? 

The `addDefault...Options` vs `compile...Options` is simply a pattern that was adopted internally early on while developing the SDK. There is no reason for one to call either of these without calling the other.

mjameswh: I think the original justification for this pattern was to make the code easier to understand and review… `addDefault` would use an object spread operator to set all missing default options, and `compile` do conversion/object initializations where required. That's clean.

Yet, as time passes, I find this pattern sometime makes it harder to introduce changes to options without breaking backward compatibility, and consequently, concerns that ought to be specific to either one are starting to bleed on the other one. Hence, I'm inclined toward hiding this duality from the caller's perspective, and simply make this an internal implementation detail of the `compile...Options` functions, like I did here.

packages/worker/src/logger.ts
mjameswh: I think the

```
    // Flatten recusive LoggerWithMetadata instances
    if (parent instanceof LoggerWithMetadata) {
```

lines just above already make this clear, no?

packages/worker/src/worker.ts
mjameswh: This is an Activity Worker lifecycle event, not specific to a single activity. Hence, no pertinent `logAttributes` to report.

packages/worker/src/worker.ts
mjameswh: This is an Workflow Worker lifecycle event, not specific to a single workflow. Hence, no pertinent `logAttributes` to report.

packages/workflow/src/logs.ts
mjameswh: Well observed… but we don't. The `getTimeDay()` function provided to the `activator`'s constructor already do `timeOfDayToBigint.

mjameswh:
> Apart from the debate about the workflow and activity lifecycle events being emitted as sdkComponent worker

I'll stick to that for now and move on.

I really want the `workflow` and `activity` tags to only be used on messages that come directly from user code. I know that some users are emitting their own lifecycle messages from custom interceptors (let's just accept for now that they have sufficient reasons to do that), and want to be able to ignore our wf/act life cycle messages. That's why I originally used a tag `worker/activity` and `worker/workflow` for these messages. But given recent discussions at introducing some kind of "log message configuration", there might soon be another way of doing this, at which point I wouldn't object for reassigning these to "workflow" and "activity".

Anyway, documentation is clear to the fact that which message is tagged with which component name may change in the future anyway, so further refinement to this remain possible.

packages/common/src/logger.ts
```diff
@@ -13,3 +13,44 @@ export interface Logger {
...
+   * Component name for messages emited from a Temporal Worker instance.
+   *
+   * This notably includes:
+   * - Issues with Worker or runtime configuration, or the JS execution environment;
+   * - Worker's, Activity's, and Workflow's lifecycle events;
+   * - Workflow Activation and Activity Task processing events;
+   * - Workflow bundling messages;
+   * - Sink processing issues;
+   * - Issues when interacting with the Core SDK library.
+   */
```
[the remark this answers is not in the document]
mjameswh: Removed the sentence. This was about a very specific case, which somehow fit into the "Issues with Worker" item anyway.

packages/worker/src/logger.ts
antlai-temporal: I was referring to `this.parentLogger.parentLogger.`.... so the recursion can be several levels deep not just one

PR #1411, opened 2024-05-02
feat(worker): Add support for HTTP CONNECT Proxy

mjameswh:
## What was changed

- Added support for HTTP CONNECT Proxy.
- At this point, client side support is only possible by setting an environment variable; this will be improved in a future PR.

## Why?

- https://github.com/temporalio/features/issues/323

## Example

Add the `proxy` property on `Connection` or `NativeConnection` options. For example:

```js
  const connection = await NativeConnection.connect({
    address: 'localhost:7233',
    proxy: {
      type: 'http-connect',
      targetHost: '127.0.0.1:8888',
      basicAuth: {
        username: 'user',
        password: 'password',
      }
    },
  });
```


cretz:
As part of this, you may want some HTTP CONNECT feature tests. See https://github.com/temporalio/features/pull/448 for the Go and Java ones, and https://github.com/temporalio/features/compare/main...http-connect-proxy-python for the Python one (not yet merged because Python hasn't released the feature yet).

packages/client/src/connection.ts
```diff
@@ -35,6 +44,13 @@ export interface ConnectionOptions {
    */
   tls?: TLSConfig | boolean | null;
 
+  /**
+   * Proxying configuration.
+   *
+   * @experimental
+   */
+  proxy?: ProxyConfig;
```
cretz, line 52: My concern here is it's unclear that this is HTTP CONNECT proxy (as opposed to SOCKS or something). That's why in Core and Python I was explicit w/ the field name that it's for HTTP CONNECT proxy. Granted other places just assume this is what is meant by "proxy" but I think it's better if we make it clear.

cretz (commented):
Just some behavior comments, I'll let others review the actual TS code

packages/client/src/connection.ts
mjameswh: Hence, `ProxyConfig` has a `type` property, with value `http-connect`. If we are to add support for other proxy protocols at some point (e.g. socks5), I'd prefer not to pollute the connection options with a different property for each of them.

cretz: I see. Arguably a bit rough to have to provide a string literal in addition to the field knowing we probably will never support another type. Basically we traded willingness for our connection options to have multiple fields for a user to have to provide multiple fields, but no biggie. So long as clear the to user, works for me.

packages/client/src/connection.ts
```diff
@@ -291,11 +304,23 @@ export class Connection {
     optionsWithDefaults.metadata['client-name'] ??= 'temporal-typescript';
     optionsWithDefaults.metadata['client-version'] ??= pkg.version;
 
+    const originalGrpcProxy = process.env.grpc_proxy;
+    if (normalizedOptions.proxy) {
+      const { targetHost: target, basicAuth: auth } = normalizedOptions.proxy;
+      const { scheme, hostname: host, port } = parseHttpConnectProxyAddress(target);
+      const authString = auth ? `${auth.username}:${auth.password}@` : '';
+      process.env.grpc_proxy = `${scheme}://${authString}${host}:${port}`;
```
cretz, line 312: Hrmm, I still think we should leave the env alone and just tell people they should use it if they want proxying. Note, this is what happens in Go. Granted we have no synchronous guarantees that our code doesn't get interrupted so we couldn't do this there if we wanted. But still, users are used to the idea of needing an environment variable for this and if our gRPC library chooses to make it more explicit later, we can do the same.

cretz (approved):
Sorry to make for so much backtracking. I personally like that grpc-js's env-var-only limitation is also a user one (if they do better in the future we can too I think).

PR #1414, opened 2024-05-07
fix(bundler): Pass Webpack output to logger in a single record

ikonst:
## What was changed

- Don't include color control characters when logging bundler output.
- Log bundler multi-line output as a single log record.

## Why?

Since we take this output and give it to a logging mechanism:
- Color should be omitted since the log message might not be written to a color-enabled console or encapsulated in JSON thus turning the control characters into a nuisance for reading.
- Multi-line message should be logged as a single record to maintain context.

## Checklist

1. Closes: n/a

2. How was this tested: Locally.

3. Any docs updates needed? None.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,6 +28,7 @@
         "@types/ms": "^0.7.34",
         "@types/node": "^20.10.8",
         "@types/stack-utils": "^2.0.3",
+        "@types/supports-color": "^8.1.3",
         "@typescript-eslint/eslint-plugin": "^6.18.1",
         "@typescript-eslint/parser": "^6.18.1",
         "arg": "^5.0.2",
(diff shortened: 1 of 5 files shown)


mjameswh:
Hello @ikonst!

First of all, sorry for the delay, and thanks for your contribution.

I'm definitely down for emitting Webpack's logs as a single message, rather than split into multiple lines.

We have been a bit hesitant regarding disabling Webpack colors because:
- We feel that it's preferable to bundle workflow ahead of time for production environment;
- Cutting down colors completely would be a regression for developers that are working in non-CI environment;
- We are looking into completely revisit Webpack logging in the near future.

Still, I think we could find a reasonable compromise by only enabling colors if the logger is connected to stderr AND stderr supports colors.

I pushed a quick attempt at that. Would it be possible for you to please confirm that this resolves the pain point from your side?

ikonst:
Yep, certainly

mjameswh (approved):

PR #1417, opened 2024-05-10
Evolve update long-poll implementation

### LINKED ISSUE #1372: [Feature Request] SDK should not return an update handle if the update has not reached the desired state
See https://github.com/temporalio/features/issues/432

dandavison:
Fixes #1372

- Repeatedly send `UpdateWorkflowExecution` until update is >= Accepted or >= requested stage; thereafter send `PollWorkflowExecutionUpdate` if necessary.
- Prevent SDK returning handle to update before it has reached Accepted state.
- Don't swallow gRPC timeout errors from server.

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -1068,8 +1068,10 @@ export class WorkflowClient extends BaseClient {
         output.outcome
       );
       if (
+        !output.outcome &&
         waitForStage ===
-        temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED
+          temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage
+            .UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_COMPLETED
       ) {
         await this._pollForUpdateOutcome(handle.updateId, input.workflowExecution);
       }
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -778,7 +778,7 @@ export class WorkflowClient extends BaseClient {
       },
     };
 
-    // Repeatedly send UpdateWorkflowExecution until update is >= Accepted or >= requested stage (if
+    // Repeatedly send UpdateWorkflowExecution until update is >= Accepted or >= `waitForStage` (if
     // the server receives a request with an update ID that already exists, it responds with
     // information for the existing update).
     let response: temporal.api.workflowservice.v1.UpdateWorkflowExecutionResponse;
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -777,10 +777,19 @@ export class WorkflowClient extends BaseClient {
         },
       },
     };
-    let response: temporal.api.workflowservice.v1.UpdateWorkflowExecutionResponse;
 
+    // Repeatedly send UpdateWorkflowExecution until update is >= Accepted or >= requested stage (if
+    // the server receives a request with an update ID that already exists, it responds with
+    // information for the existing update).
+    let response: temporal.api.workflowservice.v1.UpdateWorkflowExecutionResponse;
     try {
-      response = await this.workflowService.updateWorkflowExecution(req);
+      do {
+        response = await this.workflowService.updateWorkflowExecution(req);
+      } while (
+        response.stage < waitForStage &&
+        response.stage <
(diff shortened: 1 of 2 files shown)


packages/test/src/test-integration-update.ts
```diff
@@ -3,14 +3,39 @@ import { isGrpcServiceError } from '@temporalio/client';
...
 import { helpers, makeTestFunction } from './helpers-integration';
 
-const test = makeTestFunction({ workflowsPath: __filename });
+// Use a reduced server long-poll expiration timeout, in order to confirm that client
+// polling/retry strategies result in the expected behavior, in particular:
+// - An update that is cut off by this timeout but eventually completes should be seen by the client
+//   to complete eventually and seamlessly.
+// - For an update that never completes (e.g. because worker is offline), startUpdate never returns
+//   a handle to the client.
+const LONG_POLL_EXPIRATION_INTERVAL_SECONDS = 1.0;
```
dandavison, line 12: We could reduce this further, but currently this will fail due to a server bug: the server should allow `UpdateWorkflowExecution` requests against completed workflows: https://github.com/temporalio/temporal/pull/6083

packages/test/src/test-integration-update.ts
```diff
@@ -416,3 +433,32 @@ test('Update/Signal/Query example in WorkflowHandle docstrings works', async (t)
...
+  const timeoutPromise = new Promise<string>((f) =>
+    setTimeout(() => f('timeout'), 500 + LONG_POLL_EXPIRATION_INTERVAL_SECONDS * 1000)
+  );
+
+  t.is(
+    await Promise.race([updatePromise, timeoutPromise]),
+    'timeout',
+    'The update call should never return, since it should be waiting until Accepted, yet there is no worker.'
+  );
+});
```
dandavison, line 450: This test fails on `main`.

packages/test/src/test-integration-update.ts
```diff
@@ -416,3 +433,32 @@ test('Update/Signal/Query example in WorkflowHandle docstrings works', async (t)
...
+  await worker.runUntil(async () => {
+    const wfHandle = await startWorkflow(workflowWithUpdates);
+    const arg = 'wait-for-longer-than-server-long-poll-timeout';
+    const updateResult = await wfHandle.executeUpdate(update, { args: [arg] });
+    t.deepEqual(updateResult, [arg]);
+    await wfHandle.executeUpdate(doneUpdate);
+    const wfResult = await wfHandle.result();
+    t.deepEqual(wfResult, [arg, 'done', '$']);
+  });
+});
```
dandavison, line 464: This test passes on `main`, but it's a test worth having.

packages/client/src/workflow-client.ts
```diff
@@ -777,10 +777,19 @@ export class WorkflowClient extends BaseClient {
         },
       },
     };
-    let response: temporal.api.workflowservice.v1.UpdateWorkflowExecutionResponse;
 
+    // Repeatedly send UpdateWorkflowExecution until update is >= Accepted or >= requested stage (if
```
Sushisource, line 781:
```suggestion
    // Repeatedly send UpdateWorkflowExecution until update is >= Accepted or >= `waitForStage` (if
```

Currently reads a bit like "requested" is the name of a stage

packages/test/src/test-integration-update.ts
```diff
@@ -3,14 +3,31 @@ import { isGrpcServiceError } from '@temporalio/client';
...
+// polling/retry strategies result in the expected behavior
+const LONG_POLL_EXPIRATION_INTERVAL_SECONDS = 1.0;
+
+const test = makeTestFunction({
+  workflowsPath: __filename,
+  workflowEnvironmentOpts: {
+    server: {
+      extraArgs: [
+        '--dynamic-config-value',
+        `history.longPollExpirationInterval="${LONG_POLL_EXPIRATION_INTERVAL_SECONDS}s"`,
```
Sushisource, line 16: You don't need to change this - changing the grpc timeout of the call itself will also make the server respond just before that with the empty reply.

That said, if this is easier, works too.

Sushisource (approved):

packages/test/src/test-integration-update.ts
dandavison: > changing the grpc timeout of the call itself will also make the server respond just before that with the empty reply.

I haven't experimented to confirm the behavior, but that's not my understanding: the server responds with an empty response (indicating the most advanced stage known to have been reached) if it hits its own self-imposed deadline. But if the client's request timeout is reached then the server responds with the timeout error.

https://github.com/temporalio/temporal/blob/e7c9edd9732099770538a4ea2c4210847fcea6e1/service/history/api/updateworkflow/api.go#L272-L273

PR #1428, opened 2024-05-23
fix(core): build native libs against glibc 2.24

mjameswh:
## What changed

- In CI, build native artifacts for Linux targets in Docker containers, using the `manylinux-2.24` image; this effectively brings the required glibc version down to 2.18 (we're technically building against glibc 2.24, but it appears that the resulting library file is only binding to 2.18 and older symbols).

## Context

- In the recent CI refactor, we stopped using Buildjet's Ubuntu 18.04 runners, as they are incompatible with some recent and upcoming GHA changes. This resulted in increasing the required glibc version from 2.17 (in 1.9) to 2.29 (in 1.10 on x64) and 2.35 (on arm64). This jump is problematic to several users who have internal requirements of using specific base images (e.g. RHE8).

- Some of the GHA changes that are making this more complicated: Node 16 actions will automatically be updated to use Node 20 in June, but Node 20 require at least glibc 2.28, so older runners will no longer be usable; there's an environment variable that can be set to prevent auto-update, but only til October; at the same time, `actions/*-artifact` v1 and v2 are being completely retired in June, and v3 (which is the last version compatible with Node 16) is being retired in October.

- Our Python and .NET SDKs are already built against manylinux images.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,13 +31,14 @@ jobs:
         include:
           # For Linux targets, use Buildjet's builders to get the oldest supported glibc version
           - platform: linux-x64
-            runner: buildjet-4vcpu-ubuntu-2004
+            runner: buildjet-4vcpu-ubuntu-2204
             target: x86_64-unknown-linux-gnu
-            container: quay.io/pypa/manylinux_2_28_x86_64
+            container: quay.io/pypa/manylinux_2_24_x86_64
             out-file: libtemporal_sdk_typescript_bridge.so
           - platform: linux-arm
             runner: buildjet-4vcpu-ubuntu-2204-arm
             target: aarch64-unknown-linux-gnu
+            container: quay.io/pypa/manylinux_2_24_aarch64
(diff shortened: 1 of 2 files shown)


antlai-temporal:
I see many git actions downgraded from v4 to v3 because the base docker image needed for node 20 does not support them. What is the impact of that? are v3 deprecated already? Is there an alternative base image that could work with node20 and v4? What is the long term plan?

antlai-temporal (approved):

mjameswh:
> I see many git actions downgraded from v4 to v3 because the base docker image needed for node 20 does not support them. What is the impact of that? are v3 deprecated already? Is there an alternative base image that could work with node20 and v4? What is the long term plan?

I'm working on a list of upcoming GHA feature deprecations/removal for all of our projects, and some will indeed be problematic very soon (starting in June, then others in October).

We're not the only concerned by these upcoming changes. See [this comment](https://github.com/actions/runner/issues/2906#issuecomment-2105397058) for example.

PR #1429, opened 2024-05-27
fix(workflow): Don't propagate cancellation from non-cancellable scopes to children

### LINKED ISSUE #1423: [Bug] Unexpected behaviour related to CancellationScope.nonCancellable

Use condition with timeout inside nonCancellable scope, and wait a signal to unlock condition await

### Describe the bug

Hello, I recently saw a weird behaviour which I don’t know if it’s normal or not because I didn’t find any documentation related to this subject. Here is a test code I used to reproduce this issue

```
export async function runCancellableActivity(): Promise<void> {
  try {
    await sleep(1000000);
  } catch (err) {
    await CancellationScope.nonCancellable(async () => {
      try {
        console.log('condition');
        await condition(() => false, 10000); // worked good for => await condition(() => false);
      } catch (error) {
        console.log('still error', error);
      }
    });
    throw err;
  }
}
```
### Minimal Reproduction
my condition is wrapped inside nonCancellable scope, if I understood well, CancelledFailure should never propagated into this scope but it’s not always the case, if I decide to give a timeout to my condition. my app did printed still error on console with CancelledFailure exception.

### Environment/Versions

- OS and processor:  Mac Intel
- Temporal Version: SDK 1.8.6 -> 1.9.2
- Are you using Docker or Kubernetes or building Temporal from source? I used docker and locally host cluster

mjameswh:
## What changed

- Previously, cancellation of a non-cancellable `CancellationScope` would incorrectly propagate to children scopes. This would notably make it impossible to use `condition(fn, timeout)` inside a non-cancellable scope to execute rollback actions on workflow cancellation (see #1423). This PR introduces a fix for this incorrect behavior.

- Fixed `CancellationScope.withTimeout()` to cancel the timer if the function completes before the timeout, similar to how `condition(fn, timeout)` works.

- Timers created by the mocked `setTimeout()` function (part of the Workflow sandbox) would previously bypass Workflow Interceptors. This was both unintentional and undesirable. This PR fixes that issue.

- To avoid breaking existing Workflow histories, all of the preceeding fixes are gated with an SDK flag (i.e. an internal marker written down to history by the SDK, vaguely similar the `patched()` API, but for SDK's internal usage). The cancellation-propagation fix is _disabled by default_ in this PR. It will be turned on by default in a second PR.

## Misc

- Fixes #1423.


packages/workflow/src/internals.ts
```diff
@@ -304,12 +306,15 @@ export class Activator implements ActivationHandler {
...
+  private readonly knownPresentPatches = new Set<string>();
 
   /**
    * Patches we sent to core {@link patched}
    */
-  public readonly sentPatches = new Set<string>();
+  private readonly sentPatches = new Set<string>();
+
+  private readonly knownFlags = new Set<number>();
+  private readonly newlyUsedFlags = new Set<number>();
```
Sushisource, line 317: You don't need to figure out what's newly used or not, Core does that for you. You can just blindly send anything you used in this activation

Sushisource (commented):
Just looked at the patch stuff. Looks fine except the one comment about work you don't need to do.

packages/workflow/src/internals.ts
```diff
@@ -394,10 +399,11 @@ export class Activator implements ActivationHandler {
     }
   }
 
-  getAndResetCommands(): coresdk.workflow_commands.IWorkflowCommand[] {
-    const commands = this.commands;
-    this.commands = [];
-    return commands;
+  concludeActivation(): ActivationCompletion {
+    const commands = this.commands.splice(0);
+    const usedInternalFlags = [...this.newlyUsedFlags];
```
bergundy, line 404: Do you really need to distinguish the newly used ones from the rest? Once a flag is on, it's on for good.

bergundy (approved):

packages/test/src/test-integration-workflows.ts
```diff
@@ -399,3 +402,299 @@ test('Build Id appropriately set in workflow info', async (t) => {
...
+    checks.childScope_localActivityCancelled = expectCancellation(sleepLA(2000));
+    checks.childScope_signalExtWorkflowCancelled = expectCancellation(signalTargetWorkflow.signal('signal'));
+  });
+  checks.childScope_cancelRequestedCancelled = expectCancellation(childScope.cancelRequested);
+
+  parentScope.cancel();
+
+  // Wait for all cancellations to be processed
+  await sleepLA(1);
+  await Promise.resolve();
```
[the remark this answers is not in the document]
mjameswh: We need to flush commands to Core, because its Core that will "confirm" cancellation of the commands.

packages/test/src/helpers-integration.ts
```diff
@@ -47,36 +50,50 @@ const defaultDynamicConfigOptions = [
...
+      : undefined;
     const workflowBundle = await bundleWorkflowCode({
       ...bundlerOptions,
       workflowInterceptorModules: [...defaultWorkflowInterceptorModules, ...(opts.workflowInterceptorModules ?? [])],
       workflowsPath: opts.workflowsPath,
     });
     t.context = {
-      env,
+      envLocal,
+      envTimeSkipping,
```
bergundy, line 88: I was imagining every test run would be parameterized to run either with the CLI or the time skipping server instead of creating both.

packages/test/src/helpers-integration.ts
```diff
@@ -163,3 +180,98 @@ export function helpers(t: ExecutionContext<Context>): Helpers {
     },
   };
 }
+
+export interface HelpersTimeSkipping extends Helpers {
```
bergundy, line 184: I'd just add this to `Helpers` instead of having two different interfaces.
You want the entire suite to be able to run against either implementation (or an external server).

bergundy: Also, if you're extending the interface there's no need to redefine all of the methods.

packages/test/src/mock-internal-flags-interceptor.ts
bergundy, line 1: nit: this isn't an interceptor and shouldn't be name as if it were.

packages/test/src/workflows/index.ts
```diff
@@ -16,7 +16,7 @@ export * from './cancel-workflow';
 export * from './cancellation-error-is-propagated';
 export * from './cancellation-scopes';
 export * from './cancellation-scopes-with-callbacks';
-export * from './child-and-shield';
+export * from './child-and-noncancellable';
```
bergundy, line 19: Might as well refactor all of these tests to use the new format...

packages/test/src/workflows/stack-tracer.ts
```diff
@@ -12,29 +12,29 @@ const { queryOwnWf } = wf.proxyActivities<typeof activities>({
 });
 
 export async function stackTracer(): Promise<[string, string]> {
+  const { executeChild, sleep } = wf;
```
bergundy, line 15: why is this important?

packages/workflow/src/cancellation-scope.ts
```diff
@@ -31,38 +34,56 @@ export interface CancellationScopeOptions {
...
+ *       if (isCancellation(e)) {
+ *         // Run cleanup activity in a non-cancellable scope
+ *         await CancellationScope.nonCancellable(async () => {
+ *           await cleanupActivity();
+ *         }
+ *       } else {
+ *         throw e;
+ *       }
+ *     }
+ *   }
```
bergundy, line 76: Should this be less indented?

packages/workflow/src/cancellation-scope.ts
```diff
@@ -31,38 +34,56 @@ export interface CancellationScopeOptions {
 }
 
 /**
- * In the SDK, Workflows are represented internally by a tree of scopes where the `execute` function runs in the root scope.
- * Cancellation propagates from outer scopes to inner ones and is handled by catching {@link CancelledFailure}s
- * thrown by cancellable operations (see below).
+ * Cancellation Scopes provide the mechanic by which a Workflow may gracefully handle incoming requests for cancellation
```
bergundy, line 37: Awesome docstring!

packages/workflow/src/flags.ts
```diff
@@ -0,0 +1,30 @@
...
+   *   of the timeout, similar to how `condition(fn, timeout)` works.
+   * - Timers created using setTimeout can now be intercepted.
+   *
+   * @since Introduced in 1.10.2/1.10.3.
+   */
+  NonCancellableScopesAreShieldedFromPropagation: defineFlag(1, false),
+} as const;
+
+function defineFlag(id: number, def: boolean): SdkFlag {
+  const flag = { id, default: def };
```
bergundy, line 23: nit: I think the defaults will always be false but this is fine.

bergundy: nit: you might want to prevent duplicate registration for the same ID.

packages/workflow/src/internals.ts
```diff
@@ -785,6 +790,41 @@ export class Activator implements ActivationHandler {
...
+      assertValidFlag(flag);
+      this.knownFlags.add(flag);
+    }
+  }
+
+  public hasFlag(flag: SdkFlag): boolean {
+    if (this.knownFlags.has(flag.id)) {
+      return true;
+    }
+    if (!this.info.unsafe.isReplaying && flag.default) {
```
bergundy, line 821: Looks like `default` is footgun since we wouldn't add it to the set of known flags by default.
I recommend removing that layer of complexity and assume all flags are off by default.

packages/test/src/helpers-integration.ts
mjameswh: Yes, but both must be created and terminated from the before/after hooks.

packages/test/src/mock-internal-flags-interceptor.ts
mjameswh: Good catch. It was at some point, but evolved to what it is now. I'll rename the file.

packages/test/src/workflows/stack-tracer.ts
mjameswh: Without this, I sometime get `Object.executeChild` and `Object.sleep` in stack traces, instead of `executeChild`/`sleep` (but not always). Not sure where the variations come from, but I don't think that we have control on this anyway.

packages/workflow/src/flags.ts
mjameswh: > nit: I think the defaults will always be false but this is fine.

No. Assuming introduction of flags in two distinct releases (let say 1.10.2/1.10.3), the first release would default to false, and the second release would default to true.

packages/workflow/src/internals.ts
mjameswh: See my other comment.

PR #1433, opened 2024-06-04
Expose resource-based autotuning

Sushisource:
## What was changed
Exposed the new resourced based autotuning configuration

Also added worker validate call

## Why?
Get it into the hands of people!

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1430

2. How was this tested:
Added some integration tests

3. Any docs updates needed?
Yes, will need docs on what this is for and how to use it, eventually.

### DIFF
--- a/packages/core-bridge/src/conversions.rs
+++ b/packages/core-bridge/src/conversions.rs
@@ -7,7 +7,6 @@ use neon::{
 };
 use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
 use temporal_client::HttpConnectProxyOptions;
-use temporal_sdk_core::api::worker::{SlotKind, SlotSupplier};
 use temporal_sdk_core::{
     api::telemetry::{Logger, MetricTemporality, TelemetryOptions, TelemetryOptionsBuilder},
     api::{
@@ -21,9 +20,9 @@ use temporal_sdk_core::{
         TestServerConfigBuilder,
     },
     telemetry::{build_otlp_metric_exporter, start_prometheus_metric_exporter},
-    ClientOptions, ClientOptionsBuilder, ClientTlsConfig, FixedSizeSlotSupplier, RealSysInfo,
-    ResourceBasedSlots, ResourceBasedSlotsOptions, ResourceBasedSlotsOptionsBuilder,
-    ResourceSlotOptions, RetryConfig, TlsConfig, TunerBuilder, Url,
+    ClientOptions, ClientOptionsBuilder, ClientTlsConfig, ResourceBasedSlotsOptions,
+    ResourceBasedSlotsOptionsBuilder, ResourceSlotOptions, RetryConfig, SlotSupplierOptions,
+    TlsConfig, TunerHolderOptionsBuilder, Url,
 };
 
 pub enum EphemeralServerConfig {
@@ -66,7 +65,11 @@ pub(crate) trait ObjectHandleConversionsExt {
         &self,
         cx: &mut FunctionContext,
     ) -> NeonResult<HashMap<String, String>>;
-    fn as_slot_supplier(&self, cx: &mut FunctionContext) -> NeonResult<SlotSupplierOptions>;
+    fn as_slot_supplier(
+        &self,
+        cx: &mut FunctionContext,
+        rbo: &mut Option<ResourceBasedSlotsOptions>,
+    ) -> NeonResult<SlotSupplierOptions>;
 }
 
 impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
@@ -400,57 +403,35 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
             js_value_getter!(cx, self, "nonStickyToStickyPollRatio", JsNumber) as f32;
 
         let tuner = if let Some(tuner) = js_optional_getter!(cx, self, "tuner", JsObject) {
-            let mut tuner_holder = TunerBuilder::default();
-            let mut rbslots = None;
-
-            fn to_slot_supplier<SK>(
-                opts: SlotSupplierOptions,
-                maybe_rbslots: &mut Option<Arc<ResourceBasedSlots<RealSysInfo>>>,
-            ) -> Arc<dyn SlotSupplier<SlotKind = SK> + Send + Sync>
-            where
-                SK: SlotKind + Send + Sync + 'static,
-            {
-                match opts {
-                    SlotSupplierOptions::FixedSize(num_slots) => {
-                        Arc::new(FixedSizeSlotSupplier::new(num_slots))
-                    }
-                    SlotSupplierOptions::Resource(opts, rbs) => {
-                        let rbslots_ref = if let Some(ref rbslots) = maybe_rbslots {
-                            rbslots
-                        } else {
-                            let rbslots = Arc::new(ResourceBasedSlots::new_from_options(rbs));
-                            maybe_rbslots.insert(rbslots)
-                        };
-                        rbslots_ref.as_kind(opts)
-                    }
-                }
-            }
+            let mut tuner_holder = TunerHolderOptionsBuilder::default();
+            let mut rbo = None;
 
             if let Some(wf_slot_supp) =
                 js_optional_getter!(cx, &tuner, "workflowTaskSlotSupplier", JsObject)
             {
-                tuner_holder.workflow_slot_supplier(to_slot_supplier(
-                    wf_slot_supp.as_slot_supplier(cx)?,
-                    &mut rbslots,
-                ));
+                tuner_holder.workflow_slot_options(wf_slot_supp.as_slot_supplier(cx, &mut rbo)?);
             }
             if let Some(act_slot_supp) =
                 js_optional_getter!(cx, &tuner, "activityTaskSlotSupplier", JsObject)
             {
-                tuner_holder.activity_slot_supplier(to_slot_supplier(
-                    act_slot_supp.as_slot_supplier(cx)?,
-                    &mut rbslots,
-                ));
+                tuner_holder.activity_slot_options(act_slot_supp.as_slot_supplier(cx, &mut rbo)?);
             }
             if let Some(local_act_slot_supp) =
                 js_optional_getter!(cx, &tuner, "localActivityTaskSlotSupplier", JsObject)
             {
-                tuner_holder.local_activity_slot_supplier(to_slot_supplier(
-                    local_act_slot_supp.as_slot_supplier(cx)?,
-                    &mut rbslots,
-                ));
+                tuner_holder.local_activity_slot_options(
+                    local_act_slot_supp.as_slot_supplier(cx, &mut rbo)?,
+                );
             }
-            tuner_holder.build()
+            if let Some(rbo) = rbo {
+                tuner_holder.resource_based_options(rbo);
+            }
+            match tuner_holder.build_tuner_holder() {
+                Err(e) => {
+                    return cx.throw_error(format!("Invalid tuner options: {:?}", e));
+                }
+                Ok(th) => Arc::new(th),
+            }
         } else {
             return cx.throw_error("Missing tuner");
         };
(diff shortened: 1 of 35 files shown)


Sushisource:
I don't love how kind of annoying it is to pass all these types through to the bridge and convert everything properly. It's a bit annoying. I might make some more obvious options on the core side to allow for more direct conversion such that the bridge doesn't actually build these slot suppliers, but rather can just pass options through to Core for "known" supplier implementations.

If people agree that'd be nicer I'll make the Core changes and update this

packages/worker/src/worker-options.ts
```diff
@@ -744,20 +796,21 @@ export function compileWorkerOptions(rawOpts: WorkerOptions, logger: Logger): Co
...
-    opts.maxConcurrentWorkflowTaskExecutions = opts.maxCachedWorkflows;
-  }
-  if (opts.maxCachedWorkflows > 0 && opts.maxConcurrentWorkflowTaskExecutions < 2) {
-    logger.warn(
-      "maxConcurrentWorkflowTaskExecutions can't be lower than 2 if maxCachedWorkflows is non-zero. Defaulting to 2."
-    );
-    opts.maxConcurrentWorkflowTaskExecutions = 2;
-  }
+  // if (opts.maxCachedWorkflows > 0 && opts.maxConcurrentWorkflowTaskExecutions > opts.maxCachedWorkflows) {
+  //   logger.warn(
```
Sushisource, line 800: This could technically stay, but would require some slightly annoying checks around looking to see if they're only using the "old" options. I can restore it if we think it's worth keeping.

mjameswh: What will happen if they provide settings that don't respect these invariants? I think that should stay. However, it would be possible to move these to the corresponding if block in `addDefaultWorkerOptions`…

packages/core-bridge/src/conversions.rs
```diff
@@ -533,4 +585,43 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
             )),
         }
     }
+
+    fn as_slot_supplier(&self, cx: &mut FunctionContext) -> NeonResult<SlotSupplierOptions> {
+        if let Some(num_slots) = js_optional_value_getter!(cx, self, "numSlots", JsNumber) {
```
mjameswh, line 590: As a general rule, I'd prefer to avoid duck-typing the config, and even more so from the Rust side. Let's add a `type` discriminator property in `SlotSupplier`, similar to what we have in `EphemeralServerConfig`. I'd even be inclined to expose that property to users (i.e. they would have to specify the `type` property when explicitly passing a `WorkerOptions.tunner`).

packages/core-bridge/ts/worker-tuner.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * @experimental
+ *
+ * Options for a {@link ResourceBasedTuner} to control target resource usage
+ */
+export interface ResourceBasedTunerOptions {
+  // A value between 0 and 1 that represents the target (system) memory usage. It's not recommended
+  // to set this higher than 0.8, since how much memory a workflow may use is not predictable, and
+  // you don't want to encounter OOM errors.
+  targetMemoryUsage: number;
```
mjameswh, line 52: I'm ok with this description as long as resource-based tuners are experimental.

But before we reach GA, we'll need to better understand what is accounted as "memory usage" and "available memory" for the purpose of this tuner, and adjust the description accordingly.

Among other things:
- How does "memory usage" relates with Node's heap? e.g. `--max-old-space-size`?
- Is this memory usage/available memory of the process or the system? 
- What's the impact of Worker Threads?
- Are we sure that available memory is properly determined in containerized environments?

packages/test/src/test-worker-tuner.ts
```diff
@@ -0,0 +1,73 @@
...
+      tuner: {
+        tunerOptions: resourceBasedTunerOptions,
+        activityTaskSlotOptions: {
+          minimumSlots: 2,
+          maximumSlots: 10,
+          rampThrottle: 20,
+        },
+      },
+    });
+    const client = new WorkflowClient();
```
mjameswh, line 30: Use `new Client()`, then `client.workflow`.

packages/worker/src/worker-options.ts
```diff
@@ -185,17 +187,31 @@ export interface WorkerOptions {
...
+  tuner?: WorkerTuner;
+
   /**
    * Maximum number of Activity tasks to execute concurrently.
    * Adjust this to improve Worker resource consumption.
-   * @default 100
+   *
+   * Mutually exclusive with the {@link tuner} option.
+   *
+   * @default 100 if no {@link tuner} is set
```
mjameswh, line 204: Just a note that once `WorkerTuner` reaches GA, we can consider deprecating these options, promoting explicitly using a Fixed Size tuner instead.

packages/worker/src/worker-tuner.ts
```diff
@@ -0,0 +1,131 @@
...
+export interface ResourceBasedTuner {
+  // Options for the tuner
+  tunerOptions: ResourceBasedTunerOptions;
+  // Options for workflow task slots. Defaults to a minimum of 2 slots and a maximum of 1000 slots
+  // with no ramp throttle
+  workflowTaskSlotOptions?: ResourceBasedSlotOptions;
+  // Options for activity task slots. Defaults to a minimum of 1 slots and a maximum of 2000 slots
+  // with 50ms ramp throttle
+  activityTaskSlotOptions?: ResourceBasedSlotOptions;
+  // Options for local activity task slots. Defaults to a minimum of 1 slots and a maximum of 2000
```
mjameswh, line 55: I know this might a bit tricky to implement, but I get the feeling that if `localActivityTaskSlotSupplier` is not specified, then both regular and activities and local activities should be controlled by a single tuner…

packages/worker/src/worker-options.ts
Sushisource: They actually will get checked by Core anyway, if the supplier provides a total slots number which fixed size ones do

packages/worker/src/worker-tuner.ts
Sushisource: Yeah, I could see that. It's not hard to implement if we do it in Core. So I could do it that way.

If I make that change though I also might change how the options are passed through generally to make things a bit easier. Let me know if you have thoughts on passing through just some known option sets rather than constructing the whole tuner in the bridge.

mjameswh:
> I don't love how kind of annoying it is to pass all these types through to the bridge and convert everything properly. It's a bit annoying. I might make some more obvious options on the core side to allow for more direct conversion such that the bridge doesn't actually build these slot suppliers, but rather can just pass options through to Core for "known" supplier implementations.
> 
> If people agree that'd be nicer I'll make the Core changes and update this

Personally, I'm more annoyed by all the `js_optional_getter!` gymnastic in bridge, and duplicated definitions between TS and Rust side that we need to do in bridge, but you can't really do much about it…

For the slot supplier thing, yes, moving that to the Core side would make sense, though my argument for doing so would really be about reducing impact on every lang bridges while the Worker Tuner APIs evolves.

packages/core-bridge/ts/worker-tuner.ts
mjameswh: > So nothing specific to node should really matter that much, all that matters is how the system usage is being reported.

Yeah… and that's why I'm concerned that this might actually lead to very unintuitive behaviors.

For context, Node (and I think that's partially true of all of our core-based SDKs languages) has its own notion of "heap", which gets allocated in large blocks, and then heap objects get allocated from that heap. That's the memory that is GC managed. The problem is that "system memory usage" is often not representative (i.e. not proportional to) of "heap usage", which I think could result in a few pathological cases.

For example, let's imagine that I run a worker with `targetMemoryUsage` set to 80%, on a system with 64 GB of RAM, without configuring Node's `--max-old-space-size` (or set it to a value <60%). IIRC, Node will by default limit its heap size to the smallest of (25% of available system memory or 4GB). With that config, system memory usage will never grow even near the 80% (4GB / 64GB, plus some overhead for non-heap stuff would put us at ~10%). Consequently, the `ResourceBasedAllocator` will very likely raise all `maxXxxSlot` parameters to their max values, but then the process dies from running out of heap…

I think the opposite direction is also possible. For various reasons, many GC implementations have tend to be reluctant to hand back heap blocks to the system. So it might be possible for memory usage to quickly grow to the target system memory usage, yet heap usage to be relatively low.

All I'm saying is that we'll need to play with this more to see how well that works. Of course, some of this may result in refinement to the engine itself, but we'll most certainly also need to clarify documentation…

packages/worker/src/worker-tuner.ts
```diff
@@ -0,0 +1,131 @@
...
+  ResourceBasedTunerOptions,
+  WorkerTuner as NativeWorkerTuner,
+  SlotSupplier as NativeSlotSupplier,
+} from '@temporalio/core-bridge';
+import { Duration, msToNumber } from '@temporalio/common/lib/time';
+
+export type WorkerTuner = ResourceBasedTuner | TunerHolder;
+
+type ResourceBasedSlotsForType = ResourceBasedSlotOptions & {
+  tunerOptions: ResourceBasedTunerOptions;
```
cretz, line 12: Haven't looked at TS, but is it common for us to reference types from the bridge project in user-facing code? Sure I know they are interfaces, but seems a bit off

packages/worker/src/worker-tuner.ts
```diff
@@ -0,0 +1,131 @@
...
+  // slots with 50ms ramp throttle
+  localActivityTaskSlotOptions?: ResourceBasedSlotOptions;
+}
+
+/**
+ * @experimental
+ *
+ * This tuner allows for different slot suppliers for different slot types.
+ */
+export interface TunerHolder {
```
cretz, line 65: How is this interface different than `WorkerTuner`?

packages/worker/src/worker-tuner.ts
```diff
@@ -0,0 +1,131 @@
...
+  workflowTaskSlotSupplier: SlotSupplier;
+  activityTaskSlotSupplier: SlotSupplier;
+  localActivityTaskSlotSupplier: SlotSupplier;
+}
+
+export function asNativeTuner(tuner: WorkerTuner): NativeWorkerTuner {
```
cretz, line 71: Does this need to be exported. I am thinking we want the smallest user-facing API surface we can get away with.

packages/core-bridge/ts/worker-tuner.ts
Sushisource: Yeah, that makes good sense. The usage determination interface is pluggable in Core, so we can implement different ways of measuring.

packages/worker/src/worker-tuner.ts
Sushisource: It does so it can be used from the other parts of the package. It is not publicly exported at a top level.

packages/worker/src/worker-tuner.ts
Sushisource: The slot supplier types used inside are different. The core layer one uses a number instead of a duration

packages/worker/src/worker-tuner.ts
Sushisource: It's not that common but there's no meaningful issue. They would be exact dupes. If they ever need to diverge it can be done compatibly.

packages/core-bridge/src/conversions.rs
Sushisource: Hmm, I did have this earlier and then I stopped doing it specifically because forcing the user to type it seemed a little annoying. If we're fine with that, then I'll add it back.

packages/worker/src/worker-options.ts
Sushisource: I've added the check back, but, my inclination is to move all of this, including the checks for cached workflow settings, into Core since there's no reason to dupe this everywhere. I'll follow up on that in another Core PR.

mjameswh: > my inclination is to move all of this, including the checks for cached workflow settings, into Core since there's no reason to dupe this everywhere. I'll follow up on that in another Core PR.

I'd also be in favor of avoiding that kind of dupe, but I must point out these checks were _intentionally_ added here (knowing they were dup'ed) because that allowed better warning/error reporting, and made it easier to handle default values (after all, these checks are pretty much linked to the ones just above that will determine a default value for `maxCachedWorkflows`, number of pollers, etc). So I'm not against moving this out to Core, but that might be trickier to do properly than it seems at first.

packages/core-bridge/src/conversions.rs
mjameswh: I'm ok with making `type` optional on the user facing API. However, I'd definitely want to avoid doing duck-typing in bridge. That's not that bad, really, but I strive at making the TS/Rust interface easier to maintain/less error-prone, and I feel that having Rust code guess the type of objects based on the presence or not of some property kind goes against that goal.

packages/worker/src/worker-tuner.ts
mjameswh: > Haven't looked at TS, but is it common for us to reference types from the bridge project in user-facing types? Sure I know they are interfaces, but seems a bit off
AFAIK, only for `TelemetryOptions` and things under it.

> but there's no meaningful issue. [...] If they ever need to diverge it can be done compatibly.
At the very least, to avoid future problems, we must reexport those types from the worker package, and discourage users from ever importing them directly from the core package. But even then, that might still be a concern on the long term due to incoming links to the generated API docs, and therefore, we should generally try to avoid doing so. Now, in this specific case, given that the API is experimental, I don't mind just doing a reexport.

packages/core-bridge/src/conversions.rs
```diff
@@ -401,14 +402,47 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
...
+                tuner_holder.activity_slot_options(act_slot_supp.as_slot_supplier(cx, &mut rbo)?);
+            }
+            if let Some(local_act_slot_supp) =
+                js_optional_getter!(cx, &tuner, "localActivityTaskSlotSupplier", JsObject)
+            {
+                tuner_holder.local_activity_slot_options(
+                    local_act_slot_supp.as_slot_supplier(cx, &mut rbo)?,
+                );
+            }
+            dbg!(&rbo);
```
mjameswh, line 426: dbg?

mjameswh:
Sorry, didn't have time for a second round of review today. Monday.

packages/core-bridge/ts/worker-tuner.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * @experimental
```
mjameswh, line 2: nit: move the `@experiemntal` tag line at the end of the comment. Same for other comments in this file.

packages/test/src/test-worker-tuner.ts
```diff
@@ -0,0 +1,154 @@
+import test from 'ava';
+import { v4 as uuid4 } from 'uuid';
+import { Client } from '@temporalio/client';
+import { ResourceBasedTunerOptions } from '@temporalio/core-bridge';
+import { defaultOptions } from './mock-native-worker';
+import { RUN_INTEGRATION_TESTS, Worker } from './helpers';
+import { successString } from './workflows';
+
+if (RUN_INTEGRATION_TESTS) {
```
mjameswh, line 9: No worry for this file, but in the future, please uses `test-integration-workflows.ts` as a base for integration tests. We plan to completely remove the older pattern.

packages/worker/src/worker-options.ts
```diff
@@ -185,17 +187,31 @@ export interface WorkerOptions {
    */
   dataConverter?: DataConverter;
 
+  /**
```
mjameswh, line 190: Add `@experimental`

packages/worker/src/worker-tuner.ts
```diff
@@ -0,0 +1,169 @@
...
+  WorkerTuner as NativeWorkerTuner,
+  SlotSupplier as NativeSlotSupplier,
+} from '@temporalio/core-bridge';
+import { Duration, msToNumber } from '@temporalio/common/lib/time';
+
+export { FixedSizeSlotSupplier, ResourceBasedTunerOptions };
+
+export type WorkerTuner = ResourceBasedTuner | TunerHolder;
+
+type ResourceBasedSlotsForType = ResourceBasedSlotOptions & {
```
mjameswh, line 13: `@experimental`

packages/worker/src/worker-tuner.ts
```diff
@@ -0,0 +1,169 @@
...
+  type: 'resource-based';
+  tunerOptions: ResourceBasedTunerOptions;
+};
+
+/**
+ * @experimental
```
mjameswh, line 19: Move `@experimental` to the last line.

mjameswh (approved):
A few nits to fix. After that, feel free to merge.

PR #1436, opened 2024-06-06
Expose next retry delay

Sushisource:
## What was changed
Expose the new next retry delay field for activity failures

## Why?
Parity

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1349

2. How was this tested:
Added test verifying property is sent to Core, where functionality is already tested

3. Any docs updates needed?

### DIFF
--- a/packages/common/src/time-base.ts
+++ b/packages/common/src/time-base.ts
@@ -1,6 +1,0 @@
-import { StringValue } from 'ms';
-
-/**
- * A duration, expressed either as a number of milliseconds, or as a {@link https://www.npmjs.com/package/ms | ms-formatted string}.
- */
-export type Duration = StringValue | number;
--- a/packages/common/src/time-base.ts
+++ b/packages/common/src/time-base.ts
@@ -1,0 +1,6 @@
+import { StringValue } from 'ms';
+
+/**
+ * A duration, expressed either as a number of milliseconds, or as a {@link https://www.npmjs.com/package/ms | ms-formatted string}.
+ */
+export type Duration = StringValue | number;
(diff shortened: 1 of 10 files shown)


packages/common/src/time-base.ts
```diff
@@ -0,0 +1,6 @@
+import { StringValue } from 'ms';
+
+/**
+ * A duration, expressed either as a number of milliseconds, or as a {@link https://www.npmjs.com/package/ms | ms-formatted string}.
+ */
+export type Duration = StringValue | number;
```
mjameswh, line 6: I don't mind the refactor, but why?

packages/test/src/test-worker-activities.ts
```diff
@@ -246,3 +254,31 @@ test('Non ApplicationFailure TemporalFailures thrown from Activity are wrapped w
     t.is(result?.failed?.failure?.applicationFailureInfo?.type, 'TemporalFailure');
   });
 });
+
+test('nextRetryDelay in activity failures is propagated to Core', async (t) => {
```
mjameswh, line 258: Please also add an integration test for this.

packages/common/src/time-base.ts
Sushisource: Circular import

mjameswh: I see. Can you try instead to move the `WorkflowExecutionAlreadyStartedError` class out of `errors.ts` to `failure.ts`? I think that should break the cycle in a more natural way, given that `WorkflowExecutionAlreadyStartedError` is indeed a `TemporalFailure`, not a generic Error.

mjameswh (approved):
Looks great, thanks!

Issue #1440, opened 2024-06-12
[Bug] Emitted client metrics missing temporal_ prefix


tsurdilo:
With TS release 1.10.1
looks like client metrics like request_failure, long_request_failure etc are missing the default temporal_ prefix in the metric name

Request to add this back. Thanks.

PR #1441, opened 2024-06-12
Fix clients being created without prefixed metric meter

Sushisource:
## What was changed
Client metrics were being emitted without the prefix

## Why?
Bug

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1440

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/test/src/test-prometheus.ts
+++ b/packages/test/src/test-prometheus.ts
@@ -106,7 +106,6 @@ test.serial('Exporting Prometheus metrics from Core works with lots of options',
       // Verify 'total' suffix
       t.assert(text.includes('temporal_worker_start_total'));
       // Verify prefix exists on client request metrics
-      console.log(text);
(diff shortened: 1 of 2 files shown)


cretz (approved):

packages/test/src/test-prometheus.ts
```diff
@@ -105,6 +105,9 @@ test.serial('Exporting Prometheus metrics from Core works with lots of options',
       );
       // Verify 'total' suffix
       t.assert(text.includes('temporal_worker_start_total'));
+      // Verify prefix exists on client request metrics
+      console.log(text);
```
cretz, line 109: Can probably remove this log

mjameswh (approved):
Please remove the `console.log()` line before merging.
Thanks, @Sushisource, for fixing this.

PR #1449, opened 2024-06-22
Instrument signals with OTel

iravid:
## What was changed

Add instrumentation to `signal` in the client interceptor and `handleSignal` in the workflow inbound interceptor.

## Why?

We're starting activities from signal handlers and would like to see them on the same trace.

## Checklist

2. How was this tested:

Haven't seen tests in this module - lmk if needed! Will test this soon on our own project.

4. Any docs updates needed?

Don't think so!


mjameswh:
Thanks @iravid for sharing this.

There's a few things that needs to happen before we can merge this:
- Handle `WorkflowOutboundCallsInterceptor.signalWorkflow` so that spans propagate when a workflow signals another workflow;
- Add a corresponding check in [test-otel.ts](https://github.com/temporalio/sdk-typescript/blob/main/packages/test/src/test-otel.ts#L113).

Also, if you feel up to the task, we should do the same with `handleUpdate` and friends.

iravid:
Thanks @mjameswh! Will happily address those two; thanks for the pointer to the tests.

I'm a bit short on time this week so if alright by you will save instrumentation of the other operations for another PR?

mjameswh:
> I'm a bit short on time this week so if alright by you will save instrumentation of the other operations for another PR?

Sure, no problem.

iravid:
Hey @mjameswh apologies for the delay here (😅) - added the requested adjustments and tested locally by removing the `.skip` on the integration test and running `npm run test` with `RUN_INTEGRATION_TESTS=true`

mjameswh (approved):
This looks good.

Thanks a lot for adding tests, @iravid, and sorry for the delays!

Issue #1317, opened 2023-12-11
[Feature Request] Expose `updateId` to Update handlers


mjameswh:
### Describe the solution you'd like

A Workflow Update carries an `updateId` field used to dedup updates. However, server side dedup is per-workflow-execution. Workflow implementers may have legitimate reasons to extend that guarantee across continue-as-new boundary.

Doing that effectively require access to the `updateId`, which is not exposed to the update handler at the moment.

We should make this information accessible through an `updateInfo()` function.

### Extra thoughts

`updateInfo()` would rely on an `asyncLocalStorage` set just before invoking interceptors. It might be interesting to make that `asyncLocalStorage` more general, so that it also get set while handling queries and signals, thus providing more general execution context tracking, with only marginal extra difficulty. This could possibly allow providing similar `signalInfo()` and `queryInfo()` APIs (though there seems to be very limited use cases for that ATM), but also allow detecting stuff like workflow commands started from a query handler.

mjameswh:
Note that, even though the update handler don’t receive the `updateId`, workflow interceptors do. Therefore, in the mean time, one can capture the `updateId` form an interceptor, and make it accessible to your actual update handler code using an async local storage...

Issue #1432, opened 2024-05-30
[Feature Request] Simply proper usage of `AsyncLocalStorage` in Workflow context


mjameswh:
### Description

`AsyncLocalStorage` are frequently used inside of Workflow context, for example to implement correlation/tracing/context propagation mechanisms. However, most users don't know (or forget) that ALS must be "disabled" when they are no longer required (i.e. when the workflow execution gets evicted from cache). Otherwise, they do retain heap and still get called for every single `AsyncResource` (e.g. mostly `Promise` in this context) creation.

Given that we already control access to the `AsyncLocalStorage` in Workflow context, it would be easy for the SDK to track instantiation of `AsyncLocalStorage`, and automatically dispose them when the execution context is about to be destroyed.

We also need to investigate either usage of ALS inside of Workflow Context is efficient. Since ALS hooks work at the Worker Thread level, each ALS instance “tags” (i.e. adds a `[kResourceStore]` property on) every single `AsyncResource` that gets created in that Worker Thread, without regard to which VM the ALS or the AsyncResource was created in.

This could possibly mean that, even when ALS get disposed properly, using ALS inside workflows may currently add a memory and performance penalty that grows exponentially with the number of currently cached workflows. Based on memory snapshots provided by a customer, the constant multiplier here would be relatively small (48 bytes per “tag” in >99% of cases), and therefore, this problem may not be immediately apparent, but could still have sizable impact on Workflow Worker where `maxCachedWorkflow` is high.

Again, given that we already control access to the `AsyncLocalStorage` in Workflow context, it would be possible for the SDK to provide a custom implementation of ALS that is workflow-aware and/or that scales better.

neelance:
Please see #1557.

PR #1450, opened 2024-06-25
feat(workflow): Expose updateId to update handlers

antlai-temporal:
## What was changed

A method `currentUpdateInfo()` has been added to workflow to obtain the name and id of a currently executing update handler from within. The solution uses AsyncLocalStorage to propagate  this information across asynchronous invocations. Calling `currentUpdateInfo()` within a workflow, but outside the async scope of an update handler invocation, returns `undefined`.

## Why?

Facilitate deduplication of updates, particularly across CAN boundaries.

## Checklist

1. Closes 
#1317

### DIFF
--- a/packages/workflow/src/interfaces.ts
+++ b/packages/workflow/src/interfaces.ts
@@ -190,6 +190,21 @@ export interface UnsafeWorkflowInfo {
   readonly isReplaying: boolean;
 }
 
+/**
+ * Information about a workflow update.
+ */
+export interface UpdateInfo {
+  /**
+   *  A workflow-unique identifier for this update.
+   */
+  readonly id: string;
+
+  /**
+   *  The update type name.
+   */
+  readonly name: string;
+}
+
 export interface ParentWorkflowInfo {
   workflowId: string;
   runId: string;
--- a/packages/workflow/src/interfaces.ts
+++ b/packages/workflow/src/interfaces.ts
@@ -192,6 +192,8 @@ export interface UnsafeWorkflowInfo {
 
 /**
  * Information about a workflow update.
+ *
+ * @experimental
  */
 export interface UpdateInfo {
   /**
--- a/packages/workflow/src/internals.ts
+++ b/packages/workflow/src/internals.ts
@@ -21,6 +21,7 @@ import { checkExtends } from '@temporalio/common/lib/type-helpers';
 import type { coresdk, temporal } from '@temporalio/proto';
 import { alea, RNG } from './alea';
 import { RootCancellationScope } from './cancellation-scope';
+import { UpdateScope } from './update-scope';
 import { DeterminismViolationError, LocalActivityDoBackoff, isCancellation } from './errors';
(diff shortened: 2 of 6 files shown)


packages/workflow/src/interfaces.ts
```diff
@@ -190,6 +190,21 @@ export interface UnsafeWorkflowInfo {
   readonly isReplaying: boolean;
 }
 
+/**
+ * Information about a workflow update.
```
mjameswh, line 194: Add `@experimental` here

packages/workflow/src/internals.ts
```diff
@@ -629,33 +630,40 @@ export class Activator implements ActivationHandler {
...
           } else {
             throw error;
           }
-        })
+        });
+      untrackPromise(res);
+      return res;
+    };
+    untrackPromise(
+      UpdateScope.updateWithInfo(updateId, name, doUpdateImpl).catch((error: any) => {
```
mjameswh, line 664: That catch statement is useless.

I think in this case, we have sufficient certainty that the promise can't fail. `doUpdateImpl` won't let a throw bubble up. Therefore, you don't even need the `catch()` statement.

packages/workflow/src/workflow.ts
```diff
@@ -868,6 +870,19 @@ export function workflowInfo(): WorkflowInfo {
...
 
+/**
+ * Get information about the current update if any.
+ *
+ * @return Info for the current update handler the code calling this is executing
+ * within if any.
+ *
+ * @experimental
+ */
+export function currentUpdateInfo(): UpdateInfo | undefined {
```
mjameswh, line 881: I'd call it `updateInfo()`, similar to `workflowInfo()` and `activityInfo()`.

mjameswh (changes requested):
See review comments.

packages/workflow/src/interfaces.ts
antlai-temporal: Done

packages/workflow/src/workflow.ts
antlai-temporal: I'm following the convention discussed in https://github.com/temporalio/sdk-python/pull/544 where @cretz suggested to use `current` because it is only available within the scope of a handler execution. But I'm happy to change that, if it is ok to be different across languages...

packages/workflow/src/internals.ts
antlai-temporal: I was worried about storage.run throwing after calling `storage.disable()` but I think you are right, it does not throw, it just creates another context...

mjameswh: TBH, I don't like the fact that we create yet another ASL, due to [a memory issue](https://github.com/temporalio/sdk-typescript/issues/1432). But that's ok for now.

packages/workflow/src/workflow.ts
mjameswh: Arguable, but fine... Let's align with python.

mjameswh (approved):

PR #1452, opened 2024-06-27
feat(workflow): Unify GRPC::CANCELED and DEADLINE_EXCEEDED in an SDK Timeout exception for Update

antlai-temporal:
## What was changed

Introduced a new exception `WorkflowUpdateRPCTimeoutOrCancelledError` that unifies grpc exceptions `DEADLINE_EXCEEDED` and `CANCELLED` during an update call for starting or polling for a result. 

## Checklist

1. Closes 
#1422 
2. How was this tested:
Integration tests

### DIFF
--- a/packages/client/src/errors.ts
+++ b/packages/client/src/errors.ts
@@ -50,6 +50,20 @@ export class WorkflowUpdateFailedError extends Error {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


mjameswh (approved):
LGTM

Issue #1453, opened 2024-07-01
Refactor stack trace queries to use new protos


twin-drill:
Found in:
https://github.com/temporalio/api/commit/29b1f4eebbf90d5ef034d7cb75b30b6ae0d623ab

Small reminder to refactor TS query protos regarding stack traces -- we have to rename a few objects and fields + add the `internal_code` flag to `FileLocation`s.

Files that need to be changed at first glance:
packages/workflow/src/internals.ts

PR #1454, opened 2024-07-02
Update stack trace interfaces to match new protos

twin-drill:
## What was changed
Update stack trace interfaces to match new protos (JSON-equivalent)

Closes https://github.com/temporalio/sdk-typescript/issues/1453


twin-drill:
`prettier` 3.1.1 doesn't like some things that 3.3.2 is okay with. Should I upgrade the deps or rerun it on 3.1.1?

Sushisource (approved):
In this case I think this looks good. I might be worth adding a test to verify that you can interpret the JSON output from these hand-written structs as the real protos, but we don't have to block the PR on it.

Up to you if you want to do it to be sure the translation is correct.

twin-drill:
Ready to be merged, CI is still flaky but failures shouldn't be caused directly by this code.

PR #1455, opened 2024-07-02
client: Make gRPC INTERNAL status retryable

mjameswh:
## What was changed and why?

- In client, some variants of gRPC's `INTERNAL` status code are now considered as retryable (more specifically: errors where message contain `RST_STREAM with code 0` or `RST_STREAM with code 2`, except TLS related issues). Users are seeing these specific errors quite regularly, apparently due to transient issues at the HTTP2 connection level. After discussion with both SDK and Server team, we believe that in most case retrying these errors would be beneficial, and in the worst case, should only result in minor waste of resource on the server.

- In WorkflowClient, `list().withHistories()` no longer swallow client errors (they would previously result in an entry where `history` is `undefined`, with no way to know the underlying issue).

### DIFF
--- a/packages/client/src/grpc-retry.ts
+++ b/packages/client/src/grpc-retry.ts
@@ -100,6 +100,7 @@ export function defaultGrpcRetryOptions(options: Partial<BackoffOptions> = {}): 
 const retryableCodes = new Set([
   grpc.status.UNKNOWN,
   grpc.status.RESOURCE_EXHAUSTED,
+  grpc.status.INTERNAL,
   grpc.status.UNAVAILABLE,
   grpc.status.ABORTED,
   grpc.status.DATA_LOSS,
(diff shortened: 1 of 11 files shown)


packages/client/src/grpc-retry.ts
```diff
@@ -100,6 +100,7 @@ export function defaultGrpcRetryOptions(options: Partial<BackoffOptions> = {}):
 const retryableCodes = new Set([
   grpc.status.UNKNOWN,
   grpc.status.RESOURCE_EXHAUSTED,
+  grpc.status.INTERNAL,
   grpc.status.UNAVAILABLE,
   grpc.status.ABORTED,
```
Sushisource, line 105: Wait, if it's supposed to be becoming _retryable_, and this is the list of _retryable_ codes...

Quinn-With-Two-Ns:
>gRPC's INTERNAL status code is now considered as retryable. That's what we do in all other SDKs.

I don't think this is true. All SDKs do **not** consider INTERNAL retryable.

packages/client/src/grpc-retry.ts
Sushisource: I can't read. Ignore me.

Sushisource (approved):

cretz:
:+1: to @Quinn-With-Two-Ns - internal is not meant to be retried IIUC (except on older servers based on capability, but that was a long time ago, can probably just not retry).

Issue #1435, opened 2024-06-04
Provide methods to wait for all handlers to finish


drewhoskins-temporal:
Users want to easily await (or ignore) all their remaining handlers when they continue as new or exit the workflow.  Since we can't tell what they want, they should specify.

dandavison:
Closed by https://github.com/temporalio/sdk-typescript/pull/1459

PR #1459, opened 2024-07-08
Warn on unfinished handlers and provide `workflow.allHandlersFinished()` API

### LINKED ISSUE #1434: Warn or error when update handlers dangle across CAN or workflow exit
We can recommend the await-all-handlers or drop-all-handlers methods. TBD whether error or warning.

dandavison:
Fixes #1434 and #1435.

- new workflow API to wait for all signal and update handlers to be finished:
  `await workflow.condition(() => workflow.allHandlersFinished())`
- By default, warn when workflow exits with unfinished handlers, but not when a handler finishes due to workflow cancellation or workflow failure.
- Warnings controlled by `HandlerUnfinishedPolicy`

### DIFF
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -8,11 +8,16 @@ export type WorkflowUpdateType = (...args: any[]) => Promise<any> | any;
 export type WorkflowUpdateValidatorType = (...args: any[]) => void;
 export type WorkflowUpdateAnnotatedType = {
   handler: WorkflowUpdateType;
+  unfinishedPolicy: HandlerUnfinishedPolicy;
   validator?: WorkflowUpdateValidatorType;
   description?: string;
 };
 export type WorkflowSignalType = (...args: any[]) => Promise<void> | void;
-export type WorkflowSignalAnnotatedType = { handler: WorkflowSignalType; description?: string };
+export type WorkflowSignalAnnotatedType = {
+  handler: WorkflowSignalType;
+  unfinishedPolicy: HandlerUnfinishedPolicy;
+  description?: string;
+};
 export type WorkflowQueryType = (...args: any[]) => any;
 export type WorkflowQueryAnnotatedType = { handler: WorkflowQueryType; description?: string };
 
@@ -117,4 +122,23 @@ export type UntypedActivities = Record<string, ActivityFunction>;
 export interface HistoryAndWorkflowId {
   workflowId: string;
   history: temporal.api.history.v1.History | unknown | undefined;
+}
+
+/**
+ * Policy defining actions taken when a workflow exits while update or signal handlers are running.
+ * The workflow exit may be due to successful return, failure, cancellation, or continue-as-new.
+ */
+export enum HandlerUnfinishedPolicy {
+  /**
+   * Issue a warning in addition to abandoning the handler execution.
+   */
+  WARN_AND_ABANDON = 1,
+
+  /**
+   * Abandon the handler execution.
+   *
+   * In the case of an update handler this means that the client will receive an error rather than
+   * the update result.
+   */
+  ABANDON = 2,
 }
--- a/packages/common/src/interfaces.ts
+++ b/packages/common/src/interfaces.ts
@@ -130,7 +130,7 @@ export interface HistoryAndWorkflowId {
  */
 export enum HandlerUnfinishedPolicy {
   /**
-   * Issue a warning in addition to abandoning the handler execution.
+   * Issue a warning in addition to abandoning the handler execution. The warning will not be issued if the workflow fails.
    */
   WARN_AND_ABANDON = 1,
 
--- a/packages/test/src/test-integration-workflows-with-recorded-logs.ts
+++ b/packages/test/src/test-integration-workflows-with-recorded-logs.ts
@@ -284,23 +284,10 @@ class UnfinishedHandlersWorkflowTerminationTypeTest {
 
     return await worker.runUntil(async () => {
       if (this.handlerType === 'update') {
-        switch (this.workflowTerminationType) {
-          case 'cancellation': {
-            await this.assertWorkflowUpdateFailedError(executeUpdate);
-            break;
-          }
-          case 'continue-as-new': {
-            await this.assertWorkflowNotFoundError(executeUpdate);
-            break;
-          }
-          case 'failure': {
-            await this.assertWorkflowNotFoundError(executeUpdate);
-            break;
-          }
-          case 'return': {
-            await this.assertWorkflowNotFoundError(executeUpdate);
-            break;
-          }
+        if (this.workflowTerminationType === 'cancellation') {
+          await this.assertWorkflowUpdateFailedError(executeUpdate);
+        } else {
+          await this.assertWorkflowNotFoundError(executeUpdate);
         }
       }
 
--- a/packages/test/src/test-integration-workflows-with-recorded-logs.ts
+++ b/packages/test/src/test-integration-workflows-with-recorded-logs.ts
@@ -177,7 +177,7 @@ class UnfinishedHandlersTest {
   }
 }
 
-export const unfinishedHandlersWorkflowTerminationTypeUpdate = workflow.defineUpdate<void>(
+export const unfinishedHandlersWorkflowTerminationTypeUpdate = workflow.defineUpdate<string>(
   'unfinishedHandlersWorkflowTerminationTypeUpdate'
 );
 export const unfinishedHandlersWorkflowTerminationTypeSignal = workflow.defineSignal(
@@ -185,38 +185,66 @@ export const unfinishedHandlersWorkflowTerminationTypeSignal = workflow.defineSi
 );
 
 export async function runUnfinishedHandlersWorkflowTerminationTypeWorkflow(
-  workflowTerminationType: 'cancellation' | 'continue-as-new' | 'failure' | 'return'
+  workflowTerminationType: 'cancellation' | 'continue-as-new' | 'failure' | 'return',
+  waitAllHandlersFinished?: 'wait-all-handlers-finished'
 ): Promise<void> {
+  let handlerMayReturn = false;
+
   workflow.setHandler(unfinishedHandlersWorkflowTerminationTypeUpdate, async () => {
-    await workflow.condition(() => false);
-    throw new Error('unreachable');
+    await workflow.condition(() => handlerMayReturn);
+    return 'update-result';
   });
 
   workflow.setHandler(unfinishedHandlersWorkflowTerminationTypeSignal, async () => {
-    await workflow.condition(() => false);
-    throw new Error('unreachable');
+    await workflow.condition(() => handlerMayReturn);
   });
 
   switch (workflowTerminationType) {
     case 'cancellation':
       await workflow.condition(() => false);
     case 'continue-as-new':
+      if (waitAllHandlersFinished) {
+        handlerMayReturn = true;
+        await workflow.condition(workflow.allHandlersFinished);
+      }
       await workflow.continueAsNew('return');
     case 'failure':
       throw new workflow.ApplicationFailure('Deliberately failing workflow with an unfinished handler');
     case 'return':
-      break;
+      if (waitAllHandlersFinished) {
+        handlerMayReturn = true;
+        await workflow.condition(workflow.allHandlersFinished);
+      }
   }
 }
 
 // These tests confirm that the warning is issued / not issued as appropriate for workflow
-// termination via cancellation, continue-as-new, failure, and return.
+// termination via cancellation, continue-as-new, failure, and return, and that there is no warning
+// when waiting on allHandlersFinished before return or continue-as-new.
 test('unfinished update handler with workflow return', async (t) => {
   await new UnfinishedHandlersWorkflowTerminationTypeTest(t, 'update', 'return').testWarningIsIssued(true);
 });
 
+test('unfinished update handler with workflow return waiting for all handlers to finish', async (t) => {
+  await new UnfinishedHandlersWorkflowTerminationTypeTest(
+    t,
+    'update',
+    'return',
+    'wait-all-handlers-finished'
+  ).testWarningIsIssued(false);
+});
+
 test('unfinished signal handler with workflow return', async (t) => {
   await new UnfinishedHandlersWorkflowTerminationTypeTest(t, 'signal', 'return').testWarningIsIssued(true);
+});
+
+test('unfinished signal handler with workflow return waiting for all handlers to finish', async (t) => {
+  await new UnfinishedHandlersWorkflowTerminationTypeTest(
+    t,
+    'signal',
+    'return',
+    'wait-all-handlers-finished'
+  ).testWarningIsIssued(false);
 });
 
 test('unfinished update handler with workflow cancellation', async (t) => {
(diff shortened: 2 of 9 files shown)


packages/test/src/test-integration-workflows-with-recorded-logs.ts
```diff
@@ -0,0 +1,298 @@
...
+
+test('unfinished signal handler with workflow failure', async (t) => {
+  await new UnfinishedHandlersWithCancellationOrFailureTest(t, 'signal', 'failure').testWarningIsIssued();
+});
+
+class UnfinishedHandlersWithCancellationOrFailureTest {
+  constructor(
+    private readonly t: ExecutionContext<Context>,
+    private readonly handlerType: 'update' | 'signal',
+    private readonly workflowTerminationType: 'cancellation' | 'failure'
```
dandavison, line 223: TODO (blocking PR): this should test continue-as-new

packages/workflow/src/internals.ts
```diff
@@ -651,10 +676,21 @@ export class Activator implements ActivationHandler {
...
+        return Promise.reject(new IllegalStateError(`No registered update handler for update: ${name}`));
+      }
+      const { unfinishedPolicy } = entry;
+
       const execute = composeInterceptors(this.interceptors.inbound, 'handleUpdate', this.updateNextHandler.bind(this));
       this.acceptUpdate(protocolInstanceId);
+      this.inProgressUpdates.set(updateId, { name, unfinishedPolicy, id: updateId });
       const res = execute(input)
         .then((result) => this.completeUpdate(protocolInstanceId, result))
+        .then(() => {
```
bergundy, line 691: Should this be in a `finally()` block?

bergundy (commented):
I'm concerned with `allHandlersFinished` returning `true` between handling of a signal and an update.

I noticed that updates aren't reordered an put in the first activation batch:

https://github.com/temporalio/sdk-typescript/blob/9d31dbd6f1bb6f2afa4bc2193365140aca149314/packages/worker/src/workflow/vm-shared.ts#L315

This means that this workflow won't get a chance to see an update:

```ts

const fooSignal = workflow.defineSignal('foo');
const fooUpdate = workflow.defineUpdate('foo');

export async function http(): Promise<string> {
  var numFoos = 0;

  workflow.setHandler(fooSignal, () => {
    numFoos++
  });

  workflow.setHandler(fooUpdate, () => {
    numFoos++
  });

  await workflow.condition(workflow.allHandlersFinished);
  return `foos: ${numFoos}`;
}
```

Starter:

```ts
  const handle = await client.workflow.start(workflowType, {
    workflowId,
    taskQueue: 'test',
    args,
  });
  await handle.signal('foo');
  await handle.startUpdate('foo', {
    waitForStage: WorkflowUpdateStage.ACCEPTED,
  });
```

Fixing this would require to break history compatibility or find a safe way to patch outside of the sandbox, which we haven't done before.

packages/workflow/src/internals.ts
dandavison: I actually placed it in a `finally()` block initially. However, if the handler is interrupted by cancellation, then we want that to count as an unfinished handler. But if it were in a `finally()` block, a cancellation exception would cause the handler to  be considered completed. There are tests for all these cases. I agree we need to be careful with this logic; do you see a bug with it?

bergundy: > But if it were in a finally() block, a cancellation exception would cause the handler to be considered completed.

But it is completed, why not?

dandavison:
> I'm concerned with allHandlersFinished returning true between handling of a signal and an update.

@bergundy and I discussed this. There is indeed a bug, though not introduced in this PR and therefore not blocking this PR. We've opened https://github.com/temporalio/sdk-typescript/issues/1474

packages/workflow/src/internals.ts
dandavison: Discussed in other channels:

An update handler execution interrupted by a workflow cancellation exception should be considered "finished", since the effect of a cancellation exception is to cancel the update and return an appropriate error to the update caller.

It is not obvious whether a signal handler execution interrupted by a workflow cancellation exception should be considered "finished", but, as a product decision, we don't want to issue unfinished handler warnings on workflow cancellation since the unfinishedness in that case is not something that can be addressed by improving the workflow implementation (it's simply because something cancelled the workflow).

So, the tests need updating here and, yes, the in-progress state transition should be in a `finally()` callback, as it was initially. cc @cretz re .NET

dandavison: Updated implementation and tests: we now issue the warnings neither on cancellation nor on workflow failure.

packages/workflow/src/internals.ts
```diff
@@ -651,8 +676,16 @@ export class Activator implements ActivationHandler {
...
+        return Promise.reject(new IllegalStateError(`No registered update handler for update: ${name}`));
+      }
+      const { unfinishedPolicy } = entry;
+
       const execute = composeInterceptors(this.interceptors.inbound, 'handleUpdate', this.updateNextHandler.bind(this));
       this.acceptUpdate(protocolInstanceId);
+      this.inProgressUpdates.set(updateId, { name, unfinishedPolicy, id: updateId });
       const res = execute(input)
         .then((result) => this.completeUpdate(protocolInstanceId, result))
         .catch((error) => {
```
dandavison, line 691: @mjameswh @bergundy is there a reason we catch `TemporalFailure` rather than `ApplicationFailure` here? (The test suite passes with either.)

Sushisource:
Should also close https://github.com/temporalio/sdk-typescript/issues/1435

packages/workflow/src/internals.ts
bergundy: Hmm... we should reject the update in case of any error AFAICT, not just Temporal or Application failures.

dandavison: To answer my own question:

> is there a reason we catch `TemporalFailure` rather than `ApplicationFailure` here?

Yes, `Cancelledfailure` extends `TemporalFailure`, and we want workflow cancellation to result in rejection of the update, in the same way as workflow failure.

 > Hmm... we should reject the update in case of any error AFAICT, not just Temporal or Application failures.
 
There would be no point in rejecting the update in this case: the WFT is going to fail, so any messages/commands emit by the workflow will be dropped.

packages/test/src/test-integration-workflows-with-recorded-logs.ts
```diff
@@ -0,0 +1,388 @@
...
+});
+
+test('unfinished update handler with continue-as-new waiting for all handlers to finish', async (t) => {
+  await new UnfinishedHandlersWorkflowTerminationTypeTest(
+    t,
+    'update',
+    'continue-as-new',
+    'wait-all-handlers-finished'
+  ).testWarningIsIssued(false);
+});
```
dandavison, line 269: This test is flaky. Fixing that is the last thing needed in this PR.

dandavison: I have papered over that test failure in this PR; I believe it is a server bug. Here's a full description (no issue yet):

--

I have a workflow that CANs in the same WFT as an update is completed.
I ensure that the update is sent in the first WFT and, most of the time, I see what I expect:

1. worker receives WFT: `[doUpdate, startWorkflow]`
2. worker sends WFT completion: `[accept&completeUpdate, CAN]`
3. update caller gets successful update response
4. worker receives WFT after CAN: `[startWorkflow]`
5. workflow completes

However, sometimes (~1/20) I see

1. \<as above\>
2. \<as above>
3. server error when handling UpdateWorkflowExecution: [`unable to locate current workflow execution`](https://github.com/temporalio/temporal/blob/99f15c4dabf9a22763aaa13653c83a832bd0fffa/service/history/statemachine_environment.go#L167) (update caller does not get a response)
4. WFT after CAN contains the update for a second time, in addition to startWorkflow

I assume what's happening is that the server error causes an internal retry (FE => History) of the UpdateWorkflowExecution, and this retry lands on the 2nd run, after CAN. But I haven't yet dug into why the server error happens
```
	// for close workflow we need to check if it is still the current run
	// since it's possible that the workflowID has a newer run before it's locked 

wcache.GetCurrentRunID(...) != wfContext.GetWorkflowKey().RunID
```

packages/test/src/test-integration-workflows-with-recorded-logs.ts
dandavison: I've added coverage for CAN now. The tests confirm that the warnings, and waiting on `allHandlersFinished` behave as expected in the various cases of handlers being interrupted by workflow return, CAN, failure, and cancellation.

packages/test/src/test-integration-workflows-with-recorded-logs.ts
```diff
@@ -0,0 +1,397 @@
...
+
+test('unfinished update handler with workflow cancellation', async (t) => {
+  await new UnfinishedHandlersWorkflowTerminationTypeTest(t, 'update', 'cancellation').testWarningIsIssued(false);
+});
+
+test('unfinished signal handler with workflow cancellation', async (t) => {
+  await new UnfinishedHandlersWorkflowTerminationTypeTest(t, 'signal', 'cancellation').testWarningIsIssued(false);
```
dandavison, line 266: TODO (blocking PR): we are going back to warning on cancellation
```suggestion
  await new UnfinishedHandlersWorkflowTerminationTypeTest(t, 'signal', 'cancellation').testWarningIsIssued(true);
```

packages/workflow/src/internals.ts
```diff
@@ -651,8 +676,16 @@ export class Activator implements ActivationHandler {
         this.rejectUpdate(protocolInstanceId, error);
         return;
       }
+
+      const entry = this.updateHandlers.get(name);
```
mjameswh, line 680: We shouldn't need this if statement; we just checked that `updateHandlers` indeed contains key `name`, and null values should not be allowed in that map.

In fact, for coherency, we should extract `entry` above, and use `validator`/`handler` from that `entry` rather than looking into `updateHandlers` every single time; i.e. once we start handling a specific incoming update request, everything relating to that update request should be handled using the same "update handler entry".

packages/workflow/src/workflow.ts
```diff
@@ -1434,6 +1439,11 @@ export function upsertMemo(memo: Record<string, unknown>): void {
   });
 }
 
+export function allHandlersFinished(): boolean {
```
mjameswh, line 1442: Docstring?

packages/workflow/src/internals.ts
dandavison: We changed our minds. We are now issuing the warning on cancellation (because the workflow author can avoid leaving unfinished handlers) but not on failure (although there too, the workflow author could avoid it, so it's not clear what the correct choice is).

I've update this PR accordingly, adding tests showing that the warnings _are not_ issued when workflow cancellation propagates to a handler (since then the handler completes before workflow completion) but _are_ issued when the handler shields itself from cancellation and thus does not complete before the main workflow routine terminates due to the cancellation.

packages/workflow/src/workflow.ts
dandavison: Oops, thanks! Added docstring.

packages/common/src/interfaces.ts
```diff
@@ -118,3 +123,22 @@ export interface HistoryAndWorkflowId {
   workflowId: string;
   history: temporal.api.history.v1.History | unknown | undefined;
 }
+
+/**
+ * Policy defining actions taken when a workflow exits while update or signal handlers are running.
+ * The workflow exit may be due to successful return, failure, cancellation, or continue-as-new.
```
mjameswh, line 129: You mention the different ways that a workflow may exit. We should make it clear (probably below, on `WARN_AND_ABANDON`) that failure dont result in warn.

dandavison: Good call, done.

packages/workflow/src/internals.ts
```diff
@@ -638,7 +664,7 @@ export class Activator implements ActivationHandler {
     const doUpdateImpl = async () => {
       let input: UpdateInput;
       try {
-        if (runValidator && this.updateHandlers.get(name)?.validator) {
+        if (runValidator && entry.validator) {
```
mjameswh, line 667: Just below:

```
this.validateUpdateNextHandler.bind(this, entry.validator)
```

And add the corresponding argument on `validateUpdateNextHandler`. Then remove the access on `this.updateHandlers` from `validateUpdateNextHandler`.

packages/workflow/src/internals.ts
```diff
@@ -651,8 +677,10 @@ export class Activator implements ActivationHandler {
         this.rejectUpdate(protocolInstanceId, error);
         return;
       }
-      const execute = composeInterceptors(this.interceptors.inbound, 'handleUpdate', this.updateNextHandler.bind(this));
       this.acceptUpdate(protocolInstanceId);
+      const execute = composeInterceptors(this.interceptors.inbound, 'handleUpdate', this.updateNextHandler.bind(this));
```
mjameswh, line 681: Same. Bind `entry.handler` as argument on `updateNextHandler`.

packages/workflow/src/internals.ts
```diff
@@ -651,8 +677,10 @@ export class Activator implements ActivationHandler {
...
+      const { unfinishedPolicy } = entry;
+      this.inProgressUpdates.set(updateId, { name, unfinishedPolicy, id: updateId });
       const res = execute(input)
         .then((result) => this.completeUpdate(protocolInstanceId, result))
```
mjameswh, line 685: Combine the `.then` and `.catch` clauses in a single call to `.then`. That will create a single promise instead of two. Then make sure to `untrackPromise()` that promise, before calling `.finally()` on it (which also needs to be `untrackPromise()`, but that one is already there).

packages/workflow/src/internals.ts
```diff
@@ -741,7 +774,9 @@ export class Activator implements ActivationHandler {
       args: arrayFromPayloads(this.payloadConverter, activation.input),
       signalName,
       headers: headers ?? {},
-    }).catch(this.handleWorkflowFailure.bind(this));
+    })
+      .catch(this.handleWorkflowFailure.bind(this))
```
mjameswh, line 778: We need to `untrackPromise()` both the promise resulting from the call on `.catch()`, and the one from `.finally()`.

packages/workflow/src/internals.ts
dandavison: Let's address this in https://github.com/temporalio/sdk-typescript/issues/1486

packages/test/src/test-integration-workflows-with-recorded-logs.ts
dandavison: Opened https://github.com/temporalio/temporal/issues/6375

packages/workflow/src/internals.ts
dandavison: Thanks! Tightened all that up in recent commits.

mjameswh (approved):
Still a few nits, but I'm ok to fix those in a follow-up PR.

packages/workflow/src/internals.ts
dandavison: Beautiful. I don't have a ton of JS background and I always forget that bind can be used for general partial application like this. I really like this improvement. 77240fe49fddf2fc8c5b91cd9a2c2daf9d34550f

dandavison:
Thanks very much for reviewing @mjameswh! 

> Still a few nits, but I'm ok to fix those in a follow-up PR.

The only thing I've pushed to a follow-up is the task of generating example stack traces to review our use of `untrackPromise`: https://github.com/temporalio/sdk-typescript/issues/1486, which was needed even before this PR.

PR #1463, opened 2024-07-11
fix: Fix flakes in Worker Tuner and Workflow Update tests

mjameswh:
## What was changed and Why?

- Remove CI's pinning on CLI 0.12.0.
  - This was causing multiple CI flakes due to bugs that had already been fixed in server.
- Capture Dev Server logs.
  - Hopefully, this may help understand some other flakes in the future.
- In SDK's internal tests, if using the `TEST_LOG_LEVEL` env variable, we now also set Core's logging to that same level.
- A few other very minor refactors and improvements here and there.
- In one of the Worker Tuner tests, disable the workflow cache, as the Resource Based Tuner seems not to prioritize allocation of permits to at least one non-sticky poller, which was causing flakes.

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -806,13 +806,15 @@ export class WorkflowClient extends BaseClient {
     waitForStage: WorkflowUpdateStage,
     input: WorkflowStartUpdateInput
   ): Promise<WorkflowStartUpdateOutput> {
+    waitForStage = waitForStage >= WorkflowUpdateStage.ACCEPTED ? waitForStage : WorkflowUpdateStage.ACCEPTED;
+    const waitForStageProto = workflowUpdateStage.toProtoEnum(waitForStage);
     const updateId = input.options?.updateId ?? uuid4();
     const req: temporal.api.workflowservice.v1.IUpdateWorkflowExecutionRequest = {
       namespace: this.options.namespace,
       workflowExecution: input.workflowExecution,
       firstExecutionRunId: input.firstExecutionRunId,
       waitPolicy: {
-        lifecycleStage: workflowUpdateStage.toProtoEnum(waitForStage),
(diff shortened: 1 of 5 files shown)


packages/test/src/test-worker-tuner.ts
```diff
@@ -1,154 +1,141 @@
...
+        minimumSlots: 2,
+        maximumSlots: 10,
+        rampThrottle: 20,
+      },
+    },
+    // FIXME(JWH): Resource Based Tuner does not guarantee that at least one permit would be
+    // allocated to a non-sticky pollers, which means the worker may not be able to poll for a
+    // WFT from the normal TQ before this test times out. For now, let's completely disable usage
+    // of sticky TQ. Rmeove this once https://github.com/temporalio/sdk-core/issues/775 gets fixed.
+    maxCachedWorkflows: 0,
```
Sushisource, line 29: :+1: Thanks again for investigating that

Sushisource (approved):

Sushisource:
> In SDK's internal tests, if using the TEST_LOG_LEVEL env variable, we now also set Core's logging to that same level.

That is very nice, been meaning to do something like that for a while

packages/client/src/workflow-client.ts
```diff
@@ -806,13 +806,15 @@ export class WorkflowClient extends BaseClient {
     waitForStage: WorkflowUpdateStage,
     input: WorkflowStartUpdateInput
   ): Promise<WorkflowStartUpdateOutput> {
+    waitForStage = waitForStage >= WorkflowUpdateStage.ACCEPTED ? waitForStage : WorkflowUpdateStage.ACCEPTED;
```
dandavison, line 809: Is this a bug fix -- I was allowing clients to submit `Admitted`? Thanks if so! (I should add test coverage for it.)

PR #1473, opened 2024-07-23
fix(worker): Don't fail Worker on Activity Task decode failures

### LINKED ISSUE #1404: [Bug] Some TS Worker errors while processing a Task may cause complete Worker failure
### Describe the bug

In a recent incident, an error in processing an Activity Task (due to `ActivityInfo.startTime` being `null`, which would result in throwing a `TypeError`) led to complete Worker failure.

This kind of unexpected errors must be caught and handled on TS side, as Core can't properly handle them.

mjameswh:
## What was changed and why?

- Catch all unexpected exceptions while decoding Activity Tasks in TS Worker. That notably includes things such as: protobuf fields that are unexpectedly null, errors while decoding heartbeat details, errors while instantiating activity interceptors, etc.

  - These exceptions would previously result in failure of the Worker.

- Also, include field names in error messages reported by `tsToMs()` and `tsToDate()` when the `ts` value is null or undefined.

  - This should make it easier to triage and diagnose a specific category of issues we have been observing recently due to a change in the server-side protobuf library.

Fixes #1404.


packages/worker/src/worker.ts
```diff
@@ -854,85 +856,79 @@ export class Worker {
...
+                      this.options.interceptors.activity
+                    );
+                    output = { type: 'run', activity, input };
                     break;
-                  }
-                  let args: unknown[];
-                  try {
-                    args = await decodeArrayFromPayloads(this.options.loadedDataConverter, task.start?.input);
-                  } catch (err) {
+                  } catch (e) {
```
antlai-temporal, line 920: Can we just move the break after the catch, and just have one `break`

packages/worker/src/worker.ts
```diff
@@ -854,85 +856,79 @@ export class Worker {
...
+                    const error = ensureApplicationFailure(e);
                     output = {
```
antlai-temporal, line 922: What happens if it is not an application failure? Can you add a comment?

packages/worker/src/worker.ts
```diff
@@ -1746,6 +1742,14 @@ async function extractActivityInfo(
   const { taskToken } = task as NonNullableObject<coresdk.activity_task.IActivityTask>;
   const start = task.start as NonNullableObject<coresdk.activity_task.IStart>;
   const activityId = start.activityId;
+  let heartbeatDetails = undefined;
+  try {
+    heartbeatDetails = await decodeFromPayloadsAtIndex(dataConverter, 0, start.heartbeatDetails);
+  } catch (e) {
+    throw ApplicationFailure.fromError(e, {
+      message: `Failed to parse heartbeat details for activity ${activityId}: ${errorMessage(e)}`,
+    });
```
antlai-temporal, line 1751: Is this non-retriable?

antlai-temporal (commented):
High level comments:
-Is this better for a "DEBUG" mode, or we want workers never to fail in production? In general, you want to fail fast when an unrecoverable error happens in prod, and not continue unless you know for sure the state has not been messed up... We are catching all exceptions, can we be sure of that?
-Are activities going to keep retrying, or just fail the workflow straight away (better the later).
- Can other things apart from Date be unexpectedly null?

packages/worker/src/worker.ts
```diff
@@ -854,85 +856,79 @@ export class Worker {
...
+                      });
+                    }
+                    let args: unknown[];
+                    try {
+                      args = await decodeArrayFromPayloads(this.options.loadedDataConverter, task.start?.input);
+                    } catch (err) {
+                      throw ApplicationFailure.fromError(err, {
+                        message: `Failed to parse activity args for activity ${activityType}: ${errorMessage(err)}`,
+                        nonRetryable: false,
+                      });
```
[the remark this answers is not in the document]
mjameswh: Payload conversion error could be due to a transient issue (e.g. an error accessing KMS, or old worker getting a task that would work in newer worker).

I agree that retrying might sometime be problematic, but it was already that way previously, and redefining that behavior is outside the scope of this PR.

packages/worker/src/worker.ts
bergundy: the default is retryable.

packages/worker/src/worker.ts
```diff
@@ -1746,6 +1742,14 @@ async function extractActivityInfo(
...
   const activityId = start.activityId;
+  let heartbeatDetails = undefined;
+  try {
+    heartbeatDetails = await decodeFromPayloadsAtIndex(dataConverter, 0, start.heartbeatDetails);
+  } catch (e) {
+    throw ApplicationFailure.fromError(e, {
+      message: `Failed to parse heartbeat details for activity ${activityId}: ${errorMessage(e)}`,
+    });
+  }
   return {
```
bergundy, line 1753: Can you make sure that if this code throws we log and error and dump the task we got so we can inspect this better?

packages/worker/src/worker.ts
```diff
@@ -854,85 +856,79 @@ export class Worker {
...
+                    );
+                    output = { type: 'run', activity, input };
                     break;
-                  }
-                  let args: unknown[];
-                  try {
-                    args = await decodeArrayFromPayloads(this.options.loadedDataConverter, task.start?.input);
-                  } catch (err) {
+                  } catch (e) {
+                    const error = ensureApplicationFailure(e);
```
bergundy, line 921: I'd want to ensure we have error logs in addition to failing the tasks. Didn't verify that this exists.

packages/worker/src/worker.ts
mjameswh: That's arguable, but intentional from my part. I personally feel that duplicating the break statement here makes it more obvious that we reached the conclusion of these logic branches. The `output = { ... }` statements don't stand out enough by themselves.

packages/worker/src/worker.ts
mjameswh: `ensureApplicationFailure` is an already existing internal helper function. 

```
/**
 * If `error` is already an `ApplicationFailure`, returns `error`.
 *
 * Otherwise, converts `error` into an `ApplicationFailure` with:
 *
 * - `message`: `error.message` or `String(error)`
 * - `type`: `error.constructor.name` or `error.name`
 * - `stack`: `error.stack` or `''`
 */
export function ensureApplicationFailure(error: unknown): ApplicationFailure {
```

I don't there's more to say than this. WDYT?

packages/worker/src/worker.ts
antlai-temporal: Sorry, I meant whether we should make it non-retryable and fail the workflow

mjameswh: Payload conversion error could be due to transient issues.

packages/worker/src/worker.ts
mjameswh: I added log, including `proto.toJSON()` and the unaltered protbuf-encoded buffer (in base64 notation).

```
2024-07-24T04:45:57.463Z [ERROR] Error while processing ActivityTask.start: Expected scheduledTime to be a timestamp, got null {
  sdkComponent: 'worker',
  taskQueue: 'test',
  error: TypeError: Expected scheduledTime to be a timestamp, got null
      at requiredTsToMs (/Users/jwatkins/Development/Temporal/TypeScriptSDK/sdk-typescript-2/packages/common/src/time.ts:37:11)
      at Worker.extractActivityInfo (/Users/jwatkins/Development/Temporal/TypeScriptSDK/sdk-typescript-2/packages/worker/src/worker.ts:1064:43)
      at runNextTicks (node:internal/process/task_queues:60:5)
      at listOnTimeout (node:internal/timers:540:9)
      at process.processTimers (node:internal/timers:514:7)
      at <anonymous> (/Users/jwatkins/Development/Temporal/TypeScriptSDK/sdk-typescript-2/packages/worker/src/worker.ts:870:28),
  task: '{"taskToken":"ODQxYjEyM2EtYjc5MS00ZmE3LThmODctMTI3YjgxOWZlZWUw","start":{"workflowExecution":{"workflowId":"wfid","runId":"runId"},"activityType":"dummy","currentAttemptScheduledTime":{"seconds":"0","nanos":0},"startedTime":{"seconds":"0","nanos":0},"attempt":1,"scheduleToCloseTimeout":{"seconds":"60","nanos":0},"startToCloseTimeout":{"seconds":"60","nanos":0},"heartbeatTimeout":{"seconds":"60","nanos":0}}}',
  taskEncoded: 'CiQ4NDFiMTIzYS1iNzkxLTRmYTctOGY4Ny0xMjdiODE5ZmVlZTAaNhoNCgR3ZmlkEgVydW5JZCoFZHVtbXlSBAgAEABaBAgAEABgAWoECDwQAHIECDwQAHoECDwQAA=='
}
```

packages/worker/src/worker.ts
mjameswh: See my other comment.

mjameswh:
> Can other things apart from Date be unexpectedly null?

Of course. But for various reasons, timestamps appear more likely to pose issues than other data types in our specific use case.

packages/worker/src/worker.ts
antlai-temporal: I thought that was an assertion, never mind...

bergundy (approved):

PR #1482, opened 2024-08-02
Add test of workflow cancellation while update handler is executing

dandavison:
Adds some missing test coverage.


dandavison:
This PR contains a test for the behavior that update should have.
The test does not currently pass in CI, because the desired behavior only occurs with newer server versions.

mjameswh:
> The test does not currently pass in CI, because the desired behavior only occurs with newer server versions.

Not sure why we are dependent on server's behavior in this specific test. And just to be clear, I agree with the fact that server should fail pending updates on workflow cancellation.

But in this specific test case, the `condition()` call made from the update handler belongs to the root cancellation scope, so it should throw a cancellation failure. Shouldn't that result in rejecting the update, without the server's intervention?

dandavison:
> But in this specific test case, the condition() call made from the update handler belongs to the root cancellation scope, so it should throw a cancellation failure. Shouldn't that result in rejecting the update, without the server's intervention?

Yes, it does throw `CancellationFailure`. And, since that error class extends `TemporalFailure`, this causes the worker to include an `update-rejected` message in the WFT completion. However, the workflow is completing in the same WFT. Empirically, what I'm finding is that with server v1.24 the update caller gets gRPC NOT_FOUND, whereas with latest server `main`, the update caller gets `WorkflowUpdateFailedError` (with cause `CancelledFailure`). I will try to find out why that is.

mjameswh:
Just to be sure, I'd change your that above to two distinct test cases, one that test what happens when update handler throws a cancellation failure and the workflow doesn't, and another one that check what if the update handler doesn't throw cancellation, but the workflow does. It should be possible to just do `await new Promise(() => {})` instead of `await condition()`.

dandavison:
OK thanks that sounds like a good idea, I'll look into that.

mjameswh (approved):

PR #1487, opened 2024-08-07
Upgrade core to fix cancel-before-started abanadon activity

Sushisource:
Upgrade core & add test showing this bug is fixed


mjameswh (approved):
LGTM

PR #1488, opened 2024-08-07
fix(workflow): process all activation jobs as a single batch

### LINKED ISSUE #1474: [Bug] Workflows can be constructed in which update handlers do not, but should, execute
If the workflow below is started, and then a signal and an update are sent in the 2nd Workflow task, the workflow exits without the update handler ever being executed. See @bergundy's original comment https://github.com/temporalio/sdk-typescript/pull/1459#pullrequestreview-2184277781

```typescript
export async function workflow(): Promise<string> {
  var numFoos = 0;

  wf.setHandler(fooSignal, () => {
    numFoos++;
  });

  wf.setHandler(fooUpdate, () => {
    numFoos++;
  });

  await wf.condition(() => numFoos > 0);
  return `foos: ${numFoos}`;
}
```

The fix will be to group update activation jobs together with signals here:

 https://github.com/temporalio/sdk-typescript/blob/ce0532f9be2d47db48c00ebbf17a39063ae42753/packages/worker/src/workflow/vm-shared.ts#L314-L317

Fixing this will be backwards incompatible w.r.t. replay of existing workflows, so will need to use an SDK flag.

We did not update that code when we implemented update because core orders the activation jobs and we were planning to eliminate this typescript code, but this decision overlooked the fact that update and signal activation jobs need to be placed into the same batch in order for them all to be executed before giving the chance for an awaitable such as the `workflow.condition()` above to be unblocked.

mjameswh:
## What was changed

- Previously, when processing a Workflow activation, the SDK would execute `notifyHasPatch` and `signalWorkflow` jobs in distinct phases, before other types of jobs. The primary reason behind that multi-phase algorithm was to avoid the possibility that a Workflow execution might complete before all incoming signals have been dispatched (at least to the point that the _synchronous_ part of the handler function has been executed).

  This PR replaces that multi-phase algorithm with a simpler one where jobs are simply sorted as `signals and updates -> others`, but without processing them as distinct batches (i.e. without leaving/reentering the VM context between each group, which automatically triggers the execution of all outstanding microtasks). That single-phase approach resolves some rare, edge-case quirks of the former algorithm (including the one described in #1474), and yet still satisfies to the original requirement of ensuring that every `signalWorkflow` jobs - and now `doUpdate` jobs as well - have been given a proper chance to execute before the Workflow main function might complete.

  This change is gated by a new SDK flag, to preserve compatibility with workflow execution history produced by previous releases. However, there is no intention at this point of making that that flag rollback-safe.

- Lang no longer stops processing jobs for the current activation after the workflow code emits a workflow completion command; instead, lang continues to process remaining jobs until the next stop point, after which it reports all generated commands to Core.


packages/test/src/test-workflows.ts
```diff
@@ -130,7 +130,24 @@ async function createWorkflow(
...
 async function activate(t: ExecutionContext<Context>, activation: coresdk.workflow_activation.IWorkflowActivation) {
   const { workflow, runId } = t.context;
-  const completion = await workflow.activate(activation);
+
+  // Core guarantees the following jobs ordering: patches -> signals+update -> others
+  // Tests are likely to fail if we artifically pass an activation that does not follow that order
+  const jobs: coresdk.workflow_activation.IWorkflowActivationJob[] = activation.jobs ?? [];
+  function getPriority(job: coresdk.workflow_activation.IWorkflowActivationJob) {
+    if (job.notifyHasPatch) return 0;
+    if (job.signalWorkflow || job.queryWorkflow) return 1;
```
dandavison, line 139: It looks like you meant `doUpdate` instead of `queryWorkflow` here?

mjameswh: Indeed, thanks

packages/test/src/workflows/signal-update-ordering.ts
```diff
@@ -0,0 +1,18 @@
+import * as wf from '@temporalio/workflow';
+
+export const fooSignal = wf.defineSignal('foo');
+export const fooUpdate = wf.defineUpdate<number>('foo');
+
+// Repro for https://github.com/temporalio/sdk-typescript/issues/1474
+export async function signalUpdateOrderingWorkflow(): Promise<number> {
```
Sushisource, line 7: Why not just re-use this one from test-integration-update?

packages/test/src/workflows/signal-update-ordering.ts
```diff
@@ -0,0 +1,18 @@
...
+  let numFoos = 0;
+  wf.setHandler(fooSignal, () => {
+    numFoos++;
+  });
+  wf.setHandler(fooUpdate, () => {
+    numFoos++;
+    return numFoos;
+  });
```
Sushisource, line 15: Also maybe slightly overkill, but, just to make sure somehow a handler didn't get double-run it might make sense to store `numFoos` as `totalHandlersRun` and then separately `signalsRun` and `updatesRun` and verify all three things

packages/test/src/test-workflows.ts
```diff
@@ -1853,7 +1894,7 @@ test('conditionRacer', async (t) => {
         makeFireTimerJob(1)
       )
     );
-    compareCompletion(t, completion, makeSuccess([{ cancelTimer: { seq: 1 } }]));
+    compareCompletion(t, completion, makeSuccess([], [SdkFlags.ProcessWorkflowActivationJobsAsSingleBatch]));
```
Sushisource, line 1897: Why did this cancel timer go away?

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -270,78 +271,122 @@ export const globalHandlers = new GlobalHandlers();
...
+    for (const jobs of batches) {
+      if (jobs.length === 0) continue;
       this.workflowModule.activate(
         coresdk.workflow_activation.WorkflowActivation.fromObject({ ...activation, jobs }),
         batchIndex++
       );
-      if (internals.shouldUnblockConditions(jobs[0])) {
-        this.tryUnblockConditions();
-      }
+      this.tryUnblockConditionsAndMicrotasks();
```
Sushisource, line 356: Even though it would take up more lines, I'd kinda prefer for this to read as

```
if doSingleBatch {
    // new code with no loop
} else {
    oldMultiBatchProcessing();
}
```

Which would allow the new path to be free of some of the clutter of the old path. Up to you though.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -270,78 +271,122 @@ export const globalHandlers = new GlobalHandlers();
...
     const completion = this.workflowModule.concludeActivation();
+
     // Give unhandledRejection handler a chance to be triggered.
     await new Promise(setImmediate);
     if (this.unhandledRejection) {
       return {
         runId: this.activator.info.runId,
+        // FIXME: Calling `activator.errorToFailure()` directly from outside the VM is unsafe, as it
+        // depends on the `failureConverter` and `payloadConverter`, which may be customized and
+        // therefore ain't guaranteed not to access `global` or to cause scheduling microtasks.
```
Sushisource, line 368:
```suggestion
        // therefore aren't guaranteed not to access `global` or to cause scheduling microtasks.
```

Didn't know you were from the south James :)

Sushisource (approved):
I think this makes good sense. Nothing blocking, but a few suggestions.

packages/test/src/test-workflows.ts
mjameswh: That CancelTimer command was bogus. We just sent a `FireTimer(1)`. Why would we cancel an already fired timer?

That timer comes from the `condition(fn, timeout)` line. In this case, `fn` will return true as soon as the signal handler is invoked. Since signals were procesed in a distinct phase, and timer events in a second phase, the microtasks created internally by `condition()` would have the opportunity to settle on a state where "`fn` returns true, because the signal has been received, but timer has not yet fired, because the job has not yet been processed".

`ProcessWorkflowActivationJobsAsSingleBatch` fixes that. As soon as the `fn` returns true, the internal condition promise (the one without a timer) gets resolved, but by the time the `then` microtask gets executed, the `TimerFired` event has also been processed, and so the attempt to cancel the timer is known to be a noop.

Just for extra safety, I modified that test to return the boolean from `condition(fn, timeout)`, so we know that we didn't break the documented semantic of that API.

packages/test/src/workflows/signal-update-ordering.ts
mjameswh: I added a distinct, more comprehensive unit test that traces important events as perceived by the workflow. I'll this test as is.

packages/worker/src/workflow/vm-shared.ts
mjameswh: Good point.

Issue #1392, opened 2024-04-11
Support for WorkflowIdConflictPolicy


josh-berry:
See https://github.com/temporalio/features/issues/437

stephanos:
NOTE: Please don't forget SignalWithStart API.

hugo082:
Hi there, any updates on this feature? I see the PR is inactive since several month.
AFAIK TS sdk is the only one missing this feature

stephanos:
Good question; the [PR seems to still be blocked](https://github.com/temporalio/sdk-typescript/pull/1490); I'll ask again.

PR #1490, opened 2024-08-13
feat(workflow): Support workflow ID conflict policy

antlai-temporal:
## What was changed
Added `workflowIdConflictPolicy` to workflow start and SignalWithStart

## Checklist

1. Closes 
#1392 
2. How was this tested:

Added two integration tests


packages/common/src/workflow-options.ts
```diff
@@ -39,13 +39,57 @@ export enum WorkflowIdReusePolicy {
...
   /**
    * Terminate the current workflow if one is already running.
+   *
+   * This option belongs in `WorkflowIdConflictPolicy` but it is here for backwards compatibility.
+   * With no Running Workflow, the behavior is the same as `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE`,
+   * but with a Running Workflow, it behaves like `WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING`.
+   *
+   * *Note: When setting this option, `WorkflowIdConflictPolicy` should have value `WORKFLOW_ID_CONFLICT_POLICY_UNSPECIFIED`*
    */
   WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING = 4,
```
dandavison, line 49: I had a go at making this more direct (I used the word "deprecated" since I don't think people should be using this in new code, right?)

```suggestion
  /**
   * Terminate the current workflow if one is already running, otherwise allow a duplicate ID.
   *
   * Deprecated: use `WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING` and `WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE` instead.
   *
   * This option belongs in `WorkflowIdConflictPolicy` but it is here for backwards compatibility.
   * *Note: If setting this option, `WorkflowIdConflictPolicy` must be left unspecified.*
   */
  WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING = 4,
```

packages/common/src/workflow-options.ts
```diff
@@ -39,13 +39,57 @@ export enum WorkflowIdReusePolicy {
...
+// Avoid importing the proto implementation to reduce workflow bundle size
+// Copied from temporal.api.enums.v1.WorkflowIdConflictPolicy
+/**
+ * Defines what should happen when trying to start a Workflow with the same Workflow Id of a Running workflow.
+ *
+ * See `WorkflowIdReusePolicy` for handling Workflow Id duplication of a *Closed* workflow.
+ *
+ * *Note: It is never valid to have two actively running instances with the same Workflow Id.*
+ */
+export enum WorkflowIdConflictPolicy {
```
dandavison, line 64: Just minor suggestions for consistency with the other one:

```suggestion
/**
 * Defines what happens when trying to start a Workflow with the same ID as a *Running* Workflow.
 *
 * See {@link WorkflowOptions.workflowIdReusePolicy} for what happens when trying to start a Workflow with the same ID as a *Closed* Workflow.
 *
 * *Note: It is not possible to have two actively running Workflows with the same ID.*
 */
export enum WorkflowIdConflictPolicy {
```

packages/common/src/workflow-options.ts
```diff
@@ -56,6 +100,15 @@ export interface BaseWorkflowOptions {
...
   workflowIdReusePolicy?: WorkflowIdReusePolicy;
 
+  /**
+   * Defines what should happen when trying to start a Workflow with the same Workflow Id of a Running workflow.
+   *
+   * *Note: It is never valid to have two actively running instances with the same Workflow Id.*
+   *
+   * @default {@link WorkflowIdConflictPolicy.WORKFLOW_ID_CONFLICT_POLICY_UNSPECIFIED}
+   */
+  workflowIdConflictPolicy?: WorkflowIdConflictPolicy;
```
dandavison, line 110:
```suggestion
  /**
   * Defines what happens when trying to start a Workflow with the same ID as a *Running* Workflow.
   *
   * *Note: It is not possible to have two actively running Workflows with the same ID.*
   *
   * @default {@link WorkflowIdConflictPolicy.WORKFLOW_ID_CONFLICT_POLICY_UNSPECIFIED}
   */
  workflowIdConflictPolicy?: WorkflowIdConflictPolicy;
```

packages/common/src/workflow-options.ts
```diff
@@ -56,6 +100,15 @@ export interface BaseWorkflowOptions {
    */
   workflowIdReusePolicy?: WorkflowIdReusePolicy;
```
dandavison, line 101: Copying from above:

```typescript
  /**
   * Defines what happens when trying to start a Workflow with the same ID as a *Closed* Workflow.
   *
   * *Note: It is not possible to have two actively running Workflows with the same ID.*
   *
   * @default {@link WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE}
   */
  workflowIdReusePolicy?: WorkflowIdReusePolicy;
```

dandavison (approved):
LGTM! I made some minor suggestions to the public-facing dosctrings to make them even more direct and consistent with each other.

packages/common/src/workflow-options.ts
```diff
@@ -39,13 +39,57 @@ export enum WorkflowIdReusePolicy {
...
+   *  This is the default so that we can set the policy
+   * `WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING` in `WorkflowIdReusePolicy`, which is incompatible with setting any other
+   *  `WorkflowIdConflictPolicy` values.
+   *
+   * The actual default behavior is `WORKFLOW_ID_CONFLICT_POLICY_FAIL` for a start Workflow request, and `WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING` for a signal with start Workflow request.
+   */
+  WORKFLOW_ID_CONFLICT_POLICY_UNSPECIFIED = 0,
+
+  /**
+   * Do not start a new Workflow. Return instead a `WorkflowExecutionAlreadyStartedError`.
```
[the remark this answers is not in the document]
antlai-temporal: Done

packages/common/src/workflow-options.ts
```diff
@@ -39,13 +39,57 @@ export enum WorkflowIdReusePolicy {
...
+ * *Note: It is never valid to have two actively running instances with the same Workflow Id.*
+ */
+export enum WorkflowIdConflictPolicy {
+  /**
+   *  This is the default so that we can set the policy
+   * `WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING` in `WorkflowIdReusePolicy`, which is incompatible with setting any other
+   *  `WorkflowIdConflictPolicy` values.
+   *
+   * The actual default behavior is `WORKFLOW_ID_CONFLICT_POLICY_FAIL` for a start Workflow request, and `WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING` for a signal with start Workflow request.
+   */
```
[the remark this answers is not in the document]
antlai-temporal: Unfortunately, this is the only way users can know what the actual default is, there is a discussion in the Python PR why we need to make that explicit in comments, and why the default cannot be directly set, I just followed that...

packages/common/src/workflow-options.ts
```diff
@@ -39,13 +39,57 @@ export enum WorkflowIdReusePolicy {
```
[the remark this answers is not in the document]
antlai-temporal: That's much more consistent, thanks!

antlai-temporal:
@dandavison Thanks a lot for a careful review of the docstrings, it made a huge difference!

stephanos:
@antlai-temporal 👋 what's standing in the way of merging? Asking since this is the last PR before we can publish the documentation.

antlai-temporal:
> @antlai-temporal 👋 what's standing in the way of merging? Asking since this is the last PR before we can publish the documentation.
Sorry @stephanos just saw your comment. I was waiting for @mjameswh to give me the go ahead, I don't think this will be controversial...

mjameswh:
Sorry @stephanos and @antlai-temporal for the delay. I was on FTO.

It just so happens that I was reviewing this PR tonight, and I actually do have concerns that I want to resolve before this API reaches a TS release. Please hold on just a bit longer.

packages/common/src/workflow-options.ts
```diff
@@ -38,24 +41,76 @@ export enum WorkflowIdReusePolicy {
...
+// Avoid importing the proto implementation to reduce workflow bundle size
+// Copied from temporal.api.enums.v1.WorkflowIdConflictPolicy
+/**
+ * Defines what happens when trying to start a Workflow with the same ID as a *Running* Workflow.
+ *
+ * See {@link WorkflowOptions.workflowIdReusePolicy} for what happens when trying to start a Workflow with the same ID as a *Closed* Workflow.
+ *
+ * *Note: It is not possible to have two actively running Workflows with the same ID.*
+ */
+export enum WorkflowIdConflictPolicy {
```
mjameswh, line 66: I really want to remove the `WORKFLOW_ID_CONFLICT_POLICY_` prefix on these values.

I know this is inconsistent with the Reuse Policy enum just above, but `WorkflowIdReusePolicy` is the broken one and we'll need to fix it in a backward-compatible way. I'm aware of three other enums that follow that pattern.

But in all other TS APIs introduced in the last two years, we specifically avoided those prefixes; and more recently, [we even agreed](https://github.com/temporalio/sdk-typescript/pull/1156/files#r1256407894) that we'd try to transition to string unions as much as possible, though we didn't make progress on achieving that.

packages/common/src/workflow-options.ts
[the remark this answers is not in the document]
mjameswh: I understand and agree with the rationale, but I'm also uncomfortable with the current phrasing. It is trying to justify why we came up with that technical decision rather than simply stating what this value is.

What about something like this?

```
Don't explicitly specify a Workflow ID Conflict Policy.

The server will determine the effective behavior as follow:
1. If `workflowIdReusePolicy` is set to `TERMINATE_IF_RUNNING`,
   then behave the same as `WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING`;
2. If request is `StartWorkflow`, then behave the same as `WORKFLOW_ID_CONFLICT_POLICY_FAIL`;
3. If request is `SignalWithStartWorkflow`, then behave the same as `WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING`.

Note: this is the only acceptable value for Conflict Policy if the Reuse Policy is set to `TERMINATE_IF_RUNNING`.
```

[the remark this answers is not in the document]
mjameswh: Combining removal of the `WORKFLOW_ID_CONFLICT_POLICY_` prefix AND removal of the `UNSPECIFIED` value, you can update the `checkExtends` assertions to something like this:

```
checkExtends<
  `WORKFLOW_ID_CONFLICT_POLICY_${(keyof typeof WorkflowIdConflictPolicy) | 'UNSPECIFIED'}`,
  keyof typeof temporal.api.enums.v1.WorkflowIdConflictPolicy
>();
checkExtends<
  keyof typeof temporal.api.enums.v1.WorkflowIdConflictPolicy,
  `WORKFLOW_ID_CONFLICT_POLICY_${(keyof typeof WorkflowIdConflictPolicy) | 'UNSPECIFIED'}`
>();
```

And look at what we did for either `WorkflowUpdateStage` or `VersioningIntent` to convert those public enum values to the corresponding protobuf value (they use two distinct approaches, both are valid).

packages/common/src/workflow-options.ts
```diff
@@ -38,24 +41,76 @@ export enum WorkflowIdReusePolicy {
   WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE = 3,
 
   /**
-   * Terminate the current workflow if one is already running.
```
mjameswh, line 41: The deprecation comment needs the `@deprecated` annotation. Also should come last… though arguably, the fact that this is incompatible with `WorkflowIdConflictPolicy` should be part of the deprecation note.

mjameswh: Suggestion:

```
  /**
   * Terminate the current Workflow if one is already running; otherwise allow reusing the Workflow ID.
   *
   * @deprecated Use {@link WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE} instead, and
   *             set `WorkflowOptions.workflowIdConflictPolicy` to {@link WorkflowIdConflictPolicy.WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING}.
   *             When using this option, `WorkflowOptions.workflowIdConflictPolicy` must be left unspecified.
   */
```

packages/common/src/workflow-options.ts
```diff
@@ -7,11 +7,14 @@ import { checkExtends } from './type-helpers';
 // Avoid importing the proto implementation to reduce workflow bundle size
 // Copied from temporal.api.enums.v1.WorkflowIdReusePolicy
 /**
+ * Defines what happens when trying to start a Workflow with the same ID as a *Closed* Workflow.
+ *
+ * See {@link WorkflowOptions.workflowIdConflictPolicy} for what happens when trying to start a Workflow with the same ID as a *Running* Workflow.
```
mjameswh, line 12: Please break long comment lines. prettier won't do it for you.

The project is not consistent on this regard (because of no tooling to automate that), but I personally strive to get good readability at 100 columns, allowing occasional overflows… Whatever that keeps the code comfortable to read. Note however that you can't never break a line in the middle of a `{@link ...}` and similar tags; the parser won't handle that. So having extra long lines with those tags trailing on the right side is ok.

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,124 @@ test('Start of workflow is delayed', async (t) => {
...
+      client.workflow.start(conflictId, {
+        taskQueue,
+        workflowId: wfid,
+      }),
+      {
+        instanceOf: WorkflowExecutionAlreadyStartedError,
+      }
+    );
+
+    t.true(err instanceof WorkflowExecutionAlreadyStartedError && err.message === 'Workflow execution already started');
```
mjameswh, line 247: Add `message` to the `throwsAsync()` checks above…

```
      {
        instanceOf: WorkflowExecutionAlreadyStartedError,
        message: 'Workflow execution already started',
      }
```

mjameswh: Or you can even just not check for the error message. Given that it is an instance of the correct error class, that should be enough for this specific test.

mjameswh: (same for all the tests below)

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,124 @@ test('Start of workflow is delayed', async (t) => {
...
+    const handle2 = await client.workflow.start(conflictId, {
+      taskQueue,
+      workflowId: wfid,
+      workflowIdConflictPolicy: WorkflowIdConflictPolicy.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING,
+    });
+
+    const desc = await handleWithRunId.describe();
+    const desc2 = await handle2.describe();
+
+    t.deepEqual(desc.runId, desc2.runId);
```
mjameswh, line 275: nit: Use `t.is()` for simple strings

packages/test/src/test-integration-workflows.ts
```diff
@@ -216,6 +216,124 @@ test('Start of workflow is delayed', async (t) => {
...
+    // Confirm terminates and starts new
+    const handle3 = await client.workflow.start(conflictId, {
+      taskQueue,
+      workflowId: wfid,
+      workflowIdConflictPolicy: WorkflowIdConflictPolicy.WORKFLOW_ID_CONFLICT_POLICY_TERMINATE_EXISTING,
+    });
+
+    const descWithRunId = await handleWithRunId.describe();
+    const desc3 = await handle3.describe();
+    t.true(descWithRunId.runId !== desc3.runId);
```
mjameswh, line 288:
```suggestion
    t.not(descWithRunId.runId, desc3.runId);
```

mjameswh (changes requested):
The only blockers are my comments regarding the `WorkflowIdConflictPolicy` enum, as that affects public APIs.

Everything else can technically be dealt in a distinct PR, though those other things should really not take long to fix anyway.

stephanos:
@mjameswh no worries; I was actually informed that the docs are only blocked on the Go SDK and are therefore unblocked now!

packages/common/src/workflow-options.ts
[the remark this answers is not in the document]
antlai-temporal: I think the challenge is that we have  `WORKFLOW_ID_REUSE_POLICY_UNSPECIFIED`, again an optional field, and we could have got rid of the prefix. It is a bit too late for that without creating major deprecation issues. And we typically specify both reuse_policy and conflict_policy together, so it looks very confusing if one has the prefix, the other does not, and one can be set with an UNSPECIFIED field and the other cannot...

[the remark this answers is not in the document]
antlai-temporal: So in that sense, if we were starting from scratch, I completely agree with you. But I think for consistency we need to keep the prefix and the unspecified field...

packages/common/src/workflow-options.ts
antlai-temporal: Much better thanks

packages/common/src/workflow-options.ts
antlai-temporal: I wonder if we could use this https://prettier.io/docs/en/options.html#print-width but that could break all the`{@link}`...

packages/test/src/test-integration-workflows.ts
antlai-temporal: Done

stephanos:
@antlai-temporal is there an update here?

packages/common/src/workflow-options.ts
dandavison: @mjameswh @antlai-temporal is this comment the only thing blocking this PR? Is there a path forwards where we get this PR merged now, without blocking on https://github.com/temporalio/sdk-typescript/pull/1534?

mjameswh (approved):
I will refactor the `WorkflowIdConflictPolicy` enum in #1534.

antlai-temporal:
> I will refactor the `WorkflowIdConflictPolicy` enum in #1534.

Thanks @mjameswh !

PR #1497, opened 2024-08-14
fix(ci): Revert enabling file-based persistence on dev server for stress tests

mjameswh:
## What changed

- Disable file-based persistence on dev server for stress tests, which we previously enabled in #1480. Though desirable in some aspects, enabling file-based persistence resulted in reducing tests performance, causing recurring timeout errors during nightly stress tests, as well as flakes during regular PR/main branch commit stress tests.

- For now, we can hopefully resolve the original issue by scaling up the GHA runners (which we already did in #1492). As a more stable solution, we can investigate at transitioning to Temporal Cloud.

### DIFF
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -113,7 +113,7 @@ jobs:
       - name: Run Temporal CLI
         shell: bash
         run: |
-          temporal server start-dev --headless --db-filename /tmp/temporal.sqlite3 &> /tmp/devserver.log &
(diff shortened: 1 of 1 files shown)


Sushisource (approved):
Wow, fascinating.

Is the issue here basically a bunch of writes are hitting disk instead of memory and that just slowed down the server a bunch?

Good on you for figuring that one out. I needed to be a little more permissive in terms of looking outside of the commit range where it started failing, which I did a little, but not enough, lol.

PR #1498, opened 2024-08-14
fix(client): Reexport WorkflowExecutionAlreadyStartedError from `@temporalio/client`

mjameswh:
## What was changed

- `WorkflowExecutionAlreadyStartedError` was previously in `@temporalio/client`, but got recently moved to `@temporalio/common`. Add a reexport to avoid build issues.

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -22,6 +22,7 @@ export {
   TemporalFailure,
   TerminatedFailure,
   TimeoutFailure,
+  WorkflowExecutionAlreadyStartedError,
 } from '@temporalio/common';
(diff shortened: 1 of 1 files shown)


mnahkies:
@mjameswh this was previously exported by `@temporalio/workflow` as well. I've changed our imports to `@temporalio/common` for now, but I wondered if this should be fixed as well.

Also curious if there's any advice on which packages we should prefer to import things like this, is it always `common` where available?

mjameswh:
> Also curious if there's any advice on which packages we should prefer to import things like this, is it always common where available?

In general, yes, the export from `common` is the correct one.

PR #1500, opened 2024-08-17
feat: mark run activity span as consumer span

remitamine:
## What was changed

the arguments passed to `instrument` function in OpenTelemetryActivityInboundInterceptor's execute method has been expended which would be passed to the Open Telemetry `startActiveSpan` method.

## Why?

this marks the run activity spans as [consumer spans](https://opentelemetry.io/docs/concepts/signals/traces/#consumer) and adds a value to the [messaging system field](https://opentelemetry.io/docs/specs/semconv/attributes-registry/messaging/).
the information in itself is useful and it's needed in some cases for observability software such as Elastic Observability when [apm-agent-nodejs](https://github.com/elastic/apm-agent-nodejs) is used with [OpenTelemetry bridge](https://www.elastic.co/guide/en/apm/agent/nodejs/current/opentelemetry-bridge.html) where those fields are checked to properly [map the span](https://github.com/elastic/apm-agent-nodejs/blob/2031a263befadc16f00e3584f26932f03b4e7d92/lib/opentelemetry-bridge/OTelSpan.js#L212) to a [messaging transaction](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md).

### DIFF
--- a/packages/interceptors-opentelemetry/src/instrumentation.ts
+++ b/packages/interceptors-opentelemetry/src/instrumentation.ts
@@ -58,6 +58,7 @@ async function wrapWithSpan<T>(
 export interface InstrumentOptions<T> {
   tracer: otel.Tracer;
   spanName: string;
+  spanOptions?: otel.SpanOptions;
   fn: (span: otel.Span) => Promise<T>;
   context?: otel.Context;
   acceptableErrors?: (err: unknown) => boolean;
@@ -69,14 +70,23 @@ export interface InstrumentOptions<T> {
 export async function instrument<T>({
   tracer,
   spanName,
+  spanOptions = {},
   fn,
   context,
   acceptableErrors,
 }: InstrumentOptions<T>): Promise<T> {
   if (context) {
     return await otel.context.with(context, async () => {
-      return await tracer.startActiveSpan(spanName, async (span) => await wrapWithSpan(span, fn, acceptableErrors));
+      return await tracer.startActiveSpan(
+        spanName,
+        spanOptions,
+        async (span) => await wrapWithSpan(span, fn, acceptableErrors)
+      );
     });
   }
-  return await tracer.startActiveSpan(spanName, async (span) => await wrapWithSpan(span, fn, acceptableErrors));
+  return await tracer.startActiveSpan(
+    spanName,
+    spanOptions,
+    async (span) => await wrapWithSpan(span, fn, acceptableErrors)
+  );
 }
--- a/packages/interceptors-opentelemetry/src/worker/index.ts
+++ b/packages/interceptors-opentelemetry/src/worker/index.ts
@@ -29,7 +29,13 @@ export class OpenTelemetryActivityInboundInterceptor implements ActivityInboundC
   async execute(input: ActivityExecuteInput, next: Next<ActivityInboundCallsInterceptor, 'execute'>): Promise<unknown> {
     const context = await extractContextFromHeaders(input.headers);
     const spanName = `${SpanName.ACTIVITY_EXECUTE}${SPAN_DELIMITER}${this.ctx.info.activityType}`;
-    return await instrument({ tracer: this.tracer, spanName, fn: () => next(input), context });
+    return await instrument({
+      tracer: this.tracer,
+      spanName,
+      spanOptions: { attributes: { 'messaging.system': 'temporal' }, kind: otel.SpanKind.CONSUMER },
+      fn: () => next(input),
+      context,
+    });
   }
 }
 


mjameswh:
@remitamine Can you please sign the CLA? I'll have to decline this PR otherwise.

mjameswh:
> one issue I'm having with this is that the @CLAassistant is requesting permissions that i don't understand why would they be necessary, especially about the access to private contributions(even if it's partial) and the ability to modify the contributions.

Oh, that's unexpected. I totally understand your concern.

For context, [CLA Assistant](https://cla-assistant.io/) is developed and operated by SAP. We are mere users of that service. The list of permissions the service is requesting are those that are normally required from a project's owner (i.e. the user or organization owning the project to be covered by the CLA). **As a contributor, you should only be required for permission to read your email address.**

Now, there seems to be some long outstanding issues in CLA Assistant that makes it sometime request for more permissions than it should. One known case of this is [having cookie disabled](https://github.com/cla-assistant/cla-assistant/issues/566#issuecomment-635116981). Could this be your case? Also, make sure to access the CLA signature page using [this URL](https://cla-assistant.io/temporalio/sdk-typescript?pullRequest=1500) (which includes the repo name AND the PR number); I believe any other route will get you to the project administrator interface, which will trigger a request for extended GH permissions.

remitamine:
> > one issue I'm having with this is that the @CLAassistant is requesting permissions that i don't understand why would they be necessary, especially about the access to private contributions(even if it's partial) and the ability to modify the contributions.
> 
> Oh, that's unexpected. I totally understand your concern.
> 
> For context, [CLA Assistant](https://cla-assistant.io/) is developed and operated by SAP. We are mere users of that service. The list of permissions the service is requesting are those that are normally required from a project's owner (i.e. the user or organization owning the project to be covered by the CLA). **As a contributor, you should only be required for permission to read your email address.**
> 
> Now, there seems to be some long outstanding issues in CLA Assistant that makes it sometime request for more permissions than it should. One known case of this is [having cookie disabled](https://github.com/cla-assistant/cla-assistant/issues/566#issuecomment-635116981). Could this be your case? Also, make sure to access the CLA signature page using [this URL](https://cla-assistant.io/temporalio/sdk-typescript?pullRequest=1500) (which includes the repo name AND the PR number); I believe any other route will get you to the project administrator interface, which will trigger a request for extended GH permissions.

i guess it might be related to the cookies, although they are not completely disabled i do have add-ons that control the cookies.
anyway, using the URL that you've posted worked this time without additional permissions.

mjameswh:
Thanks for signing in the CLA!

Now, regarding the changes you propose in this PR, I think that `SpanKind.CLIENT` and `SpanKind.Server` would better reflect the relationship between a Workflow and an Activity. The thing here is that, from the Workflow's perspective, the Activity executes _synchronously_ (i.e. the Workflow is generally waiting for the Activity to complete, see comments [here](https://opentelemetry.io/docs/concepts/signals/traces/#client)).

What do you think?

remitamine:
> Thanks for signing in the CLA!
> 
> Now, regarding the changes you propose in this PR, I think that `SpanKind.CLIENT` and `SpanKind.Server` would better reflect the relationship between a Workflow and an Activity. The thing here is that, from the Workflow's perspective, the Activity executes _synchronously_ (i.e. the Workflow is generally waiting for the Activity to complete, see comments [here](https://opentelemetry.io/docs/concepts/signals/traces/#client)).
> 
> It is true that, under the hood, the Temporal server will actually queue up a task for later execution, but that's an internal detail.
> 
> What do you think?

based on my understanding of [temporal](https://docs.temporal.io/workers#task-queue) and the semantic conventions for both [HTTP](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) and [Messaging](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/) spans, i think that Messaging spans fit more into Temporal's having a server(intermediary) that receives messages from a producer and forwards the messages into consumers, in the this case the activity is processing the messages forwarded from the temporal server after receiving start activity messages from the producer, and the processing is not necessary immediate(it's queued), the queuing of the messages is actually crucial to make the distinction between HTTP/RPC spans(that fit into client and server types where the messages are supposed to be process immediately) and the messaging spans(that fit into the producer and consumer types where the messages can be delayed based on the queue).

chris-olszewski:
Our Opentelemetry integration has changed significantly since this PR was opened. Since there seems to be open questions on the proper span options, I would rather this decision get made in a GitHub issue than a PR. Please open a GitHub issue if you still think a consumer span is appropriate here.

PR #1513, opened 2024-09-12
chore: Update Core to a8150d5c (+job ordering update)

### LINKED ISSUE #1511: [Feature Request] Update Core SDK: expect `initializeWorkflow` instead of `startWorkflow` jobs
### Description

Core recently renamed the `startWorkflow` activation job to `initializeWorkflow`, with a different ordering and expectations (see [Core's PR](https://github.com/temporalio/sdk-core/pull/789)). We need to update activation handling logic.

mjameswh:
## What was changed

- Update Core to a8150d5c.
- Updated Workflow Activation handling to expect `InitializeWorkflow` jobs, instead of `StartWorkflow` (fix #1511).
- Delay payload conversion of workflow's memo/SA/lastfailure/etc until we're inside the sandbox, as there's no guarantee that converted payloads are safe to transfer through the V8 message port to the worker thread.

### DIFF
--- a/packages/worker/src/workflow/vm-shared.ts
+++ b/packages/worker/src/workflow/vm-shared.ts
@@ -328,6 +328,11 @@ export abstract class BaseVMWorkflow implements Workflow {
     }));
     this.activator.addKnownFlags(activation.availableInternalFlags ?? []);
 
+    // Initialization of the workflow must happens before anything else. Yet, keep the init job in
+    // place in the list as we'll use it as a marker to know when to start the workflow function.
+    const initWorkflowJob = activation.jobs.find((job) => job.initializeWorkflow != null)?.initializeWorkflow;
+    if (initWorkflowJob) this.workflowModule.initialize(initWorkflowJob);
+
     const hasSignals = activation.jobs.some(({ signalWorkflow }) => signalWorkflow != null);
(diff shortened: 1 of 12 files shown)


packages/worker/src/workflow/vm-shared.ts
```diff
@@ -328,6 +328,11 @@ export abstract class BaseVMWorkflow implements Workflow {
     }));
     this.activator.addKnownFlags(activation.availableInternalFlags ?? []);
 
+    // Initialization of the workflow must happens before anything else. Yet, keep the init job in
+    // place in the list as we'll use it as a marker to know when to start the workflow function.
+    const initWorkflowJob = activation.jobs.find((job) => job.initializeWorkflow != null)?.initializeWorkflow;
+    if (initWorkflowJob) this.workflowModule.initialize(initWorkflowJob);
```
Sushisource, line 334:
```suggestion
    // Initialization of the workflow must happen before anything else. Yet, keep the init job in
    // place in the list as we'll use it as a marker to know when to start the workflow function.
    const initWorkflowJob = activation.jobs.find((job) => job.initializeWorkflow != null)?.initializeWorkflow;
    if (initWorkflowJob) this.workflowModule.initialize(initWorkflowJob);
```

Sushisource (approved):
Makes sense to me!

PR #1519, opened 2024-09-12
fix(workflow): Fix multiple context leaks in `reuseV8Context` executor

### LINKED ISSUE #1476: [Bug] Reusable VM allows context leak due to global variable reassignment
### Describe the bug

When the `reuseV8Context` option is enabled, we freeze global objects to prevent modifications from leaking from one Workflow execution context to another one.

However, we do not prevent reassignment of global variables. And since context switching is based solely on global variable names, the engine will leave those reassigned global variable untouched.

For example, the following test fails:

```
export async function sharedGlobalReassignment(): Promise<[string, string]> {
  type ConsoleExtended = Console & { wfid: string };
  globalThis.console = { ...console, wfid: workflowInfo().workflowId } as ConsoleExtended;
  await sleep(1);
  return [workflowInfo().workflowId, (console as ConsoleExtended).wfid];
}

test('Shared global state', withReusableContext, async (t) => {
  const { createWorker, taskQueue, env } = t.context;
  const worker = await createWorker();
    const [res1, res2] = await Promise.all([
      env.client.workflow.execute(sharedGlobalReassignment, { taskQueue, workflowId: randomUUID() }),
      env.client.workflow.execute(sharedGlobalReassignment, { taskQueue, workflowId: randomUUID() }),
    ]);
    t.deepEqual(res1[0], res1[1]);
    t.deepEqual(res2[0], res2[1]);
  });
});
```

Deleting a global variable also fails, as the variable "reappears" on the next context switch. For example, the following test fails:

```
export async function globalMutatorAndDestructor(): Promise<number> {
  const global = globalThis as { a?: number };
  global.a = (global.a || 0) + 1;
  await sleep(1);
  delete global.a;
  await sleep(1);
  global.a = (global.a || 0) + 1;
  return global.a;
}

test('Set then Delete a global property', withReusableContext, async (t) => {
  const { createWorker, taskQueue, env } = t.context;
  const worker = await createWorker();
    const res = await env.client.workflow.execute(globalMutatorAndDestructor, { taskQueue, workflowId: randomUUID() });
    t.is(res, 1);
  });
});
```

### LINKED ISSUE #1592: [Bug] Reusable VM allows context leak due to global symbol properties
### Describe the bug

When the `reuseV8Context` option is enabled, the logic that swaps global variables in and out for a specific Workflow Execution only look at "regular" properties on `globalThis`, but ignores `symbol` properties. Symbol properties may therefore leak from one Workflow Execution to another.

In a [recent investigation](https://github.com/izakfilmalter/temporal-effect-oom), this was identified as the cause of a memory leak when using [Effect](https://github.com/Effect-TS/effect) in Workflow Code, as Effect was storing its own "global store" as a symbol property on `globalThis`. Though this specific case has just been [fixed on Effect's side](https://github.com/Effect-TS/effect/pull/4196), we recognize that there was nothing wrong in them using global symbol properties this way, and that other libraries are likely to do so. We must therefore ensure that Temporal SDK is able to properly handle global symbol properties.

Huge thanks to @izakfilmalter for his work in preparing a repro project for this issue, capturing heap dumps, and coordinating efforts between Effect's and Temporal's teams, as well as to @mikearnaldi from [Effect](https://github.com/Effect-TS/effect) for his precious collaboration in investigating this through.

mjameswh:
## What was changed

- Fix multiple context leaks and bugs in the `reuseV8Context` executor:

  - …by reassigning a new object to an existing shared global variable: `globalThis.console = { ...globalThis.console, wfId: workflowInfo().workflowId }` (fixes #1476);

  - …by modifying one of Node's built in global objects: `globalThis.Number.a = workflowInfo().workflowId`;

  - …by deleting a previously set global variable: `globalThis.a = 1; await sleep(1) ; delete globalThis.a ; await sleep(1) ; globalThis.a = (globalThis.a || 0) + 1; /* globalThis.a is 2 rather than 1 */`

  - …by defining global symbol properties: `const mySymbol = Symbol.for('...'); globalThis[mySymbol] = ...` (fixes #1592).


packages/common/src/type-helpers.ts
```diff
@@ -141,8 +141,35 @@ export function SymbolBasedInstanceOfError<E extends Error>(markerName: string):
...
+
+  if (object.constructor?.name === 'Map' || object instanceof Map) {
+    const map = object as unknown as Map<any, any>;
+    for (const [k, v] of map.entries()) {
+      deepFreeze(k, visited, `${path}[key:${k}]`);
+      deepFreeze(v, visited, `${path}[value:${k}]`);
+    }
+    map.set = frozenMapSet;
+    map.delete = frozenMapDelete;
+    map.clear = frozenMapClear;
```
antlai-temporal, line 160: Do you need to freeze the map itself too so that they don't override .set .delete .clear again?

packages/common/src/type-helpers.ts
```diff
@@ -141,8 +141,35 @@ export function SymbolBasedInstanceOfError<E extends Error>(markerName: string):
...
+  }
+
+  if (object.constructor?.name === 'Set' || object instanceof Set) {
+    const set = object as unknown as Set<any>;
+    for (const v of set.values()) {
+      deepFreeze(v, visited, `${path}.${v}`);
+    }
+    set.add = frozenSetAdd;
+    set.delete = frozenSetDelete;
+    set.clear = frozenSetClear;
```
antlai-temporal, line 170: Ditto for the set object, so that .add .delete .clear are not changed afterwards

antlai-temporal:
I'm a bit confused about what is the ultimate goal and why this is needed. If we want to eliminate all side channels this is not going to work, e.g., you could have timing side-channels, and the v8 VM is not designed to eliminate side-channels. If the goal is that other untrusted code is running in the same reused v8 context and attacking other users by modifying the context, I'm not sure we are protecting all the primordial classes (a String, Number,...) either, and in any case, that's a very hard problem...

Is this just to help prevent some mistakes, nothing to do with security?

mjameswh:
> Is this just to help prevent some mistakes, nothing to do with security?

No, that's really not a security thing (that's simply impossible AFAIK with Node's `vm`). And we don't have a goal of preventing _all_ possible escapes out of the VM.

The `reuseV8Context` is a performance optimization (roughly 66% reduction on RAM usage for cached workflows, and up to 50% on CPU usage). The trick is that instead of creating a new vm context for each workflow (which is slow and allocates a large number of objects), we use a single vm context, swapping global variables in and out before and after executing each activation.

However, there's been some reports of users doing uncommon but still legitimate and reasonable things in their workflow code, that would result in context from one workflow execution leaking into another workflow execution. Depending on use cases, that may result in non-deterministic behaviors, memory leaks, or some other oddities.

For example, one user recently reported that they were mocking functions in the `console` object inside their workflow code to redirect messages to the workflow log. They were doing that because their workflow was calling some outside library, not meant specifically for Temporal Workflows, which would print things its own logging through `console.log`. Agreed, that's not the best way of doing this, but still something I would have expected to be supported by our sandbox. Their code looked something like this:

```
async function myWorkflow() {
  const workflowId = workflowInfo().workflowId;
  globalThis.console = {
    ...console,
    log: (message) => {
      log(`${workflowId}: message`)
    }
  }
  // ...
```

The problem here is that previously, when leaving a workflow's execution context, the Reusable VM's would not touch any global variable that existed right after start. In this case, that means that it would leave the `console` object in place, even though it was no longer the original, pristine console object! That would result in them seeing log messages with an incorrect `workflowId`.

Once you know it, it's quite easy to fix their code to avoid this issue (they did), but it's also an opportunity for us to make our engine more reliable.

mjameswh:
There's also one test, "Shared global state is frozen", that existed before this PR that was actually not testing what we thought it was testing. That is, the test was asserting attempting to add a custom property to some global object from Node's default vm context would be rejected.

However, it was doing so by adding something to the `setTimetout` function, but it turns out that's insufficient as the `globalThis.setTimeout` is overridden by the SDK itself, resulting in it being present in the `globals` object exposed outside of the vm, where global variables that would not been modified inside the vm would not appear in the `globals` object (seems to be some kind of internal optimization done by Node/V8).

This PR fixes that bug; the test is now asserting the proper behavior against the `Array` global function instead.

packages/common/src/type-helpers.ts
```diff
@@ -152,8 +179,11 @@ export function deepFreeze<T>(object: T): T {
 
     if (value && typeof value === 'object') {
       try {
-        deepFreeze(value);
+        deepFreeze(value, visited, `${path}.${name}`);
       } catch (err) {
+        if (typeof value !== 'object' || value.constructor.name !== 'Uint8Array') {
```
antlai-temporal, line 184: You are checking that `if (value && typeof value === 'object') {` already
do you need to check again because deepFreeze could change it?
Also, a comment why Uint8Array is special may help

packages/test/src/test-isolation.ts
```diff
@@ -118,3 +142,46 @@ test('Shared global state is frozen', withReusableContext, async (t) => {
...
+  const worker = await createWorker();
+  await worker.runUntil(async () => {
+    const [res1, res2] = await Promise.all([
+      env.client.workflow.execute(sharedGlobalReassignment, { taskQueue, workflowId: randomUUID() }),
+      env.client.workflow.execute(sharedGlobalReassignment, { taskQueue, workflowId: randomUUID() }),
+    ]);
+    t.deepEqual(res1[0], res1[1]);
+    t.deepEqual(res1[0], res1[2]);
+    t.deepEqual(res2[0], res2[1]);
+    t.deepEqual(res2[0], res2[2]);
```
antlai-temporal, line 166: Do we want to check res1[0] !== res2[0] for completeness

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -73,13 +70,31 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
...
+        for (const name of Object.getOwnPropertyNames(globalThis)) {
+          const descriptor = Object.getOwnPropertyDescriptor(globalThis, name);
+          pristine.set(name, descriptor);
+        }
+      `,
+      this.context
+    );
+    for (const [k, v] of this.pristine.entries()) {
+      if (k !== 'globalThis') {
+        deepFreeze(v.value);
```
antlai-temporal, line 92: Do we need to deepFreeze the key `k` too?

antlai-temporal (approved):

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -73,13 +70,31 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
     this._context = vm.createContext(globals, { microtaskMode: 'afterEvaluate' });
     this.injectConsole();
     script.runInContext(this.context);
-    this.contextKeysToPreserve = new Set(Object.keys(this.context));
+
+    this.pristine = new Map<string, PropertyDescriptor>();
+    (globals as any).__pristine = this.pristine;
+    // The V8 context is really composed of two distinct objects: the `globals` object defined above
+    // on the outside, and another internal object to which we only have access from the inside, which
+    // defines the built-in global properties. We need
```
antlai-temporal, line 78: fix comment

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -104,39 +119,62 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
...
+    const { isolateExecutionTimeoutMs } = this;
     const workflowModule: WorkflowModule = new Proxy(
       {},
       {
         get(_: any, fn: string) {
           return (...args: any[]) => {
-            Object.assign(context, bag);
+            for (const [pname, pdesc] of bag.entries()) {
+              Object.defineProperty(context, pname, pdesc);
+            }
```
antlai-temporal, line 132: I'm a bit confused about`bag` initialization, in particular if the first time we call 
` return vm.runInContext(`__TEMPORAL__.api.${fn}``
the bag would be empty, and whether that matters.
I see the bag is filled inside  `finally{...` after `bag.clear()` but not sure about the first call...

packages/common/src/type-helpers.ts
mjameswh: It is. After replacing these methods, we continue with the normal deepfreeze logic below, which will freeze the set/delete/clear methods, then the map itself.

packages/common/src/type-helpers.ts
```diff
@@ -152,8 +179,11 @@ export function deepFreeze<T>(object: T): T {
...
+          console.log(`Failed to freeze ${path}.${name}`, err);
```
bergundy, line 185: Please remove this before merging.

packages/common/src/type-helpers.ts
```diff
@@ -141,8 +141,35 @@ export function SymbolBasedInstanceOfError<E extends Error>(markerName: string):
   };
 }
 
-// Thanks MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze
-export function deepFreeze<T>(object: T): T {
+// This implementation is based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze.
+// Note that there are limits to this approach, as traversing using getOwnPropertyXxx doesn't allow
+// reaching variables defined using internal scopes. This implementation specifically look for and deal
```
bergundy, line 146:
```suggestion
// reaching variables defined using internal scopes. This implementation specifically looks for and deals
```

bergundy: Why is dealing with these types important? Have you seen any shared maps and sets that are exposed to users?

packages/test/src/test-isolation.ts
```diff
@@ -102,9 +102,33 @@ test('Module state is isolated and maintained between activations', async (t) =>
...
+export async function moduleMapMutator(): Promise<[number, number]> {
+  someMap.set('test', (someMap.get('test') || 0) + 1);
+  const middle = someMap.get('test') as number;
+  await sleep(1);
+  someMap.set('test', (someMap.get('test') || 0) + 1);
+  const final = someMap.get('test') as number;
+  return [middle, final];
+}
+
+test('Deep Freeze reaches Map values', async (t) => {
```
bergundy, line 116: Do we actually freeze the map? We're not supposed to freeze user's modules.

packages/test/src/test-isolation.ts
```diff
@@ -102,9 +102,33 @@ test('Module state is isolated and maintained between activations', async (t) =>
...
+    ]);
+    t.deepEqual(res1, [1, 2]);
+    t.deepEqual(res2, [1, 2]);
+  });
+});
+
 export async function sharedGlobalMutator(): Promise<void> {
   try {
-    (setTimeout as any).a = 1;
+    (Array as any).a = 1;
```
bergundy, line 131: Why did you change this?

packages/test/src/test-isolation.ts
```diff
@@ -118,3 +142,46 @@ test('Shared global state is frozen', withReusableContext, async (t) => {
...
+  return [workflowInfo().workflowId, middle, (globalThis.console as ConsoleExtended).wfid];
+}
+
+test('Reassign shared global state', withReusableContext, async (t) => {
+  const { createWorker, taskQueue, env } = t.context;
+  const worker = await createWorker();
+  await worker.runUntil(async () => {
+    const [res1, res2] = await Promise.all([
+      env.client.workflow.execute(sharedGlobalReassignment, { taskQueue, workflowId: randomUUID() }),
+      env.client.workflow.execute(sharedGlobalReassignment, { taskQueue, workflowId: randomUUID() }),
```
bergundy, line 161: This may be racy in CI and may not actually provide the test coverage you want. There's nothing guaranteeing that the workflows will run concurrently.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -26,11 +26,8 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
    *
    * Use the {@link context} getter instead
    */
-  _context?: vm.Context;
-  /**
-   * Store the global object keys we want to share between contexts
-   */
-  readonly contextKeysToPreserve: Set<string>;
+  private _context?: vm.Context;
+  private pristine?: Map<string, PropertyDescriptor>;
```
bergundy, line 30: Docstring.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -104,39 +119,62 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
...
+    const pristine = this.pristine;
+    const bag: Map<string, PropertyDescriptor> = new Map();
+    const { isolateExecutionTimeoutMs } = this;
     const workflowModule: WorkflowModule = new Proxy(
       {},
       {
         get(_: any, fn: string) {
           return (...args: any[]) => {
-            Object.assign(context, bag);
+            for (const [pname, pdesc] of bag.entries()) {
```
bergundy, line 130: nit: would have liked to prevent constructing the entries array for every time we call into the VM.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -104,39 +119,62 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
...
                 }
               }
-              for (const k in keysToDelete) {
-                delete context[k];
-              }
+
+              // Looks like Node/V8 is not properly syncing deletion of keys on the outter context
+              // object to the inner globalThis object. Hence, we delete them from inside the context.
+              context.__TEMPORAL_ARGS__ = keysToCleanup;
+              vm.runInContext(
```
bergundy, line 162: This will add a bit of overhead for every time we enter the context, I would try to find an alternative way to do this.

packages/common/src/type-helpers.ts
mjameswh: No longer pertinent

packages/test/src/test-isolation.ts
mjameswh: That test was presumably testing that v8's built-in functions were not mutable, but the SDK redefines `setTimeout`, so that test was actually useless. Turns out in fact that built-in functions _were_ unexpectedly mutable.

packages/worker/src/workflow/reusable-vm.ts
mjameswh: No, k would be a string or a symbol, which can't be mutated.

mjameswh:
The code changed significantly since the first round of reviews.

I'm closing this PR and opening a new one.

PR #1525, opened 2024-09-17
Document OtelCollectorExporter.otel property

ikonst:
Clarifying the expected format and the semantics.

Context: https://temporalio.slack.com/archives/C02MTL0GEKH/p1726577174636549


mjameswh (approved):
Definitely useful. Thanks!

PR #1527, opened 2024-09-18
Document timestamps in activity Info

ikonst:
Using this as an opportunity to confirm my understanding is correct w.r.t `scheduledTimestampMs`.


mjameswh (approved):
Thanks, that's pertinent.

Issue #1309, opened 2023-12-05
[Bug] Comment unclear on when activities are cancelled after calling `Worker.shutdown`


bergundy:
It should mention that the activities are cancelled after `shutdownGraceTime`.

https://github.com/temporalio/sdk-typescript/blob/28d4de33a2697c949b863d1c541063384c4fcc7c/packages/worker/src/worker.ts#L733

gaurav7019:
Can I pick this up? I am new to this repo, let me know where I should start.

lorensr:
@gaurav7019 sure, thanks—you can submit a PR that edits the comment linked to above

nikhilbhatia08:
@lorensr I would like to submit a PR that edits the comment

PR #1528, opened 2024-09-20
fix(worker): clarify activity cancellation after shutdownGraceTime

nikhilbhatia08:
## What was changed

Updated the shutdown process description to mention that ongoing activities will be canceled after the shutdownGraceTime has passed. This ensures the behavior is clear in the worker shutdown flow.

## Checklist

1. Closes  [#1309](https://github.com/temporalio/sdk-typescript/issues/1309)

3. How was this tested:

4. Any docs updates needed?

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -746,7 +746,8 @@ export class Worker {
   }
 
   /**
-   * Start shutting down the Worker. The Worker stops polling for new tasks and sends
+   * Start shutting down the Worker. The Worker stops polling for new tasks and sends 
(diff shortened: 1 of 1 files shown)


mjameswh:
Thanks @nikhilbhatia08, but I'm not comfortable with that one.

That whole comment, and probably other places, need to be completely rewritten to correctly explain the shutdown process. There's much more to it than just questions about activity cancellations.

Given the complexity of the shutdown logic and the fact that we don't have any up-to-date, correct and comprehensive source for that knowledge, I don't think that task can be achieved by external contributors at this point. I modified the issue ticket to remove the "Good first issue" tag.

Again, thanks anyway 😉

PR #1530, opened 2024-09-21
fix(workflow): Workflow Activation Encoder was discarding SDK flags

### LINKED ISSUE #1533: [Bug] NDE while replaying Workflows in 1.11.1 with same code
### Describe the bug

Using TS SDK 1.11.0 and 1.11.1, some Workflows may result in NDE on replay, even though there's been no change to Workflow Code.

The NDE will generally involves:
 
    A TimerCanceled command; or
    A change in the ordering of some commands.

mjameswh:
## What was changed

- The logic that applies payload encoding on Workflow Activation Completions was missing some object spread operators, causing lang's SDK flags to be discarded. This PR fixes that logic.

- Introduced a mechanism that allows retroactively enabling SDK Flags on Workflow Task being replayed, based on the SDK version that produced that originally processed the WFT.

## Why?

- Due to SDK flags not being properly recorded to the Workflow Task, this bug would result in Non Determinism Errors on attempts to replay a Workflow where the presence or absence of a flag would effectively impact subsequent commands produced by the Workflow.

- The most common symptoms of this issue would be an NDE that involves:
  - A `TimerCanceled` command; or
  - A change in the ordering of some commands.

- Problematic Workflow Tasks will have been produced specifically by either TS SDK 1.11.0 or 1.11.1. Similar symptoms on Workflow Tasks emitted by other releases of the TS SDK would not be related to this bug.

- Given that existing Workflow Histories may already be affected by this omission, there's a high probability that updating Workers to an SDK release that fixes this issue would cause storms of NDEs errors, for which the only way forward would be to reset or terminate those workflows. To avoid that situation, we introduced a mechanism to retroactively enable the affected SDK Flags on WFT emitted by SDK 1.11.0 or 1.11.1.

Fixes #1533.


ikonst:
Consider adding a comment to `Activator.hasFlag` to describe how flags work in that regard:
```
+   // During replays, sdkMetadata should dictate flags, as to respect their state in previous runs
    if (!this.info.unsafe.isReplaying && flag.default) {
      this.knownFlags.add(flag.id);
      return true;
    }
```

packages/workflow/src/flags.ts
```diff
@@ -33,17 +38,46 @@ export const SdkFlags = {
...
 
-function defineFlag(id: number, def: boolean): SdkFlag {
-  const flag = { id, default: def };
+/**
+ *
+ * @param id
+ * @param def
+ * @param otherConditions
+ * @returns
+ */
```
ikonst, line 55: meant to document?

packages/workflow/src/flags.ts
```diff
@@ -14,9 +17,11 @@ export const SdkFlags = {
    *   of the timeout, similar to how `condition(fn, timeout)` works.
    * - Timers created using setTimeout can now be intercepted.
    *
-   * @since Introduced in 1.10.2/1.11.0.
+   * @since Introduced in 1.10.2/1.11.0. However, due to an SDK bug, SDKs v1.11.0 and v1.11.1 were not
+   *        properly writing back the flags to history, possibly resulting in NDE on replay. We therefore
+   *        consider that a WFT emitted by Worker v1.11.0 or v1.11.1 to implicitely have this flag on.
```
Sushisource, line 22:
```suggestion
   *        consider that a WFT emitted by Worker v1.11.0 or v1.11.1 to implicitly have this flag on.
```

packages/workflow/src/internals.ts
```diff
@@ -879,18 +879,47 @@ export class Activator implements ActivationHandler {
   }
 
   /**
-   * Check if a flag is known to the Workflow Execution; if not, enable the flag if workflow
-   * is not replaying and the flag is configured to be enabled by default.
+   * Check if an SDK Flag may be considered as enabled for the current Workflow Task.
+   *
+   * SDK flags play a role similar to the `patched()` API, but are meant for internal usage by the
+   * SDK itself. They make it possible for the SDK to evolves its behaviors over time, while still
```
Sushisource, line 885:
```suggestion
   * SDK itself. They make it possible for the SDK to evolve its behaviors over time, while still
```

Sushisource (approved):
A little unfortunate to have to look at version strings, but, makes sense to me.

Issue #1327, opened 2023-12-27
[Feature Request] Validate input


lorensr:
### Is your feature request related to a problem? Please describe.

`client.workflow.start(type, { args: {} })` results in an opaque error message:

![image](https://github.com/temporalio/sdk-typescript/assets/251288/0958b186-fbf0-49d0-9c21-839c2b84a8fa)

### Describe the solution you'd like

Throw an error like `args must be an array`

mjameswh:
No longer pertinent.

PR #1531, opened 2024-09-23
Fix argument validation in WorkflowHandle for issue #1327

Vignesh18280:
#1327 

**What was changed**
This PR fixes the argument validation for WorkflowHandle. It ensures that the args parameter is always validated as an array and throws an appropriate error if it is not.

**Why?**
To address issue #1327, improving error handling and clarity for users by ensuring the args passed to workflows are always arrays.

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -308,6 +308,9 @@ function ensureArgs<W extends Workflow, T extends WorkflowStartOptions<W>>(
   opts: T
 ): Omit<T, 'args'> & { args: unknown[] } {
   const { args, ...rest } = opts;
+  if(args && !Array.isArray(args)) {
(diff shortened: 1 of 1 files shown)


mjameswh:
Can you please clarify the use case that would result in seeing this error message?

The TypeScript signature for that function no longer allows passing a non-array to `args`, and we generally do not add type checking for pure JS use cases.

Vignesh18280:
@mjameswh 
Thanks for the response.
It's true that TypeScript enforces array inputs for args.
However, since the issue remained open, I assumed that Temporal wanted to implement additional error handling for potential edge cases, particularly in JavaScript environments where type safety isn't as strict. I aimed to add that validation to ensure better runtime safety. 
Let me know if this approach aligns with the goal!

mjameswh:
I see. Then thanks again, but I'd prefer not changing this.

PR #1534, opened 2024-09-25
Normalize construction of user facing enums

mjameswh:
## What was changed

- This PR normalizes conventions around user facing `enums` in the TS SDK.

- More specifically, it features the following conventions:

  - No more redundant prefixes on enum values.
  
    For example:
    ```diff
      await startChild(sleep, {
    -   parentClosePolicy: ParentClosePolicy.PARENT_CLOSE_POLICY_ABANDON,
    +   parentClosePolicy: ParentClosePolicy.ABANDON,
      });
    ```

  - No more 'UNSPECIFIED' values in user facing enums; TypeScript's `undefined` already means that.
  
  - "Enum values" are now of type string, and the "enum" themselves are now implemented using `const` objects, with a string union type defined from all values of the const object. That makes it possible for users to simply pass the string representation of an enum value, without having to import and refer to the enum definition object. That's much more TS idiomatic and is better supported by TS code editors.

    The preceding example can therefore be further reduced down to:
    ```diff
      await startChild(sleep, {
    -   parentClosePolicy: ParentClosePolicy.ABANDON,
    +   parentClosePolicy: 'ABANDON',
      });
    ```

  - The const object is preserved (i.e not completely replaced by the union of strings type), for backward compatibility reasons, but also because it plays better in regard to documentation of each enum values.

  - We preserve build-time backward compatibility for enum values that existed before.

    For example, the following code will still compile, but will report a deprecation on `ParentClosePolicy.PARENT_CLOSE_POLICY_ABANDON`.

    ```typescript
      await startChild(sleep, {
        parentClosePolicy: ParentClosePolicy.PARENT_CLOSE_POLICY_ABANDON,
      });
    ```

  - We maintain the SCREAMING_SNAKE_CASE format for enum values

  - For "user facing enums" to/from "proto enums" conversions, we normalize on a `encodeEnumName(string): ProtoEnumValue` and a `decodeEnumName(ProtoEnumValue): string` functions, where `ProtoEnumValue` really is the _numerical value_ of the proto enum value. This ensures uniformity for all enum types, without risk of getting the proto definitions imported in the Workflow sandbox (which would significantly increase the per-workflow memory footprint).

### DIFF
--- a/packages/common/src/internal-workflow/enums-helpers.ts
+++ b/packages/common/src/internal-workflow/enums-helpers.ts
@@ -158,6 +158,7 @@ export function makeProtoEnumConverters<
   const reverseTable: Record<ProtoEnumValue, ShortStringEnumKey> = Object.fromEntries(
     Object.entries(mapTable).map(([k, v]) => [v, k])
   );
+  const hasUnspecified = (mapTable as any)['UNSPECIFIED'] === 0 || (mapTable as any)[`${prefix}UNSPECIFIED`] === 0;
 
   function isShortStringEnumKeys(x: unknown): x is ShortStringEnumKey {
     return typeof x === 'string' && x in mapTable;
@@ -192,6 +193,10 @@ export function makeProtoEnumConverters<
     if (input == null) {
       return undefined;
     } else if (typeof input === 'number') {
+      if (hasUnspecified && input === 0) {
+        return undefined;
+      }
+
       if (isNumericEnumValue(input)) {
         return reverseTable[input];
       }
--- a/packages/common/src/internal-workflow/enums-helpers.ts
+++ b/packages/common/src/internal-workflow/enums-helpers.ts
@@ -1,0 +1,289 @@
+import { Exact, RemovePrefix, UnionToIntersection } from '../type-helpers';
+
+/**
+ * Create encoding and decoding functions to convert between the numeric `enum` types produced by our
+ * Protobuf compiler and "const object of strings" enum values that we expose in our public APIs.
+ *
+ * ### Usage
+ *
+ * Newly introduced enums should follow the following pattern:
+ *
+ * ```ts
+ *     type ParentClosePolicy = (typeof ParentClosePolicy)[keyof typeof ParentClosePolicy];
+ *     const ParentClosePolicy = {
+ *       TERMINATE: 'TERMINATE',
+ *       ABANDON: 'ABANDON',
+ *       REQUEST_CANCEL: 'REQUEST_CANCEL',
+ *     } as const;
+ *
+ *     const [encodeParentClosePolicy, decodeParentClosePolicy] = //
+ *       makeProtoEnumConverters<
+ *         coresdk.child_workflow.ParentClosePolicy,
+ *         typeof coresdk.child_workflow.ParentClosePolicy,
+ *         keyof typeof coresdk.child_workflow.ParentClosePolicy,
+ *         typeof ParentClosePolicy,
+ *         'PARENT_CLOSE_POLICY_'  // This may be an empty string if the proto enum doesn't add a repeated prefix on values
+ *       >(
+ *         {
+ *           [ParentClosePolicy.TERMINATE]: 1, // These numbers must match the ones in the proto enum
+ *           [ParentClosePolicy.ABANDON]: 2,
+ *           [ParentClosePolicy.REQUEST_CANCEL]: 3,
+ *
+ *           UNSPECIFIED: 0,
+ *         } as const,
+ *         'PARENT_CLOSE_POLICY_'
+ *       );
+ * ```
+ *
+ * `makeProtoEnumConverters` supports other usage patterns, but they are only meant for
+ * backward compatibility with former enum definitions and should not be used for new enums.
+ *
+ * ### Context
+ *
+ * Temporal's Protobuf APIs define several `enum` types; our Protobuf compiler transforms these to
+ * traditional (i.e. non-const) [TypeScript numeric `enum`s](https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums).
+ *
+ * For various reasons, this is far from ideal:
+ *
+ *  - Due to the dual nature of non-const TypeScript `enum`s (they are both a type and a value),
+ *    it is not possible to refer to an enum value from code without a "real" import of the enum type
+ *    (i.e. can't simply do `import type ...`). In Workflow code, such an import would result in
+ *    loading our entire Protobuf definitions into the workflow sandbox, adding several megabytes to
+ *    the per-workflow memory footprint, which is unacceptable; to avoid that, we need to maintain
+ *    a mirror copy of each enum types used by in-workflow APIs, and export these from either
+ *    `@temporalio/common` or `@temporalio/workflow`.
+ *  - It is not desirable for users to need an explicit dependency on `@temporalio/proto` just to
+ *    get access to these enum types; we therefore made it a common practice to reexport these enums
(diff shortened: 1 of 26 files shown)


packages/common/src/internal-workflow/enums-helpers.ts
```diff
@@ -0,0 +1,294 @@
+import { Exact, RemovePrefix, UnionToIntersection } from '../type-helpers';
+
+/**
+ * Create encoding and decoding functions to convert between the numeric `enum` types produced by our
+ * Protobuf compiler and "const object of strings" enum values that we expose in our public APIs.
```
Sushisource, line 5: Now that's a docstring

packages/common/src/internal-workflow/enums-helpers.ts
```diff
@@ -0,0 +1,294 @@
...
+ *
+ * ### Usage
+ *
+ * Newly introduced enums should follow the following pattern:
+ *
+ * ```ts
+ *     type ParentClosePolicy = (typeof ParentClosePolicy)[keyof typeof ParentClosePolicy];
```
Sushisource, line 12: Yeah, kinda like we talked about in person, a macro really would make this dramatically easier.

I think probably some of this boilerplate could be reduced potentially, needing to pass the type, `typeof`, and `keyof` all separately, and possibly auto-assigning the enum numbers... but I don't think it's necessarily worth doing to save a few lines either.

packages/common/src/internal-workflow/enums-helpers.ts
```diff
@@ -0,0 +1,294 @@
...
+      }
+
+      if (isNumericEnumValue(input)) {
+        return reverseTable[input];
+      }
+
+      // FIXME: What should we do with unknown values?
+      // Throwing may end up causing more problems than it would solve if the decoded values
+      // would actually never get read anwyay. An alternative would be to return a string value
+      // containing the numeric value, something like "unknown_23"...
```
Sushisource, line 207: I think returning the string with the number in it makes sense to me since it mirrors the behavior you'd get with the pure numerical approach.

packages/common/src/type-helpers.ts
```diff
@@ -17,6 +17,84 @@ export function checkExtends<_Orig, _Copy extends _Orig>(): void {
...
+
+type KeysOfUnion<ObjectType> = ObjectType extends unknown ? keyof ObjectType : never;
+
+type ArrayElement<T> = T extends readonly unknown[] ? T[0] : never;
+
+type ExactObject<ParameterType, InputType> = {
+  [Key in keyof ParameterType]: Exact<ParameterType[Key], ObjectValue<InputType, Key>>;
+} & Record<Exclude<keyof InputType, KeysOfUnion<ParameterType>>, never>;
+
+export type Exact<ParameterType, InputType> =
```
Sushisource, line 71: Wowza

Sushisource (approved):
Cool! This should be a big QoL improvement

packages/common/src/internal-workflow/enums-helpers.ts
```diff
@@ -0,0 +1,294 @@
...
+      return undefined;
+    } else if (typeof input === 'string') {
+      let shorten: string = input;
+      if (shorten.startsWith(prefix)) {
+        shorten = shorten.slice(prefix.length);
+      }
+      if (isShortStringEnumKeys(shorten)) {
+        return mapTable[shorten];
+      }
+      throw new TypeError(`Invalid enum value: '${input}'`);
```
dandavison, line 184: I think this one's a `ValueError` rather than a `TypeError`.

packages/common/src/internal-workflow/enums-helpers.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    throw new TypeError(`Invalid proto enum value: '${input}' of type ${typeof input}`);
+  }
+
+  return [encode, decode] as const;
+}
+
+/**
+ * Given the exploded parameters of a proto enum (i.e. short keys, prefix, and short key of the
+ * unspecified value), make a type that _exactly_ correspond to the const object of strings enum,
```
dandavison, line 218:
```suggestion
 * unspecified value), make a type that _exactly_ corresponds to the const object of strings enum,
```

packages/common/src/internal-workflow/enums-helpers.ts
```diff
@@ -0,0 +1,294 @@
...
+      ? {
+          // e.g.: "UNSPECIFIED"?: undefined
+          [k in `${Unspecified}`]?: undefined;
+        }
+      : object)
+>;
+
+/**
+ * Given the exploded parameters of a proto enum (i.e. short keys, prefix, and short key of the
+ * unspecified value), make a type that _exactly_ correspond to the mapping table that the user is
```
dandavison, line 268:
```suggestion
 * unspecified value), make a type that _exactly_ corresponds to the mapping table that the user is
```

dandavison (approved):
Wow! There's quite a lot going on... but this should be a really nice improvement.

I'll just ask one question for now. If we take something like
```typescript
decodeStartChildWorkflowExecutionFailedCause(activation.failed.cause)
```

It would be nice if this looked more like
```typescript
decode(activation.failed.cause)
```

Do you see any possibilities in that direction?

packages/common/src/internal-workflow/enums-helpers.ts
mjameswh: >Yeah, kinda like we talked about in person, a macro really would make this dramatically easier.

Definitely. Unfortunately, TypeScript is only about the type system. It can only be used to define type constraints, but can't define new code in any way. Hence this very 

> I think probably some of this boilerplate could be reduced potentially, needing to pass the type, typeof, and keyof all separately, and possibly auto-assigning the enum numbers... but I don't think it's necessarily worth doing to save a few lines either.

I thought similarly, and made several iterations already on that type definition, but for various reasons, that would each time end up in less strict validations.

mjameswh:
> Wow! There's quite a lot going on... but this should be a really nice improvement.
> 
> I'll just ask one question for now. If we take something like
> 
> ```ts
> decodeStartChildWorkflowExecutionFailedCause(activation.failed.cause)
> ```
> 
> It would be nice if this looked more like
> 
> ```ts
> decode(activation.failed.cause)
> ```
> 
> Do you see any possibilities in that direction?

You would need to pass something that represents the enum-const-object itself, e.g. `decode(StartChildWorkflowExecutionFailedCause, activation.failed.cause)`. Same on `encode()`.

I think that might be possible, by adding a symbol property on the enum-const-object to record the mapping and reverse-mapping tables.

I wouldn't dig into that possibility at this point though.

packages/common/src/internal-workflow/enums-helpers.ts
```diff
@@ -0,0 +1,294 @@
...
+    ShortStringEnumKey
+  >,
+>(
+  mapTable: MapTable,
+  prefix: Prefix
+): [
+  (
+    input: ShortStringEnumKey | `${Prefix}${ShortStringEnumKey}` | ProtoEnumValue | null | undefined
+  ) => ProtoEnumValue | undefined, //
+  (input: ProtoEnumValue | null | undefined) => ShortStringEnumKey | undefined, //
```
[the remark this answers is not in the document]
mjameswh: The protobuf-to-ts transpiler makes every field `MyProtoType | null | undefined`. That's just the way it is, unfortunately, so `decode(...)` must accept that.

Now, I don't think there's any legitimate use case for `encode(...)` to receive `null` as input, but just for symetry, most of our `encode` functions have followed that pattern anyway.

PR #1539, opened 2024-10-07
fix(worker): Gracefully shut down Worker on unexpected errors

### LINKED ISSUE #1536: [Bug] Worker keeps accepting WFT after Worker Thread terminates due to ERR_WORKER_OUT_OF_MEMORY
### Describe the bug

A user reported the following sequence of events:

- They have a TS Workflow Worker running normally;
- At some point, they get the following ERROR-level log message: `{ code:"ERR_WORKER_OUT_OF_MEMORY" }`.
- After that point, it appears that the Worker is no longer making any progress on pending Workflow Tasks, despite the fact that the Worker process is still running.
- From the Server side, metrics indicate a considerable increase in calls to `PollWorkflowExecutionHistory` and a considerable reduction in calls to `PollWorkflowTaskQueue`. Provided data doesn't allow identifying the exact provenance of those calls (they normally operate ~12 active Workers on that NS).
- The situation continues until the Worker process is restarted, after which Workflow progress resumes.

### Analysis

`ERR_WORKER_OUT_OF_MEMORY` is an error from Node itself. That means that Node terminated a Worker Thread due to running out of memory.

Given the symptoms, it is sensible to assume that the Worker Thread killed by Node would be the Workflow Worker Thread, which means the Temporal Worker will no longer be able to process incoming Workflow Task.

What happens next is not clear, but the correct behavior would be for the Worker to initiate shutdown. There’s very little we can do at that point, but at least, let’s not pretend that everything’s all right. Should probably print a clear CRITICAL level message to the log, and terminate the process ASAP.

mjameswh:
## What changed

- Properly detect unexpected termination of a Workflow Worker Thread, e.g. due to `ERR_WORKER_OUT_OF_MEMORY` or an Unhandled Promise Rejection that could not be associated with a specific Workflow run. We were previously monitoring the Worker Thread's `exit` event, but empirical testing demonstrated that this would sometime not be sufficient, requiring that we also listen for the `error` event; monitoring the `error` event also has the advantage that get details on why the Worker Thread failed, so we can better report that to the user.

- When facing an "Unexpected Error", the Worker now attempts to perform a graceful shutdown, similar to calling `worker.shutdown()`. Previously, the Worker would immediately fall into the FAILED state, essentially performing no cleanup, not even stopping Core's polers; this could possibly result in various "zombie states", e.g. the Worker's `run` promise not resolving/failing even though the Worker has reached FAILED state; Core continuing to poll for activity tasks; etc.

- Until now, `Worker.runUntil(...)` would wait indefinitely for both the Worker's `run` promise and the provided promise to resolve or fail, _even in error cases_. In most practical use cases, that would create a possibility for the worker to hang indefinitely if the Worker was stopped due to "unexpected" factors.
 
    For example, in this very common test idiom, sending a SIGINT to the process would initiate shutdown of the Worker, potentially resulting in termination of the Worker before the Workflow completes; in that case, the Workflow would never complete, and consequently, the `runUntil` promise would never resolve, leaving the process in a hang state.

    ```
    await Worker.runUntil(() => client.workflow.execute(...));
    ```

    The behavior of `Worker.runUntil(...)` has therefore been changed so that if the worker shuts down before the inner promise completes, `runUntil` will allow no more than a certain delay, `promiseCompletionTimeout`, for the inner promise to complete, after which a `PromiseCompletionTimeoutError` is thrown.

    At this point, it is believed that a delay is very rarely pertinent, and would most certainly be indicative of a bug in user code; `promiseCompletionTimeout` therefore defaults to 0 seconds, which means we shouldn't wait. If needed, that duration can be adjusted through a new optional `RunUntilOptions` argument on the `runUntil(fnOrPromise, options)`.

## Misc

- Fixes #1536


packages/worker/src/worker.ts
```diff
@@ -1761,3 +1852,33 @@ async function extractActivityInfo(
     ),
   };
 }
+
+/**
+ * A utility function to await a promise with a timeout.
+ *
+ * This function properly clean up the timer when the provided promise resolves or rejects.
```
Sushisource, line 1859:
```suggestion
 * This function properly cleans up the timer when the provided promise resolves or rejects.
```

packages/test/src/test-worker-lifecycle.ts
```diff
@@ -54,6 +75,59 @@ if (RUN_INTEGRATION_TESTS) {
...
+    const client = new Client();
+    const worker = await Worker.create({
+      ...defaultOptions,
+      taskQueue,
+    });
+
+    // This workflow will allocate large block of memory, hopefully causing a ERR_WORKER_OUT_OF_MEMORY.
+    // Note that due to the way Node/V8 optimize byte code, its possible that this may trigger
+    // other type of errors, including some that can't be intercepted cleanly.
+    client.workflow
```
Sushisource, line 94: Hmm... is there some kind of externally generated exception that would work that isn't this? Forcing an OOM is a bit annoying and the comments are already referencing some unreliability. Maybe some kind of forced segfault or something?

Sushisource: Or alternatively could we start a VM with a much smaller max allowed heap size?

Sushisource (approved):
Nothing blocking, a few suggestions.

packages/worker/src/worker.ts
```diff
@@ -1550,49 +1586,83 @@ export class Worker {
...
+    if (innerResult === undefined && promiseCompletionTimeoutMs > 0) {
+      await timeoutPromise(innerPromise, promiseCompletionTimeoutMs);
+    }
+    if (innerResult === undefined)
+      innerResult = {
+        status: 'rejected',
+        reason: new PromiseCompletionTimeoutError(
+          `Promise did not resolve within ${promiseCompletionTimeoutMs}ms after Worker completed shutdown`
+        ),
+      };
```
antlai-temporal, line 1647: Do you want to use {} with the if to clarify?

packages/worker/src/worker.ts
```diff
@@ -1550,49 +1586,83 @@ export class Worker {
...
+
+    if (workerError === undefined) {
+      if (innerResult.status === 'fulfilled') {
+        return innerResult.value;
+      } else {
+        throw innerResult.reason;
+      }
```
antlai-temporal, line 1654: Can `innerResult.reason` be missing? i.e., innerResult not undefined but the field is...

packages/worker/src/worker.ts
```diff
@@ -1630,6 +1710,16 @@ export class Worker {
               })
             )
           ).pipe(
+            // Reinject fatalError inside the stream, if any,
+            // with precedence over the error from the pipe
```
antlai-temporal, line 1714: A bit more info on the "why" could be useful here...
RxJS is just complicating things here...

packages/worker/src/worker.ts
```diff
@@ -1644,13 +1734,11 @@ export class Worker {
         );
       } finally {
         Runtime.instance().deregisterShutdownSignalCallback(shutdownCallback);
+
+        await this.nativeWorker.finalizeShutdown();
       }
```
antlai-temporal, line 1739: Can Rust hang here with corrupted state?

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -74,21 +79,25 @@ export class WorkerThreadClient {
 
       completion.resolve(result.output);
     });
-    workerThread.on('exit', () => {
+    workerThread.on('error', (err) => {
+      logger.error(`Workflow Worker Thread failed: ${err}`, err);
```
antlai-temporal, line 83: Can this fire multiple times, or it is just once? If the former you may want to lock the first exitError...

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -74,21 +79,25 @@ export class WorkerThreadClient {
...
+      this.exitError = new UnexpectedError(`Workflow Worker Thread exited prematurely: ${err}`, err);
+      // Node will automatically terminate the Worker Thread, immediately after this event.
+    });
+    workerThread.on('exit', (exitCode) => {
+      logger.trace(`Workflow Worker Thread exited with code ${exitCode}`, { exitError: this.exitError });
       this.workerExited = true;
```
antlai-temporal, line 89: Is "exit" guaranteed to be called after "error"? Do we know this.exitError will be set by the error handler first?

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -255,6 +268,10 @@ export class VMWorkflowThreadProxy implements Workflow {
    * Proxy request to the VMWorkflow instance
    */
   async dispose(): Promise<void> {
-    await this.workerThreadClient.send({ type: 'dispose-workflow', runId: this.runId });
+    try {
+      await this.workerThreadClient.send({ type: 'dispose-workflow', runId: this.runId });
+    } catch (e) {
+      // Ignore errors when disposing
+    }
```
antlai-temporal, line 275: Do you want to log.trace?

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -28,12 +29,15 @@ export function setUnhandledRejectionHandler(getWorkflowByRunId: (runId: string)
         return;
       }
     }
-    // The user's logger is not accessible in this thread,
-    // dump the error information to stderr and abort.
-    console.error('Unhandled rejection', { runId }, err);
-    process.exit(1);
+
+    console.error('An Unhandled Promise rejection could not be associated to a Workflow Run', { runId, error: err });
+    throw new UnhandledRejectionError(
```
antlai-temporal, line 34: This error message is a bit cryptic, it could help say why this happens and how to fix it, or point to docs...

antlai-temporal (approved):

packages/test/src/test-worker-lifecycle.ts
mjameswh: We also have a test for Uncaught Promise Rejection elsewhere.

But I absolutely want to cover the ERR_WORKER_OUT_OF_MEMORY case, even if I can't guarantee reproduction 100% of the time. It is an error that our users see in practice, some more than once a day, and until now, might leaving the Worker in a kind of zombie state.

> Or alternatively could we start a VM with a much smaller max allowed heap size?

I did at some point, and was considering exposing new worker settings for these limits. But surprisingly, that actually turned out to be no more reliable than the present test, and I decided not to expose those settings, or at least not at this time, as figuring out appropriate values turned out to be significantly harder than what it would seem.

And by the way, with the current settings, the ERR_WORKER_OUT_OF_MEMORY error happens > 95% of the time.

packages/test/src/test-worker-lifecycle.ts
```diff
@@ -54,6 +75,59 @@ if (RUN_INTEGRATION_TESTS) {
...
+            'Worker terminated due to reaching memory limit: JS heap out of memory',
+        }),
+      ]);
+      if (res !== false) {
+        t.is(worker.getState(), 'FAILED');
+      } else {
+        if (worker.getState() === 'RUNNING') {
+          worker.shutdown();
+          await workerRunPromise;
+          t.log('Non concluent result: Worker did not fail as expected');
```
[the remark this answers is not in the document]
mjameswh: Pay attention: no.

But I still want a way of grepping at test output to determine if the OOM case was replicated or not.

packages/test/src/test-workflow-unhandled-rejection-crash.ts
```diff
@@ -17,13 +17,17 @@ if (RUN_INTEGRATION_TESTS) {
...
-        'Worker thread shut down prematurely, this could be caused by an' +
-        ' unhandled rejection in workflow code that could not be' +
-        ' associated with a workflow run',
-    });
-    await handle.terminate();
+    try {
+      await t.throwsAsync(worker.run(), {
+        instanceOf: UnexpectedError,
+        message:
+          'Workflow Worker Thread exited prematurely: UnhandledRejectionError: ' +
```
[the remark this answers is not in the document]
mjameswh: I would say that's quite intentional and reads clearly to me, but if you have suggestions, go ahead.

For context: "Worker Thread" is the name of the Node.js' threading API, so just saying "Worker Thread exited prematurely" would be overly generic. Also, only Workflow Tasks get run in Worker Threads, because that's part of the sandboxing mechanism. So being specific on this specific term feels appropriate to me. I want this to be understood as "a thread used by your Temporal Worker for running Workflows".

packages/test/src/workflows/fill-memory.ts
```diff
@@ -0,0 +1,19 @@
+export async function fillMemory(): Promise<number> {
+  // It looks like JIT compilation of the following code affects the way the out of memory error
+  // will get reported. That is, once the code has been optimized, the out of memory condition is
+  // very likely to trigger a low level (native-like) error, which may terminate the whole process,
+  // rather than just the Worker Thread.
+  //
+  // Wrapping code in an eval block delays JIT compilation, thus reducing the probability of flakes.
+
+  // Allocate 32GB of memory
```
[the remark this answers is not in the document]
mjameswh: I removed that comment, it is misleading. How much we allocate is not really important.

There are different ways that Node may fail due to an "out of memory" condition, and the criterias for each of them are totally undocumented. This test is about triggering a very specific type of OOM, the `ERR_WORKER_OUT_OF_MEMORY` error, which seems to be related to specific heap spaces (probably the young gen) and/or to JIT compilation.

This code has been resulting in a ERR_WORKER_OUT_OF_MEMORY error in all setups I have tried; that includes all platforms, for Node 16, 18 and 20, with various quantity of RAM, more than 95% of the time. The error happens almost immediately, way before reaching the order of GBs. In fact, if I add a `console.log`, it appears that the process crashes on the first execution of the loop!

Yet, if I remove the loop, or reduce the higher limit on the loop (threshold seems to be at ~64), then it no longer fails, but execution becomes much slower. Similarly, if I change the code to allocate larger arrays, then it tends to fail early by throwing an error (rather than killing the process). And if I add an `await wf.sleep();` in the loop, then it can allocate much more memory without crashing.

All of this suggests interractions of V8's JIT compiler and/or GC memory allocator. The higher limit on the loop probably triggers an early JIT optimization (which would not be deemed useful if the loop is expected to run less than a certain number of time). Similarly, when allocating larger arrays, it's possible that the heap allocator decides to allocate directly from the perm gen space (which would have a significantly higher capacity) rather than from the young gen (which has a default capacity of 16 MB according to Node's documentation).

Now, given all of those uncertainties, this test is likely to be fragile. Why then do I care that much about reproducing this very specific error pattern? For some unknown reason, that's an error pattern that users are observing regularly in practice, while I have seen no report of other types of OOM errors in a long time (except for docker/pods memory limit misconfiguration, but there's nothing more that the SDK can do about those).

packages/worker/src/worker.ts
mjameswh: It can't. The type mandates that `innerResult.reason` be defined if status is `rejected` (which is implied by `status !== 'fulfilled'`).

packages/test/src/test-worker-lifecycle.ts
Sushisource: Works for me

packages/worker/src/workflow/threaded-vm.ts
mjameswh: That would only be noise IMHO.

packages/worker/src/errors.ts
```diff
@@ -8,6 +8,63 @@ import { ShutdownError, TransportError, UnexpectedError } from '@temporalio/core
...
+  }
+}
+
+/**
+ * Error thrown by {@link Worker.runUntil} if the provided Promise does not resolve within the specified
+ * {@link RunUntilOptions.promiseCompletionTimeout|timeout period} after the Worker has stopped.
+ */
+@SymbolBasedInstanceOfError('PromiseCompletionTimeoutError')
+export class PromiseCompletionTimeoutError extends Error {}
+
```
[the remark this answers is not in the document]
mjameswh: That's an interesting observation, though I don't think there would be a need for both a generic "promise didn't complete in time" and a worker-specific variant.

That is, let's not change the name; if we ever see a need for this, we may simply _reuse_ that error in a different context, by changing its api doc to not be worker-specific.

packages/worker/src/worker.ts
```diff
@@ -1483,7 +1512,14 @@ export class Worker {
...
       mergeMap(this.handleWorkflowActivations.bind(this)),
       mergeMap(async (completion) => {
-        await this.nativeWorker.completeWorkflowActivation(completion.buffer.slice(completion.byteOffset));
+        try {
+          await this.nativeWorker.completeWorkflowActivation(completion.buffer.slice(completion.byteOffset));
+        } catch (error) {
+          this.logger.error('Core reported failure in completeWorkflowActivation(). Initiating Worker shutdown.', {
+            error,
+          });
+          this.unexpectedErrorSubject.error(error);
```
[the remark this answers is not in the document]
mjameswh: I'll pass for now, but may reconsider for a subsequent PR.

packages/worker/src/worker.ts
mjameswh: I think there's a very slight possibly that it may happen, but in most "corrupted internal state" situations, Core will just panic, which will terminate the process anyway.

On the opposite, there's a non-negligeable probability that not calling `finalizeShutdown()` might result in difficulties shuting down the process.

packages/worker/src/workflow/threaded-vm.ts
mjameswh: Node.js' doc is not clear on guarantees here, but it is quite clear from their code that no other event would be received after `exit`, so if we are receive some `error`, that will be before `exit`.

Also, our code kind of deal gracefully with the case where we'd would have unexpectedly received an `exit` event without an `error` first, by just making this an `UnexpectedError` (see below).

packages/worker/src/workflow/threaded-vm.ts
mjameswh: Node.js' doc is not clear on guarantees here, but I don't think we could possibly receive multiple `error` events. And assuming we would, that would likely indicate that prior errors were not fatal, and consequently, we might really be interested only in the _latest_ error, not the first.

packages/worker/src/workflow/vm-shared.ts
mjameswh: That's a totally unexpected case. If a user ever see this, that's very likely a bug in the SDK. There's nothing more to document.

PR #1548, opened 2024-10-18
feat(client): Add utility functions isGrpcDeadlineError and isGrpcCancelledError

mjameswh:
## What was changed

- Added utility functions `isGrpcDeadlineError` and `isGrpcCancelledError` .

## Why?

- Setting a deadline on gRPC calls (including all "high level" client calls made using our WorkflowClient) is a best practice, as not doing so may possibly result in waiting forever for a response that will never come. Yet, doing so implies that the user will need to check thrown errors, which turns this into a non-trivial exercise.

See for example the following snippet provided by a user. Note that they needed to add an explicit dependency on `@grpc/grpc-js` only for this purpose, as they need access to the `status` enum.

```
import { status as grpcStatus } from "@grpc/grpc-js";
import { isGrpcServiceError, ServiceError, WorkflowClient } from "@temporalio/client";

async function isWorkflowCompleted(workflowClient: WorkflowClient, workflowId: string, ms: number): Promise<boolean> {
    const handle = workflowClient.getHandle(workflowId);
    try {
      await workflowClient.withDeadline(Date.now() + ms, () => handle.result());
      return true;
    } catch (e) {
      if (
        e instanceof ServiceError &&
        isGrpcServiceError(e.cause) &&
        e.cause.code === grpcStatus.DEADLINE_EXCEEDED
      ) {
        return false;
      }
      throw e;
    }
}
```

After this PR, this snippet can be reduced to:

```
import { isGrpcDeadlineError, WorkflowClient } from "@temporalio/client";

async function isWorkflowCompleted(workflowClient: WorkflowClient, workflowId: string, ms: number): Promise<boolean> {
    const handle = workflowClient.getHandle(workflowId);
    try {
      await workflowClient.withDeadline(Date.now() + ms, () => handle.result());
      return true;
    } catch (e) {
      if (isGrpcDeadlineError(e)) return false;
      throw e;
    }
}
```

… and the user probably no longer needs an explicit dependency on `@grpc/grpc-js`.

### DIFF
--- a/packages/client/src/errors.ts
+++ b/packages/client/src/errors.ts
@@ -1,4 +1,4 @@
-import { ServiceError as GrpcServiceError } from '@grpc/grpc-js';
+import { ServiceError as GrpcServiceError, status } from '@grpc/grpc-js';
 import { RetryState, TemporalFailure } from '@temporalio/common';
 import { isError, isRecord, SymbolBasedInstanceOfError } from '@temporalio/common/lib/type-helpers';
 
@@ -79,12 +79,29 @@ export class WorkflowContinuedAsNewError extends Error {
   }
 }
 
+/**
+ * Returns true if the provided error is a {@link GrpcServiceError}.
+ */
 export function isGrpcServiceError(err: unknown): err is GrpcServiceError {
   return (
     isError(err) &&
     typeof (err as GrpcServiceError)?.details === 'string' &&
     isRecord((err as GrpcServiceError).metadata)
   );
+}
+
+/**
+ * Returns true if the provided error is a {@link GrpcServiceError} with code DEADLINE_EXCEEDED.
+ */
+export function isGrpcDeadlineError(err: unknown): err is GrpcServiceError & { code: status.DEADLINE_EXCEEDED } {
+  return isGrpcServiceError(err) && (err as GrpcServiceError).code === status.DEADLINE_EXCEEDED;
+}
+
+/**
+ * Returns true if the provided error is a {@link GrpcServiceError} with code CANCELLED.
+ */
+export function isGrpcCancelledError(err: unknown): err is GrpcServiceError & { code: status.CANCELLED } {
+  return isGrpcServiceError(err) && (err as GrpcServiceError).code === status.CANCELLED;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/client/src/errors.ts
```diff
@@ -87,6 +90,36 @@ export function isGrpcServiceError(err: unknown): err is GrpcServiceError {
...
+export function isGrpcCancelledError(err: unknown): err is Error {
+  while (isError(err)) {
+    if (isGrpcServiceError(err) && (err as GrpcServiceError).code === status.CANCELLED) {
+      return true;
+    }
+    err = (err as any).cause;
+  }
+  return false;
+}
+
```
antlai-temporal, line 122: The logic is exactly the same for both, it just changes the constant, do you think is worthwhile to factor out that logic in a separate internal method that takes the constant as argument?

packages/client/src/errors.ts
```diff
@@ -87,6 +90,36 @@ export function isGrpcServiceError(err: unknown): err is GrpcServiceError {
...
+ *
+ * @see {@link Connection.withDeadline}
+ */
+export function isGrpcDeadlineError(err: unknown): err is Error {
+  while (isError(err)) {
+    if (isGrpcServiceError(err) && (err as GrpcServiceError).code === status.DEADLINE_EXCEEDED) {
+      return true;
+    }
+    err = (err as any).cause;
+  }
```
antlai-temporal, line 104: There is no error type that includes the "cause" field? Is that the reason you cast to `any`?

antlai-temporal (approved):

packages/client/src/errors.ts
mjameswh: Yeah… But not yet.

The length and complexity of the duplicated portion is very low, whereas the shared internal function we'd introduce to dedup this (something like `hasGrpcErrorAncestor(code)`? or maybe `hasErrorAncestor(predicateFn)`?) would add an unnatural abstraction with very limited potential of reuse afterward. I don't think that's a win at this point. Let's delay.

packages/client/src/errors.ts
mjameswh: > There is no error type that includes the "cause" field?

That's it. Or to be very exact, there is no _generic_ error that includes it. The `cause` field in JS is a wide-spread convention, but the TS standard lib definition doesn't formally include it. Now, I could cast the error to `TemporalFailure` just to get the field, but I think in this case that would be even worse than casting to `any`.

PR #1553, opened 2024-10-25
User-implementable SlotSupplier

Sushisource:
## What was changed
Add user-implementable slot suppliers

## Why?
Parity. Cool feature.

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?
Yes, will be updating docs pages after all Core langs are done

### DIFF
--- a/packages/core-bridge/Cargo.toml
+++ b/packages/core-bridge/Cargo.toml
@@ -20,7 +20,7 @@ incremental = false
 async-trait = "0.1.83"
 futures = { version = "0.3", features = ["executor"] }
 log = "0.4"
-neon = { version = "1.0.0", default-features = false, features = ["napi-6"] }
+neon = { version = "1.0.0", default-features = false, features = ["napi-6", "futures"] }
 opentelemetry = "0.24"
 parking_lot = "0.12"
 prost = "0.13"
--- a/packages/core-bridge/Cargo.toml
+++ b/packages/core-bridge/Cargo.toml
@@ -28,8 +28,6 @@ prost-types = "0.13"
 serde_json = "1.0"
 tokio = "1.13"
 once_cell = "1.19"
-temporal-sdk-core = { version = "*", path = "../../../sdk-core/core", features = ["ephemeral-server"] }
-temporal-client = { version = "*", path = "../../../sdk-core/client" }
-#temporal-sdk-core = { version = "*", path = "./sdk-core/core", features = ["ephemeral-server"] }
-#temporal-client = { version = "*", path = "./sdk-core/client" }
+temporal-sdk-core = { version = "*", path = "./sdk-core/core", features = ["ephemeral-server"] }
+temporal-client = { version = "*", path = "./sdk-core/client" }
 tokio-stream = "0.1"
--- a/packages/core-bridge/Cargo.toml
+++ b/packages/core-bridge/Cargo.toml
@@ -17,6 +17,7 @@ lto = true
 incremental = false
 
 [dependencies]
+async-trait = "0.1.83"
 futures = { version = "0.3", features = ["executor"] }
 log = "0.4"
 neon = { version = "1.0.0", default-features = false, features = ["napi-6"] }
@@ -27,6 +28,8 @@ prost-types = "0.13"
 serde_json = "1.0"
 tokio = "1.13"
 once_cell = "1.19"
-temporal-sdk-core = { version = "*", path = "./sdk-core/core", features = ["ephemeral-server"] }
-temporal-client = { version = "*", path = "./sdk-core/client" }
+temporal-sdk-core = { version = "*", path = "../../../sdk-core/core", features = ["ephemeral-server"] }
+temporal-client = { version = "*", path = "../../../sdk-core/client" }
+#temporal-sdk-core = { version = "*", path = "./sdk-core/core", features = ["ephemeral-server"] }
+#temporal-client = { version = "*", path = "./sdk-core/client" }
 tokio-stream = "0.1"
--- a/packages/core-bridge/src/conversions.rs
+++ b/packages/core-bridge/src/conversions.rs
@@ -1,5 +1,5 @@
 use crate::helpers::*;
-use log::error;
+use log::{error, warn};
 use neon::{
     context::Context,
     handle::Handle,
@@ -7,6 +7,7 @@ use neon::{
     types::{JsBoolean, JsNumber, JsString},
 };
 use prost::Message;
+use std::cell::RefCell;
 use std::marker::PhantomData;
 use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
 use temporal_client::HttpConnectProxyOptions;
@@ -31,6 +32,7 @@ use temporal_sdk_core::{
     ResourceBasedSlotsOptionsBuilder, ResourceSlotOptions, RetryConfig, SlotSupplierOptions,
     TlsConfig, TunerHolderOptionsBuilder, Url,
 };
+use tokio::sync::oneshot;
 
 pub enum EphemeralServerConfig {
     TestServer(TestServerConfig),
@@ -659,6 +661,27 @@ struct BridgePermitData {
     permit: Arc<Root<JsObject>>,
 }
 
+struct CallAbortOnDrop {
+    chan: Channel,
+    aborter: oneshot::Receiver<Root<JsFunction>>,
+}
+
+static ABORT_STR: &str = "__aborted by core__";
+
+impl Drop for CallAbortOnDrop {
+    fn drop(&mut self) {
+        if let Ok(aborter) = self.aborter.try_recv() {
+            let _ = self.chan.try_send(move |mut cx| {
+                let cb = aborter.to_inner(&mut cx);
+                let this = cx.undefined();
+                let abort_str = cx.string(ABORT_STR);
+                let _ = cb.call(&mut cx, this, [abort_str.upcast()]);
+                Ok(())
+            });
+        }
+    }
+}
+
 #[async_trait::async_trait]
 impl<SK: SlotKind + Send + Sync> SlotSupplier for SlotSupplierBridge<SK> {
     type SlotKind = SK;
@@ -672,7 +695,7 @@ impl<SK: SlotKind + Send + Sync> SlotSupplier for SlotSupplierBridge<SK> {
             let worker_build_id = ctx.worker_build_id().to_string();
             let is_sticky = ctx.is_sticky();
 
-            let callback_fut = self
+            let (callback_fut, _abort_on_drop) = match self
                 .channel
                 .send(move |mut cx| {
                     let context = Self::mk_reserve_ctx(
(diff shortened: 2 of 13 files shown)


packages/worker/src/worker-tuner.ts
```diff
@@ -73,15 +87,127 @@ export interface ResourceBasedTuner {
...
+
+export interface SlotPermit {
+  // Use this field to associate any data you like with a particular slot
+  data?: any;
+}
+
+export interface SlotReserveContext {
+  // The name of the task queue for which this reservation request is associated
+  taskQueue: string;
+  // The build id of the worker that is requesting the reservation
```
mjameswh, line 161: Docs for `workerIdentity` and `workerBuildId` are reversed.

Also, make all of these `/** ... */` comments, so they would be visible in API doc.

packages/worker/src/worker-tuner.ts
```diff
@@ -73,15 +87,127 @@ export interface ResourceBasedTuner {
...
+   */
+  releaseSlot(slot: SlotReleaseContext<SI>): void;
+}
+
+export interface SlotPermit {
+  // Use this field to associate any data you like with a particular slot
+  data?: any;
+}
+
+export interface SlotReserveContext {
```
mjameswh, line 158: No field for worker type? e.g. workflow, activity local activity

packages/worker/src/worker-tuner.ts
```diff
@@ -73,15 +87,127 @@ export interface ResourceBasedTuner {
...
+   * finished, whether successfully or not, or because the slot was no longer needed (ex: the number
+   * of active pollers decreased). This call should be non-blocking.
+   *
+   * @param ctx The context for releasing a slot.
+   */
+  releaseSlot(slot: SlotReleaseContext<SI>): void;
+}
+
+export interface SlotPermit {
+  // Use this field to associate any data you like with a particular slot
```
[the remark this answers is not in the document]
Sushisource: Ah, good point. I just get so stuck in "types actually mean something" instead of "anything goes in JS" lol

packages/worker/src/worker-tuner.ts
```diff
@@ -73,15 +87,127 @@ export interface ResourceBasedTuner {
...
+  taskQueue: string;
+  // The build id of the worker that is requesting the reservation
+  workerIdentity: string;
+  // The identity of the worker that is requesting the reservation
+  workerBuildId: string;
+  // True iff this is a reservation for a sticky poll for a workflow task
+  isSticky: boolean;
+
+  // Returns the number of currently outstanding slot permits, whether used or un-used.
+  numIssuedSlots(): number;
```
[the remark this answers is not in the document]
Sushisource: It's because this value could change over the course of a call to reserve a slot if that call is awaiting on other stuff.

packages/worker/src/worker-tuner.ts
Sushisource: Thanks. I thought // still counted for API docs. Will fix

packages/worker/src/worker-tuner.ts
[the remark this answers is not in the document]
mjameswh: But doesn't the provider itself already know about the current number of slots it allowed? Why do we even need Core to provide that information?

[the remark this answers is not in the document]
Sushisource: Oh - yes. It could technically know that. Just in practice this is probably something you pretty commonly want and it's nice to have handled for you. That said I'm fine w/ leaving it off initially and adding it in if requested.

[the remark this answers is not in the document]
mjameswh: I'd rather leave it out for now, and see while progressing either that's useful.

I foresee some potential synchronization issues that users may struggle with; I can think this more in depth if needed, but my intuition is that anyway, this function would rarely be pertinent in anything more than a trivial sample `SlotSupplier`.

Sushisource:
Requires https://github.com/temporalio/sdk-typescript/pull/1553 to merge before it can compile

packages/core-bridge/ts/worker-tuner.ts
```diff
@@ -69,3 +87,102 @@ export interface FixedSizeSlotSupplier {
...
+  /**
+   * This function is called when trying to reserve slots for "eager" workflow and activity tasks.
+   * Eager tasks are those which are returned as a result of completing a workflow task, rather than
+   * from polling. Your implementation must not block, and If a slot is available, return a permit
+   * to use that slot.
+   *
+   * @param ctx The context for slot reservation.
+   * @returns Maybe a permit to use the slot which may be populated with your own data.
+   */
+  tryReserveSlot(ctx: SlotReserveContext): SlotPermit | undefined;
```
cretz, line 120: I think in the JS world it's a bit clearer to ask a user to explicitly return null instead of undefined if it's by intention like in this case.

packages/core-bridge/ts/worker-tuner.ts
```diff
@@ -69,3 +87,102 @@ export interface FixedSizeSlotSupplier {
   // The maximum number of slots that can be issued
   numSlots: number;
 }
+
+/**
+ * The interface can be implemented to provide custom slot supplier behavior.
+ */
+export interface CustomSlotSupplier<SI extends SlotInfo> {
```
cretz, line 94: I noticed that a "maximum slots" getter isn't here. I think that's a good thing, just noticing the mismatch with Go/Java.

cretz (approved):
LGTM. Besides the fact that these will be called `CustomSlotSuppliers` and Go and Java get to just use `SlotSupplier`, I like that everything else is basically the same.

packages/core-bridge/Cargo.toml
```diff
@@ -17,16 +17,19 @@ lto = true
...
 prost = "0.13"
 prost-types = "0.13"
 serde_json = "1.0"
 tokio = "1.13"
 once_cell = "1.19"
-temporal-sdk-core = { version = "*", path = "./sdk-core/core", features = ["ephemeral-server"] }
-temporal-client = { version = "*", path = "./sdk-core/client" }
+temporal-sdk-core = { version = "*", path = "../../../sdk-core/core", features = ["ephemeral-server"] }
+temporal-client = { version = "*", path = "../../../sdk-core/client" }
+#temporal-sdk-core = { version = "*", path = "./sdk-core/core", features = ["ephemeral-server"] }
```
mjameswh, line 33: Leftover? lines 33 and 34

packages/core-bridge/src/conversions.rs
```diff
@@ -603,7 +612,285 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
...
+                    channel: cx.channel(),
+                    _kind: PhantomData,
+                };
+                Ok(SlotSupplierOptions::Custom(Arc::new(ssb)))
+            }
             _ => cx.throw_type_error("Invalid slot supplier type"),
         }
     }
 }
+
```
mjameswh, line 649: Everything from this point on should be in a different file.

packages/core-bridge/src/runtime.rs
```diff
@@ -111,6 +111,14 @@ pub enum RuntimeRequest {
         key: String,
         callback: Root<JsFunction>,
     },
+    // GenericCallback {
```
mjameswh, line 114: leftover?

packages/core-bridge/ts/worker-tuner.ts
```diff
@@ -69,3 +87,102 @@ export interface FixedSizeSlotSupplier {
...
+   *
+   * @param ctx The context for slot reservation.
+   * @param abortSignal The SDK may decide to abort the reservation request if it's no longer
+   *   needed. Implementations may clean up and then must reject the promise with AbortError.
+   * @returns A permit to use the slot which may be populated with your own data.
+   */
+  reserveSlot(ctx: SlotReserveContext, abortSignal: AbortSignal): Promise<SlotPermit>;
+
+  /**
+   * This function is called when trying to reserve slots for "eager" workflow and activity tasks.
```
mjameswh, line 112: There's no case of eager workflow task in TS at this point.

packages/worker/src/worker-tuner.ts
```diff
@@ -33,23 +44,41 @@ type ResourceBasedSlotsForType = ResourceBasedSlotOptions & {
...
+  /**
+   * Amount of slots that will be issued regardless of any other checks.
+   * Defaults to 2 for workflow tasks and 1 for activity tasks.
+   */
   minimumSlots?: number;
-  // Maximum amount of slots permitted
-  // Defaults to 1000 for workflow tasks and 2000 for activity tasks.
+  /**
+   * Maximum amount of slots permitted
+   * Defaults to 1000 for workflow tasks and 2000 for activity tasks.
```
mjameswh, line 75: I know this is not new from this PR, but reading this again makes me think… Isn't 1000/2000 ridiculously high as defaults? Or will this effectively get capped by some other limit anyway (I mean, something else than only resource usage).

IIRC, the default namespace config in Cloud is to limit the _total_ number of polers to 2000. Could we get in a situation where the RB provider would allow that many slots in _polling state_ at the same time, from the same worker?

Should we have a distinct limit for the number of slots that have been allowed but haven't yet been "marked as in-use"?

packages/core-bridge/Cargo.toml
Sushisource: I've left it here on purpose until https://github.com/temporalio/sdk-core/pull/838 is merged

packages/core-bridge/ts/worker-tuner.ts
Sushisource: I can never keep this straight. It's almost like having two kinds of nothing is a bad idea.

packages/core-bridge/ts/worker-tuner.ts
Sushisource: I think I'll leave it as is for when we maybe reintroduce

packages/worker/src/worker-tuner.ts
Sushisource: Yeah, it is, but the idea is it's supposed to be capped by the resource tuner long before it gets to that point. So, not something else, but that.

Good question about the polling, but right now no way, and when I do poller tuning I'll have to consider that of course. Probably there is also just a different limit on max polling ceiling which will in effect restrict this (or this will be bound to that)

packages/worker/src/worker-tuner.ts
```diff
@@ -141,11 +167,60 @@ const isResourceBasedTuner = (tuner: WorkerTuner): tuner is ResourceBasedTuner =
...
+
+  constructor(
+    private readonly supplier: CustomSlotSupplier<SI>,
+    private readonly logger: Logger
+  ) {}
+
+  async reserveSlot(ctx: SlotReserveContext, registerAbort: any): Promise<SlotPermit> {
+    const abortController = new AbortController();
+    registerAbort(() => abortController.abort(abortString));
+    return await this.supplier.reserveSlot(ctx, abortController.signal).catch((err) => {
```
mjameswh, line 188: How bad will it be back in Rust world if the user function actually returned undefined or null rather than a real object? TS types are unfortunately not guarantees at runtime.

We usually don't get paranoid with these things in TS SDK, but since the contract is kind of inverted here, I'm wondering if we may need to be a tiny bit more careful.

Sushisource: It shouldn't matter. Any returned value counts as a "permit" and gets used as the data for it. However I will double check that with a test. It's worth calling that out in the doc string too probably.

mjameswh: If the rust-side don't ever look into the returned object, there me be absolutely problem here.

Sushisource: Thank you! Fixed

mjameswh (approved):
Awesome!

PR #1559, opened 2024-11-07
Added limit to number of 'available activities' list when calling an unregistered activity

KaiStevenson:
## What was changed

When emitting an 'Activity function xyz is not registered on this Worker' error, the number of activities listed in 'available activities are:' has been capped to a maximum of 10. It will be of the format 'Activity function xyz is not registered on this Worker, available activities: activity1, activity2 ... activity10 and 37 others.'

## Why?

'Available activities are:' can be a useful hint as to which worker picked up the activity, and can immediately highlight problems if there are, in fact, _no_ registered activities! But for workers with hundreds or thousands of activities registered, these logs quickly become huge, taking up too much storage and memory.

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -917,11 +917,12 @@ export class Worker {
                     const { activityType } = info;
                     const fn = this.options.activities.get(activityType);
                     if (typeof fn !== 'function') {
+                      const keys = [...this.options.activities.keys()];
                       throw ApplicationFailure.create({
                         type: 'NotFoundError',
-                        message: `Activity function ${activityType} is not registered on this Worker, available activities: ${JSON.stringify(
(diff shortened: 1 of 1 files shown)


mjameswh:
Hey Kai! Sorry for delaying feedback on this PR. I was hesitant on either to merge it.

My concern is that cutting down the list of activities of defeats the purpose of returning that list, to start with. I mean, printing out the list of known activities helps users figuring out why the activity they are trying to schedule doesn't exist. Are they running the wrong version of the activity worker? Some spelling errors? Wrong task queue? Some form of misconfiguration? If the list is incomplete, then it can't serve that purpose, and should be omitted entirely.

At this point, given that the related memory leak issue has been resolved, I'd prefer leaving this error message as-is. We may reconsider this at a later point if this seems to be a recurring issue.

PR #1561, opened 2024-11-09
fix(worker): abort reason should be CancellationFailure

ikonst:
## What was changed

- The activity cancellation's abort signal now has a `CancellationFailure` object as its `reason` property. This is useful when checking the `reason` as well as being what's thrown when `throwIfAborted()` is called (currently it's a `DOMException` as per [the default behavior](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason)).
- Unit test was added.

## Why?

- Useful to know the Temporal reason (e.g. "TIMED_OUT").
- Easier to identify the abort as coming from Temporal.
- Same behavior whether you call `throwIfAborted()` or `await cancelled`.

### DIFF
--- a/packages/test/src/activities/index.ts
+++ b/packages/test/src/activities/index.ts
@@ -55,8 +55,16 @@ export async function throwAnError(useApplicationFailure: boolean, message: stri
   }
 }
 
-export async function waitForCancellation(): Promise<void> {
-  await Context.current().cancelled;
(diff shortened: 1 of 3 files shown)


ikonst:
@mjameswh can take a look?

mjameswh (approved):
That totally makes sense, indeed. Good catch, @ikonst!

PR #1563, opened 2024-11-14
fix(worker): Fix memory leak on non-existant activity

### LINKED ISSUE #1359: [Feature request] WorkerStatus does not have numInFlightLocalActivities

I want to determine current load of the worker. We have 3 types of task slots (workflow / activities / local activities) in the worker each with a concurrency limit. However, I have no way of finding out how many concurrent local activities are running.

We have a similar field in WorkerStatus -> numInFlightActivities

mjameswh:
## What was changed

- In case where an Activity Task would fail before the activity function was actually called, e.g. because there's no activity function registered for the activity name or because of a Payload Codec/Converter error, then the activity-specific RxJS Observer would never get closed, resulting in a memory leak.

- Also, added `numInFlightNonLocalActivities` and `numInFlightLocalActivities` in `WorkerStatus` (closes #1359).

## How was this tested

- Tested locally with a custom load testing environment specifically meant to trigger the original memory leak.

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -351,8 +351,20 @@ export interface WorkerStatus {
   numInFlightWorkflowActivations: number;
   /**
    * Number of in-flight (currently actively processed) Activities
+   *
+   * This includes both local and non-local Activities.
+   *
+   * See {@link numInFlightNonLocalActivities} and {@link numInFlightLocalActivities} for a breakdown.
    */
   numInFlightActivities: number;
+  /**
+   * Number of in-flight (currently actively processed) non-Local Activities
+   */
+  numInFlightNonLocalActivities: number;
+  /**
+   * Number of in-flight (currently actively processed) Local Activities
+   */
+  numInFlightLocalActivities: number;
   /**
    * Number of Workflow executions cached in Worker memory
    */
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

packages/worker/src/worker.ts
```diff
@@ -1060,7 +1081,12 @@ export class Worker {
             return { taskToken, result };
           }),
           filter(<T>(result: T): result is Exclude<T, undefined> => result !== undefined),
-          map((rest) => coresdk.ActivityTaskCompletion.encodeDelimited(rest).finish())
+          map((rest) => coresdk.ActivityTaskCompletion.encodeDelimited(rest).finish()),
+          tap({
+            next: () => {
+              group$.close();
+            },
```
antlai-temporal, line 1088: A comment about why it is not enough to close() in the finally{} would be useful here...

antlai-temporal (approved):

packages/worker/src/worker.ts
mjameswh: Simply because we may never go through that try-finally block... 😉

That might not be obvious by looking at the diff just now, but there are a few code paths above that throws or returns.

PR #1571, opened 2024-11-19
chore: Update Core to latest

mjameswh:
## What was changed

- Updated core to dd37e29

### DIFF
--- a/packages/client/src/async-completion-client.ts
+++ b/packages/client/src/async-completion-client.ts
@@ -1,6 +1,5 @@
 import { status as grpcStatus } from '@grpc/grpc-js';
 import { ensureTemporalFailure } from '@temporalio/common';
-import type { temporal } from '@temporalio/proto';
 import {
   encodeErrorToFailure,
   encodeToPayloads,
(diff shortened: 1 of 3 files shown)


packages/client/src/async-completion-client.ts
```diff
@@ -216,23 +216,23 @@ export class AsyncCompletionClient extends BaseClient {
     const payloads = await encodeToPayloads(this.dataConverter, details);
     let cancelRequested = false;
     try {
-      let response: temporal.api.workflowservice.v1.RecordActivityTaskHeartbeatResponse;
       if (taskTokenOrFullActivityId instanceof Uint8Array) {
-        response = await this.workflowService.recordActivityTaskHeartbeat({
+        const response = await this.workflowService.recordActivityTaskHeartbeat({
```
mjameswh, line 220: If anyone wonders, the `Response` interface for both of these calls was previously identical, so assigning both to a single `response` var was ok. But a new prop has been added to one of them.

antlai-temporal (approved):

PR #1573, opened 2024-11-21
feat(client): add `client.workflow.count` high level API

### LINKED ISSUE #1069: [Feature Request] Support countWorkflowExecutions

I'm trying to check if a workflow execution exists by using the workflowId.

I was checking the types from the TypeScript packages and I was trying to implement the below solution:
```ts
async checkWorkflowExecutionExists(workflowId: string): Promise<boolean> {
  const request = new temporal.api.workflowservice.v1.DescribeWorkflowExecutionRequest({
    execution: new temporal.api.common.v1.WorkflowExecution({ workflowId }),
    namespace: 'default',
  });
  const { count } = await this.client.workflowService.countWorkflowExecutions(request);
  return count.greaterThan(0);
}
```

### Additional context

But I'm getting this error:
```
│ [Nest] 40  - 03/02/2023, 10:15:15 PM   ERROR [ExceptionsHandler] proto_1.temporal.api.workflowservice.v1.D │
│ TypeError: proto_1.temporal.api.workflowservice.v1.DescribeWorkflowExecutionRequest is not a constructor   │
│     at CheckoutService.checkWorkflowExecutionExists (/baxus-services/dist/apps/api/main.js:2066:25)        │
│     at CheckoutController.checkWorkflowExecutionExists (/baxus-services/dist/apps/api/main.js:1889:37)     │
│     at /baxus-services/node_modules/@nestjs/core/router/router-execution-context.js:38:29                  │
│     at processTicksAndRejections (node:internal/process/task_queues:96:5)                                  │
│     at /baxus-services/node_modules/@nestjs/core/router/router-execution-context.js:46:28                  │
│     at /baxus-services/node_modules/@nestjs/core/router/router-proxy.js:9:17  
```

I was inspecting the code of the '@temporalio/proto' package but I don't see any reference to that `DescribeWorkflowExecutionRequest` constructor, so why the types are wrong?

Thanks for your support! <3

THardy98:
## What was changed
Added a `count` method to the workflow client, a higher-level, user-friendlier option to using the corresponding GRPC method directly.

1. Closes #1069

2. How was this tested:
Added an integration test.

3. Any docs updates needed?
Not sure

### DIFF
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -12,9 +12,6 @@ import { decodeMapFromPayloads } from '@temporalio/common/lib/internal-non-workf
 import { temporal, google } from '@temporalio/proto';
 import {
   CountWorkflowExecution,
-  CountWorkflowExecutionsAggregationGroup,
-  RawCountWorkflowExecutions,
-  RawCountWorkflowExecutionsAggregationGroup,
   RawWorkflowExecutionInfo,
   WorkflowExecutionInfo,
   WorkflowExecutionStatusName,
@@ -89,21 +86,19 @@ export async function executionInfoFromRaw<T>(
   };
 }
 
-export function decodeCountWorkflowExecutionsResponse(raw: RawCountWorkflowExecutions): CountWorkflowExecution {
+export function decodeCountWorkflowExecutionsResponse(
+  raw: temporal.api.workflowservice.v1.ICountWorkflowExecutionsResponse
+): CountWorkflowExecution {
   return {
     // Note: lossy conversion of Long to number
     count: raw.count!.toNumber(),
-    groups: raw.groups!.map((group) => executionCountAggregationGroupFromRaw(group)),
-  };
-}
-
-export function executionCountAggregationGroupFromRaw(
-  raw: RawCountWorkflowExecutionsAggregationGroup
-): CountWorkflowExecutionsAggregationGroup {
-  return {
-    // Note: lossy conversion of Long to number
-    count: raw.count!.toNumber(),
-    groupValues: raw.groupValues!.map((group_value) => searchAttributePayloadConverter.fromPayload(group_value)),
+    groups: raw.groups!.map((group) => {
+      return {
+        // Note: lossy conversion of Long to number
+        count: group.count!.toNumber(),
+        groupValues: group.groupValues!.map((value) => searchAttributePayloadConverter.fromPayload(value)),
+      };
+    }),
   };
 }
 
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -10,7 +10,15 @@ import { Replace } from '@temporalio/common/lib/type-helpers';
 import { optionalTsToDate, requiredTsToDate } from '@temporalio/common/lib/time';
 import { decodeMapFromPayloads } from '@temporalio/common/lib/internal-non-workflow/codec-helpers';
 import { temporal, google } from '@temporalio/proto';
-import { RawWorkflowExecutionInfo, WorkflowExecutionInfo, WorkflowExecutionStatusName } from './types';
+import {
+  CountWorkflowExecution,
+  CountWorkflowExecutionsAggregationGroup,
+  RawCountWorkflowExecutions,
+  RawCountWorkflowExecutionsAggregationGroup,
+  RawWorkflowExecutionInfo,
+  WorkflowExecutionInfo,
+  WorkflowExecutionStatusName,
+} from './types';
 
 function workflowStatusCodeToName(code: temporal.api.enums.v1.WorkflowExecutionStatus): WorkflowExecutionStatusName {
   return workflowStatusCodeToNameInternal(code) ?? 'UNKNOWN';
@@ -78,6 +86,24 @@ export async function executionInfoFromRaw<T>(
         }
       : undefined,
     raw: rawDataToEmbed,
+  };
+}
+
+export function countWorkflowExecutionFromRaw(raw: RawCountWorkflowExecutions): CountWorkflowExecution {
+  return {
+    // Note: lossy conversion of Long to number
+    count: raw.count!.toNumber(),
+    groups: raw.groups!.map((group) => executionCountAggregationGroupFromRaw(group)),
+  };
+}
+
+export function executionCountAggregationGroupFromRaw(
+  raw: RawCountWorkflowExecutionsAggregationGroup
+): CountWorkflowExecutionsAggregationGroup {
+  return {
+    // Note: lossy conversion of Long to number
+    count: raw.count!.toNumber(),
+    group_values: raw.groupValues!.map((group_value) => searchAttributePayloadConverter.fromPayload(group_value)),
   };
 }
 
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -89,7 +89,7 @@ export async function executionInfoFromRaw<T>(
   };
 }
 
-export function countWorkflowExecutionFromRaw(raw: RawCountWorkflowExecutions): CountWorkflowExecution {
+export function decodeCountWorkflowExecutionsResponse(raw: RawCountWorkflowExecutions): CountWorkflowExecution {
   return {
     // Note: lossy conversion of Long to number
     count: raw.count!.toNumber(),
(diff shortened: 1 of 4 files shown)


packages/test/src/test-integration-workflows.ts
```diff
@@ -1264,3 +1264,37 @@ export const interceptors: workflow.WorkflowInterceptorsFactory = () => {
...
+export async function completableWorkflow(completes: boolean): Promise<void> {
+  await workflow.condition(() => completes);
+}
+
+test('Count workflow executions', async (t) => {
+  const { taskQueue, createWorker, executeWorkflow, startWorkflow } = helpers(t);
+  const worker = await createWorker();
+  const client = t.context.env.client;
+
+  await worker.runUntil(async () => {
```
mjameswh, line 1277: In this specific case, it would be easy to design this test to not even need to have an actual Workflow function and to start a Workflow Worker.

Instead of passing a Workflow function to `startWorkflow`, you may provide a raw string (when passing a workflow function, the SDK will internally take the name of that function). Then instead of (or in addition to?) grouping by `ExecutionStatus`, group by `WorkflowType`.

That will make the test slightly faster. Not a huge difference, but those small things eventually cumulate to something significant.

packages/client/src/types.ts
```diff
@@ -52,6 +55,16 @@ export interface WorkflowExecutionInfo {
...
 }
 
+export interface CountWorkflowExecution {
+  count: number;
+  groups: CountWorkflowExecutionsAggregationGroup[];
+}
+
+export interface CountWorkflowExecutionsAggregationGroup {
+  count: number;
+  group_values: SearchAttributeValue[];
```
mjameswh, line 65:
```suggestion
  groupValues: SearchAttributeValue[];
```

packages/test/src/test-integration-workflows.ts
THardy98: Turns out we can only `GROUP BY` on `ExecutionStatus`, this is what I get when I try to `GROUP BY` on `WorkflowType`:
```
invalid query: operation is not supported: \'group by\' clause is only supported for ExecutionStatus search attribute'
```

Because of this, I'm going to keep the test case as is, its seems to be the only way to test that we get the groupings & group values we expect.

packages/test/src/test-integration-workflows.ts
```diff
@@ -1264,3 +1264,37 @@ export const interceptors: workflow.WorkflowInterceptorsFactory = () => {
...
+    // Run 3 workflows that complete.
+    for (let i = 0; i < 3; i++) {
+      await executeWorkflow(completableWorkflow, { args: [true] });
```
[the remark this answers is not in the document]
THardy98: Done.

packages/test/src/test-integration-workflows.ts
```diff
@@ -1,7 +1,8 @@
 import { randomUUID } from 'crypto';
 import { ExecutionContext } from 'ava';
 import { firstValueFrom, Subject } from 'rxjs';
-import { WorkflowFailedError } from '@temporalio/client';
+import asyncRetry from 'async-retry';
```
mjameswh, line 4: There are some lint warnings here. Make sure to execute `npm run lint` before pushing to GitHub.

packages/client/src/types.ts
```diff
@@ -14,6 +14,9 @@ export type GetWorkflowExecutionHistoryRequest =
 export type DescribeWorkflowExecutionResponse =
   proto.temporal.api.workflowservice.v1.IDescribeWorkflowExecutionResponse;
 export type RawWorkflowExecutionInfo = proto.temporal.api.workflow.v1.IWorkflowExecutionInfo;
+export type RawCountWorkflowExecutions = proto.temporal.api.workflowservice.v1.ICountWorkflowExecutionsResponse;
```
mjameswh, line 17: We have no need for a reexport of `RawCountWorkflowExecutions` and `RawCountWorkflowExecutionsAggregationGroup`

packages/client/src/helpers.ts
```diff
@@ -81,6 +89,24 @@ export async function executionInfoFromRaw<T>(
...
 
+export function decodeCountWorkflowExecutionsResponse(raw: RawCountWorkflowExecutions): CountWorkflowExecution {
+  return {
+    // Note: lossy conversion of Long to number
+    count: raw.count!.toNumber(),
+    groups: raw.groups!.map((group) => executionCountAggregationGroupFromRaw(group)),
+  };
+}
+
+export function executionCountAggregationGroupFromRaw(
```
mjameswh, line 100: No need to export that function. And anyway, I would inline that function into `decodeCountWorkflowExecutionsResponse`.

packages/client/src/helpers.ts
```diff
@@ -10,7 +10,15 @@ import { Replace } from '@temporalio/common/lib/type-helpers';
 import { optionalTsToDate, requiredTsToDate } from '@temporalio/common/lib/time';
 import { decodeMapFromPayloads } from '@temporalio/common/lib/internal-non-workflow/codec-helpers';
 import { temporal, google } from '@temporalio/proto';
-import { RawWorkflowExecutionInfo, WorkflowExecutionInfo, WorkflowExecutionStatusName } from './types';
+import {
+  CountWorkflowExecution,
+  CountWorkflowExecutionsAggregationGroup,
+  RawCountWorkflowExecutions,
```
mjameswh, line 16: Don't import `RawCountWorkflowExecutions*` from `./types.ts`. Instead, use the type as `temporal.api....CountWorkflowExecution*`.

packages/client/src/helpers.ts
```diff
@@ -81,6 +89,24 @@ export async function executionInfoFromRaw<T>(
...
+  };
+}
+
+export function executionCountAggregationGroupFromRaw(
+  raw: RawCountWorkflowExecutionsAggregationGroup
+): CountWorkflowExecutionsAggregationGroup {
+  return {
+    // Note: lossy conversion of Long to number
+    count: raw.count!.toNumber(),
+    groupValues: raw.groupValues!.map((group_value) => searchAttributePayloadConverter.fromPayload(group_value)),
```
mjameswh, line 106:
```suggestion
    groupValues: raw.groupValues!.map((value) => searchAttributePayloadConverter.fromPayload(value)),
```

packages/test/src/test-integration-workflows.ts
THardy98: I'd thought I had run this prior to pushing, does it not make code changes, just warnings?

packages/test/src/test-integration-workflows.ts
THardy98: I'd run this prior to pushing, does it not make code changes, just warnings?
In any case, I've addressed the warnings.

packages/client/src/types.ts
THardy98: Done.

packages/client/src/helpers.ts
THardy98: Done. I've inlined the AggregationGroup interface as well, we don't use it anymore.

packages/client/src/helpers.ts
THardy98: Done.

packages/client/src/helpers.ts
THardy98: Done.

packages/client/src/workflow-client.ts
```diff
@@ -1308,6 +1309,28 @@ export class WorkflowClient extends BaseClient {
     };
   }
 
+  /**
+   * Return workflow execution count by given `query`.
```
mjameswh, line 1313: I see you copied the comment from `list()` above, but it might not have been that great of an example to follow. Let's try to improve this doc a bit (and the one above, while we're at it). Maybe start with the suggestion below.

```suggestion
   * Return the number of Workflow Executions matching the given `query`. If no `query` is provided, then return the total number of Workflow Executions for this namespace.
```

We should also point out clearly that counts are approximates and eventually consistent.

packages/client/src/workflow-client.ts
```diff
@@ -1308,6 +1309,28 @@ export class WorkflowClient extends BaseClient {
...
+   *
+   * ⚠️ To use advanced query functionality, as of the 1.18 server release, you must use Elasticsearch based visibility.
```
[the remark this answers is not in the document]
THardy98: Removed, added Visibility API semantics to this as well

mjameswh (approved):
Great! Thanks a lot, @THardy98!

PR #1574, opened 2024-11-26
chore: Fix CI after GHA's drop of node16 actions

mjameswh:
## What changed

- Modified our GHA workflows so that when building the release-grade native artifacts on Linux targets, only the actual Rust compilation step runs inside a GLIBC 2.24-based Docker container; all other operations are performed directly on the runner, which can now be any recent Ubuntu.

  This was required because GHA no longer allows actions based on Node16 (i.e. actions are forced to use Node20 instead), but Node20 requires at least GLIBC 2.28. It is therefore not possible to have a single execution environment that can both execute Node20-based actions, yet produces native libraries that can then execute in the GLIBC 2.24-based environments that some of our users have.

- Given that the build-step-in-docker approach is much slower and less likely to benefit from the build cache, we only use that approach when building release-grade native artifacts. That is, we use a fast, non-docker and non-glibc-24-compatible build approach in 'Continuous Integration', but a longer, optimized and glibc-24 compatible build in 'Build Packages'.


.github/workflows/ci.yml
```diff
@@ -206,26 +183,29 @@ jobs:
...
   integration-tests:
+    timeout-minutes: 10
     needs:
-      - compile-native-binaries
+      - compile-native-binaries-debug
     strategy:
       fail-fast: false
       matrix:
-        node: [16, 18, 20]
+        node: [16, 18, 20, 22]
```
josh-berry, line 192: Are you planning to drop Node 16 separately?

Sushisource (approved):
I do wonder if we really need this combinatorial explosion of matrix options, but, if it's free :man_shrugging:

.github/workflows/release.yml
```diff
@@ -0,0 +1,505 @@
...
+      - name: Put native files into place
+        working-directory: ./packages/core-bridge/releases
+        run: |
+          mv tmp/corebridge-*/* ./
+          rm -rf tmp
+
+      - name: Install Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
josh-berry, line 180: Any reason not to use Node 22?

.github/workflows/release.yml
```diff
@@ -0,0 +1,505 @@
...
+
+  # Run integration tests. Uses the native binaries built in compile-native-binaries-release,
+  # but build `@temporalio/*` packages locally.
+  integration-tests:
+    needs:
+      - compile-native-binaries-release
+    strategy:
+      fail-fast: false
+      matrix:
+        node: [16, 18, 20]
```
josh-berry, line 227:
```suggestion
        node: [18, 20, 22]
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,505 @@
...
+          name: integration-tests-${{ matrix.platform }}-node${{ matrix.node }}-${{ matrix.server }}-${{ matrix.reuse-v8-context && 'reuse' || 'noreuse' }}-devserver-logs
+          path: /tmp/devserver.log
+
+  # Tests that npm init @temporalio results in a working worker and client
+  test-npm-init:
+    needs: build-packages
+    strategy:
+      fail-fast: false
+      matrix:
+        node: [16, 18, 20]
```
josh-berry, line 331:
```suggestion
        node: [18, 20, 22]
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,505 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Install Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
josh-berry, line 467: 22?

.github/workflows/ci.yml
josh-berry: nm, discussed offline & you told me this was intentional.

PR #1577, opened 2024-11-29
Propagate otel context when scheduling local activities

### LINKED ISSUE #1057: [Feature Request] Propagate opentelemetry context to local activities
Currently missing, only non-local activities and child workflows traces are attached to the workflow's span.

THardy98:
## What was changed
Added otel context when we schedule local activities

1. Closes #1057 

2. How was this tested:
Added to an existing integration test

3. Any docs updates needed?
Unsure


mjameswh (approved):
That all seems good. Thanks!

PR #1585, opened 2024-12-11
Update-with-start

dandavison:
This PR adds an API for "update-with-start", using the MultiOperation gRPC API.

The test suite is incomplete, but please feel free to review the API, implementation, and tests that are present.

The following code demonstrates the API and is taken from the [early-return sample](https://github.com/temporalio/samples-typescript/pull/401):

```typescript
  const startWorkflowOperation = WithStartWorkflowOperation.create(transactionWorkflow, {
    workflowId,
    args: [transactionID],
    taskQueue: 'early-return',
    workflowIdConflictPolicy: 'FAIL',
  });

  const earlyConfirmation = await client.workflow.executeUpdateWithStart(getTransactionConfirmation, {
    startWorkflowOperation,
  });
  const wfHandle = await startWorkflowOperation.workflowHandle();
  const finalReport = await wfHandle.result();
```

Public API

- New client interceptor `startUpdateWithStart` with interfaces `WorkflowStartUpdateWithStartInput` and `WorkflowStartUpdateWithStartOutput`
- New class `WithStartWorkflowOperation`
- `WorkflowClient.executeUpdateWithStart` and `WorkflowClient.startUpdateWithStart`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19187,6 +19187,9 @@
         "@temporalio/proto": "file:../proto",
         "abort-controller": "^3.0.0"
       },
+      "devDependencies": {
+        "protobufjs": "^7.2.5"
+      },
       "engines": {
         "node": ">= 18.0.0"
       }
@@ -19196,7 +19199,6 @@
       "version": "1.11.5",
       "license": "MIT",
       "dependencies": {
-        "@grpc/grpc-js": "^1.12.4",
         "@temporalio/proto": "file:../proto",
         "long": "^5.2.3",
         "ms": "^3.0.0-canary.1",
@@ -21531,13 +21533,13 @@
         "@temporalio/client": "file:../client",
         "@temporalio/common": "file:../common",
         "@temporalio/proto": "file:../proto",
-        "abort-controller": "^3.0.0"
+        "abort-controller": "^3.0.0",
+        "protobufjs": "^7.2.5"
       }
     },
     "@temporalio/common": {
       "version": "file:packages/common",
       "requires": {
-        "@grpc/grpc-js": "^1.12.4",
         "@temporalio/proto": "file:../proto",
         "long": "^5.2.3",
         "ms": "^3.0.0-canary.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -19196,6 +19196,7 @@
       "version": "1.11.5",
       "license": "MIT",
       "dependencies": {
+        "@grpc/grpc-js": "^1.12.4",
         "@temporalio/proto": "file:../proto",
         "long": "^5.2.3",
         "ms": "^3.0.0-canary.1",
@@ -21536,6 +21537,7 @@
     "@temporalio/common": {
       "version": "file:packages/common",
       "requires": {
+        "@grpc/grpc-js": "^1.12.4",
         "@temporalio/proto": "file:../proto",
         "long": "^5.2.3",
         "ms": "^3.0.0-canary.1",
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -1,6 +1,6 @@
 import { ServiceError as GrpcServiceError } from '@grpc/grpc-js';
 import {
-  CustomGrpcServiceError as CustomGrpcServiceError,
+  CustomGrpcServiceError,
   LoadedDataConverter,
   mapFromPayloads,
   NamespaceNotFoundError,
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -5,6 +5,7 @@ import {
   NamespaceNotFoundError,
   searchAttributePayloadConverter,
   SearchAttributes,
+  WorkflowExecutionAlreadyStartedError,
 } from '@temporalio/common';
 import { Replace } from '@temporalio/common/lib/type-helpers';
 import { optionalTsToDate, requiredTsToDate } from '@temporalio/common/lib/time';
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -134,7 +134,10 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
         for (const status of statuses) {
           const detail = status.details?.[0];
           const statusType = detail?.type_url?.replace(/^type.googleapis.com\//, '') as FailureName | undefined;
-          if (statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted' || status.code === grpcStatus.OK) {
+          if (
+            statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted' ||
+            status.code === grpcStatus.OK
+          ) {
             continue;
           }
           err.message = status.message ?? err.message;
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -139,7 +139,7 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
             continue;
           }
           throw new CustomGrpcServiceError(
-            status.message || err.message,
+            status.message ?? err.message,
             status.code || err.code,
             detail?.value?.toString() || err.details,
             err.metadata.getMap()
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -1,4 +1,5 @@
 import { ServiceError as GrpcServiceError } from '@grpc/grpc-js';
+import { Status } from '@grpc/grpc-js/build/src/constants';
 import {
   LoadedDataConverter,
   mapFromPayloads,
@@ -17,7 +18,6 @@ import {
   WorkflowExecutionInfo,
   WorkflowExecutionStatusName,
 } from './types';
-import { Status } from '@grpc/grpc-js/build/src/constants';
 
 function workflowStatusCodeToName(code: temporal.api.enums.v1.WorkflowExecutionStatus): WorkflowExecutionStatusName {
   return workflowStatusCodeToNameInternal(code) ?? 'UNKNOWN';
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -126,7 +126,7 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
         throw new NamespaceNotFoundError(namespace);
       }
       case 'temporal.api.errordetails.v1.MultiOperationExecutionFailure': {
-        // MultiOperationExecutionFailure contains errorstatuses for multiple
+        // MultiOperationExecutionFailure contains error statuses for multiple
         // operations. A MultiOperationExecutionAborted error status means that
         // the corresponding operation was aborted due to an error in one of the
         // other operations. We rethrow the first operation error that is not
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -17,6 +17,7 @@ import {
   WorkflowExecutionStatusName,
 } from './types';
 import { CustomGrpcServiceError } from './errors';
+import { Status } from '@grpc/grpc-js/build/src/constants';
 
 function workflowStatusCodeToName(code: temporal.api.enums.v1.WorkflowExecutionStatus): WorkflowExecutionStatusName {
   return workflowStatusCodeToNameInternal(code) ?? 'UNKNOWN';
@@ -135,7 +136,7 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
         for (const status of statuses) {
           const detail = status.details?.[0];
           const statusType = detail?.type_url?.replace(/^type.googleapis.com\//, '') as FailureName | undefined;
-          if (statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted') {
+          if (statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted' || status.code === Status.OK) {
             continue;
           }
           throw new CustomGrpcServiceError(
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -1,12 +1,10 @@
-import { ServiceError as GrpcServiceError } from '@grpc/grpc-js';
-import { Status } from '@grpc/grpc-js/build/src/constants';
+import { ServiceError as GrpcServiceError, status as grpcStatus } from '@grpc/grpc-js';
 import {
   LoadedDataConverter,
   mapFromPayloads,
   NamespaceNotFoundError,
   searchAttributePayloadConverter,
   SearchAttributes,
-  WorkflowExecutionAlreadyStartedError,
 } from '@temporalio/common';
 import { Replace } from '@temporalio/common/lib/type-helpers';
 import { optionalTsToDate, requiredTsToDate } from '@temporalio/common/lib/time';
@@ -136,7 +134,7 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
         for (const status of statuses) {
           const detail = status.details?.[0];
           const statusType = detail?.type_url?.replace(/^type.googleapis.com\//, '') as FailureName | undefined;
-          if (statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted' || status.code === Status.OK) {
+          if (statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted' || status.code === grpcStatus.OK) {
             continue;
           }
           err.message = status.message ?? err.message;
--- a/packages/client/src/helpers.ts
+++ b/packages/client/src/helpers.ts
@@ -16,7 +16,6 @@ import {
   WorkflowExecutionInfo,
   WorkflowExecutionStatusName,
 } from './types';
-import { CustomGrpcServiceError } from './errors';
 import { Status } from '@grpc/grpc-js/build/src/constants';
 
 function workflowStatusCodeToName(code: temporal.api.enums.v1.WorkflowExecutionStatus): WorkflowExecutionStatusName {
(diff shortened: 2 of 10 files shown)


packages/client/src/workflow-client.ts
```diff
@@ -834,26 +952,118 @@ export class WorkflowClient extends BaseClient {
...
+    // requested wait stage is COMPLETED, further polling is done before returning the UpdateHandle.
     let response: temporal.api.workflowservice.v1.UpdateWorkflowExecutionResponse;
     try {
       do {
-        response = await this.workflowService.updateWorkflowExecution(req);
-      } while (response.stage < waitForStageProto);
+        response = await this.workflowService.updateWorkflowExecution(request);
+      } while (
+        response.stage < UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED
+      );
```
dandavison, line 986: This is a bug fix against basic Update. Previously, if `waitStage=Completed` we were issuing `UpdateWorkflow` requests after the update has reached `Accepted`. But that is not what we should do: we should issue `UpdateWorkflow` until `Accepted` and then switch to `PollWorkflowUpdate`. All other SDKs do this.

packages/client/src/helpers.ts
```diff
@@ -1,5 +1,6 @@
 import { ServiceError as GrpcServiceError } from '@grpc/grpc-js';
 import {
+  CustomGrpcServiceError as CustomGrpcServiceError,
```
mjameswh, line 3:
```suggestion
  CustomGrpcServiceError,
```

packages/common/package.json
```diff
@@ -12,6 +12,7 @@
   "author": "Temporal Technologies Inc. <sdk@temporal.io>",
   "license": "MIT",
   "dependencies": {
+    "@grpc/grpc-js": "^1.12.4",
```
mjameswh, line 15: Let's not add a dependency to gRPC here if we can avoid.

packages/common/src/errors.ts
```diff
@@ -52,3 +54,18 @@ export class NamespaceNotFoundError extends Error {
     super(`Namespace not found: '${namespace}'`);
   }
 }
+
+/**
+ * Thrown when it's necessary to rethrow a grpc-js ServiceError with modifications.
+ */
+@SymbolBasedInstanceOfError('CustomGrpcServiceError')
+export class CustomGrpcServiceError extends Error {
```
mjameswh, line 62: Put that error in `client/src/errors.ts` instead.

mjameswh: Also, not a fan of the name. "custom" in this context means nothing to the end user. Could we have more targetted errors? Or merge this one with `client.ServiceError`?

packages/test/src/helpers-integration.ts
```diff
@@ -3,10 +3,15 @@ import { status as grpcStatus } from '@grpc/grpc-js';
 import { ErrorConstructor, ExecutionContext, TestFn } from 'ava';
 import {
   isGrpcServiceError,
+  WithStartWorkflowOperation,
```
mjameswh, line 6: Several leftover imports.

packages/test/src/test-integration-update.ts
```diff
@@ -17,6 +25,8 @@ const test = makeTestFunction({
       extraArgs: [
         '--dynamic-config-value',
         `history.longPollExpirationInterval="${LONG_POLL_EXPIRATION_INTERVAL_SECONDS}s"`,
+        '--dynamic-config-value',
```
mjameswh, line 28: I don't think you need that one, you already added it to defaultDynamicConfigOptions.

packages/client/src/workflow-client.ts
```diff
@@ -451,6 +457,33 @@ export interface IntoHistoriesOptions {
...
 
+/**
+ * Defines how to start a workflow when using {@link WorkflowClient.startUpdateWithStart} and
+ * {@link WorkflowClient.executeUpdateWithStart}. `workflowIdConflictPolicy` is required in the options.
+ */
+export class WithStartWorkflowOperation<T extends Workflow> {
+  private _workflowHandle: Promise<WorkflowHandle<T>>;
+  private _resolveWorkflowHandle!: (handle: WorkflowHandle<T>) => void;
+  public _executed = false;
+
```
mjameswh, line 468: Don't prefix private propsA few things here:

1. Let's avoid this "underscore prefix" convention if we can. In TS, `_xxx` usually means an unused variable/parameter (e.g. to silence linters). Using that as a visibility marker is unusual. And yes, I know this file contains many such occurences, but that's something we should depart from, not replicate.

2. You can expose the handle as `public readonly workflowHandle`, so you'll no longer need a method below.

3. Use symbols to define protected methods/fields; for example:

```
const resolveWorkflowHandle: unique symbol = Symbol();
const operationExecuted: unique symbol = Symbol();

export class WithStartWorkflowOperation<T extends Workflow> {
  public readonly workflowHandle: Promise<WorkflowHandle<T>>;
  public [operationExecuted]: boolean = false;

  constructor(
    public workflowTypeOrFunc: string | T,
    public options: WorkflowStartOptions<T> & { workflowIdConflictPolicy: WorkflowIdConflictPolicy }
  ) {
    this.workflowHandle = new Promise<WorkflowHandle<T>>((resolve) => {
      this[resolveWorkflowHandle] = resolve;
    });
  }
}
```

mjameswh: Also, we should probably `reject` the `WorkflowHandle` promise if the Workflow failed to start.

packages/client/src/workflow-client.ts
```diff
@@ -834,26 +962,121 @@ export class WorkflowClient extends BaseClient {
...
+        // e.g. startWorkflow / updateWorkflow keys
+        startResp = multiOpResp.responses?.[0]
+          ?.startWorkflow as temporal.api.workflowservice.v1.IStartWorkflowExecutionResponse;
+        onStart(startResp);
+        updateResp = multiOpResp.responses?.[1]
+          ?.updateWorkflow as temporal.api.workflowservice.v1.IUpdateWorkflowExecutionResponse;
+        reachedStage =
+          updateResp.stage ??
+          UpdateWorkflowExecutionLifecycleStage.UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_UNSPECIFIED;
+      } catch (err) {
```
mjameswh, line 1065: I think moving the while-loop (and probably the var definitions and the return statement) _inside_ of the try-block would be easier to read here.

packages/client/src/helpers.ts
```diff
@@ -123,6 +125,27 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
...
+        // other operations. We rethrow the first operation error that is not
+        // MultiOperationExecutionAborted.
+        const { statuses } = temporal.api.errordetails.v1.MultiOperationExecutionFailure.decode(entry.value);
+        for (const status of statuses) {
+          const detail = status.details?.[0];
+          const statusType = detail?.type_url?.replace(/^type.googleapis.com\//, '') as FailureName | undefined;
+          if (statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted') {
+            continue;
+          }
+          throw new CustomGrpcServiceError(
```
mjameswh, line 141: What kind of details do we have here? Could we map these statuses to error classes that we already have elsewhere?

packages/client/src/helpers.ts
```diff
@@ -123,6 +125,27 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
...
+            status.message || err.message,
```
mjameswh, line 142: `??` is generally what you want.

mjameswh:
Please make sure to add `@experimental` annotations on all publicly exposed APIs introduced in this PR.

packages/client/src/workflow-client.ts
```diff
@@ -302,11 +305,14 @@ function defaultWorkflowClientOptions(): WithDefaults<WorkflowClientOptions> {
   };
 }
 
-function assertRequiredWorkflowOptions(opts: WorkflowOptions): void {
-  if (!opts.taskQueue) {
+function assertRequiredWorkflowOptions(opts?: {
```
mjameswh, line 308: Why did you have to make `opts` optional? Should we check specifically for `!opts` and give a distinct error message if it is undefined?

packages/client/src/workflow-client.ts
```diff
@@ -451,6 +457,33 @@ export interface IntoHistoriesOptions {
...
+ * {@link WorkflowClient.executeUpdateWithStart}. `workflowIdConflictPolicy` is required in the options.
+ */
+export class WithStartWorkflowOperation<T extends Workflow> {
+  private _workflowHandle: Promise<WorkflowHandle<T>>;
+  private _resolveWorkflowHandle!: (handle: WorkflowHandle<T>) => void;
+  public _executed = false;
+
+  constructor(
+    public workflowTypeOrFunc: string | T,
+    public options: WorkflowStartOptions<T> & { workflowIdConflictPolicy: WorkflowIdConflictPolicy }
```
mjameswh, line 471: Can live with this temporarily, but we should avoid ad-hoc types in public facing APIs.

packages/test/src/test-integration-update.ts
```diff
@@ -45,6 +55,151 @@ export async function workflowWithUpdates(): Promise<string[]> {
   return state;
 }
 
+test('UWS happy path', async (t) => {
```
mjameswh, line 58: Please expand the `UWS` abbreviation here and below.

```suggestion
test('UpdateWithStart - Happy path', async (t) => {
```

packages/client/src/workflow-client.ts
dandavison: Yes, good call, agreed!

packages/client/src/workflow-client.ts
dandavison: Thanks, great suggestions here.

> You can expose the handle as `public readonly workflowHandle`, so you'll no longer need a method below.

We'll still want a callable won't we? I.e. this would look odd:


```typescript
const wfHandle = await startOp.workflowHandle
```

packages/client/src/workflow-client.ts
dandavison: Good question. That `?` looks like a mistake to me. Removed.

packages/test/src/test-integration-update.ts
dandavison: Removed

packages/test/src/helpers-integration.ts
dandavison: Thanks, removed

packages/common/src/errors.ts
dandavison: Reverted all changes to `common/src`. Still not sure how to rethrow `GrpcServiceError`. One option seems to be just to mutate the received error and re-throw it.

packages/common/package.json
dandavison: Thanks, done.

packages/client/src/helpers.ts
dandavison: Thanks, done.

packages/test/src/test-integration-update.ts
dandavison: Thanks, done.

packages/common/src/errors.ts
dandavison: I am:

- Identifying `WorkflowExecutionAlreadyStartedError` and using that as the error if appropriate
- Otherwise mutating the received error so that it acquires the more specific GrpcServiceError fields (code, message) as unpacked from the MultiOp error, and rethrowing the altered GrpcServiceError error

packages/client/src/workflow-client.ts
dandavison: Resolved as per group decision to use a promise-returning `.workflowhandle()`, reworked class

packages/client/src/helpers.ts
dandavison: See above https://github.com/temporalio/sdk-typescript/pull/1585#discussion_r1890975131

dandavison:
This is ready for another pass @mjameswh

dandavison:
One uncertainty I have is exactly how we should construct the `WithStartWorkflowOperation`, i.e. whether via `new` or via a factory function.

package-lock.json
```diff
@@ -19196,6 +19196,7 @@
       "version": "1.11.5",
       "license": "MIT",
       "dependencies": {
+        "@grpc/grpc-js": "^1.12.4",
```
mjameswh, line 19199: That should no longer be here. Make sure to run `npm install` to update the lock file.

dandavison: Oops, 303cd122f5446d8715812c388e610436ad536dbe

packages/test/src/test-integration-update-interceptors.ts
```diff
@@ -1,6 +1,20 @@
...
+  WorkflowStartUpdateWithStartInput,
+  WorkflowStartUpdateWithStartOutput,
+  WorkflowUpdateStage,
+} from '@temporalio/client';
 import * as wf from '@temporalio/workflow';
-import { Next, UpdateInput, WorkflowInboundCallsInterceptor, WorkflowInterceptors } from '@temporalio/workflow';
+import {
+  Next,
+  UpdateInput,
+  WorkflowIdConflictPolicy,
```
mjameswh, line 14: Unused import

dandavison: Thanks, removed. (I'm used to `ruff` removing them on save, and also would expect a linter violation.)

packages/client/src/workflow-client.ts
```diff
@@ -447,6 +453,44 @@ export interface IntoHistoriesOptions {
...
+ * Defines how to start a workflow when using {@link WorkflowClient.startUpdateWithStart} and
+ * {@link WorkflowClient.executeUpdateWithStart}. `workflowIdConflictPolicy` is required in the options.
+ *
+ * @experimental
+ */
+export class WithStartWorkflowOperation<T extends Workflow> {
+  public [withStartWorkflowOperationUsed]: boolean = false;
+  public [withStartWorkflowOperationResolve]: ((handle: WorkflowHandle<T>) => void) | undefined = undefined;
+  public [withStartWorkflowOperationReject]: ((error: any) => void) | undefined = undefined;
+  private workflowHandlePromise: Promise<WorkflowHandle<T>>;
```
mjameswh, line 470: I think you should be able to make these `private`.

packages/client/src/workflow-client.ts
```diff
@@ -447,6 +453,44 @@ export interface IntoHistoriesOptions {
...
+
+  private constructor(
+    public workflowTypeOrFunc: string | T,
```
mjameswh, line 473: Sorry for the uncertainty. Let's revert to have a public constructor.

packages/client/src/helpers.ts
```diff
@@ -123,6 +126,25 @@ export function rethrowKnownErrorTypes(err: GrpcServiceError): void {
...
+        // other operations. We rethrow the first operation error that is not
+        // MultiOperationExecutionAborted.
+        const { statuses } = temporal.api.errordetails.v1.MultiOperationExecutionFailure.decode(entry.value);
+        for (const status of statuses) {
+          const detail = status.details?.[0];
+          const statusType = detail?.type_url?.replace(/^type.googleapis.com\//, '') as FailureName | undefined;
+          if (statusType === 'temporal.api.failure.v1.MultiOperationExecutionAborted' || status.code === Status.OK) {
+            continue;
+          }
+          err.message = status.message ?? err.message;
```
mjameswh, line 142: Need to check for some specific types of inner errors and throw the appropriate classes in those cases.

packages/test/src/test-integration-update-interceptors.ts
mjameswh: Maybe you are just missing some local config in vscode. I get linter errors here, and can choose either to auto-remove extra imports on save (I used to do that, but now prefer to keep it disabled)

packages/client/src/helpers.ts
```diff
@@ -1,10 +1,12 @@
 import { ServiceError as GrpcServiceError } from '@grpc/grpc-js';
+import { Status } from '@grpc/grpc-js/build/src/constants';
 import {
   LoadedDataConverter,
   mapFromPayloads,
   NamespaceNotFoundError,
   searchAttributePayloadConverter,
   SearchAttributes,
+  WorkflowExecutionAlreadyStartedError,
```
mjameswh, line 9: Unused import

dandavison: Removed

packages/client/src/helpers.ts
dandavison: I believe we agreed that these are all done.

packages/client/src/helpers.ts
```diff
@@ -1,10 +1,12 @@
 import { ServiceError as GrpcServiceError } from '@grpc/grpc-js';
```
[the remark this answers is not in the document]
dandavison: Done (the lower case name is a trick to avoid a clash?)

packages/client/src/workflow-client.ts
dandavison: Done

packages/client/src/workflow-client.ts
dandavison: Done

packages/client/src/helpers.ts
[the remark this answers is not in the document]
mjameswh: Lowercase: No, or at least, not on out side. That's just how it is exported publicly by their package.

dandavison:
Thanks very much for the expert reviewing @mjameswh, as always.

Issue #1334, opened 2024-01-10
[Feature Request] Run temporal worker in bun


Code2Life:
### Bun worker can not work Issue

I tried to use bun as temporal worker, when running temporal client it looked perfect, the workflow created with HTTP2+gRPC, which just supported in bun.js last month.

But when I ran temporal worker it can not work, even bun.sh had implement v8 compatible vm context sandbox. Firstly I found this issue:
![image](https://github.com/temporalio/sdk-typescript/assets/14833440/9bae6fe2-28bc-4587-beb5-196f5dfdaabb)

Then, i tried to comment out the createHook code block, this error disappeared, but the core-bridge was broken

```
2024-01-10T08:41:16.131Z [INFO] webpack 5.89.0 compiled successfully in 486 ms
2024-01-10T08:41:16.132Z [INFO] Workflow bundle created { size: '0.74MB' }
2024-01-10T08:41:16.284520Z  INFO temporal_sdk_core::worker: Initializing worker task_queue=hello-world namespace=default
thread '<unnamed>' panicked at 'Hash table capacity overflow', /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hashbrown-0.13.2/src/raw/mod.rs:90:40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-01-10T08:41:16.286Z [INFO] Worker state changed { state: 'RUNNING' }
2024-01-10T08:41:16.290075Z  INFO next_workflow_activation: temporal_sdk_core::worker: Initiated shutdown task_queue=hello-world namespace=default task_queue=hello-world
2024-01-10T08:41:16.291Z [INFO] Worker state changed { state: 'DRAINED' }
2024-01-10T08:41:16.291Z [INFO] Worker state changed { state: 'STOPPED' }
thread 'core' panicked at 'Workflow processing terminates cleanly: Error joining workflow processing thread: Ok("Hash table capacity overflow")', sdk-core/core/src/worker/mod.rs:409:14
 96 |         }
 97 |         this.shutDownRequested = true;
 98 |         await this.send({ type: 'destroy' });
 99 |         const exitCode = await this.workerThread.terminate();
100 |         if (exitCode !== exports.TERMINATED_EXIT_CODE) {
101 |             throw new core_bridge_1.UnexpectedError(`Failed to terminate Worker thread, exit code: ${exitCode}`);
                                      ^
```

### Additional context

[...]

PR #1586, opened 2024-12-13
chore: Upgrade to Node 18

### LINKED ISSUE #1584: [Bug] mTLS NativeConnectionOptions tonic::transport::Error(Transport, NativeCertsNotFound)

Use mTLS within `NativeConnectionOptions` to connect to a cluster

### Describe the bug

Moving from 1.11.3 to 1.11.5 broke the `NativeConnectionOptions` / connect call with a TransportError

```
/app/node_modules/@temporalio/worker/lib/connection.js:58
                throw new core_bridge_1.TransportError(err.message);
                      ^
TransportError: tonic::transport::Error(Transport, NativeCertsNotFound)
    at NativeConnection.connect (/app/node_modules/@temporalio/worker/lib/connection.js:58:23)
```

Able to confirm it's an issue with 1.11.5 after reverting back to 1.11.3

### Minimal Reproduction

**tlsConfig option**

```
connectionOptions.tls = {
      serverNameOverride: config.TEMPORAL_TLS_SERVER_NAME,
      serverRootCACertificate: fs.readFileSync(config.TEMPORAL_TLS_CA),
      clientCertPair: {
        crt: fs.readFileSync(config.TEMPORAL_TLS_CERT),
        key: fs.readFileSync(config.TEMPORAL_TLS_KEY),
      },
    }
```

### Environment/Versions

The application is built and deployed as a docker image

- Docker / `node:22.11.0-bookworm-slim`
- Hosting environment / Ubuntu 22

mjameswh:
## What was changed

- Upgraded minimum Node requirement to 18;

- Removed patch in `create-project` to deal with pre-TS5 configurations;

- Ignore errors thrown while trying to use the `v8.promiseHooks.createHook` API in the Workflow Worker, as an attempt to avoid one known issue preventing usage of Temporal with bun and deno (see #1334). Note that there are probably other issues involved in getting Temporal Workers running on bun or deno.

- Add test to confirm that ``NativeConnection` can connect to server using `mtls` (closes #1584).

- Homogenize and clarify that default address is `localhost` everywhere in public exposed code (that's the correct thing to do and should be non-breaking everywhere except on very badly configured dual-stack machines, in which case they would have had errors anyway due to inconsistencies between `Client` and `NativeClient`).

- Add a workaround to temporarily fix test failures observed in CI due to a regression in Node 22.12.0, causing incorrect resolution of `localhost` on Windows.


CONTRIBUTING.md
```diff
@@ -50,9 +50,9 @@ See [sdk-structure.md](./docs/sdk-structure.md)
 
 ### Environment setup
 
-- The TS SDK can be executed on 16, 18 and 20. However, for SDK development, we recommend
-  using Node 20. For easier testing, you may want to use a version manager, such as:
-  [nvm](https://github.com/nvm-sh/nvm/blob/master/README.md) or [volta](https://volta.sh/).
+- The TS SDK can be executed on 18, 20 or 22. However, we recommend using Node 22 for SDK development.
+  For easier testing while development. you may want to use a version manager, such as
```
dandavison, line 54:
```suggestion
  For easier testing during development you may want to use a version manager, such as
```

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -265,7 +259,13 @@ export class GlobalHandlers {
           store.childToParent.delete(promise);
           store.promiseToStack.delete(promise);
         },
-      });
+      }) as () => void;
+    } catch (_) {
+      // v8.promiseHooks.createHook is not available in bun and Node.js < 16.14.0.
+      // That's ok, collecting stack trace is an optional feature anyway.
+      //
+      // FIXME: This should be sent to logs, not the console… but we don't have acces to it here.
```
dandavison, line 267: I appreciate the unicode ellipsis…

```suggestion
      // FIXME: This should be sent to logs, not the console… but we don't have access to it here.
```

dandavison (approved):
LGTM (but build failing)

PR #1595, opened 2025-01-08
Include update info in workflow logging output

### LINKED ISSUE #1532: Show update handler and ID in logging context
See https://github.com/temporalio/features/issues/466 for more details

THardy98:
This change includes update info (update id and name) in workflow logging output when logging within an update handler.
1. Closes #1532

2. How was this tested:
Added small integration test

3. Any docs updates needed?
Not sure


packages/workflow/src/logs.ts
```diff
@@ -72,12 +73,16 @@ export const log: WorkflowLogger = Object.fromEntries(
...
         const getLogAttributes = composeInterceptors(activator.interceptors.outbound, 'getLogAttributes', (a) => a);
+        const updateInfo = currentUpdateInfo();
         return loggerSink[level](message, {
           // Inject the call time in nanosecond resolution as expected by the worker logger.
           [LogTimestamp]: activator.getTimeOfDay(),
           sdkComponent: SdkComponent.workflow,
           ...getLogAttributes(workflowLogAttributes(activator.info)),
           ...attrs,
+          // Add update info if it exists
+          ...(updateInfo && { updateId: updateInfo.id }),
```
mjameswh, line 84: Let's use a single object spread operation for both id and name.

Something like that:
```suggestion
          ...(updateInfo && { updateId: updateInfo.id, updateName: updateInfo.name }),
```

mjameswh: Thinking more about this, I'd say we should do this in `workflowLogAttributes()`, for coherency. That behavior will be visible to user interceptors.

Only problem though is that `workflowLogAttributes()` may be called from outside the workflow sandbox, so you will need to add a call to `inWorkflowContext()` first to not attempt this in other cases.

That could give something like this:

```
export function workflowLogAttributes(info: WorkflowInfo): Record<string, unknown> {
  const attributes = {
    namespace: info.namespace,
    taskQueue: info.taskQueue,
    workflowId: info.workflowId,
    runId: info.runId,
    workflowType: info.workflowType,
  };

  // Add details about the current update if appropriate
  if (inWorkflowContext()) {
    const updateInfo = currentUpdateInfo();
    if (updateInfo) {
      attributes['updateId'] = updateInfo.id;
      attributes['updateName'] = updateInfo.name;
    }
  }

  return attributes;
}
```

THardy98: Sounds good, updated with the snippet above.

> Only problem though is that workflowLogAttributes() may be called from outside the workflow sandbox, so you will need to add a call to inWorkflowContext() first to not attempt this in other cases.

From the code references, does this only occur when a worker creates a new workflow?
https://github.com/temporalio/sdk-typescript/blob/sdk-t-add-update-info-to-logging-context/packages/worker/src/worker.ts#L1313

THardy98:
Not sure what is causing CI to fail, its unrelated to the code changes

packages/workflow/src/logs.ts
mjameswh: > From the code references, does this only occur when a worker creates a new workflow?

Probably, yeah. This is used to provide some basic attributes very early during workflow initialization and in case starting the workflow fail for whatever reason.

Arguably, we could duplicate that code, once in Worker knowing that we are not in workflow context, and once in workflows/logs.ts where we'd know we are in workflow context, but I think that in this case it is best to keep this in a single place, to avoid eventual inconsistencies whenever we add extra workflow log attributes.

mjameswh:
> Not sure what is causing CI to fail, its unrelated to the code changes

That's unrelated. I'm already on this: #1596.

mjameswh (approved):

PR #1598, opened 2025-01-14
fix: remove custom error

albertlai431:
## What was changed
Remove the custom `MODULE_NOT_FOUND` error catching

## Why?
I was running into an issue earlier where I was getting this error: `ValueError: Could not find a file at the specified payloadConverterPath: '/workspace/typescript/src/payload-converter.ts'.`

However, the error message was actually `Cannot find module './generated/root'`, which was a downstream `MODULE_NOT_FOUND` error. So, the `payload-converter.ts` file existed, but the `root` file it was importing didn't exist. However, the error message was misleading because it made me believe that `payload-converter.ts` couldn't be found.

I think it would be best to remove this custom catching logic and just throw an error if it does (in the else case anyways).


mjameswh:
I see. Yeah, the current error handling can indeed be misleading. Yet, I think the hint about which file was being loaded might be pertinent.

I will need to think a bit more about this.

PR #1604, opened 2025-01-17
docs: Use numbered list for env setup in Contribution Guide

chadxz:
## What was changed
* Updated the environment setup portion of the contribution guide to use a numbered list instead of bullets.

## Why?

When following this guide to setup my environment, I found myself skimming over the bulleted list instead of treating it as a set of tasks to complete. Using a numbered list for this section helps readability.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,34 +50,39 @@ See [sdk-structure.md](./docs/sdk-structure.md)
 
 ### Environment setup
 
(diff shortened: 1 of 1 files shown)


mjameswh:
Makes total sense. Thanks!

mjameswh (approved):

chadxz:
@mjameswh Thanks! I saw that it had a lint error so I pushed a fix for that. Hopefully g2g now.

PR #1605, opened 2025-01-17
fix(workflow): Fix multiple context leaks in `reuseV8Context` executor

### LINKED ISSUE #1476: [Bug] Reusable VM allows context leak due to global variable reassignment

### LINKED ISSUE #1592: [Bug] Reusable VM allows context leak due to global symbol properties

mjameswh:
## What was changed

- Fix multiple context leaks and bugs in the `reuseV8Context` executor:

  - …by reassigning a new object to an existing shared global variable: `globalThis.console = { ...globalThis.console, wfId: workflowInfo().workflowId }` (fixes #1476);

  - …by modifying one of Node's built in global objects: `globalThis.Number.a = workflowInfo().workflowId`;

  - …by deleting a previously set global variable: `globalThis.a = 1; await sleep(1) ; delete globalThis.a ; await sleep(1) ; globalThis.a = (globalThis.a || 0) + 1; /* globalThis.a is 2 rather than 1 */`

  - …by defining global symbol properties: `const mySymbol = Symbol.for('...'); globalThis[mySymbol] = ...` (fixes #1592).

Note: this is a second round over #1519. The code has changed considerably since the first previous review.


packages/common/src/type-helpers.ts
```diff
@@ -219,23 +219,37 @@ export function SymbolBasedInstanceOfError<E extends Error>(markerName: string):
...
+// We also do not attempt to visit the prototype chain, as this would make it much harder to load
+// polyfills, and it is extremely unlikely anyway that one would modify the prototype of a built-in
+// object in a way that would have undesirable consequences (i.e. a polyfill function may actually
+// leak to another workflow context, but it wouldn't carry anything Workflow specific).
+export function deepFreeze<T>(object: T, visited = new WeakSet<any>()): T {
+  if (object == null || visited.has(object) || (typeof object !== 'object' && typeof object !== 'function'))
+    return object;
+  visited.add(object);
+  if (Object.isFrozen(object)) return object;
+
```
antlai-temporal, line 237: Can an object be frozen but not deepFrozen? If not, add a comment....

packages/test/src/test-reusev8context-perf.ts
```diff
@@ -0,0 +1,170 @@
+// DONOTMERGE -- DO NOT REVIEW THIS FILE
+
```
antlai-temporal, line 2: Is this comment correct? Do you need to remove this file?

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -93,7 +97,29 @@ export function injectConsole(context: vm.Context): void {
...
+
+export function injectGlobals(context: vm.Context): void {
+  const globals = {
+    AsyncLocalStorage,
+    URL,
+    URLSearchParams,
+    assert,
+    TextEncoder,
+    TextDecoder,
+    AbortController,
```
antlai-temporal, line 117: Do any of these need to be deep frozen, or we assume they won't be abused...?

antlai-temporal (approved):
It seems we don't deep freeze Sets and Maps. Is this a concern? Do we need to document this somehow?

mjameswh:
> It seems we don't deep freeze Sets and Maps. Is this a concern? Do we need to document this somehow?

That limitation is documented in `deepFreeze()`'s typedoc. Not a concern at this point because there should be no instance of Map or Set in `globalThis` after V8 context initialization, nor directly in one of our preloaded modules (i.e. the SDK itself). I had previously added coverage for those because I used a Map as part of the V8 context swapping algorithm, but no longer need it.

packages/common/src/type-helpers.ts
mjameswh: Not sure what needs to be clarified here. The official typedoc on `Object.freeze()` already indicates that freezing is shallow, and generally well known to advanced JS devs.

packages/worker/src/workflow/vm-shared.ts
mjameswh: They will get deep frozen at the same time as all other global variables, after loading the script.

mjameswh:
I moved `deepFreeze()` to `reusable-vm.ts` (rather than the `common` package) as some implementation decisions (e.g. freezing functions, not freezing prototypes, not handling Maps and Sets, etc) are specific to the Reusable VM Workflow Sandbox use case, and may not be appropriate for other use cases. It therefore doesn't seem appropiate to consider this as a reusable utility function.

PR #1606, opened 2025-01-17
fix(workflow): Clearly diffentiate WFT failures resulting from unhandle Promise rejection

### LINKED ISSUE #1039: [Bug] Unhandled promise rejections results in hard to diagnose WFT failure
### Describe the bug

Unhandled promise rejections causes WFT failures. This is reasonable given that Node 16's default policy is to crash the process on unhandled promise rejections.

However, when this happens, the WFT failure actually carries very little context information, making it difficult for the user to 1) understand the reason why the WFT failed and 2) the origin of the failing Promise.

It is therefore proposed that WFT failures caused by unhandled promise rejections should:
1. use an error message that clearly indicates "Unhandled promise rejection";
2. that the WFT failure should carry the stack trace to the place where failing promise was created; 
3. that a message with "Unhandled promise rejection" and the stack trace to the promise creation should be reported to the Worker's log.

mjameswh:
## What changed

- The sandbox was previously converting all non-TemporalFailure errors thrown while processing a Workflow Activation into unhandled promise rejections, to be caught elsewhere. This was done intentionally and technically worked correctly most of the time, but made it impossible to differentiate errors that were indeed uncaught promise rejections, occasionally resulting in hard to diagnose errors (see #1039). It also resulted in some incorrect behaviors in uncommon/rare use cases.

  This PR introduces a cleaner way to propagate non-TemporalFailures errors up, thus avoiding the intentional use of Unhandled Promise Rejection for that purpose.

  Also, errors resulting from actual unhandled promise rejections now get wrapped in a `UnhandledRejectionError`, making this fact clear in error messages that are reported in logs, Workflow History, and UI.

Fixes #1039.


packages/workflow/src/internals.ts
```diff
@@ -964,6 +970,9 @@ export class Activator implements ActivationHandler {
         },
         true
       );
+    } else {
+      // Only keep the first error that bubbles up; subsequent errors will be ignored.
+      if (this.workflowTaskError === undefined) this.workflowTaskError = error;
```
bergundy, line 975: Wouldn't it be better to throw a special error type and handle in the unhandled rejection path?

packages/workflow/src/worker-interface.ts
```diff
@@ -188,15 +188,20 @@ export function concludeActivation(): coresdk.workflow_completion.IWorkflowActiv
  * @returns number of unblocked conditions.
  */
 export function tryUnblockConditions(): number {
+  const activator = getActivator();
+
+  // If a Workflow Task error was captured, bubble it up now
+  if (activator.workflowTaskError) throw activator.workflowTaskError;
```
bergundy, line 194: This won't cover exceptions that are thrown in the condition functions AFAIU.

packages/workflow/src/internals.ts
mjameswh: I considered that avenue, but personally can't see any strong point in favor of it. And without a strong argument, I'd prefer just avoiding perusing unhandled rejections, because they make it harder for new SDK devs to figure out what's happening.

packages/workflow/src/worker-interface.ts
mjameswh: Condition functions are sync. If a condition throws synchronously, the error will bubble up out of `tryUnblockConditions()`, just the same as if they had been thrown from that line.

Microtasks that may get unblocked by the resolving the condition will not execute synchronously, so throwing from there would therefore belong to whatever is waiting on the corresponding Promise, or will result in an unhandled promise rejection, which would be correct in that case.

bergundy: 👍

packages/workflow/src/internals.ts
bergundy: Yeah, my reasoning behind it is that there's now two paths where workflow task errors may be surfaced, making the code harder to follow. The fact that we throw on `tryUnblockConditions` isn't very intuitive to me.

bergundy (approved):
Leaving it up to you to make a call here. Functionally I don't see a problem with this code.

THardy98 (approved):
Left a couple comments, nothing blocking

packages/test/src/test-payload-converter.ts
```diff
@@ -193,12 +193,18 @@ if (RUN_INTEGRATION_TESTS) {
...
     });
-    const logger = new DefaultLogger('ERROR', (entry) => {
-      if (entry.meta?.error.message === 'Unknown encoding: json/protobuf') {
-        markErrorThrown();
-      }
+
+    Runtime.install({
+      logger: new DefaultLogger('WARN', (entry) => {
+        if (
+          entry.message.includes('Failing workflow task') &&
```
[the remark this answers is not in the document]
mjameswh: It comes from Core.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -323,97 +323,100 @@ export abstract class BaseVMWorkflow implements Workflow {
...
+      const doSingleBatch = !hasSignals || this.activator.hasFlag(SdkFlags.ProcessWorkflowActivationJobsAsSingleBatch);
 
-    const [patches, nonPatches] = partition(activation.jobs, ({ notifyHasPatch }) => notifyHasPatch != null);
-    for (const { notifyHasPatch } of patches) {
-      if (notifyHasPatch == null) throw new TypeError('Expected notifyHasPatch to be set');
-      this.activator.notifyHasPatch(notifyHasPatch);
-    }
+      const [patches, nonPatches] = partition(activation.jobs, ({ notifyHasPatch }) => notifyHasPatch != null);
+      for (const { notifyHasPatch } of patches) {
+        if (notifyHasPatch == null) throw new TypeError('Expected notifyHasPatch to be set');
```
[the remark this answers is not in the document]
mjameswh: Good catch, that's leftover from a previous refactor. No longer makes any sense.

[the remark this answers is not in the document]
mjameswh: Good catch! That's really leftover from a previous refactor, but was kept just to make the compiler happy, as TSC has no way of figuring out that `notifyHasPatch` can't be undefined in this specific case. A non-null type assertion could be used instead, but will skip that just now.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -434,14 +437,20 @@ export abstract class BaseVMWorkflow implements Workflow {
...
+    }
     this.unhandledRejection = err;
   }
 
   /**
    * Call into the Workflow context to attempt to unblock any blocked conditions and microtasks.
    *
-   * This is performed in a loop allowing microtasks to be processed between
-   * each iteration until there are no more conditions to unblock.
+   * This is performed in a loop, going and out of the VM, allowing microtasks to be processed
```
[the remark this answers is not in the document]
mjameswh: Actually, `going in and out`. Fixed.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -323,97 +323,100 @@ export abstract class BaseVMWorkflow implements Workflow {
...
+      if (doSingleBatch) {
+        // updateRandomSeed requires the same special handling as patches (before anything else, and don't
+        // unblock conditions after each job). Unfortunately, prior to ProcessWorkflowActivationJobsAsSingleBatch,
+        // they were handled as regular jobs, making it unsafe to properly handle that job above, with patches.
+        const [updateRandomSeed, rest] = partition(nonPatches, ({ updateRandomSeed }) => updateRandomSeed != null);
+        if (updateRandomSeed.length > 0)
+          this.activator.updateRandomSeed(updateRandomSeed[updateRandomSeed.length - 1].updateRandomSeed!);
         this.workflowModule.activate(
-          coresdk.workflow_activation.WorkflowActivation.fromObject({ ...activation, jobs }),
-          batchIndex++
```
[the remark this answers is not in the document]
mjameswh: I'm puzzled, it looks like that diff is not from the present PR.

But in any case, there is indeed two code paths in this function. The "single batch" implementation and the "legacy" implementation. The former handle all jobs in a single batch, so having a "batch index" makes no sense. The `this.workflowModule.activate()` function will default that to 0, which is appropriate.

PR #1612, opened 2025-01-24
Typed search attributes

### LINKED ISSUE #1232: [Feature Request] Typed Search Attributes
Implementation of [typed search attributes](https://github.com/temporalio/proposals/blob/master/all-sdk/sdk-typed-search-attributes.md).

THardy98:
## What was changed
Introduces `TypedSearchAttributes` to the Typescript SDK, allowing users to supply and receive type-enforced search attributes. `TypedSearchAttributes` are intended to replace the existing (not type-enforced) `SearchAttributes` which have been marked as `@deprecated` in its existing usages.

## Why?
Brings parity with other SDKs which have a typed search attribute implementation.
Brings parity to how the server represents search attributes.
Provides a typed, structured interface to use search attributes.

## For reviewers

**Points of interest:**
- `typed-search-attributes.ts` contains the API
- `test-typed-search-attributes.ts` contains tests and demonstrates API usage
- `worflow.ts` for changes to `upsertSearchAttributes` (mutations to search attributes in workflow info)
- `payload-search-attributes.ts` contains decoding/encoding logic for search attributes (both legacy and typed)

---
1. Closes #1232 
2. How was this tested:
Added tests, see `test-typed-search-attributes.ts`

3. Any docs updates needed?
Likely as we are deprecating existing `SearchAttributes`


packages/client/src/schedule-helpers.ts
```diff
@@ -326,14 +343,9 @@ export async function decodeScheduleAction(
...
       retry: decompileRetryPolicy(pb.startWorkflow.retryPolicy),
-      searchAttributes: Object.fromEntries(
-        Object.entries(
-          mapFromPayloads(
-            searchAttributePayloadConverter,
-            pb.startWorkflow.searchAttributes?.indexedFields ?? {}
-          ) as SearchAttributes
-        )
-      ),
+      // TODO(thomas): is there a reason why we didn't filter out empty arrays here?
```
mjameswh, line 346: IIRC Schedules were introduced in 1.18, at which point the server would no longer be sending empty arrays anyway.

packages/client/src/schedule-helpers.ts
```diff
@@ -397,3 +426,48 @@ export function decodeScheduleRecentActions(
     }
   );
 }
+
+/*
```
mjameswh, line 430: Just make sure to delete those notes before your PR moves to non-draft.

packages/client/src/schedule-types.ts
```diff
@@ -70,9 +70,19 @@ export interface ScheduleOptions<A extends ScheduleOptionsAction = ScheduleOptio
...
+   *
+   * @deprecated Use {@link typedSearchAttributes} instead.
    */
   searchAttributes?: SearchAttributes;
 
+  /**
+   * Additional indexed information attached to the Schedule. More info:
+   * https://docs.temporal.io/docs/typescript/search-attributes
+   *
+   * Values are always converted using {@link JsonPayloadConverter}, even when a custom Data Converter is provided.
```
mjameswh, line 82: We should also mention that SA don't get encoded (i.e. they don't go through PayloadCodec). That's important as people may not realize that there are security implication in using SA. Please make sure that's fact is clear everywhere that a user would supply search attributes.

packages/client/src/schedule-helpers.ts
```diff
@@ -342,6 +354,7 @@ export async function decodeScheduleAction(
   throw new TypeError('Unsupported schedule action');
 }
 
+// TODO(thomas): move to internal package (or at least to the helpers.ts file)
```
mjameswh, line 357: Please also create `encodeSearchAttributes` and `encodeTypedSearchAttributes` help functions. Move those four helper functions, and anything else that is specific to encoding/decoding SAs, to `common/src/converter/payload-search-attributes.ts`. And make sure to use those helpers everywhere (e.g. workflow-client, schedule-client, workflow/src/internals.ts, etc).

packages/common/src/converter/payload-converter.ts
```diff
@@ -319,6 +333,60 @@ export class SearchAttributePayloadConverter implements PayloadConverter {
...
+      // Convert Date to ISO string
+      value = (value as Date).toISOString();
+    }
+
+    const payload = this.jsonConverter.toPayload(value);
+    if (payload === undefined) {
+      throw new ValueError('Could not convert typed search attribute to payload');
+    }
+    // Note: this shouldn't be the case but the compiler complains without this check.
+    if (payload.metadata === undefined || payload.metadata === null) {
```
mjameswh, line 356: Please you the following idiom whenever you get this:

```suggestion
    if (payload.metadata == null) {
```

For reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#difference_between_null_and_undefined

packages/common/src/converter/payload-converter.ts
```diff
@@ -319,6 +333,60 @@ export class SearchAttributePayloadConverter implements PayloadConverter {
...
+  jsonConverter = new JsonPayloadConverter();
+
+  public toPayload<T>(typedSearchAttribute: T): Payload {
+    if (!isTypedSearchAttribute(typedSearchAttribute)) {
+      throw new ValueError('Invalid typed search attribute');
+    }
+
+    const type = typedSearchAttribute[0];
+    let value = typedSearchAttribute[1];
+    if (type === 'DATETIME') {
```
mjameswh, line 346:
```suggestion
    let [type, value] = typedSearchAttribute;
    if (type === 'DATETIME') {
```

packages/common/src/converter/payload-converter.ts
```diff
@@ -319,6 +333,60 @@ export class SearchAttributePayloadConverter implements PayloadConverter {
...
+    }
+
+    // TODO(thomas): type casting undefined values is not clear to caller
+
+    let value = this.jsonConverter.fromPayload(payload);
+    // If no 'type' metadata field or no given value, we skip.
+    if (payload.metadata.type === undefined || value === undefined) {
+      return undefined as T;
+    }
+    const type = toSearchAttributeType(decode(payload.metadata.type));
```
mjameswh, line 376: In that case, better make the return type of this function `T | undefined`, to make it clear that the caller has the responsibility of handling that possibility.

packages/client/src/schedule-helpers.ts
mjameswh: Actually, it might even be desirable to have an `encodeSearchAttributes` function, which accepts both untyped and typed SAs. There's some little details in how to properly do that; duplicating that code is undesirable.

packages/client/src/schedule-types.ts
```diff
@@ -70,9 +70,19 @@ export interface ScheduleOptions<A extends ScheduleOptionsAction = ScheduleOptio
...
+   */
+  typedSearchAttributes?: TypedSearchAttributePair[];
```
mjameswh, line 84: I think that one will be tricky on update. We'd be receiving existing SAs, which would be an object rather than an array of Pair. We'll discuss that on Monday.

THardy98: Changed to `TypedSearchAttributePair[] | TypedSearchAttributes` as discussed offline.

I like this much more.

packages/common/src/converter/payload-converter.ts
THardy98: I can't change the type of this function because it breaks the interface . The `PayloadConverter` interface definition seems very restrictive, the way it's typed makes it difficult to get much type-safety in the `toPayload` and `fromPayload` definitions.

In any case, I left a comment noting why this was the case and what the suggested usage is. Not clear to me how to make this better.

packages/common/src/converter/payload-converter.ts
THardy98: Done.

packages/client/src/schedule-helpers.ts
THardy98: Done.
Yeah that reduced a lot of the duplicated code, didn't realize how much was being re-written.
Wrote `encodeUnifiedSearchAttributes` as a single encoding function for both forms of search attributes.

THardy98:
Updated the PR description for reviewers. Some remaining points of uncertainty:

- a couple `TODO` comments in places where I follow an existing pattern but am unsure if this is still the case (particularly those in `payload-search-attributes.ts` and `internals.ts`
- I serialized expected responses in tests (i.e. `JSON.parse(JSON.stringify(...))`) to handle JSON converted responses from `query`. Figured this was okay, but just want to double-check

THardy98:
Who do I need to get in touch with to update docs? (considering we're deprecating a user-facing feature in exchange for another)

THardy98:
Naming question (probably/possibly bike-shedding):

Do we prefer `TypedSearchAttributes` or `SearchAttributesTyped`?

packages/workflow/src/internals.ts
```diff
@@ -478,8 +478,12 @@ export class Activator implements ActivationHandler {
     // Most things related to initialization have already been handled in the constructor
     this.mutateWorkflowInfo((info) => ({
       ...info,
-      searchAttributes:
-        (mapFromPayloads(searchAttributePayloadConverter, searchAttributes?.indexedFields) as SearchAttributes) ?? {},
+
+      // TODO(thomas): do we not filter out undefined values?
```
mjameswh, line 482: Not sure what you meant exactly by that comment, but current impl looks ok to me. On `initializeWorkflow`, it safe to assume that the server wouldn't send us "undefined" SA values.

packages/workflow/src/workflow.ts
```diff
@@ -380,9 +390,12 @@ function startChildWorkflowExecutionNextHandler({
         workflowIdReusePolicy: encodeWorkflowIdReusePolicy(options.workflowIdReusePolicy),
         parentClosePolicy: encodeParentClosePolicy(options.parentClosePolicy),
         cronSchedule: options.cronSchedule,
-        searchAttributes: options.searchAttributes
-          ? mapToPayloads(searchAttributePayloadConverter, options.searchAttributes)
-          : undefined,
+        searchAttributes:
+          options.searchAttributes || options.typedSearchAttributes
+            ? {
+                indexedFields: encodeUnifiedSearchAttributes(options.searchAttributes, options.typedSearchAttributes),
```
mjameswh, line 396: There's no `indexFields` indirection on Workflow commands _that we emit_, because the Core's API define those fields as a map of Payload, rather than reusing the SearchAttributes message from upstream's proto (arguably, that's an error in Core's proto definitions, but I wouldn't mind changing that at this point).

packages/test/src/test-integration-split-one.ts
```diff
@@ -723,6 +728,10 @@ test('Workflow can read WorkflowInfo', configMacro, async (t, config) => {
     runId: handle.firstExecutionRunId,
     taskQueue,
     searchAttributes: {},
+    // Ensure serialized data structure is correct
+    typedSearchAttributes: JSON.parse(JSON.stringify(new TypedSearchAttributes())),
```
mjameswh, line 732: It was not evident to me at first sight that the `TypedSearchAttributes` object would actually serialize to JSON properly. By spec, `JSON.stringify(obj)` only serializes enumerable own properties, so an object's private fields shouldn't be listed.

But turns out it does, as TS predates JS' EcmaScript 5 class model, so the TypeScript compiler translates "private fields" to regular properties on the JS object. So we get something like this:

```
{
    "searchAttributes": {
        "CustomIntField": ["INT", 42],
        "CustomKeywordField": ["KEYWORD", "test-value2"]
    }
}
```

Still, that feels fragile, especially given that this is what sink functions will see, and what users might see in logs if they print out Workflow Info or Search Attributes. I particularly don't like that we have this inner `searchAttributes` object, and that we serialize using our internal storage format rather than something that adheres to the "user visible" API (i.e. the format that a user could pass as input to the `TypedSearchAttributes` class).

So two suggestions:
1. On the `TypedSearchAttributes` class, redefine the `toJSON()` function to return the precise object as we'd like it to get serialized, i.e. probably a `TypedSearchAttributePair[]`;
2. In at least some of your tests, rather than doing `JSON.parse(JSON.stringify(someSearchAttributeObject))` for the purpose of `deepEquals()`, be explicit on what you would expect to get (e.g. `[["CustomIntField": ["INT", 42]]]`), so that we don't break that at a later point (that would be just too easy to miss otherwise).

packages/test/src/test-sinks.ts
```diff
@@ -118,6 +118,7 @@ if (RUN_INTEGRATION_TESTS) {
       memo: {},
       parent: undefined,
       searchAttributes: {},
+      typedSearchAttributes: JSON.parse(JSON.stringify(new TypedSearchAttributes())),
```
mjameswh, line 121: Looks like we might have more and more discrepencies between the actual `WorkflowInfo` interface and that sink functions see of it. We already had the `unsafe.now()`, mentionned below, and now, `typedSearchAttributes` will be plain objects/arrays rather than an instance of the `TypedSearchAttributes` class.

We could consider rehydrating the `WorkflowInfo` objects before actually passing them to sink functions, but I honestly don't think there's a strong use case for it enough to justify the extra CPU cycle and memory churning that would involve. So I think we may simply create a variant of the `WorkflowInfo` interface that corresponds to what we actually get in sinks.

Don't fix this just now, but please add a FIXME and/or open a ticket for that.

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1372,98 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+        // Update if it's a valid typed search attribute pair.
+        if (isTypedSearchAttributePair(pair)) {
+          typedSearchAttributes = typedSearchAttributes.updateAttributes([pair]);
+        }
+      }
+      return {
+        ...info,
+        searchAttributes: {
+          ...info.searchAttributes,
+          // Note that we keep empty arrays in the search attributes.
```
mjameswh, line 1460: I think that was a previous oversight and would prefer we actually fix that. We should prune out "empty" search attributes on upsert.

packages/workflow/src/workflow.ts
```diff
@@ -1347,7 +1362,8 @@ export function setDefaultSignalHandler(handler: DefaultSignalHandler | undefine
  *
  * @param searchAttributes The Record to merge. Use a value of `[]` to clear a Search Attribute.
  */
-export function upsertSearchAttributes(searchAttributes: SearchAttributes): void {
+
+export function upsertSearchAttributes(searchAttributes: SearchAttributes | TypedSearchAttributeUpdatePair[]): void {
```
mjameswh, line 1366: Need to update the typedoc above.

packages/workflow/src/workflow.ts
```diff
@@ -948,6 +962,7 @@ export function makeContinueAsNewFunc<F extends Workflow>(
  *
  *```ts
  *import { continueAsNew } from '@temporalio/workflow';
+import { SearchAttributeValue } from '../../common/lib/interfaces';
```
mjameswh, line 965: Why add this `import` to this specific sample? I'd remove.

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1372,98 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+    });
+
+    activator.mutateWorkflowInfo((info: WorkflowInfo): WorkflowInfo => {
+      // Use updateSearchAttributes to get a deep copy of the current state.
+      let typedSearchAttributes = info.typedSearchAttributes.updateAttributes([]);
+      for (const [k, v] of Object.entries(searchAttributes)) {
+        const typedKey = TypedSearchAttributes.getKeyFromUntyped(k, v);
+
+        // Unable to discern the typing of the key, skip.
+        if (!typedKey) {
```
mjameswh, line 1422: I think in this case it would be preferable to see if there is already an SA named `k` in `typedSearchAttributes`, and if so:
1. Try to use type of SA `k` in `typedSearchAttributes` as the type for `v`;
2. If the type is incompatible, remove entry `k` from `typedSearchAttributes`.

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1372,98 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
       },
-    };
-  });
+    });
+
+    activator.mutateWorkflowInfo((info: WorkflowInfo): WorkflowInfo => {
+      // Use updateSearchAttributes to get a deep copy of the current state.
+      let typedSearchAttributes = info.typedSearchAttributes.updateAttributes([]);
+      for (const [k, v] of Object.entries(searchAttributes)) {
+        const typedKey = TypedSearchAttributes.getKeyFromUntyped(k, v);
```
mjameswh, line 1419: Seems like the code here (`getKeyFromUntyped` and below) wants to be permissive, kind of allowing going through with `v` being a single primitive, rather than an array, but then you'll get until line 1442 and fail on `newValue = v[0]`. You may also get weird behaviors if `v` is set to the single value `false` or `0`, because of the `!v` check at line 1427.

Technically speaking, the former API only allowed `v` to be an array or undefined, so let's just _throw_ early if `v` anything else than that.

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1372,98 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+
+        // Unable to discern the typing of the key, skip.
+        if (!typedKey) {
+          continue;
+        }
+
+        // Delete the key if the value is undefined/null/empty list.
+        if (!v || (Array.isArray(v) && v.length === 0)) {
```
mjameswh, line 1427:
```suggestion
        if (v != null || (Array.isArray(v) && v.length === 0)) {
```

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1372,98 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+          if (isTypedSearchAttributeUpdatePair(pair)) {
+            typedSearchAttributes = typedSearchAttributes.updateAttributes([pair]);
+          }
+          continue;
+        }
+
+        let newValue: unknown = v;
+        if (typedKey.type !== SearchAttributeType.KEYWORD_LIST) {
+          // Skip if we found multiple values for a non-list key.
+          if (Array.isArray(v) && v.length > 1) {
```
mjameswh, line 1438: Remove from `typedSearchAttributes` in that case. That's better than having incohenrencies.

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1372,98 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
+          // Convert to single-value.
+          newValue = v[0];
+        }
+
+        if (typedKey.type === SearchAttributeType.DATETIME) {
+          newValue = new Date(newValue as string);
+        }
+
+        const typedValue: unknown = [typedKey.type, newValue];
+        const pair: unknown = [typedKey, typedValue];
```
mjameswh, line 1450: That seems incorrect. Both `typedKey` and `typedValue` contains the type.

packages/common/src/converter/payload-converter.ts
```diff
@@ -55,6 +55,20 @@ export function mapToPayloads<K extends string>(converter: PayloadConverter, map
   ) as Record<K, Payload>;
 }
 
+/**
+ * Run {@link PayloadConverter.toPayload} on each value in the map.
+ *
+ * @throws {@link ValueError} if conversion of any value in the map fails
+ */
+export function typedMapToPayloads<K extends string, T>(
+  converter: PayloadConverter,
```
mjameswh, line 64: AFAICS, you should be able to just add a `T = any` generic parameter, like so:

```
export function mapToPayloads<K extends string, T = any>(
  converter: PayloadConverter,
  map: Record<K, T>
): Record<K, Payload> {
  return Object.fromEntries(
    Object.entries(map).map(([k, v]): [K, Payload] => [k as K, converter.toPayload(v)])
  ) as Record<K, Payload>;
}
```

Then you no longer need the `typedMapToPayloads()` function.

packages/common/src/converter/payload-converter.ts
```diff
@@ -97,6 +111,19 @@ export function mapFromPayloads<K extends string>(
   ) as Record<K, unknown>;
 }
 
+export function typedMapFromPayloads<K extends string, T>(
+  converter: PayloadConverter,
```
mjameswh, line 115: Similarly:

```
export function mapFromPayloads<K extends string, T = unknown>(
  converter: PayloadConverter,
  map?: Record<K, Payload> | null | undefined
): Record<K, T> | undefined {
  if (map == null) return undefined;
  return Object.fromEntries(
    Object.entries(map).map(([k, payload]): [K, T] => {
      const value = converter.fromPayload<T>(payload as Payload);
      return [k as K, value];
    })
  ) as Record<K, T>;
}
```

packages/common/src/index.ts
```diff
@@ -12,6 +12,7 @@ export * from './converter/data-converter';
 export * from './converter/failure-converter';
 export * from './converter/payload-codec';
 export * from './converter/payload-converter';
+export * from './converter/payload-search-attributes';
```
mjameswh, line 15: Let's be specific in what we reexport. TBH, I don't think there's anything in that file that I would consider as public API. It's ok for our own packages to import inner files of the common package.

packages/common/src/index.ts
```diff
@@ -24,6 +25,7 @@ export type { Timestamp, Duration, StringValue } from './time';
 export * from './workflow-handle';
 export * from './workflow-options';
 export * from './versioning-intent';
+export * from './typed-search-attributes';
```
mjameswh, line 28: Again, be specific. There are things here that are meant to be public APIs, but not `*`.

mjameswh:
> Naming question (probably/possibly bike-shedding):
> 
> Do we prefer TypedSearchAttributes or SearchAttributesTyped?

Definitely `TypedSearchAttributes` for me.

packages/test/src/test-typed-search-attributes.ts
```diff
@@ -0,0 +1,427 @@
...
+  untyped_single_string: ['one'],
+  untyped_single_int: [1],
+  untyped_single_double: [1.23],
+  untyped_single_bool: [true],
+  untyped_single_date: [date],
+  untyped_multi_string: ['one', 'two'],
+};
+
+// The corresponding typed search attributes from untypedSearchAttributes.
+const typedFromUntypedInput: TypedSearchAttributePair[] = [
```
mjameswh, line 47: Though it may be useful to have a "createAtttributeValuePair()" utility function, it is not how I would expect a user to use the Typed Search Attributes API in general.

The mental model, from a user's POV must be `key(name, type) => value`.

Take this Python code:

```
    # Define TSA Keys, would usually be in some global place
    key_1 = SearchAttributeKey.for_text("python-test-schedule-sa-update-key-1")
    key_2 = SearchAttributeKey.for_keyword("python-test-schedule-sa-update-key-2")

    # Example of setting TSA Keys=>Values from scratch
    handle = await client.create_schedule(
        f"schedule-{uuid.uuid4()}",
        Schedule(action=create_action, spec=ScheduleSpec()),
        search_attributes=TypedSearchAttributes(
            [
                SearchAttributePair(key_1, val_1),
                SearchAttributePair(key_2, val_2),
            ]
        ),
    )

    # Example of reading TSA values
    assert input.description.typed_search_attributes[key_1] == val_1
    assert input.description.typed_search_attributes[key_2] == val_2

    # Creating new TSA by deriving exisitng TSA K/V
    return ScheduleUpdate(
        input.description.schedule,
        search_attributes=input.description.typed_search_attributes.updated(
            SearchAttributePair(key_1, val_1 + "-new"),
            SearchAttributePair(key_2, val_2 + "-new"),
        ),
    )
```

In TS and following existing idioms of the TS SDK, that give something like this:

```
    // Define TSA Keys, would usually be in some global place
    export const key1 = defineSearchAttribute("my-custom-sa-key-1", 'STRING');
    export const key2 = defineSearchAttribute("my-custom-sa-key-2", 'KEYWORD');
    export const key3 = defineSearchAttribute("my-custom-sa-key-3", 'INT');

    // Example of setting TSA Keys=>Values from scratch
    const handle = await client.workflow.start(
        // ...
        searchAttributes: [              // Technically no need for a builder function here as TS type 
            [key1, "My customer Name"],  // system is expressive enough to provide type safety on mappings.
            [key2, "some-keyword"],      // I wouldn't mind some utility function to make this more readable,
        ]                                // but the tuple notation should technically be enough.
    );

    // Example of reading TSA values
    assert input.description.typedSearchAttributes.get(key1) === "My customer Name";
    assert input.description.typedSearchAttributes.get(key2) "some-keyword";

    # Creating new TSA by deriving exisitng TSA K/V
    return ScheduleUpdate(
        input.description.schedule,
        search_attributes=input.description.typedSearchAttributes.updated([
            [key2, undefined],  // Delete an existing key
            [key3, 84],
        ]),
    )
```

packages/workflow/src/internals.ts
THardy98:
```
export function decodeSearchAttributes(indexedFields: Record<string, Payload> | undefined | null): SearchAttributes {
  if (!indexedFields) return {};
  return Object.fromEntries(
    Object.entries(mapFromPayloads(searchAttributePayloadConverter, indexedFields) as SearchAttributes).filter(
      ([_, v]) => v && v.length > 0
    ) // Filter out empty arrays returned by pre 1.18 servers
  );
}
```
When we `decodeSearchAttributes` from server (i.e. in `executionInfoFromRaw`) we make sure to filter out undefined values and empty arrays. I assumed that server could send undefined SA values because of this, or is this some legacy logic?

packages/test/src/test-typed-search-attributes.ts
THardy98: As discussed offline, the API now offers usage like:

```
// Define keys:
const key1 = { name: 'key_text', type: SearchAttributeType.TEXT };
const key2 = { name: 'key_int', type: SearchAttributeType.INT };

// Define pairs:
const textPair = { key: key1, value: 'even_newer_value' };
const intPair = { key: key2, value: 45 };

// Define collections
const searchAttrs = [textPair, intPair];
// or inline
const inlineAttrs = [
    { key: key1, value: 'even_newer_value' },
    { key: key2, value: 45 }
]

// Define class instance 
const typedSearchAttrs = new TypedSearchAttributes(searchAttrs);
// or inline
const typedSearchAttrs = new TypedSearchAttributes([
    { key: key2, value: 45 }
]); 
```

packages/common/src/index.ts
THardy98: Yup, lack of oversight here. I've made the exports more explicit now.

packages/common/src/index.ts
THardy98: Yeah I agree.

Haven't commit this change yet, getting some weird import issues after replacing with `@temporalio/common/lib/converter/payload-search-attributes`, but I'll add it once I fix it.

packages/common/src/converter/payload-converter.ts
THardy98: Also removed.

packages/common/src/interfaces.ts
```diff
@@ -98,7 +98,8 @@ export type WorkflowResultType<W extends Workflow> = ReturnType<W> extends Promi
  *
  * Dates are serialized as ISO strings.
  */
-export type SearchAttributes = Record<string, SearchAttributeValue | Readonly<SearchAttributeValue> | undefined>;
+export type SearchAttributeValueOrReadonly = SearchAttributeValue | Readonly<SearchAttributeValue> | undefined;
```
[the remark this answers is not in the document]
THardy98: Done.

packages/workflow/src/workflow.ts
THardy98: Removed given the different API

packages/workflow/src/workflow.ts
THardy98: Ditto the previous comment on removing from `typedSearchAttributes`.

In any case, I've removed this check and a couple other validation checks in the `upsert` logic. If we can determine a valid type for the given input then these additional checks are redundant.


I've also left a short comment in the `upsert` logic that covers the basic cases, thought it might be useful though the logic is not complicated. Can remove if you think unnecessary. Testing these cases (i.e. multi-value input for non keyword list, invalid objects/primitives) is difficult given that server will error if you try to do this.

packages/workflow/src/workflow.ts
THardy98: `v` is recognized as `SearchAttributeValueOrReadonly` within the loop, which does imply array or undefined.

I don't get any compiler issues here given that typing, but I've added an error check just in case if we want to be extra safe.

packages/workflow/src/workflow.ts
THardy98: I don't think case 1 is possible. If we get `undefined` from `TypedSearchAttributes.getKeyFromUntyped`, we've checked that the value is **not** `string`, `number`, `boolean`, `Date`, or `string[]`. Consequently, there would be no possible `SearchAttributeType` that would be valid. The value would fail a `isValidValueForType` check for any `SearchAttributeType`.

For case 2, I can't think of scenarios where `typedSearchAttributes[k]` and `searchAttributes[k]` both exists but have different values (but I very well could just not be aware of the semantics). 

If we enumerate the possibilities for attr `A`:
1. `A` belongs in both `typed` and `untyped`, same value
2. `A` belongs in both `typed` and `untyped`, different value (if this is possible)
3. `A` belongs in `typed` but not `untyped` (if this is possible)
4. `A` belongs in `untyped` but not `typed` (if this is possible)
5. `A` does not belong in `typed` or `untyped`

**deletion** makes things _more_ consistent for cases 2 (`A` will only exist in `untyped` but at least there are no differing values) and 3 (`A` will not exist in either), but makes things _less_ consistent for case 1 (which is arguably the common case). It's a no-op for cases 4 and 5.

In any case, I'm not sure if this would be _expected behavior_ for a user. I don't think a user would expect an update/insert operation to delete unless explicitly specified to do so.

WDYT?

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1372,98 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
-      ...info,
-      searchAttributes: {
-        ...info.searchAttributes,
-        ...searchAttributes,
+    activator.mutateWorkflowInfo((info: WorkflowInfo): WorkflowInfo => {
+      const newSearchAttributes: SearchAttributes = {};
+      for (const pair of searchAttributes) {
+        const [k, v] = pair;
+        if (v === null) {
+          // If the value is null (i.e. delete), set the value to an empty list.
```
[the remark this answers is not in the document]
THardy98: Ditto the previous comment on the Python SDK's implementation (removed but maybe we should clarify semantics with Python SDK and Chad)

packages/workflow/src/workflow.ts
THardy98: Not sure what happened there, probably some code I moved around that I forgot to remove.

packages/workflow/src/workflow.ts
THardy98: Done.

packages/workflow/src/workflow.ts
THardy98: Done.

I hesitated to do this before because I noticed this comment in Python's upsert impl:
```
 # Update the keys in the existing dictionary. We keep exact values
 # sent in instead of any kind of normalization. This means empty
 # lists remain as empty lists which matches what the server does.
```
It's not clear to me why this would be useful, regardless if it matches server. Maybe worth asking Chad about this ?

packages/test/src/test-integration-split-one.ts
THardy98: Yeah I agree with these points.

1. I've defined `toJSON()` on the class to return its JSON representation as `SearchAttributePair[]`
2. I've removed all usages of `JSON.stringify` in tests. I've changed `WorkflowInfo` to return `SearchAttributePair[]` instead of `TypedSearchAttributes`. This is the only case where the server response is `SearchAttributePair[]` because we don't get an opportunity to decode. I think this is okay, particularly because users will likely use `SearchAttributePair[]` when specifying search attribute options. Maybe we can change the field name to reflect that the value is a list of pairs, I have no preference. Consequently, when making assertions on `WorkflowInfo` we just need to compare lists, no need for `JSON.stringify`. All other test cases work with `TypedSearchAttributes` directly.

packages/workflow/src/workflow.ts
THardy98: Ah, good catch.

I'm surprised that the TS compiler didn't complain about this given that with the unexpected `indexedFields` field it should be `Record<string<Record<string, Payload>>` instead of `Record<string, Payload>` like it expects.

THardy98:
Just want to call out this comment I wrote in `test-typed-search-attributes.ts`, because I'm having trouble figuring out this behavior:

```
    // Note: There is a discrepancy between workflow info and workflow exec description
    // after upserts with untyped search attributes.
    //
    // The upsert in this test is done through a signal, which serializes the upsert input.
    // The serialized input converts dates to their ISO string representation. When we read untyped
    // seralized search attributes in upsert, we try to infer their key type based on the value.
    // By default, untyped string values are identified as TEXT type. As a result, serialized date strings
    // become TEXT type and attributes that may have been KEYWORD type are also inferred as TEXT type.
    //
    // When we read the search attributes from the workflow info, we expect to see these type conversions
    // in the typed search attributes.
    //
    // This does not happen with workflow exec description. It's unclear to me why this happens, given that
    // the same upsert logic applies. The only difference being that we decode search attributes from server
    // for the workflow exec description.
```

packages/test/src/test-sinks.ts
THardy98: Yup, just created a literal to satisfy the test and added a FIXME as discussed offline.

packages/test/src/test-integration-split-one.ts
THardy98: Reverted the type back to a class type `TypedSearchAttributes` as discussed offline (keeping the `toJSON` override).

packages/workflow/src/workflow.ts
THardy98: Fixed the deletion case for upsert.

Added some improvements to handle date strings and less type loss from `KEYWORD` to `TEXT`

packages/workflow/src/workflow.ts
THardy98: Removed the comment. Code is pretty readable and doesn't really warrant it.

THardy98:
> Just want to call out this comment I wrote in `test-typed-search-attributes.ts`, because I'm having trouble figuring out why this behavior occurs:
> 
> ```
>     // Note: There is a discrepancy between typed search attributes in workflow info and workflow
>     // exec description after upserts with untyped search attributes.
>     //
>     // The upsert in this test is done through a signal, which serializes the upsert input.
>     // The serialized input converts dates to their ISO string representation. When we read untyped
>     // seralized search attributes in upsert, we try to infer their key type based on the value.
>     // By default, untyped string values are identified as TEXT type. As a result, serialized date strings
>     // become TEXT type and attributes that may have been KEYWORD type are also inferred as TEXT type.
>     //
>     // When we read the search attributes from the workflow info, we expect to see these type conversions
>     // in the typed search attributes.
>     //
>     // This does not happen with workflow exec description. It's unclear to me why this happens, given that
>     // the same upsert logic applies. The only difference being that we decode search attributes from server
>     // for the workflow exec description.
> ```
> 
> I suspect the decoding from server attributes is the cause, but I don't understand how this could happen given that the type conversions due to the upsert should occur first?

We're pushing to the activator before upserting and the workflow exec description reads from server, so we don't incur this type loss.

In any case, we handle date strings and KEYWORD -> TEXT type loss a bit better now which removes this discrepancy.

Removed this comment.

packages/common/src/index.ts
THardy98: Done

packages/workflow/src/workflow.ts
```diff
@@ -1347,7 +1374,8 @@ export function setDefaultSignalHandler(handler: DefaultSignalHandler | undefine
  *
  * @param searchAttributes The Record to merge. Use a value of `[]` to clear a Search Attribute.
  */
-export function upsertSearchAttributes(searchAttributes: SearchAttributes): void {
+
```
mjameswh, line 1377: There should be no empty line here. Not sure why the linter isn't complaining.

packages/workflow/src/workflow.ts
```diff
@@ -1347,7 +1374,8 @@ export function setDefaultSignalHandler(handler: DefaultSignalHandler | undefine
  *
  * @param searchAttributes The Record to merge. Use a value of `[]` to clear a Search Attribute.
```
mjameswh, line 1375: > Use a value of `[]` to clear a Search Attribute

I think that statement is no longer true.

packages/workflow/src/workflow.ts
```diff
@@ -1322,7 +1334,9 @@ export function setDefaultSignalHandler(handler: DefaultSignalHandler | undefine
  * Updates this Workflow's Search Attributes by merging the provided `searchAttributes` with the existing Search
  * Attributes, `workflowInfo().searchAttributes`.
  *
- * For example, this Workflow code:
+ * Search attributes can be upserted using either SearchAttributes (deprecated) or SearchAttributeUpdatePair[]
```
mjameswh, line 1337: I'd only mention the new approach. Or at very least, mention it first.

packages/workflow/src/workflow.ts
```diff
@@ -1335,7 +1349,20 @@ export function setDefaultSignalHandler(handler: DefaultSignalHandler | undefine
  * });
  * ```
  *
- * would result in the Workflow having these Search Attributes:
+ * Equivalently, upserting a workflow's search attributes using SearchAttributeUpdatePair[]:
+ *
+ * ```ts
+ * upsertSearchAttributes([
+ *  { key: { name: 'CustomIntField', type: 'INT' }, value: 1 },
```
mjameswh, line 1356: Please add a `defineSearchAttribute()` helper function in the commmon package, and use it here to predefine the SAs rather than inlining the definition as the value of the `key` property. That's the usage pattern that users are expected to follow.

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1384,119 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
-  activator.pushCommand({
-    upsertWorkflowSearchAttributes: {
-      searchAttributes: mapToPayloads(searchAttributePayloadConverter, searchAttributes),
-    },
-  });
+  if (Array.isArray(searchAttributes)) {
+    // Typed search attributes
+    activator.pushCommand({
+      upsertWorkflowSearchAttributes: {
+        searchAttributes: mapToPayloads<string, TypedSearchAttributeUpdateValue<SearchAttributeType>>(
```
mjameswh, line 1391: Why are you using `mapToPayloads` here rather than `encodeSearchAttributes()`? If needed, you can make `encodeSearchAttributes()` allow undefined on either argument so that it is usable for this purpose.

Same for legacy SA below.

packages/workflow/src/workflow.ts
```diff
@@ -1356,21 +1384,119 @@ export function upsertSearchAttributes(searchAttributes: SearchAttributes): void
...
-    return {
-      ...info,
-      searchAttributes: {
-        ...info.searchAttributes,
-        ...searchAttributes,
+    activator.mutateWorkflowInfo((info: WorkflowInfo): WorkflowInfo => {
+      // Create a copy of the current state.
+      const newSearchAttributes: SearchAttributes = { ...info.searchAttributes };
+      for (const pair of searchAttributes) {
+        if (pair.value === null) {
```
mjameswh, line 1407: Here, we check if value is `null`, but in legacy SA, we check for it to be `undefined`. That may be confusing. Let's use `== null` in both cases, just in case.

THardy98: Changed to `== null`.

packages/workflow/src/workflow.ts
THardy98: Done.

packages/workflow/src/workflow.ts
THardy98: Fair, removed the old approach from the doc.

packages/workflow/src/workflow.ts
THardy98: Removed.

packages/common/src/search-attributes.ts
```diff
@@ -0,0 +1,300 @@
+import type { temporal } from '@temporalio/proto';
+import { makeProtoEnumConverters } from './internal-workflow';
+
+/**
+ * If another SDK creates a Search Attribute that's not an array, we wrap it in an array.
+ *
+ * Dates are serialized as ISO strings.
+ */
+export type SearchAttributeValueOrReadonly = SearchAttributeValue | Readonly<SearchAttributeValue> | undefined;
```
mjameswh, line 9: I think those three types should be `@deprecated`, no?

packages/common/src/search-attributes.ts
```diff
@@ -0,0 +1,300 @@
...
+    const res = new TypedSearchAttributes();
+    res.searchAttributes = state;
+    return res;
+  }
+
+  /**
+   * Return JSON representation of this class as SearchAttributePair[]
+   * Default toJSON method is not used because it's JSON representation includes private state.
+   */
+  toJSON(): SearchAttributePair[] {
```
mjameswh, line 168: Can add a `@hidden` annotation on this method, we don't need it to appear in API docs.

packages/workflow/src/workflow.ts
```diff
@@ -948,6 +956,7 @@ export function makeContinueAsNewFunc<F extends Workflow>(
  *
  *```ts
```
mjameswh, line 957: Incorrect/Incoherent indentation on this code snippet. Make sure there's a space at the begining of each line — see `function workflowInfo()` above for reference.

packages/test/src/test-typed-search-attributes.ts
```diff
@@ -0,0 +1,437 @@
...
+    typedSearchAttributes?: TypedSearchAttributes | SearchAttributePair[];
+  };
+  expected: {
+    searchAttributes?: SearchAttributes;
+    typedSearchAttributes?: TypedSearchAttributes;
+  };
+}
+
+// inputTestCases contains permutations of search attribute inputs
+const inputTestCases: TestInputSearchAttributes[] = [
```
mjameswh, line 178: nit: When doing this kind of dataset-based tests, it's generally useful to also include a "name" or "comment" on each entry, and output it in the assertion. Makes it much easier to when tests fail.

In your assertion below, you would do somehting like:

```
inputTestCases.map(async ({ input, expected, name }) => {
  t.deepEqual(desc.searchAttributes, expected.searchAttributes, name)
}
```

packages/test/src/test-typed-search-attributes.ts
```diff
@@ -0,0 +1,437 @@
...
+    // Update search attributes with untyped input.
+    await handle.signal(mutateSearchAttributes, untypedUpdateAttrs);
+    res = await handle.query(getWorkflowInfo);
+    desc = await handle.describe();
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { typed_bool, ...untypedUpdateExpected } = untypedUpdateAttrs;
+
+    assertWorkflowInfoSearchAttributes(t, res, untypedUpdateExpected, [
+      { key: { name: 'typed_text', type: SearchAttributeType.TEXT }, value: 'new_value' },
```
mjameswh, line 307: Make sure to use `defineSearchAttribute()` in at least one test (though I would personally use it in _most_ tests).

packages/common/src/search-attributes.ts
```diff
@@ -0,0 +1,300 @@
...
+        return SearchAttributeType.DATETIME;
+      case 'KeywordList':
+        return SearchAttributeType.KEYWORD_LIST;
+      default:
+        return;
+    }
+  }
+}
+
+export function defineSearchKey<T extends SearchAttributeType>(name: string, type: T): SearchAttributeKey<T> {
```
mjameswh, line 291: Sorry, I wasn't clear… In my mind, this would have been the function named `defineSearchAttribute()`, but I see that "search attribute" could equivalently mean the key or the pair, so not not ideal.

Still, I'd go for `defineSearchAttributeKey()`. "Search Attribute" is a concept, something that someone is likely to search for. Reducing it to a shorter form might make it harder for users to find docs.

mjameswh:
We now have many linter warnings about usage of deprecated APIs. You will need to silence all of them by appending a `// eslint-disable-line deprecation/deprecation` comment at the end of those lines.

packages/common/src/search-attributes.ts
```diff
@@ -0,0 +1,300 @@
...
+  return { name, type };
+}
+
+export function defineSearchAttribute<T extends SearchAttributeType>(
+  key: SearchAttributeKey<T>,
```
mjameswh, line 296: Same as before, wouldn't call this one `SearchAttribute` as that could equivalently mean either the key or an assignment pair. Also, I really want to avoid the verb `define` for this, as we have consistently used that naming pattern for things that are meant to be "defined once, then shared across your code base" (i.e. some form of constant), whereas an SA value assignement is almost always tied to a specific workflow execution.

Still debating with myself exactly how it should be named.

mjameswh: I really don't think we need this specific function. The difference with `defineSearchAttributeKey()` here is really that SA keys are something we want users to perceive as a global constant object, whereas SA key/value pairs are clearly data structures.

Still, if you want to keep this function, I don't mind, but I would call it simply `searchAttributePair`.

```
    // These would be defined in some global scope
    const UserNameSA = defineSearchAttributeKey('typed_text', 'TEXT');
    const OrderStatusSA = defineSearchAttributeKey('typed_keyword', 'KEYWORD');
    const OrderCountSA = defineSearchAttributeKey('typed_int', 'INT');
    const OrderAmountSA = defineSearchAttributeKey('typed_double', 'DOUBLE');

    // This would be pass as option on starting a workflow or some other APIs
    ...
    searchAttributes: [
      { key: UserNameSA, value: 'new_value' },
      { key: OrderStatusSA, value: 'new_keyword' },
      { key: OrderCountSA, value: 2 },
      { key: OrderAmountSA, value: 2.34 },
    ]),
    ...

    // .. or ...

    ...
    searchAttributes: [
      searchAttributePair(UserNameSA, 'new_value'),
      searchAttributePair(OrderStatusSA, 'new_keyword'),
      searchAttributePair(OrderCountSA, 2),
      searchAttributePair(OrderAmountSA, 2.34),
    ]),
    ...
```

packages/common/src/search-attributes.ts
THardy98: Done.

packages/common/src/search-attributes.ts
THardy98: Done.

packages/workflow/src/workflow.ts
THardy98: Fixed, found another instance and fixed that too. Not sure if it was due to this PR, but good to get it fixed anyway

packages/test/src/test-typed-search-attributes.ts
THardy98: Definitely useful, added.

packages/test/src/test-typed-search-attributes.ts
THardy98: Changed the tests in `test-typed-search-attributes.ts` to use the helper.

packages/common/src/search-attributes.ts
THardy98: Renamed to `defineSearchAttributeKey`

THardy98:
> We now have many linter warnings about usage of deprecated APIs. You will need to silence all of them by appending a `// eslint-disable-line deprecation/deprecation` comment at the end of those lines.

There were a lot, I suppressed them inline, but is there a reason we don't do this in the eslint config instead?

packages/common/src/search-attributes.ts
THardy98: No preference, removed them.

mjameswh:
> There were a lot, I suppressed them inline, but is there a reason we don't do this in the eslint config instead?

The "deprecation" rule plugin that we are using at the moment unfortunately doesn't have any level of configurability. And we really want to check for usage of deprecations, e.g. us using deprecated Node.js or third party libraries APIs.

packages/test/package.json
```diff
@@ -8,7 +8,11 @@
     "build:ts": "tsc --build",
     "build:protos": "node ./scripts/compile-proto.js",
     "test": "ava ./lib/test-*.js",
-    "test.watch": "ava --watch ./lib/test-*.js"
+    "test.watch": "ava --watch ./lib/test-*.js",
+    "test-split-one": "ava ./lib/test-integration-split-one.js",
```
mjameswh, line 12: Certainly useful while developing locally, but I don't think we should commit such aliases. Too specic/volatile.

mjameswh (approved):
Left one last comment. Feel free to merge after that.

PR #1623, opened 2025-02-07
feat(worker): Add multiple Worker and Telemetry options

### LINKED ISSUE #1600: [Feature Request] OTel metrics over HTTP

Expose ability to have Core OTel metrics use HTTP instead of default gRPC protocol. Would recommend a simple default-false http boolean. See https://github.com/temporalio/sdk-core/issues/820.

### LINKED ISSUE #1611: [Feature Request] Allow customization of dev server UI port

See https://github.com/temporalio/features/issues/588

mjameswh:
## What changed?

This PR groups several minor changes to the Worker/Runtime/Telemetry APIs.

### User visible changes to Worker and Telemetry options

- On `TelemetryOptions`:
  - Add the following properties:
    - `metrics.metricPrefix`
    - `metrics.attachServiceName`
    - `metrics.globalTags`
    - `metrics.histogramBucketOverrides`
    - `metrics.otel.http` — fixes #1600
  - Move `metrics.temporality` to `metrics.otel.temporality`, as Core doesn't support that option on Prometheus. This is coherent with other Core-based SDKs.
  - Deprecate property `TelemetryOption.noTemporalPrefixForMetrics`, promoting the use of the more versatile `metrics.metricPrefix` property instead

- On `DevServerConfig`:
  - Add `server.uiPort` property — fixes #1611
  - Fix the `server.dbFilename` property, which had never worked due to a naming inconsistency between the lang's interface and bridge's code

- On `NativeConnectionOptions`:
  - Add boolean property `disableErrorCodeMetricTags`. 

All of those change preserve compatibility with prior behaviors and deprecated settings.

### Misc changes

- Update Core SDK from dd37e2 to 75903f

- Fix a test flake due to visibility eventual consistency

- Added integration tests for several of the options mentioned above, as well as other options that were not tested until now.

- Transition some of the existing bridge APIs and code to adhere to a stricter TS/Rust interface, to reduce the risks of future bugs resulting from inconsistencies. All bridge code will be gradually migrated to those conventions in the near future.

## Review notes

To simplify review, every single change has been isolated in a distinct commit.

### DIFF
--- a/packages/core-bridge/src/conversions.rs
+++ b/packages/core-bridge/src/conversions.rs
@@ -240,6 +240,13 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
         if let Some(ref metrics) = js_optional_getter!(cx, self, "metrics", JsObject) {
             telemetry_opts.metric_prefix(js_value_getter!(cx, metrics, "metricPrefix", JsString));
 
+            telemetry_opts.attach_service_name(js_value_getter!(
+                cx,
+                metrics,
+                "attachServiceName",
+                JsBoolean
+            ));
+
             if let Some(ref prom) = js_optional_getter!(cx, metrics, "prometheus", JsObject) {
                 if js_optional_getter!(cx, metrics, "otel", JsObject).is_some() {
                     cx.throw_type_error(
--- a/packages/core-bridge/src/conversions.rs
+++ b/packages/core-bridge/src/conversions.rs
@@ -266,11 +266,12 @@ impl ObjectHandleConversionsExt for Handle<'_, JsObject> {
                 {
                     options.unit_suffix(unit_suffix);
                 }
-                if let Some(use_seconds_for_durations) =
-                    js_optional_value_getter!(cx, prom, "useSecondsForDurations", JsBoolean)
-                {
-                    options.use_seconds_for_durations(use_seconds_for_durations);
-                }
+                options.use_seconds_for_durations(js_value_getter!(
+                    cx,
+                    prom,
+                    "useSecondsForDurations",
+                    JsBoolean
+                ));
 
                 let options = options.build().map_err(|e| {
                     cx.throw_type_error::<_, TelemetryOptions>(format!(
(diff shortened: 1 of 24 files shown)


packages/core-bridge/src/conversions.rs
```diff
@@ -49,6 +53,17 @@ impl ArrayHandleConversionsExt for Handle<'_, JsArray> {
         }
         Ok(ret_vec)
     }
+
+    fn to_vec_of_float(&self, cx: &mut FunctionContext) -> NeonResult<Vec<f64>> {
```
Sushisource, line 57: This could be a generic that takes <JSxxx, RustType> if desired

packages/test/src/test-integration-workflows.ts
```diff
@@ -1274,32 +1275,31 @@ test('Count workflow executions', async (t) => {
...
+
+      // ... and 3 workflows that will complete
       executeWorkflow(completableWorkflow, { args: [true] }),
       executeWorkflow(completableWorkflow, { args: [true] }),
       executeWorkflow(completableWorkflow, { args: [true] }),
     ]);
   });
 
+  // FIXME: Find a better way to wait for visibility to stabilize
+  await setTimeoutPromise(1000);
```
Sushisource, line 1292: I thought there was some kind of `assertEventually` ?

packages/test/src/test-otel.ts
```diff
@@ -129,6 +153,59 @@ test.serial('Exporting OTEL metrics from Core works', async (t) => {
   }
 });
 
+test.serial('Exporting OTEL metrics using OTLP/HTTP from Core works', async (t) => {
+  let resolveCapturedRequest = (_req: http.IncomingMessage) => undefined as void;
+  const capturedRequest = new Promise<http.IncomingMessage>((r) => (resolveCapturedRequest = r));
```
Sushisource, line 158: You wrote it so great, but, IMO for things like this it's fine to just ensure the option makes it to Core since Core went through the rigamarole of having an actual server etc

Sushisource (approved):

packages/test/src/test-integration-workflows.ts
mjameswh: We do, but it is awkward because we can't use `t.deepEquals()` from it. Sleeping for 1s isn't that bad anyway, given that those tests execute concurrently anyway, so that's not blocking.

mjameswh: But we have that need of "waiting for vis to catch up" in a few places, and so I would expect users. So I want to take time at some point to see if we could somehow add an API for that as part of the Dev/Test server environment API to wait for visibility to sync.

packages/core-bridge/src/conversions.rs
mjameswh: Yeah, I'm definitely planing more refactoring of the type helpers and macros.

baptou12:
Hello, when do you plan to release this ?

PR #1624, opened 2025-02-08
Add search attributes to dev server

### LINKED ISSUE #1447: [Feature Request]  Accept search attributes in dev server startup

See https://github.com/temporalio/features/issues/494. With https://github.com/temporalio/cli/pull/593 deployed, users should be allowed to set search attributes in dev server option that translate to --search-attribute calls.

THardy98:
## What was changed
Can pass `searchAttributes` as a workflow options when creating test dev server (with or without time-skipping)

1. Closes #1447

2. How was this tested:
Added integration test

3. Any docs updates needed?
No

### DIFF
--- a/packages/core-bridge/ts/index.ts
+++ b/packages/core-bridge/ts/index.ts
@@ -459,10 +459,6 @@ export interface TimeSkippingServerConfig {
    * be supported in the future.
    */
   extraArgs?: string[];
-  /**
-   * Search attributes to be registered with the dev server.
-   */
-  searchAttributes?: SearchAttributeKey<SearchAttributeType>[];
 }
 
 /**
--- a/packages/core-bridge/ts/index.ts
+++ b/packages/core-bridge/ts/index.ts
@@ -1,6 +1,7 @@
-import { LogLevel, Duration } from '@temporalio/common';
+import { LogLevel, Duration, SearchAttributeType } from '@temporalio/common';
 import type { TLSConfig, ProxyConfig, HttpConnectProxyConfig } from '@temporalio/common/lib/internal-non-workflow';
 import { WorkerTuner } from './worker-tuner';
+import { SearchAttributeKey } from '@temporalio/common/src/search-attributes';
 
 export {
   WorkerTuner,
@@ -458,6 +459,10 @@ export interface TimeSkippingServerConfig {
    * be supported in the future.
    */
   extraArgs?: string[];
+  /**
+   * Search attributes to be registered with the dev server.
+   */
+  searchAttributes?: SearchAttributeKey<SearchAttributeType>[];
 }
 
 /**
@@ -509,6 +514,10 @@ export interface DevServerConfig {
    * be supported in the future.
    */
   extraArgs?: string[];
+  /**
+   * Search attributes to be registered with the dev server.
+   */
+  searchAttributes?: SearchAttributeKey<SearchAttributeType>[];
 }
 
 /**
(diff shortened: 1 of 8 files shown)


THardy98:
This is branched off #1612, will merge after that has been merged.

Only last commit is relevant for review on this PR.

THardy98:
cleanup existing code that registers search attributes on server startup

THardy98:
Can't cleanup all usage of existing code that registers search attributes on server startup (`registerDefaultCustomSearchAttributes`).

It's used in old tests that require a separate dev server (i.e. via CLI). Removing this would require users/CI to run dev server with search attributes specified. Alternatively, we can make an issue to refactor these tests to use the local testing environment.

packages/core-bridge/ts/index.ts
```diff
@@ -458,6 +459,10 @@ export interface TimeSkippingServerConfig {
    * be supported in the future.
    */
   extraArgs?: string[];
+  /**
```
mjameswh, line 462: I don't think the Time Skipping server supports the `--searchAttribute` option.

packages/test/src/helpers.ts
```diff
@@ -217,6 +217,7 @@ export class TestWorkflowEnvironment extends RealTestWorkflowEnvironment {
 // Some of our tests expect "default custom search attributes" to exists, which used to be the case
 // in all deployment with support for advanced visibility. However, this might no longer be true in
 // some environement (e.g. Temporal CLI). Use the operator service to create them if they're missing.
+// TODO: register search attributes when creating test environment (#1624)
```
mjameswh, line 220: Not sure what is the meaning of this FIXME. Remove?

packages/test/src/test-integration-workflows.ts
```diff
@@ -1303,3 +1311,45 @@ test('Count workflow executions', async (t) => {
     ],
   });
 });
+
+test('can register search attributes to dev server', async (t) => {
+  const { startWorkflow } = helpers(t);
+  const key = defineSearchAttributeKey('new-search-attr', SearchAttributeType.INT);
+  const newSearchAttribute: SearchAttributePair = { key, value: 12 };
+
+  // Ensure that search attribute is not registered on current env.
```
mjameswh, line 1320: I don't think checking that the SA isn't already registered is pertinent. The actual test (below) creates a new/distinct test server anyway. If you're really worried about this, you can append a random UUID to the SA's name, but even that seems overkill.

packages/test/src/helpers-integration.ts
```diff
@@ -89,12 +84,22 @@ export async function createTestWorkflowBundle({
   });
 }
 
+export const defaultSearchAttributes = [
```
mjameswh, line 87: Don't we need those typed SA keys anywhere else in our tests? I mean, they should probably be individually exposed as `const`, or all of them as properties on a `const object`.

packages/core-bridge/ts/index.ts
THardy98: Didn't realize that - removed.

packages/test/src/helpers.ts
THardy98: It was a reminder to myself to clean this up, removed.

packages/test/src/test-integration-workflows.ts
THardy98: Not sure why I did that given I'm creating a new test env right below... removed.

packages/test/src/helpers-integration.ts
THardy98: Could use them in some tests in `test-schedules.ts`, changed to an object

mjameswh (approved):

PR #1636, opened 2025-03-04
feat(client): added keepOriginalWorkflowId in SchedulePolicies

AllenWang314:
## What was changed

Implemented keepOriginalWorkflowId in the client scheduling policies in client package. This updates the schedule policies to match [documentation](https://typescript.temporal.io/api/interfaces/proto.temporal.api.schedule.v1.ISchedulePolicies#keeporiginalworkflowid).

## Why?

Change was made to update the SDK to reflect documentation. Code is analogous to ISchedulePolicies.pauseOnFailure in the client package.


mjameswh (approved):
Thanks! Everything's alright, merging.

Quinn-With-Two-Ns:
@mjameswh Did you confirm this option is actually implemented on the server?

mjameswh:
Good catch @Quinn-With-Two-Ns. It was [added to server](https://github.com/temporalio/temporal/pull/5447), but then [got reverted](https://github.com/temporalio/temporal/pull/5480) soon after that.

Seems like that option in the Protobuf API has been left in limbo until we decide how we actually want to implement that feature. I'll verify the status internally.

@AllenWang314 I'll close this PR, since there is no server side support for that option, but thanks anyway 😄.

PR #1637, opened 2025-03-04
chore(core): Upgrade Core

bergundy:
Upgraded sdk-core submodule and made adjustments in the bridge. In addition:
- Upgraded to Rust 2024 edition.
- Updated all core-bridge Rust dependencies.
- The new TTL attribute for the cached download defaults to unset (forever) - may be worth picking a different default though.

- [x] Handle new activation jobs and compilation errors.


THardy98 (commented):
lgtm, but i haven't had a lot of exposure to core-bridge

mjameswh (approved):
Looks good

.github/workflows/ci.yml
```diff
@@ -56,6 +56,10 @@ jobs:
       run:
         shell: bash
     steps:
+      - name: Install Rust toolchain
```
mjameswh, line 59: Not sure why you needed to add this. GHA's default runners now come with the Rust toolchain already. If you simply needed to update, could have done `cargo up` instead. Definitely not blocking though.

packages/common/src/internal-non-workflow/codec-types.ts
```diff
@@ -23,12 +23,14 @@ export type ReplaceNested<T, ToReplace, ReplaceWith> = T extends (...args: any[]
...
   : [keyof T] extends [never]
     ? T
-    : T extends { [k: string]: ToReplace }
-      ? {
-          [P in keyof T]: ReplaceNested<T[P], ToReplace, ReplaceWith>;
-        }
-      : T extends ToReplace
-        ? ReplaceWith | Exclude<T, ToReplace>
-        : {
+    : T extends Record<string, string> // Special exception for nexusHeader.
```
mjameswh, line 26: I vaguely remember struggling with this mapped type missing some transformations. This is probably the reason Roey had to modify it here, but I'm worried it may still be in some other aspects.

@THardy98 Can you please get a look at the [Type Fest](https://github.com/sindresorhus/type-fest) project; they probably have some type definition that does the same as this (name may differ). If they do, please copy that type definition and its dependencies into common/src/type-helpers.ts, and get rid of this one.

mjameswh:
@THardy98 Roey will be out of office for a few weeks. Can you please complete the work on this PR and get it merged?

Please get a look at my latest review comments and try to fix those first. I believe you should be able to push to Roey's branch. Ping me after that for another review if you indeed make more changes, or just merge it as it is, and we can fix remaining points at a later time.

packages/common/src/internal-non-workflow/codec-types.ts
THardy98: Took a look, the closest match is likely [MergeDeep](https://github.com/sindresorhus/type-fest/blob/ed8c987129fe180021348d8fdbd52e662ffd69a1/source/merge-deep.d.ts#L58), but I don't think it's exactly what we're looking for.

.github/workflows/ci.yml
THardy98: Yup - removed.

THardy98:
> Ping me after that for another review if you indeed make more changes, or just merge it as it is, and we can fix remaining points at a later time.

Small change removing rust installation from CI. Figure we should merge this now, no 1:1 type for `ReplaceNested`, if we need to make changes later, we can make them later.

PR #1639, opened 2025-03-06
add default query handler

THardy98:
Added support for to register a default query handler - a handler for queries that do not have registered query names/handlers.

1. Part of #1015
2. How was this tested:
Added small integration tests.

3. Any docs updates needed?
Maybe


packages/workflow/src/workflow.ts
```diff
@@ -1318,6 +1319,28 @@ export function setDefaultSignalHandler(handler: DefaultSignalHandler | undefine
...
+ *
+ * If this function is called multiple times for a given query name the last handler will overwrite any previous calls.
+ *
+ * @param handler a function that will handle queries for non-registered query names, or `undefined` to unset the handler.
+ */
+export function setDefaultQueryHandler(handler: DefaultQueryHandler | undefined): void {
+  const activator = assertInWorkflowContext(
+    'Workflow.setDefaultQueryHandler(...) may only be used from a Workflow Execution.'
+  );
+  if (typeof handler === 'function') {
```
mjameswh, line 1335: nit: You can combine the `if` and `else if` branches here.

packages/workflow/src/internals.ts
```diff
@@ -611,7 +617,12 @@ export class Activator implements ActivationHandler {
 
   // Intentionally non-async function so this handler doesn't show up in the stack trace
   protected queryWorkflowNextHandler({ queryName, args }: QueryInput): Promise<unknown> {
-    const fn = this.queryHandlers.get(queryName)?.handler;
+    let fn = this.queryHandlers.get(queryName)?.handler;
+    // Fallback to default query handler if no handler is registered.
+    if (fn === undefined) {
+      fn = this.defaultQueryHandler;
```
mjameswh, line 623: The default query handler will need an extra argument, the first one, that indicates the name of the query being called. See `signalWorkflowNextHandler()`.

mjameswh (changes requested):
The default query handler needs the name of the incoming query, just like we do for the default signal handler.

That's the only blocking change.

packages/workflow/src/internals.ts
THardy98: Fixed, improved the test to catch this case. Thanks

packages/workflow/src/workflow.ts
THardy98: Done.

packages/workflow/src/interfaces.ts
```diff
@@ -538,6 +538,11 @@ export type Handler<
  */
 export type DefaultSignalHandler = (signalName: string, ...args: unknown[]) => void | Promise<void>;
 
+/**
+ * A handler function accepting query calls for non-registered query names.
+ */
+export type DefaultQueryHandler = (queryName: string, ...args: any[]) => any;
```
mjameswh, line 544: nit: That should be `unknown` rather than `any` (both occurrences).

packages/workflow/src/internals.ts
```diff
@@ -611,7 +617,13 @@ export class Activator implements ActivationHandler {
 
   // Intentionally non-async function so this handler doesn't show up in the stack trace
   protected queryWorkflowNextHandler({ queryName, args }: QueryInput): Promise<unknown> {
-    const fn = this.queryHandlers.get(queryName)?.handler;
+    let fn = this.queryHandlers.get(queryName)?.handler;
+    // Fallback to default query handler if no handler is registered.
+    if (fn === undefined && this.defaultQueryHandler !== undefined) {
+      // Use non-null assertion because Typescript can't tell that defaultQueryHandler cannot be undefined here.
+      fn = (...args) => this.defaultQueryHandler!(queryName, ...args);
```
mjameswh, line 624: TS is technically right: even though you just checked that `this.defaultQueryHandler` is not undefined, you are then using that field inside a function call; there is no guarantee that `this.defaultQueryHandler` will still be valid by the time `fn` gets called. Well, we know it will, but the compiler can't figure this out.

You can avoid the TS error without a non-null assertion by copying `this.defaultQueryHandler` to a temporary variable inside the `if` block, or by using `this.defaultQueryHandler.bind(...)` instead:

```suggestion
      fn = this.defaultQueryHandler.bind(this, queryName);
```

mjameswh (approved):
Please address my two minor suggestions before merging.

packages/workflow/src/internals.ts
THardy98: Done

packages/workflow/src/interfaces.ts
THardy98: Done.

PR #1640, opened 2025-03-07
Add default update handler

THardy98:
Added ability to register a default update handler. The default handler will handle updates with names that have no registered handler.

1. Part of #1015

2. How was this tested:
Added a couple activation tests with some comments describing the semantics of update code.

3. Any docs updates needed?
Maybe


packages/test/src/test-workflows.ts
```diff
@@ -363,6 +363,40 @@ function makeSetPatchMarker(myPatchId: string, deprecated: boolean): coresdk.wor
...
+  protocolInstanceId: string,
+  name: string,
+  input: unknown
+): coresdk.workflow_activation.IWorkflowActivationJob {
+  return {
+    doUpdate: {
+      id,
+      protocolInstanceId,
+      name,
+      input: toPayloads(defaultPayloadConverter, input),
```
mjameswh, line 377: Shouldn't the second argument of `toPayloads` be an array? If so, the `input` argument should be `unknown[]`.

packages/test/src/workflows/updates-ordering.ts
```diff
@@ -0,0 +1,77 @@
...
+  args: unknown[];
+}
+
+/*
+    There's a surprising amount going on with the workflow below. Let's simplify it to just updateA and updateB 
+    (no updateC or the default) and walk through it.
+
+    1. setHandler for updateA
+    - this is all synchronous code until we run `UpdateScope.updateWithInfo(updateId, name, doUpdateImpl)`, 
+      which calls `doUpdateImpl` which is promise/async, so...
```
mjameswh, line 19: To be exact, it is sync until the point where we actually call the update handler (i.e. the `execute(input)` call in `doUpdateImpl()`. By specification, the update validation must be executed sync, and the workflow handler function must be called immediately after. Otherwise, the workflow's internal state could be modified between the validation and the execute functions, and the validation result could be no longer valid by the time the execute function is called.

packages/test/src/workflows/updates-ordering.ts
```diff
@@ -0,0 +1,77 @@
...
+    2. queue doUpdateImpl for A on node event queue: [doUpdateImplA]
```
mjameswh, line 20: AFAICS, in the case of this specific workflow, `doUpdateImplA` will not get queued on node's event queue because there's no promise that is awaited from that update handler. Same for `updateB`.

packages/test/src/test-workflows.ts
THardy98: Yes, but `input` is a rest parameter in `toPayloads`, its a variadic function. I've changed it anyways for consistency.

packages/test/src/workflows/updates-ordering.ts
THardy98: I'm confused.

> By specification, the update validation must be executed sync, and the workflow handler function must be called immediately after. Otherwise, the workflow's internal state could be modified between the validation and the execute functions, and the validation result could be no longer valid by the time the execute function is called.

This makes sense to me.

But how can the code be sync up until we call the `execute` call?
```
  /** Alias to `new UpdateScope({ id, name }).run(fn)` */
  static updateWithInfo<T>(id: string, name: string, fn: () => Promise<T>): Promise<T> {
    return new this({ id, name }).run(fn);
  }

UpdateScope.updateWithInfo(updateId, name, doUpdateImpl)
```
This is the async call at the end of `doUpdate`, it returns a promise.
`doUpdateImpl` itself is an async function as well.
How could these async calls guarantee sync execution till we execute the handler?

packages/test/src/workflows/updates-ordering.ts
THardy98: `doUpdateImplA` is async though, so I thought differently. Maybe we should coalesce this conversation into the other comment.

packages/workflow/src/internals.ts
```diff
@@ -676,8 +682,20 @@ export class Activator implements ActivationHandler {
...
     }
-    const entry = this.updateHandlers.get(name);
-    if (!entry) {
+
+    const entry =
+      this.updateHandlers.get(name) ??
+      (this.defaultUpdateHandler
+        ? {
+            handler: this.defaultUpdateHandler.bind(this, name),
+            validator: undefined,
```
mjameswh, line 691: That function should not be bound to `this`.

```suggestion
            handler: this.defaultUpdateHandler.bind(undefined, name),
```

Please also fix the same pattern on `defaultQueryHandler` at line 628 — my mistake, that was a code suggestion I made.

packages/test/src/workflows/updates-ordering.ts
THardy98: TLDR: only async when we yield (i.e. `await`, `then`, etc.), not when we call async... 🤦

packages/workflow/src/internals.ts
THardy98: done.

PR #1642, opened 2025-03-07
default activities

THardy98:
Users are now able to create default activity implementations by exporting a default activity function:
`export default async function(...) { activity logic...}`

1. Part of #1015

2. How was this tested:
Added simple integration tests.

3. Any docs updates needed?
Maybe


packages/worker/src/worker.ts
```diff
@@ -924,7 +924,10 @@ export class Worker {
                     );
 
                     const { activityType } = info;
-                    const fn = this.options.activities.get(activityType);
+                    // Use the corresponding activity if it exists, otherwise, fallback to default activity function (if exists)
+                    const fn = this.options.activities.get(activityType)
```
mjameswh, line 928: nit: don't do `.get()` twice for the same value.

```suggestion
                    const fn = this.options.activities.get(activityType) ??
                                   this.options.activities.get('default');
```

packages/test/src/activities/default-and-defined.ts
```diff
@@ -0,0 +1,13 @@
...
+  activityName?: string;
+  args: any[];
+}
+
+export async function definedActivity(...args: unknown[]): Promise<NameAndArgs> {
+  return { name: 'definedActivity', args };
+}
+
+export default async function (...args: unknown[]): Promise<NameAndArgs> {
+  return { name: 'default', activityName: 'definedActivity', args };
```
mjameswh, line 12: Why is this hardcoded? Seems like a bug to me. Look into `activityInfo()`

```
activityName: 'definedActivity'
```

packages/test/src/test-default-activity.ts
```diff
@@ -0,0 +1,35 @@
...
+  const worker = await createWorker({
+    activities,
+  });
+
+  await worker.runUntil(async () => {
+    const activityArgs = ['test', 'args'];
+    const res = await executeWorkflow(workflowWithMaybeDefinedActivity, {
+      args: [false, activityArgs],
+    });
+    t.deepEqual(res, { name: 'default', activityName: 'definedActivity', args: activityArgs });
```
mjameswh, line 18:
```suggestion
    t.deepEqual(res, { name: 'default', activityName: 'nonExistentActivity', args: activityArgs });
```

packages/worker/src/worker.ts
THardy98: done

packages/test/src/activities/default-and-defined.ts
THardy98: Indeed a bug (or at least not representative of what we're actually trying to test).
I suspect it was some lazy copy pasting on my part. Fixed

packages/test/src/test-default-activity.ts
THardy98: Fixed

mjameswh (approved):

PR #1644, opened 2025-03-11
[WIP] Special behaviour for temporal prefixes

### LINKED ISSUE #1599: [Feature Request] Special behavior for Temporal built-in prefixes

See https://github.com/temporalio/features/issues/576

THardy98:
## What was changed
Special behavior handling reserved temporal prefixes. Prevent users from:
- registering workflows, activities, task queues, signals/updates/queries with a reserved prefix
- calling default handlers with reserved names
- calling internal queries without going through interceptors

## Checklist
Some remaining items:
- test for default signal handler & reserved names needs to be fixed
- need to reserving prefixes from workflows,
- waiting for default update handler to be merged to prevent default update handler to be called with reserved names

1. Closes #1599

2. How was this tested:
unit/integration tests

3. Any docs updates needed?
I'm not sure.

### DIFF
--- a/packages/common/src/reserved.ts
+++ b/packages/common/src/reserved.ts
@@ -1,0 +1,20 @@
+export const TEMPORAL_RESERVED_PREFIX = '__temporal_'
+export const STACK_TRACE_RESERVED_PREFIX = '__stack_trace'
+export const ENHANCED_STACK_TRACE_RESERVED_PREFIX = '__enhanced_stack_trace'
+
+export const reservedPrefixes = [TEMPORAL_RESERVED_PREFIX, STACK_TRACE_RESERVED_PREFIX, ENHANCED_STACK_TRACE_RESERVED_PREFIX]
+
+export function throwIfReservedName(type: string, name: string) {
+    const prefix = isReservedName(name)
+    if (prefix) {
+        throw Error(`Cannot register ${type} name: '${name}', with reserved prefix: '${prefix}'`)
+    }
+}
+
+export function isReservedName(name: string): string | undefined {
+    for (const prefix of reservedPrefixes) {
+        if (name.startsWith(prefix)) {
+            return prefix
+        }
+    }
+}
--- a/packages/test/src/test-workflows.ts
+++ b/packages/test/src/test-workflows.ts
@@ -23,6 +23,7 @@ import { parseWorkflowCode } from '@temporalio/worker/lib/worker';
 import * as activityFunctions from './activities';
 import { cleanStackTrace, REUSE_V8_CONTEXT, u8 } from './helpers';
 import { ProcessedSignal } from './workflows';
+import { reservedPrefixes } from '@temporalio/common/src/reserved';
 
 export interface Context {
   workflow: VMWorkflow | ReusableVMWorkflow;
@@ -2525,6 +2526,80 @@ test('Signals/Updates/Activities/Timers - Trace promises completion order - 1.11
         ],
         [SdkFlags.ProcessWorkflowActivationJobsAsSingleBatch]
       )
+    );
+  }
+});
+
+test('Default query handler fail activations with reserved names - workflowWithDefaultHandlers', async (t) => {
+  const { workflowType } = t.context;
+  
+  await activate(
+    t,
+    makeActivation(
+      undefined, 
+      makeInitializeWorkflowJob(workflowType),
+    ),
+  );
+
+  for (const prefix of reservedPrefixes) {
+    const completion = await activate(
+      t,
+      makeActivation(
+        undefined, 
+        makeQueryWorkflowJob("1", prefix + '_query')
+      ),
+    );
+
+    compareCompletion(
+      t,
+      completion,
+      {
+        failed: {
+          failure: {
+            ...completion.failed?.failure,
+            // We only care about the error message.
+            message: `Cannot register query name: \'${prefix}_query\', with reserved prefix: \'${prefix}\'`,
+          }
+        }
+      }
+    );
+  }
+});
+
+test('Default signal handler fail activations with reserved names - workflowWithDefaultHandlers', async (t) => {
+  const { workflowType } = t.context;
+  
+  await activate(
+    t,
+    makeActivation(
+      undefined, 
+      makeInitializeWorkflowJob(workflowType),
+    ),
+  );
+
+  for (const prefix of reservedPrefixes) {
+    const job = makeSignalWorkflowJob(prefix + '_signal', []);
+
+    const completion = await activate(
+      t,
+      makeActivation(
+        undefined,
+        job
+      ),
+    );
+
+    compareCompletion(
+      t,
+      completion,
+      {
+        failed: {
+          failure: {
+            ...completion.failed?.failure,
+            // We only care about the error message.
+            message: `Cannot register signal name: \'${prefix}_signal\', with reserved prefix: \'${prefix}\'`,
+          }
+        }
+      }
     );
   }
 });
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -18,6 +18,7 @@ import { InjectedSinks } from './sinks';
 import { MiB } from './utils';
 import { WorkflowBundleWithSourceMap } from './workflow/bundler';
 import { asNativeTuner, WorkerTuner } from './worker-tuner';
+import { throwIfReservedName } from '@temporalio/common/src/reserved';
 
 export type { WebpackConfiguration };
 
@@ -822,6 +823,9 @@ export function compileWorkerOptions(rawOpts: WorkerOptions, logger: Logger): Co
   }
 
   const activities = new Map(Object.entries(opts.activities ?? {}).filter(([_, v]) => typeof v === 'function'));
+  for (const activityName of activities.keys()) {
+    throwIfReservedName('activity', activityName)
+  }
   const tuner = asNativeTuner(opts.tuner, logger);
 
   return {
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -88,6 +88,7 @@ import { ThreadedVMWorkflowCreator } from './workflow/threaded-vm';
 import { VMWorkflowCreator } from './workflow/vm';
 import { WorkflowBundleWithSourceMapAndFilename } from './workflow/workflow-worker-thread/input';
 import { CombinedWorkerRunError, GracefulShutdownPeriodExpiredError, PromiseCompletionTimeoutError } from './errors';
+import { throwIfReservedName } from '@temporalio/common/src/reserved';
 
 export { DataConverter, defaultPayloadConverter };
 
@@ -462,6 +463,7 @@ export class Worker {
    * This method initiates a connection to the server and will throw (asynchronously) on connection failure.
    */
   public static async create(options: WorkerOptions): Promise<Worker> {
+    throwIfReservedName('task queue', options.taskQueue)
     const logger = withMetadata(Runtime.instance().logger, {
       sdkComponent: SdkComponent.worker,
       taskQueue: options.taskQueue ?? 'default',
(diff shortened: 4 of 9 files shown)


THardy98:
In this change, users are prevented from registering signals/updates/queries with reserved names from the *handler*.

Users can still `defineQuery`, `defineSignal`, `defineUpdate` that have reserved names. The reason for this is because we use `defineQuery` to create the reserved queries ourselves. 

Alternatively, we can do the blocking in `define<X>` and create our internal queries/signals/updates using internal functions. Or we can do both (blocking in the `define` and the handler). I picked one because it seemed sufficient.

THardy98:
this is blocked on https://github.com/temporalio/sdk-typescript/pull/1640 so we can handle the same prefix cases for updates as well

packages/test/src/test-workflows.ts
```diff
@@ -2528,3 +2534,83 @@ test('Signals/Updates/Activities/Timers - Trace promises completion order - 1.11
     );
   }
 });
+
+test('Default query handler fail activations with reserved names - workflowWithDefaultHandlers', async (t) => {
```
mjameswh, line 2538: Why did you fallback to unit tests for default handlers?

packages/workflow/src/internals.ts
```diff
@@ -619,6 +625,8 @@ export class Activator implements ActivationHandler {
   protected queryWorkflowNextHandler({ queryName, args }: QueryInput): Promise<unknown> {
     let fn = this.queryHandlers.get(queryName)?.handler;
     if (fn === undefined && this.defaultQueryHandler !== undefined) {
+      // Do not call default query handler with reserved query name.
+      throwIfReservedName('query', queryName);
```
mjameswh, line 629: I think we should not be throwing this way here. Code below is clearly designed to make sure that we are not throwing synchronously. Also, UI expects that we'll answer back with a list of accepted query types.

Actually, I'd just skip the default handler if query name is reserved. We'll still throw below, but we'll then be going the same code path as if that wasn't a reserved type name.

packages/test/src/test-workflows.ts
THardy98: this was a while ago, i'm not sure why, i think because it seemed easier at the time to trigger these failures by sending invalid wft?

in any case, i've moved all the testing to `test-integration-workflows` under a few integration tests.

packages/workflow/src/internals.ts
```diff
@@ -797,6 +816,8 @@ export class Activator implements ActivationHandler {
     if (fn) {
       return await fn(...args);
     } else if (this.defaultSignalHandler) {
+      // Do not call default signal handler with reserved signal name.
+      throwIfReservedName('signal', signalName);
```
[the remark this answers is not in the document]
THardy98: yup, moved reserved prefix handling to `signalWorkflow`, it throws if a reserved query tries to use a default signal handler

packages/workflow/src/internals.ts
THardy98: moved reserved prefix handling to the parent function `queryWorkflow`, it throws if a reserved query tries to use a default query handler

packages/common/src/reserved.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export function throwIfReservedName(type: string, name: string): void {
+  const prefix = maybeGetReservedPrefix(name);
+  if (prefix) {
+    throw new ReservedPrefixError(type, name, prefix);
+  }
+}
+
+export function maybeGetReservedPrefix(name: string): string | undefined {
```
mjameswh: This check needs to vary based on type; not all reserved name patterns apply to all types. Also, not all reserved name patterns are prefixes; `__stack_trace` and `__enhanced_stack_trace` are exact matches.

With that in mind, I don't think using a loop makes sense. Also, I have no strong opinion about having the check logic externalized to this file, but it looks that's not what we did in other SDKs.

packages/common/src/reserved.ts
```diff
@@ -0,0 +1,31 @@
...
+export const STACK_TRACE_RESERVED_PREFIX = '__stack_trace';
+export const ENHANCED_STACK_TRACE_RESERVED_PREFIX = '__enhanced_stack_trace';
+
+export const reservedPrefixes = [
+  TEMPORAL_RESERVED_PREFIX,
+  STACK_TRACE_RESERVED_PREFIX,
+  ENHANCED_STACK_TRACE_RESERVED_PREFIX,
+];
+
+export class ReservedPrefixError extends Error {
```
mjameswh: Is there a good reason to make a custom error type for this, rather than just throwing a `TypeError`? It's not like users will want to catch that error and do `instanceof` on it.

But if we'd want to keep that error class (I don't think so), it should have the `@SymbolBasedInstanceOfError('...')` annotation, and be exported from `index.ts`.

packages/common/src/reserved.ts
```diff
@@ -0,0 +1,31 @@
...
+  constructor(type: string, name: string, prefix: string) {
+    super(`Cannot use ${type} name: '${name}', with reserved prefix: '${prefix}'`);
+    this.name = 'ReservedPrefixError';
+  }
+}
+
+export function throwIfReservedName(type: string, name: string): void {
```
mjameswh: `type` should be a union of expected values, rather than any string.

packages/worker/src/worker-options.ts
```diff
@@ -953,6 +954,9 @@ export function compileWorkerOptions(
   }
 
   const activities = new Map(Object.entries(opts.activities ?? {}).filter(([_, v]) => typeof v === 'function'));
+  for (const activityName of activities.keys()) {
```
mjameswh, line 965: We also need to check sinks.

packages/worker/src/worker.ts
```diff
@@ -467,6 +468,7 @@ export class Worker {
    * This method initiates a connection to the server and will throw (asynchronously) on connection failure.
    */
   public static async create(options: WorkerOptions): Promise<Worker> {
+    throwIfReservedName('task queue', options.taskQueue);
```
mjameswh, line 474: At this point, `taskQueue` can be undefined. Won't that cause an NPE in throwIfServedName?

packages/common/src/reserved.ts
```diff
@@ -0,0 +1,31 @@
+export const TEMPORAL_RESERVED_PREFIX = '__temporal_';
+export const STACK_TRACE_RESERVED_PREFIX = '__stack_trace';
```
mjameswh: That's not a prefix. Same for next line.

packages/common/src/reserved.ts
THardy98: removed

packages/common/src/reserved.ts
THardy98: helper was removed

packages/worker/src/worker-options.ts
THardy98: added

packages/worker/src/worker.ts
THardy98: I added a check but the WorkerOptions interface says that taskQueue can only be string

packages/common/src/reserved.ts
THardy98: renamed

packages/workflow/src/internals.ts
```diff
@@ -707,6 +719,11 @@ export class Activator implements ActivationHandler {
       throw new TypeError('Missing activation update protocolInstanceId');
     }
 
+    // If update has __temporal_ prefix but no handler exists, throw error
```
mjameswh, line 725: We'd need to skip interceptors on built-in updates, same as we did above for queries. We don't have any built-in update anyway, so no need to actually implement this just now, but we should at least leave a trace so we don't forget about this if ever we add one. Just add a comment here. And actually, you could even just remove the ` && !this.updateHandlers.get(name)` part of the condition, as we know it will never evaluate to true unless code around is modified.

packages/workflow/src/internals.ts
```diff
@@ -859,6 +876,13 @@ export class Activator implements ActivationHandler {
       throw new TypeError('Missing activation signalName');
     }
```
mjameswh, line 888: Same comment as update.

packages/workflow/src/internals.ts
THardy98: I decided to just go ahead and add the "is internal" check to skip interceptors, for both updates and signals. 

We don't have any builtin updates/signals now, but it's a trivial check to add, especially if we were planning to add a comment to remind our future selves anyway.

packages/common/src/reserved.ts
```diff
@@ -0,0 +1,26 @@
+export const TEMPORAL_RESERVED_PREFIX = '__temporal_';
+export const STACK_TRACE_RESERVED_NAME = '__stack_trace';
```
mjameswh:
```suggestion
export const STACK_TRACE_QUERY_NAME = '__stack_trace';
```

packages/common/src/reserved.ts
```diff
@@ -0,0 +1,26 @@
...
+export const ENHANCED_STACK_TRACE_RESERVED_NAME = '__enhanced_stack_trace';
```
mjameswh:
```suggestion
export const ENHANCED_STACK_TRACE_QUERY_NAME = '__enhanced_stack_trace';
```

packages/worker/src/worker-options.ts
```diff
@@ -953,6 +954,20 @@ export function compileWorkerOptions(
...
+    throwIfReservedName('activity', activityName);
+  }
+
+  // Validate sink names to ensure they don't use reserved prefixes/names
+  if (opts.sinks) {
+    for (const sinkName of Object.keys(opts.sinks)) {
+      // Allow internal sinks used by the SDK
+      if (sinkName !== '__temporal_logger' && sinkName !== '__temporal_metrics') {
```
mjameswh: This completely defeats the purpose. Instead, move the check to the beginning of the `compileOptions` function, before calling `addDefaultWorkerOptions`. At that point, we haven't yet added built-in sinks, so you will no longer need those exclusions.

packages/worker/src/worker.ts
mjameswh: > the WorkerOptions interface says that taskQueue can only be string

Ah, that's right. It is optional in the case of `ReplayWorkerOptions`, but mandatory on `WorkerOptions`. Just ignore my previous comment. Don't add a useless check for this.

packages/test/src/test-integration-workflows.ts
```diff
@@ -1440,3 +1455,232 @@ test('Workflow can return root workflow', async (t) => {
     t.deepEqual(result, 'empty test-root-workflow-length');
   });
 });
+
+const reservedNames = [TEMPORAL_RESERVED_PREFIX, STACK_TRACE_RESERVED_NAME, ENHANCED_STACK_TRACE_RESERVED_NAME];
```
mjameswh: I'm ok with what's here, but just giving some suggestions how you could have made those tests simpler:
- It's not really important either the error message is "... with reserved prefix" or "... which is a reserved name", so you could have simply made that a regexp, like:
  ```
      {
        name: 'TypeError',
        message: /Cannot use activity name: .*/
      }
  ```
- In the `reservedNames` array, you could have included `__temporal_test` (the composed form), instead of only the prefix; then, in tests below, you would no longer need the `name === TEMPORAL_RESERVED_PREFIX ? name + '_test' : name` expression.

packages/worker/src/worker-options.ts
THardy98: Ah overlooked that `addDefaultWorkerOptions` registers sinks. Moved.

mjameswh (approved):
I had one last thing, but fixed it myself.

Should be ready to merge now, just waiting for CI.

mjameswh:
I had to also remove usage of the `__temporal_logger` sink in our internal tests, as well as in the deprecated `defaultSinks()` stub.

PR #1657, opened 2025-03-21
User metadata

### LINKED ISSUE #1544: [Feature Request] Support user metadata

See https://github.com/temporalio/features/issues/486. Specifically, after https://github.com/temporalio/sdk-core/issues/830:

* Accept optional "static summary " and "static details" on all ways to start a workflow (start, signal with start, schedule, and child workflow) and populate user metadata with them
* Accept optional "summary" when creating timers and populate user metadata with it
* Accept optional "summary" when creating activities and populate user metadata with it
* Handle `__temporal_workflow_metadata` query (see features issue and https://github.com/temporalio/sdk-go/pull/1597)
* Mark experimental

THardy98:
## What was changed
Added functionality for users to provide `staticDetails` and `staticSummary` metadata to workflow start commands.
These can be retrieved via `describe` on the workflow handler and are visibile via the UI/CLI.

Users can set `currentDetails` - a mutable details string within the workflow - via `SetCurrentDetails` and retrieve it via `GetCurrentDetails`. This is visible via the UI/CLI and the `__temporal_workflow_metadata` internal query.

Users can add a `summary` (short string description) when setting a timer and when starting an activity.

A notable part of the change is the addition of `runWithOptions` to activities. Activities can now override their activity options at call time, like so:
```
* // Setup Activities from module exports
 * const { httpGet, otherActivity } = proxyActivities<typeof activities>({
 *   startToCloseTimeout: '30 minutes',
 * });
 *
 * // Use activities with default options
 * const result1 = await httpGet('http://example.com');
 *
 * // Override options for specific activity calls
 * const result2 = await httpGet.executeWithOptions({
 *   staticSummary: 'Fetches data from external API',
 *   scheduleToCloseTimeout: '5m'
 * }, ['http://api.example.com']);
 *
 * const result3 = await otherActivity.executeWithOptions({
 *   staticSummary: 'Processes the fetched data',
 *   taskQueue: 'special-task-queue'
 * }, [data]);
```

1. Closes #1544

2. How was this tested:
Couple integration tests

3. Any docs updates needed?
Maybe


packages/client/src/schedule-helpers.ts
```diff
@@ -240,6 +245,7 @@ export async function encodeScheduleAction(
   action: CompiledScheduleAction,
   headers: Headers
 ): Promise<temporal.api.schedule.v1.IScheduleAction> {
+  const jsonConverter = new JsonPayloadConverter();
```
mjameswh, line 248: Use the data converters configured on the client (i.e. custom payload converter and codec).

packages/client/src/schedule-helpers.ts
```diff
@@ -312,6 +322,8 @@ export async function decodeScheduleAction(
   pb: temporal.api.schedule.v1.IScheduleAction
 ): Promise<ScheduleDescriptionAction> {
   if (pb.startWorkflow) {
+    const jsonConverter = new JsonPayloadConverter();
```
mjameswh, line 325: Use the data converters configured on the client (i.e. custom payload converter and codec).

packages/client/src/workflow-client.ts
```diff
@@ -1196,6 +1197,7 @@ export class WorkflowClient extends BaseClient {
   protected async _signalWithStartWorkflowHandler(input: WorkflowSignalWithStartInput): Promise<string> {
     const { identity } = this.options;
     const { options, workflowType, signalName, signalArgs, headers } = input;
+    const jsonConverter = new JsonPayloadConverter();
```
mjameswh, line 1200: Use the data converters configured on the client (i.e. custom payload converter and codec).

packages/client/src/workflow-client.ts
```diff
@@ -1225,6 +1227,10 @@ export class WorkflowClient extends BaseClient {
           : undefined,
       cronSchedule: options.cronSchedule,
       header: { fields: headers },
+      userMetadata: {
+        summary: jsonConverter.toPayload(options?.staticSummary),
```
mjameswh, line 1231: You will need to use `await encodeToPayload(...)` instead here and everything else in the client package. Not in Workflows and Worker, though.

packages/workflow/src/workflow.ts
```diff
@@ -81,6 +84,27 @@ export function addDefaultWorkflowOptions<T extends Workflow>(
   };
 }
 
+function addUserMetadata(userMetadata?: UserMetadata): temporal.api.sdk.v1.IUserMetadata | undefined {
+  if (userMetadata == null) {
+    return undefined;
+  }
+
+  const jsonConverter = new JsonPayloadConverter();
```
mjameswh, line 92: Don't create a JSON payload converter. Use the payload converter from the activator.

packages/workflow/src/interceptors.ts
```diff
@@ -80,6 +80,7 @@ export interface ActivityInput {
   readonly options: ActivityOptions;
   readonly headers: Headers;
   readonly seq: number;
+  readonly cmdOpts?: WorkflowCommandOptions;
```
mjameswh, line 83: I don't think we should have this `WorkflowCommandOptions` type. I even wonder if we should inline `summary` and `details` directly in `ActivityInput` (without the extra `UserMetadata` nesting). Or add these two to `ActivityOptions`, though that might pose some minor challenge regarding the type provided to the `proxyActivities()` function.

mjameswh: What did we do in other SDKs?

packages/workflow/src/workflow.ts
```diff
@@ -127,8 +152,9 @@ function timerNextHandler(input: TimerInput) {
  *
  * @param ms sleep duration - number of milliseconds or {@link https://www.npmjs.com/package/ms | ms-formatted string}.
  * If given a negative number or 0, value will be set to 1.
+ * @param summary a short summary/description of the timer. Can serve as a timer ID.
  */
-export function sleep(ms: Duration): Promise<void> {
+export function sleep(ms: Duration, summary?: string): Promise<void> {
```
mjameswh, line 157: I'm worried this will evolve badly. What if we then need to add `description`, and then other properties? Instead, I think I'd define a `TimerOptions` type, and take this as a second argument here. Then, we can easily add whatever properties we need.

mjameswh: Out of curiosity, what do we do in other SDKs?

packages/workflow/src/workflow.ts
```diff
@@ -271,7 +308,12 @@ async function scheduleLocalActivityNextHandler({
  * Schedule an activity and run outbound interceptors
  * @hidden
  */
-export function scheduleActivity<R>(activityType: string, args: any[], options: ActivityOptions): Promise<R> {
+export function scheduleActivity<R>(
+  activityType: string,
+  args: any[],
+  options: ActivityOptions,
+  summary?: string
```
mjameswh, line 315: Yeah… So, looking at this one, I really feel this should be part of `ActivityOptions`.

packages/workflow/src/workflow.ts
```diff
@@ -499,6 +544,18 @@ export type ActivityInterfaceFor<T> = {
...
+/**
+ * Extends ActivityInterfaceFor to include the withSummaries method
+ */
+export type ActivityInterfaceWithSummaries<A> = ActivityInterfaceFor<A> & {
+  /**
+   * Provide descriptive summaries for activities
+   * @param summaries Record mapping activity names to their summary descriptions
+   * @returns A new proxy with the provided summaries
+   */
+  withSummaries(summaries: Record<string, string>): ActivityInterfaceFor<A>;
```
mjameswh, line 556: :thinking: That feels a bit awkward, and would prevent having an activity named `withSummaries` (ok, that's arguably very unlikely, but still means this approach would evolve badly).

I think we could get a better DX by using something similar to this, but at the activity-level rather than at the _activities_-level (mind the plural). That would also allow any extra option to be set at the call site, which is a recurring request.

For example, we could have:

```
const { echo } = proxyActivities<MyActivities>({ startToCloseTimeout: '5m' });

// No extra options
await echo(myArg1, myArg2);

// With extra options, higher order function style
await echo.withOptions({ summary: "..." })(myArg1, myArg2);

// Some other possible variants to be considered
await echo.withOptions({ summary: "..." }, myArg1, myArg2);
await echo.withOptions({ summary: "..." }, [myArg1, myArg2]);
```

I'll let you think you a bit about this, but we should probably include others in this discussion.

packages/common/src/workflow-options.ts
```diff
@@ -190,6 +190,21 @@ export interface BaseWorkflowOptions {
    * by {@link typedSearchAttributes}.
    */
   typedSearchAttributes?: SearchAttributePair[] | TypedSearchAttributes;
+
+  /**
+   * General fixed details for this workflow execution that may appear in UI/CLI.
+   * This can be in Temporal markdown format and can span multiple lines.
+   *
+   * @experimental
```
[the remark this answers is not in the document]
THardy98: I believe i've done so in all the relevant places

packages/client/src/schedule-helpers.ts
THardy98: yup, done

packages/client/src/schedule-helpers.ts
THardy98: yup, done

packages/client/src/workflow-client.ts
THardy98: yup, done

packages/client/src/workflow-client.ts
THardy98: yup, done

packages/workflow/src/workflow.ts
THardy98: yup, done

packages/workflow/src/interceptors.ts
THardy98: Typically they are inline, I've changed it as such (aside from using `ActivityOptions` for summary/details for an activity, and `TimerOptions` as you suggested below)

packages/workflow/src/workflow.ts
THardy98: typically we inline, but went with `TimerOptions` here given your reasoning

packages/workflow/src/workflow.ts
THardy98: yup, done

packages/workflow/src/workflow.ts
THardy98: opted for:
`await echo.runWithOptions(options, args);`

to allow users to override their default activity options

packages/workflow/src/interceptors.ts
```diff
@@ -109,10 +109,31 @@ export interface StartChildWorkflowExecutionInput {
   readonly seq: number;
 }
 
+/**
+ * User metadata that can be attached to workflow commands.
+ *
+ * Current used for:
```
mjameswh:
```suggestion
 * Currently used for:
```

packages/workflow/src/interceptors.ts
```diff
@@ -109,10 +109,31 @@ export interface StartChildWorkflowExecutionInput {
...
+ * - startTimer, scheduleActivity/scheduleLocalActivity commands
```
mjameswh:
```suggestion
 * - `StartTimer`, `ScheduleActivity` and `ScheduleLocalActivity` commands
```

packages/workflow/src/workflow.ts
```diff
@@ -112,11 +113,14 @@ function timerNextHandler(input: TimerInput) {
     }
     activator.pushCommand({
       startTimer: {
-        seq: input.seq,
-        startToFireTimeout: msToTs(input.durationMs),
+        seq,
+        startToFireTimeout: msToTs(durationMs),
+      },
+      userMetadata: options && {
```
mjameswh, line 119:
```suggestion
      userMetadata: options?.summary && {
```

Then you can remove the inner terniary

packages/workflow/src/workflow.ts
```diff
@@ -112,11 +113,14 @@ function timerNextHandler(input: TimerInput) {
...
+        summary: options.summary ? activator.payloadConverter.toPayload(options.summary) : undefined,
```
mjameswh: What if `summary` is an empty string? As it is now, it will be replaced by `undefined`. Not sure what would be the desired result.

mjameswh: If we want to keep empty strings, then it might be cleaner to use some `toOptionalPayload` helper.

mjameswh: And if we don't care about empty strings, then should instead move the terniary out of this object:

```
      userMetadata: options?.summary
        ? {
            summary: activator.payloadConverter.toPayload(options.summary),
          }
        : undefined,
```

packages/workflow/src/workflow.ts
```diff
@@ -130,8 +134,9 @@ function timerNextHandler(input: TimerInput) {
  *
  * @param ms sleep duration - number of milliseconds or {@link https://www.npmjs.com/package/ms | ms-formatted string}.
  * If given a negative number or 0, value will be set to 1.
+ * @param summary a short summary/description of the timer. Can serve as a timer ID.
```
mjameswh: `@param summary` is boggus.

packages/workflow/src/workflow.ts
```diff
@@ -501,7 +517,26 @@ export const NotAnActivityMethod = Symbol.for('__TEMPORAL_NOT_AN_ACTIVITY_METHOD
  * ```
  */
 export type ActivityInterfaceFor<T> = {
-  [K in keyof T]: T[K] extends ActivityFunction ? T[K] : typeof NotAnActivityMethod;
+  [K in keyof T]: T[K] extends ActivityFunction ? ActivityFunctionWithOptions<T[K]> : typeof NotAnActivityMethod;
+};
+
+/**
+ * Extends ActivityInterfaceFor to include the withOptions method
```
mjameswh, line 524:
```suggestion
 * Extends ActivityInterfaceFor to include the `runWithOptions` method
```

packages/workflow/src/workflow.ts
```diff
@@ -501,7 +517,26 @@ export const NotAnActivityMethod = Symbol.for('__TEMPORAL_NOT_AN_ACTIVITY_METHOD
...
+ */
+export type ActivityInterfaceWithOptions<A> = ActivityInterfaceFor<A> & {
+  runWithOptions(options: ActivityOptions, args: any[]): Promise<unknown>;
```
mjameswh, line 527: `@experimental`

packages/workflow/src/workflow.ts
```diff
@@ -330,6 +345,7 @@ export async function scheduleLocalActivity<R>(
         seq,
         attempt,
         originalScheduleTime,
+        ...(summary !== undefined && { cmdOpts: { userMetadata: { summary } } }),
```
mjameswh: Left over? I don't see any other traces of `cmdOpts`.

packages/workflow/src/workflow.ts
```diff
@@ -301,7 +315,8 @@ export function scheduleActivity<R>(activityType: string, args: any[], options:
 export async function scheduleLocalActivity<R>(
   activityType: string,
   args: any[],
-  options: LocalActivityOptions
+  options: LocalActivityOptions,
+  summary?: string
```
mjameswh: summary should be part of options, same as for regular activities

packages/workflow/src/workflow.ts
```diff
@@ -501,7 +517,26 @@ export const NotAnActivityMethod = Symbol.for('__TEMPORAL_NOT_AN_ACTIVITY_METHOD
...
  */
 export type ActivityInterfaceFor<T> = {
-  [K in keyof T]: T[K] extends ActivityFunction ? T[K] : typeof NotAnActivityMethod;
+  [K in keyof T]: T[K] extends ActivityFunction ? ActivityFunctionWithOptions<T[K]> : typeof NotAnActivityMethod;
+};
+
+/**
+ * Extends ActivityInterfaceFor to include the withOptions method
+ */
+export type ActivityInterfaceWithOptions<A> = ActivityInterfaceFor<A> & {
```
mjameswh: That type seems completely wrong. It has no reason to exist. You can't call `runWithOptions` on the activity proxy object itself.

packages/workflow/src/workflow.ts
```diff
@@ -501,7 +517,26 @@ export const NotAnActivityMethod = Symbol.for('__TEMPORAL_NOT_AN_ACTIVITY_METHOD
...
+};
+
+export type ActivityFunctionWithOptions<T extends ActivityFunction> = T & {
+  /**
+   * Run the activity, overriding its existing options with the
+   * provided options.
+   *
+   * @param options ActivityOptions
+   * @param args: list of arguments
+   * @returns return value of the activity
```
mjameswh, line 546: `@experimental`

packages/workflow/src/workflow.ts
```diff
@@ -538,29 +587,45 @@ export type ActivityInterfaceFor<T> = {
...
+
+        function activityProxyFunction(...args: unknown[]): Promise<unknown> {
+          return scheduleActivity(activityType as string, args, options);
+        }
+
+        activityProxyFunction.runWithOptions = function (
+          overrideOptions: ActivityOptions,
+          args: any[]
+        ): Promise<unknown> {
+          return scheduleActivity(activityType, args, { ...options, ...overrideOptions });
```
mjameswh: Not so fast. You will need to merge recursively on retry policy and priority.

packages/workflow/src/workflow.ts
```diff
@@ -573,25 +638,38 @@ export function proxyActivities<A = UntypedActivities>(options: ActivityOptions)
...
   if (options === undefined) {
     throw new TypeError('options must be defined');
   }
   // Validate as early as possible for immediate user feedback
   validateLocalActivityOptions(options);
-  return new Proxy(
-    {},
-    {
+
+  function createActivityProxy(options: ActivityOptions): ActivityInterfaceWithOptions<A> {
```
mjameswh: Why did you added this inner function, rather than create the proxy directly?

packages/workflow/src/workflow.ts
```diff
@@ -573,25 +638,38 @@ export function proxyActivities<A = UntypedActivities>(options: ActivityOptions)
...
+    return new Proxy({} as ActivityInterfaceWithOptions<A>, {
       get(_, activityType) {
         if (typeof activityType !== 'string') {
           throw new TypeError(`Only strings are supported for Activity types, got: ${String(activityType)}`);
         }
-        return function localActivityProxyFunction(...args: unknown[]) {
-          return scheduleLocalActivity(activityType, args, options);
+
+        function activityProxyFunction(...args: unknown[]): Promise<unknown> {
```
mjameswh: That function name may be visible in stack traces on the workflow side. Better keep the "localActivity" name

packages/workflow/src/workflow.ts
```diff
@@ -573,25 +638,38 @@ export function proxyActivities<A = UntypedActivities>(options: ActivityOptions)
...
+          return scheduleLocalActivity(activityType as string, args, options);
+        }
+
+        activityProxyFunction.runWithOptions = function (
+          overrideOptions: ActivityOptions,
+          args: any[]
+        ): Promise<unknown> {
+          return scheduleLocalActivity(activityType, args, { ...options, ...overrideOptions });
```
mjameswh: Same as before, need to handle recursive options merging on retryOptions and priority.

packages/common/src/activity-options.ts
```diff
@@ -124,6 +124,14 @@ export interface ActivityOptions {
    */
   versioningIntent?: VersioningIntent;
 
+  /**
+   * A single-line fixed summary for this workflow execution that may appear in the UI/CLI.
+   * This can be in single-line Temporal markdown format.
+   *
+   * @experimental User metadata is a new API and suspectible to change.
+   */
+  staticSummary?: string;
```
mjameswh: I think that should be only `summary`, no `static`. What's the terminology in other SDKs?

packages/common/src/activity-options.ts
```diff
@@ -192,4 +200,12 @@ export interface LocalActivityOptions {
...
    */
   cancellationType?: ActivityCancellationType;
+
+  /**
+   * A single-line fixed summary for this workflow execution that may appear in the UI/CLI.
+   * This can be in single-line Temporal markdown format.
+   *
+   * @experimental User metadata is a new API and suspectible to change.
+   */
+  staticSummary?: string;
```
mjameswh: Same

packages/workflow/src/workflow.ts
THardy98: renamed to `executeWithOptions`

packages/workflow/src/interceptors.ts
THardy98: done

packages/workflow/src/interceptors.ts
THardy98: done

packages/workflow/src/workflow.ts
THardy98: I think having an empty `summary` string replaced with `undefined` is reasonable. I don't see a reason why having an empty summary would be useful (or any more useful than undefined).

packages/workflow/src/workflow.ts
THardy98: yup from old commit
replaced with `* @param options optional timer options for additional configuration` not sure if that's entirely descriptive..

packages/workflow/src/workflow.ts
THardy98: yup, removed

packages/workflow/src/workflow.ts
THardy98: also from old commit, removed
uses staticSummary from options now

packages/workflow/src/workflow.ts
THardy98: Yup this was conflated with `ActivityFunctionWithOptions` between changes.

packages/workflow/src/workflow.ts
THardy98: added

packages/workflow/src/workflow.ts
THardy98: added a `deepMerge` utility in `object-helpers.ts` for use here (and possibly elsewhere in the future)

packages/workflow/src/workflow.ts
THardy98: no need, removed the inner function

packages/workflow/src/workflow.ts
THardy98: reverted name

packages/workflow/src/workflow.ts
THardy98: added a `deepMerge` utility in `object-helpers.ts` for use here (and possibly elsewhere in the future)

packages/common/src/activity-options.ts
THardy98: a bit of a mix, but generally `static` prefix for workflow start methods, otherwise no `static`, i'ved opted for the same

packages/common/src/activity-options.ts
THardy98: no `static` prefix for activity options, changed

packages/common/src/activity-options.ts
```diff
@@ -124,6 +124,14 @@ export interface ActivityOptions {
    */
   versioningIntent?: VersioningIntent;
 
+  /**
+   * A fixed, single-line fixed summary for this workflow execution that may appear in the UI/CLI.
```
mjameswh: You seem a bit insistant on "fixed" 😄

packages/common/src/activity-options.ts
```diff
@@ -192,4 +200,12 @@ export interface LocalActivityOptions {
    * - `ABANDON` - Do not request cancellation of the activity and immediately report cancellation to the workflow.
    */
   cancellationType?: ActivityCancellationType;
+
+  /**
+   * A fixed, single-line fixed summary for this workflow execution that may appear in the UI/CLI.
```
mjameswh: fixed twice?

THardy98: removed the 2nd one

packages/common/src/activity-options.ts
THardy98: removed the 2nd one

packages/common/src/user-metadata.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+export interface UserMetadata {
+  /** @experimental A fixed, single line summary of the command's purpose */
+  staticSummary?: string;
+  /** @experimental Fixed additional details about the command for longer-text description, can span multiple lines */
+  staticDetails?: string;
+}
+
+export function userMetadataToPayload(
+  payloadConverter: PayloadConverter,
```
mjameswh, line 17: User metadata should be converted using default payload converters, not user configured one. Otherwise, there's a risk that they might be serialized in a way that UI can't deserialize. We still need to _encode_ with the user configured codec.

Please double check what we do in other SDK, we may have made the same mistake there.

THardy98: Ah, Python has `_encode_user_metadata` and `_decode_user_metadata` functions: 

```
async def _encode_user_metadata(
    converter: temporalio.converter.DataConverter,
    summary: Optional[Union[str, temporalio.api.common.v1.Payload]],
    details: Optional[Union[str, temporalio.api.common.v1.Payload]],
) -> Optional[temporalio.api.sdk.v1.UserMetadata]:
...
    return temporalio.api.sdk.v1.UserMetadata(summary=enc_summary, details=enc_details)

async def _decode_user_metadata(
    converter: temporalio.converter.DataConverter,
    metadata: Optional[temporalio.api.sdk.v1.UserMetadata],
) -> Tuple[Optional[str], Optional[str]]:
    """Returns (summary, details)"""
   ...
 ```
 
 called with `client.dataConverter`.
 
I can make the change to default payload converters, it seems to me that it would make a lot of sense for this to be `RawValue` now that we have it.

Comment (Quinn-With-Two-Ns) [reply packages/common/src/user-metadata.ts:17]: User metadata allows private user information so it should go through the users payload converter

Comment (THardy98) [reply packages/common/src/user-metadata.ts:17]: > User metadata allows private user information so it should go through the users payload converter

Are codecs not enough for this?

Comment (mjameswh) [reply packages/common/src/user-metadata.ts:17]: Yes

Comment (cretz) [reply packages/common/src/user-metadata.ts:17]: I think in other SDKs we do use user data/payload converters (and of course payload codecs) and like any other payload our UI displays, if our UI doesn't recognize the type, it should say so. For the workflow metadata query result however, we do use a default converter IIRC.

Comment (THardy98) [reply packages/common/src/user-metadata.ts:17]: Ok - I'm going to marked this a resolved given offline discussion. We can revisit this if the need arises in the future.
@mjameswh , are you ok with this?

Comment (mjameswh) [reply packages/common/src/user-metadata.ts:17]: For the record, I still think this is not the right thing to do.

But, yes, I'm ok moving on with this given that's what we've consistently done in other SDKs.

File: packages/common/src/internal-non-workflow/codec-helpers.ts [hunk 31]
```diff
@@ -80,6 +91,18 @@ export async function encodeToPayload(converter: LoadedDataConverter, value: unk
   return await encodeSingle(payloadCodecs, payloadConverter.toPayload(value));
 }
 
+/**
+ * Run {@link PayloadConverter.toPayload} on an optional value, and then encode it.
+ */
+export function encodeOptionalToPayload(
```
mjameswh: nit: This function does not belong in `codec-helpers.ts`, and should not be called `encode`, as it is only _converting_ payloads, not running codecs.

packages/common/src/user-metadata.ts
```diff
@@ -0,0 +1,66 @@
...
+  dataConverter: LoadedDataConverter,
+  staticSummary: string | undefined,
+  staticDetails: string | undefined
+): Promise<temporal.api.sdk.v1.IUserMetadata | undefined> {
+  if (staticSummary == null && staticDetails == null) return undefined;
+
+  const { payloadConverter, payloadCodecs } = dataConverter;
+  const summary = await encodeOptionalSingle(
+    payloadCodecs,
+    await encodeOptionalToPayload(payloadConverter, staticSummary)
```
mjameswh: nit: No await needed on encodeOptionalToPayload.

packages/common/src/user-metadata.ts
```diff
@@ -0,0 +1,66 @@
...
+  );
+  const details = await encodeOptionalSingle(
+    payloadCodecs,
+    await encodeOptionalToPayload(payloadConverter, staticDetails)
```
mjameswh: No await needed on encodeOptionalToPayload

packages/common/src/user-metadata.ts
```diff
@@ -0,0 +1,66 @@
...
+  dataConverter: LoadedDataConverter,
+  metadata: temporal.api.sdk.v1.IUserMetadata | undefined | null
+): Promise<UserMetadata> {
+  const res = { staticSummary: undefined, staticDetails: undefined };
+  if (metadata == null) return res;
+
+  const staticSummary = (await decodeOptionalSinglePayload<string>(dataConverter, metadata.summary)) ?? undefined;
+  const staticDetails = (await decodeOptionalSinglePayload<string>(dataConverter, metadata.details)) ?? undefined;
+
+  if (staticSummary == null && staticDetails == null) return res;
```
mjameswh: nit: How is this condition / special case useful? At this point, `staticSummary` and `staticDetails` can be `undefined`, but they can't be `null`, so just returning the object below will do the exact same thing anyway.

packages/test/src/test-integration-split-two.ts
```diff
@@ -763,3 +780,171 @@ test.serial('default query handler is not used if requested query exists', confi
...
+      scheduleToStartTimeout: '5s',
+    },
+    []
+  );
+  // Timer (with summary)
+  await sleep(5, { summary: 'timer summary' });
+  // Set current details
+  setCurrentDetails('current wf details');
+  // Start child workflow
+  const child_handle = await startChild(completableWorkflow, {
```
mjameswh: nit

```suggestion
  const childHandle = await startChild(completableWorkflow, {
```

packages/test/src/test-integration-split-two.ts
```diff
@@ -763,3 +780,171 @@ test.serial('default query handler is not used if requested query exists', confi
...
+
+  await worker.runUntil(async () => {
+    // Start a workflow with static details
+    const handle = await startWorkflow(userMetadataWorkflow, {
+      staticSummary: 'wf static summary',
+      staticDetails: 'wf static details',
+    });
+    // Describe workflow -> static summary, static details
+    const desc = await handle.describe();
+    t.true((await desc.staticSummary()) === 'wf static summary');
```
mjameswh: nit


```suggestion
    t.is(await desc.staticSummary(), 'wf static summary');
```

packages/test/src/test-integration-split-two.ts
```diff
@@ -763,3 +780,171 @@ test.serial('default query handler is not used if requested query exists', confi
...
+    t.true((await desc.staticDetails()) === 'wf static details');
```
mjameswh: Same, and all other below.

packages/test/src/test-integration-split-two.ts
```diff
@@ -763,3 +780,171 @@ test.serial('default query handler is not used if requested query exists', confi
...
+      namespace: env.client.options.namespace,
+      execution: {
+        workflowId: res.childWorkflowId,
+        runId: res.childRunId,
+      },
+    });
+
+    for (const event of childResp.history?.events ?? []) {
+      if (event.eventType === temporal.api.enums.v1.EventType.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED) {
+        t.deepEqual(
```
mjameswh: nit: Don't use `deepEqual` if you're only comparing a primitive.

Same on all below.

```suggestion
        t.is(
```

packages/test/src/test-schedules.ts
```diff
@@ -881,4 +881,31 @@ if (RUN_INTEGRATION_TESTS) {
...
+        staticDetails: 'schedule static details',
+      },
+    });
+
+    try {
+      const describedSchedule = await handle.describe();
+      t.deepEqual(describedSchedule.spec.calendars, []);
+      t.deepEqual(describedSchedule.spec.intervals, []);
+      t.deepEqual(describedSchedule.spec.skip, []);
+      t.deepEqual(describedSchedule.action.staticSummary, 'schedule static summary');
```
mjameswh, line 905: nit

Same below

```suggestion
      t.is(describedSchedule.action.staticSummary, 'schedule static summary');
```

packages/common/src/internal-non-workflow/codec-helpers.ts
THardy98: renamed and moved to `payload-converter.ts`

packages/common/src/user-metadata.ts
THardy98: removed

packages/common/src/user-metadata.ts
THardy98: removed

packages/test/src/test-integration-split-two.ts
THardy98: renamed

packages/test/src/test-integration-split-two.ts
THardy98: done

packages/test/src/test-integration-split-two.ts
THardy98: done

packages/test/src/test-schedules.ts
THardy98: done

Issue #1630, opened 2025-02-21
[Feature Request] Built-in query responses should use "RawValue"


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/604. When #1629 is done, make sure built-in queries return a raw value

PR #1664, opened 2025-04-01
Add RawValue support for non-converted Payloads

### LINKED ISSUE #1629: [Feature Request] Support "RawValue" non-converted values

See https://github.com/temporalio/features/issues/603, need a simple payload wrapper that tells converters not to apply conversion.

THardy98:
## What was changed

Added `RawValue` support for non-converted Payloads - a thin wrapper around `Payload`.

`RawValue` is intended to be used for payloads that bypasses users custom payload converters in favour of the default payload converter.

TS does not include type hints or similar when converting from a payload back to its original representation. As such, the best we can do is just send the payload as is and have users convert it back to its original representation.

------

The existing builtin queries now return a `RawValue` that wraps their former returned values. The return types of the builtin queries however, maintain their existing type (this is because receivers of these queries will get the payload itself, not the `RawValue`). This is breaking change as consumers of these builtin queries will need to make sure they decode the builtin payloads correctly.

1. Closes #1629, #1630

2. How was this tested:
Added short integration test, modified the existing built-in metadata test.

3. Any docs updates needed?
Possibly. Maybe for the change to the built-in metadata query.


packages/workflow/src/internals.ts
```diff
@@ -306,14 +307,16 @@ export class Activator implements ActivationHandler {
...
-          return {
-            definition: {
-              type: workflowType,
-              queryDefinitions,
-              signalDefinitions,
-              updateDefinitions,
-            },
-          };
+          return new RawValue(
+            this.payloadConverter.toPayload({
```
mjameswh: IIUC, this needs to use the _default_ payload converter, not the one provided by the user.

mjameswh:
I believe we need to return `RawValue` from _all_ built-in queries, not only the user metadata query. And we should be using the _default_ payload converter, not the payload converter provided by the user.

mjameswh:
I'm really not clear how `RawValue` should work in TS on the inbound path, given that we don't have any form of type hint on functions that receive `Payload`s. Using a `rawValue` payload metadata set on the outbound feels totally wrong to me, as that makes RawValue a symetric thing (i.e. receiver will get a RawValue if and only if the caller sent a RawValue) whereas it seems to me that the reason to have `RawPayload`s is specifically to support cases where the caller and receiver would have different configurations.

For now, I'd focus on the outbound path only, simply to support return value of built-in queries. For inbound payloads, that might have to wait only we add support for type hints on receivers.

packages/workflow/src/internals.ts
THardy98: fixed

packages/test/src/workflows/stack-tracer.ts
```diff
@@ -4,6 +4,7 @@
  */
 import * as wf from '@temporalio/workflow';
 import type { EnhancedStackTrace } from '@temporalio/workflow/lib/interfaces';
+import { defaultPayloadConverter } from '@temporalio/common/lib/converter/payload-converter';
```
mjameswh: Unused import

packages/workflow/src/workflow.ts
```diff
@@ -22,6 +22,7 @@ import {
   WorkflowReturnType,
   WorkflowUpdateValidatorType,
   SearchAttributeUpdatePair,
+  RawValue,
```
mjameswh: Unused import

packages/workflow/src/internals.ts
```diff
@@ -315,14 +317,14 @@ export class Activator implements ActivationHandler {
             name,
             description: value.description,
           }));
-          return {
+          return new RawValue({
```
mjameswh: It's sad we're loosing type safety check here… There's nothing ensuring that the arg to RawValue complies with the expected type definition.

packages/common/src/converter/payload-converter.ts
```diff
@@ -100,6 +100,23 @@ export function mapFromPayloads<K extends string, T = unknown>(
   ) as Record<K, T>;
 }
 
+/**
+ * RawValue is a wrapper over a payload.
+ * A payload that belongs to a RawValue is special in that it bypasses user-defined payload converters,
+ * instead using the default payload converter. The payload still undergoes codec conversion.
+ */
+export class RawValue {
```
mjameswh: For the sake of type safety, I think we should add an optional generic parameter on RawValue, and save that generic parameter using the type brand pattern. Also, we should accept passing in an optional payload converter, i.e. in cases where user want to proce a specific payload converter (not the default one, and not the one they configured on the worker or client).

```
export declare const rawPayloadTypeBrand: unique symbol;

class RawValue<T = unknown> {
  private readonly _payload: Payload;

  // 
  private readonly [rawPayloadTypeBrand]: T = undefined as T;

  constructor(value: T, payloadConverter: PayloadConverter = defaultPayloadConverter) {
    this._payload = payloadConverter.toPayload(value);
  }

  get payload(): Payload {
    return this._payload;
  }
}
```

packages/test/src/test-integration-workflows.ts
```diff
@@ -386,7 +388,7 @@ test('Query workflow metadata returns handler descriptions', async (t) => {
 
   await worker.runUntil(async () => {
     const handle = await startWorkflow(queryWorkflowMetadata);
-    const meta = await handle.query(workflow.workflowMetadataQuery);
+    const meta = (await handle.query(workflow.workflowMetadataQuery)) as temporal.api.sdk.v1.IWorkflowMetadata;
```
mjameswh: You should not need a cast here.

packages/workflow/src/internals.ts
```diff
@@ -263,17 +263,19 @@ export class Activator implements ActivationHandler {
       '__stack_trace',
       {
         handler: () => {
-          return this.getStackTraces()
-            .map((s) => s.formatted)
-            .join('\n\n');
+          return new RawValue(
```
mjameswh:
```suggestion
          return new RawValue<string>(
```

packages/workflow/src/internals.ts
```diff
@@ -263,17 +263,19 @@ export class Activator implements ActivationHandler {
...
+          );
         },
         description: 'Returns a sensible stack trace.',
       },
     ],
     [
       '__enhanced_stack_trace',
       {
-        handler: (): EnhancedStackTrace => {
+        handler: (): RawValue => {
```
mjameswh, line 279:
```suggestion
        handler: (): RawValue<EnhancedStackTrace> => {
```

packages/workflow/src/internals.ts
```diff
@@ -293,15 +295,15 @@ export class Activator implements ActivationHandler {
...
+          return new RawValue({ sdk, stacks, sources });
         },
         description: 'Returns a stack trace annotated with source information.',
       },
     ],
     [
       '__temporal_workflow_metadata',
       {
-        handler: (): temporal.api.sdk.v1.IWorkflowMetadata => {
+        handler: (): RawValue => {
```
mjameswh, line 307:
```suggestion
        handler: (): RawValue<temporal.api.sdk.v1.IWorkflowMetadata> => {
```

packages/workflow/src/internals.ts
mjameswh: See my suggestions elsewhere that adds back type safety.

mjameswh (commented):
We're almost there, but still a bit more.

packages/workflow/src/workflow.ts
THardy98: removed

packages/common/src/converter/payload-converter.ts
THardy98: Nice, I've added the additional typing and payload converter param

packages/test/src/test-integration-workflows.ts
THardy98: removed

packages/common/src/converter/payload-converter.ts
```diff
@@ -100,6 +100,25 @@ export function mapFromPayloads<K extends string, T = unknown>(
   ) as Record<K, T>;
 }
 
+/**
+ * RawValue is a wrapper over a payload.
+ * A payload that belongs to a RawValue is special in that it bypasses user-defined payload converters,
+ * instead using the default payload converter. The payload still undergoes codec conversion.
+ */
+export declare const rawPayloadTypeBrand: unique symbol;
```
mjameswh: The typedoc must be _immediately_ followed by the `export class` line. As it is now, the typedoc attaches to the `rawPayloadTypeBrand` const.

PR #1638, opened 2025-03-05
chore: Refactor Bridge

mjameswh:
## What changed

Major refactor of Core bridge.

- Removed reactor event loop on Runtime and Worker; relying on Tokio's event loop. This greatly simplify the bridge code.

- Use Neon's Deferred/Promise support rather than callback functions + `promisify` on the TS side.

- Split code to files.

- Overall, align TS' bridge with how we do things in other Core-based SDKs. That should make it easier to port features across Core-based SDKs in the future.

- Fix some runtime-initialization hack related to metrics setup, which was in the way of implementing custom metrics support.

## Things to know

- I will continue this refactoring effort, but will do so in distinct PRs.

PR #1679, opened 2025-04-15
Worker Deployment Versioning

Sushisource:
## What was changed
Added the annotations and options for worker-deployment-based versioning

## Why?
All aboard the versioning train

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1659

2. How was this tested:
Added tests

3. Any docs updates needed?


packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,399 @@
...
+import { toCanonicalString, WorkerDeploymentVersion } from '@temporalio/common';
+import { temporal } from '@temporalio/proto';
+import { Worker } from './helpers';
+import * as activities from './activities';
+import { makeTestFunction } from './helpers-integration';
+import { unblockSignal, versionQuery } from './workflows';
+
+const test = makeTestFunction({ workflowsPath: __filename });
+
+test('Worker deployment based versioning', async (t) => {
```
THardy98, line 19: IIUC, there isn't any actual upgrading going on in this test. This test is checking that workflows with different versioning behaviours can run on workers with different deployment options?

THardy98 (commented):
Didn't really look through the core bridge stuff, maybe worth coordinating on #1638.

For your workflow test files `deployment-versioning-no-annotations...` there's a `workflows` directory in `test/src` that might make more sense for them to be located, instead of directly in `test/src`.

Just want to re-affirm how this works:
- on the worker, we have the worker's configured deployment version, and the default versioning behaviour for all its workflows
- on the workflow, we can override this versioning behaviour with the newly added workflow definition options (it's a little funky, we're making use of JS's quirk that a function is an object, so a workflow with definition options is a function with properties, as already discussed)
- workflow versioning behaviour is read into the activator (which uses it solely when concluding the activation) and workflow info

Tests were helpful, though I think you probably want someone whose more familiar with the versioning behaviour to take a look

packages/common/src/worker-deployments.ts
```diff
@@ -0,0 +1,28 @@
...
+export function toCanonicalString(version: WorkerDeploymentVersion): string {
+  return `${version.deploymentName}.${version.buildId}`;
+}
+
+/**
+ * Specifies when a workflow might move from a worker of one Build Id to another.
+ *
+ * * 'pinned' - The workflow will be pinned to the current Build ID unless manually moved.
+ * * 'auto-upgrade' - The workflow will automatically move to the latest version (default Build ID
+ *    of the task queue) when the next task is dispatched.
```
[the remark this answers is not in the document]
Sushisource: Yes

packages/worker/src/utils.ts
```diff
@@ -28,3 +29,16 @@ export function convertToParentWorkflowType(
     namespace: parent.namespace,
   };
 }
+
+export function convertDeploymentVersion(
+  v: coresdk.common.IWorkerDeploymentVersion | null | undefined
+): WorkerDeploymentVersion | undefined {
+  if (!v) {
```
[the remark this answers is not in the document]
Sushisource: Won't narrow the type properly

packages/worker/src/worker-options.ts
```diff
@@ -99,6 +105,7 @@ export interface WorkerOptions {
    * @default `@temporalio/worker` package name and version + checksum of workflow bundle's code
    *
    * @experimental The Worker Versioning API is still being designed. Major changes are expected.
+   * @deprecated Use {@link deploymentVersion} instead.
```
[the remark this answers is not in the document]
Sushisource: :man_shrugging: I suppose? But, why? We don't want people to use it and having a bunch of warnings is good for that heh

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,399 @@
...
+  const taskQueue = 'worker-deployment-based-versioning-' + randomUUID();
+  const deploymentName = 'deployment-' + randomUUID();
+  const client = t.context.env.client;
+
+  const w1DeploymentVersion = {
+    buildId: '1.0',
```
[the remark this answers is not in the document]
Sushisource: Deployment name can't, since it's first. For build id it's fine.

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,399 @@
...
+    },
+  });
+  const worker3Promise = worker3.run();
+  worker3Promise.catch((err) => {
+    t.fail('Worker 3.0 run error: ' + JSON.stringify(err));
+  });
+
+  // Wait for worker 1 to be visible and set as current version
+  const describeResp1 = await waitUntilWorkerDeploymentVisible(client, w1DeploymentVersion);
+  await setCurrentDeploymentVersion(client, describeResp1.conflictToken, w1DeploymentVersion);
```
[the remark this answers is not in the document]
Sushisource: What is "it" in that sentence? What this does is tell the server that the current version for some named deployment like `deployment` is a specific version like `deployment.1.0`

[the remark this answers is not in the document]
THardy98: "it" meant worker
oh, so this call is purely for server's awareness

packages/workflow/src/internals.ts
```diff
@@ -20,14 +20,16 @@ import {
...
+  VersioningBehavior,
 } from '@temporalio/common';
 import {
   decodeSearchAttributes,
   decodeTypedSearchAttributes,
 } from '@temporalio/common/lib/converter/payload-search-attributes';
 import { composeInterceptors } from '@temporalio/common/lib/interceptors';
 import { makeProtoEnumConverters } from '@temporalio/common/lib/internal-workflow';
-import type { coresdk, temporal } from '@temporalio/proto';
+import { coresdk, temporal } from '@temporalio/proto';
```
mjameswh, line 32: Never do a regular import to `proto` from Workflow code (that's a strong no). And as a general rule, avoid using anything from `proto` in user-facing APIs (workflow code or not).

In this case, you will need to create a pure JS mirror of the enum. Search for `makeProtoEnumConverters` in the project for examples you can copy from. That JS type should be defined in the `common` package, as you will want it to also be accessible in non-workflow context.

packages/workflow/src/interfaces.ts
```diff
@@ -592,4 +606,5 @@ export type UpdateHandlerOptions<Args extends any[]> = {
 export interface ActivationCompletion {
   commands: coresdk.workflow_commands.IWorkflowCommand[];
   usedInternalFlags: number[];
+  versioningBehavior?: temporal.api.enums.v1.VersioningBehavior;
```
mjameswh, line 609: Can't use proto enum in workflow code. See my other comment.

packages/workflow/src/internals.ts
```diff
@@ -416,6 +418,8 @@ export class Activator implements ActivationHandler {
 
   public readonly registeredActivityNames: Set<string>;
 
+  public versioningBehavior?: VersioningBehavior;
```
Quinn-With-Two-Ns, line 421: How do dynamic workflows pick their versioning behaviour per type?

packages/workflow/src/internals.ts
Sushisource: Thanks for the pointers!

packages/workflow/src/internals.ts
Sushisource: They can do it like is shown here: https://github.com/temporalio/sdk-typescript/pull/1679/files#diff-5ae1c5b6523b9493d1f852e9d303621ffdc8f817f1b80f38f572008572aec6feR14

In Python Chad didn't want to have a dynamic getter like we added in Java unless someone asks, which I can get behind, so I figured we'd do the same here.

Quinn-With-Two-Ns: Not sure it makes sense to cut features from some SDKs, but not other

Sushisource: I think the getter is a touch error prone. We could remove it from Java too, if no one ends up needing it in these other places. I didn't want to hold up being able to finish the Core ones over that, given these can get changed later anyway. Chad is who really cared: https://github.com/temporalio/sdk-python/pull/821#discussion_r2037340078

Quinn-With-Two-Ns: I think the getter is a basic requirement since versioning a per workflow type option per the design doc. I am just not sure why other SDKs users have to have reduced functionality.

Sushisource: I'm not the one who didn't want to add it :man_shrugging: . I agree with the principle. In practice, I suspect most people doing dynamic workflows would be fine with one behavior. It's hard to say. I'm fine either way, I think Chad is who you have to convince.

Sushisource: Has the getter now

packages/workflow/src/workflow.ts
```diff
@@ -1590,6 +1592,49 @@ export function allHandlersFinished(): boolean {
...
+ * ```
+ *
+ * @param options Options for the workflow defintion, or a function that returns options. If a
+ * function is provided, it will be called once just before the workflow function is called for the
+ * first time.
+ * @param fn The workflow function.
+ * @returns The same passed in workflow function, with the specified options applied. You can export
+ * this function to make it available as a workflow function.
+ */
+export function defineWorkflowWithOptions<A extends any[], RT>(
```
cretz, line 1627: Was it decided that this is only allowed for dynamic workflows and not everyone? May need to change the name and make sure it's only called (and allowed to be implemented) for dynamic workflows.

Sushisource: See my other comment, but, that's just not really possible in TS. Not without some crazy contortions. It's also less important here, though. We could fail at execution time but that feels punitive for sort of no reason

cretz: Interesting. So I see where you check `isWorkflowFunctionWithOptions` on worker init. Can you not check there if the workflow in question is a dynamic workflow and if not throw? I admit I am not familiar with this repo, but if, say I wanted to know "does a worker have a dynamic workflow registered", is that not possible? If not, how are we preventing people from creating two dynamic workflows on a worker?

Sushisource: It's naturally prevented because there can only be one top-level default export anyway, so the language does that for us basically

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,409 @@
...
+    { maxTimeout: 1000, retries: 10 }
+  );
+}
+
+async function setCurrentDeploymentVersion(
+  client: Client,
+  conflictToken: Uint8Array,
+  version: WorkerDeploymentVersion
+) {
+  return await client.workflowService.setWorkerDeploymentCurrentVersion({
```
mjameswh, line 403: Note that we'll need high level Client API support for these before the feature moves to GA. It's totally fine waiting for later PR.

I think the high level API should hide the conflict token, e.g. they would pass in a lambda function; the client would do a loop of (describe, call user function, setWorkerDeplCurVersion) until the update passes.

packages/workflow/src/workflow.ts
```diff
@@ -1590,6 +1592,49 @@ export function allHandlersFinished(): boolean {
...
+  options: WorkflowDefinitionOptionsOrGetter,
+  fn: (...args: A) => Promise<RT>
+): WorkflowFunctionWithOptions<A, RT> {
+  const wrappedFn = Object.assign(fn, {
+    options,
+    __temporal_is_workflow_function_with_options: true as const,
```
mjameswh, line 1633: Why do you need this marker? Can't you simply look for `options`? Or if this is to prevent conflicting these `options` with something else that would happen to be called the same, you can instead change the name of the `options` property to something more unique.

packages/workflow/src/worker-interface.ts
```diff
@@ -79,7 +79,21 @@ export function initRuntime(options: WorkflowCreateOptionsInternal): void {
     const workflowFn = mod[activator.info.workflowType];
     const defaultWorkflowFn = mod['default'];
 
-    if (typeof workflowFn === 'function') {
+    if (isWorkflowFunctionWithOptions(workflowFn)) {
```
mjameswh, line 82: We're loosing some important type checking here, and making the "default export with options" win over "exact named function without options", which seems undesirable.

Keep the `(typeof workflowFn === 'function')` and `(typeof defaultWorkflowFn === 'function')` at the top level, and move the check for options inside those top level branches.

Also, in the `else` branch of `workflowFn.options`, make sure that it is indeed a function before assigning it to `workflowDefinitionOptionsGetter`.

packages/test/src/test-worker-deployment-versioning.ts
Sushisource: Yeah, that's how it will work, but yes saved for later

packages/workflow/src/workflow.ts
Sushisource: Ok, I suppose I can. It just feels slightly safer to have something like this but I suppose in this case it's not like there will be other collisions

packages/workflow/src/workflow.ts
mjameswh: I'm hesitant regarding the name and signature of this function.

1. `defineWorkflowWithOptions` make it sounds like this is defining a Workflow function, which it is not. It is defining options on a workflow function, but it is the fact that you export the function from the module that actually defines it. One could notably assume incorrectly that they can do this frmo their workflow.ts:

```
// No `export` here
defineWorkflowWithOptions(async function myWorkflow () {
 ...
}, { ... })
```

2. The fact that this function returns the wrapped workflow function makes it looks like you can do:

```
export const myWorkflow = defineWorkflowWithOptions(async () => {
  // ...
})
```

which would work on the workflow side, but not on the client side, as the impl function won't have a `name` property in this case. Or it may even carry an invalid name if the user instead do something like `export const myWorkflow = defineWorkflowWithOptions(async functino differentName() { ... }`, resulting in a worker error trying to resolve the workflow type, or even worst, makes it use the wrong workflow function if there's a wokrlfow function of that name.

I understand you did this because of the default workflow handler case, and I would definitely prefer to have a single wrapper function for this, but I'm affraid that using distinct wrappers for normal workflow functions vs default workflow handler is the "safest" API, unfortunately.

So I would instead suggest this for the common case:

```
setWorkflowOptions(workflowFunction, { ... });  // No return value
```

I'm testing an alternative for the default handler, brb.

packages/workflow/src/workflow.ts
mjameswh: If we're really worried about collision, then using a symbol would be preferable. But I honestly don't think that's justified in this specific case. In fact, I know that many users do similar things for various reasons, so it is very likely that one may want to merge that functionality with their own wrapping logic.

packages/workflow/src/workflow.ts
mjameswh: Both of these works, without the need for an extra function.

```
export default async function (): Promise<string> {
  return 'dynamic';
}
setWorkflowOptions({ versioningBehavior: 'PINNED' }, module.exports.default);

... or ...

const _default = async function (): Promise<string> {
  return 'dynamic';
}
setWorkflowOptions({ versioningBehavior: 'PINNED' }, _default);

export default _default;
```

packages/worker/src/utils.ts
```diff
@@ -28,3 +29,16 @@ export function convertToParentWorkflowType(
...
+
+export function convertDeploymentVersion(
+  v: coresdk.common.IWorkerDeploymentVersion | null | undefined
+): WorkerDeploymentVersion | undefined {
+  if (v == null) {
+    return undefined;
+  }
+
+  return {
+    buildId: v.buildId ?? '',
```
mjameswh, line 41: Is it really ok that we just default to empty strings here? That means `toCanonicalString` will resolve this to `'.'`.

Should we not instead make this default to the same as if no dployment version was given? i.e. return `undefined`?

packages/test/src/helpers.ts
```diff
@@ -213,6 +214,12 @@ export class TestWorkflowEnvironment extends RealTestWorkflowEnvironment {
         : undefined),
     });
   }
+
+  static async createExistingServer(
+    opts?: ExistingServerTestWorkflowEnvironmentOptions
+  ): Promise<TestWorkflowEnvironment> {
+    return RealTestWorkflowEnvironment.createExistingServer(opts);
```
mjameswh, line 221: `createExistingServer` feels confusing. Maybe `TestWorkflowEnvironment.createFromExistingServer()`?

packages/common/src/worker-deployments.ts
```diff
@@ -0,0 +1,50 @@
+import { temporal } from '@temporalio/proto';
```
mjameswh, line 1: Make this a `type` import

packages/workflow/src/workflow.ts
Sushisource: Ok, cool, I'll recommend the first one in the docstring

packages/worker/src/utils.ts
Sushisource: yeah, good point, Build ID at least will be set here if it's set at all

Sushisource:
This will need to hang out for a bit since I want https://github.com/temporalio/api/pull/579 in before I incorporate that

packages/common/src/workflow-options.ts
```diff
@@ -253,3 +256,9 @@ export function extractWorkflowType<T extends Workflow>(workflowTypeOrFunc: stri
     `Invalid workflow type: expected either a string or a function, got '${typeof workflowTypeOrFunc}'`
   );
 }
+
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+export function isWorkflowFunctionWithOptions(obj: any): obj is WorkflowFunctionWithOptions<any[], any> {
```
mjameswh, line 261: Is there some rationale for adding this function in this file/package, even though it is only used from workflow/worker-interface.ts?

packages/common/src/worker-deployments.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export interface WorkerDeploymentVersion {
+  readonly buildId: string;
+  readonly deploymentName: string;
+}
+
+/**
+ * @returns The canonical representation of a deployment version, which is a string in the format
+ * `deploymentName.buildId`.
+ */
```
mjameswh, line 17: Please add `@internal` and `@hidden`.

packages/common/src/workflow-definition-options.ts
```diff
@@ -0,0 +1,18 @@
...
+  versioningBehavior?: VersioningBehavior;
+}
+
+type AsyncFunction<Args extends any[], ReturnType> = (...args: Args) => Promise<ReturnType>;
+export type WorkflowDefinitionOptionsOrGetter = WorkflowDefinitionOptions | (() => WorkflowDefinitionOptions);
+
+/**
+ * A workflow function that has been defined with options from {@link WorkflowDefinitionOptions}.
+ */
+export interface WorkflowFunctionWithOptions<Args extends any[], ReturnType> extends AsyncFunction<Args, ReturnType> {
```
mjameswh, line 16: Does this have any value being exported publicly? If yes, then `AsyncFunction` should also be exported. If not, then please add `@internal` and `@hidden` (I will have to experiment a bit more to figure out how to reduce this to a single annotation, but in the mean time, let's use both).

packages/core-bridge/src/worker.rs
```diff
@@ -498,6 +536,56 @@ mod config {
...
+                build_id: val.build_id,
+                deployment_name: val.deployment_name,
+            }
+        }
+    }
+
+    impl TryIntoJs for WorkerDeploymentVersion {
+        type Output = JsObject;
+
+        fn try_into_js<'cx>(self, cx: &mut impl Context<'cx>) -> JsResult<'cx, Self::Output> {
```
mjameswh, line 570: Seems like you can simply add `#derive(TryIntoJs)` on the `WorkerDeploymentVersion` struct. 

And if you really need to do it by hand, prefer using the `obj.set_property_from(cx, "buildId" seld.build_id)?;` syntax.

packages/core-bridge/src/worker.rs
```diff
@@ -453,14 +461,44 @@ mod config {
...
+
     impl BridgeWorkerOptions {
         pub(crate) fn into_core_config(self) -> Result<WorkerConfig, WorkerConfigBuilderError> {
             // Set all other options
             let mut builder = WorkerConfigBuilder::default();
             builder
                 .client_identity_override(Some(self.identity))
-                .worker_build_id(self.build_id)
-                .use_worker_versioning(self.use_versioning)
+                .versioning_strategy({
```
mjameswh, line 489: I suppose you may have written that part before the refactor…

Going forward, I really want to avoid having the bridge perform these kind of X-to-Y mappings, fallback/legacy/default value handling, etc. That kind of logic is easier to deal in lang, and that result in code that is much more easier to review. In this case, what I would do is to have bridge define an enum that replicates Core's `WorkerVersioningStrategy` enum, with the same variants and variant properties.

But feel free to disregard for now, I can revisit at a later point.

packages/core-bridge/ts/native.ts
```diff
@@ -175,6 +185,7 @@ export interface WorkerOptions {
   identity: string;
   buildId: string;
   useVersioning: boolean;
```
mjameswh, line 187: Continuing on my previous comment, remember that anything in native.ts are 100% internal APIs, so there is absolutely no reason to preserve fields that are no longer needed, or to preserve a legacy structure that no longer reflect the way they are consumed in Core. The presence of `buildId` and `useVersioning` directly in `WorkerOptions` are examples of such legacy things that we'll want to avoid from now on.

Again, I'm just explaining the direction I want to encourage, but don't feel obliged to fix in this PR. I can take care of this later.

packages/core-bridge/bridge-macros/src/lib.rs
```diff
@@ -9,6 +9,9 @@ use proc_macro::TokenStream;
 use syn::{DeriveInput, parse_macro_input};
 
 /// Procedural macro for defining bridge types with compile-time field name conversion
+///
+/// Note that enum types must all be defined on the JS side as objects with a `type` field which
```
mjameswh, line 13: Ah, yeah, that case is pretty ugly. There were too few of those to be worth optimizing.

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,409 @@
...
+      useWorkerVersioning: true,
+      version: w1DeploymentVersion,
+      defaultVersioningBehavior: 'PINNED',
+    },
+    connection: nativeConnection,
+  });
+  const worker1Promise = worker1.run();
+  worker1Promise.catch((err) => {
+    t.fail('Worker 1.0 run error: ' + JSON.stringify(err));
+  });
```
mjameswh, line 50: `JSON.stringify(err)` is very likely to simply print out `{}`. You will get more information by just concatenating `err` directly.

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,409 @@
...
+    workerDeploymentOptions: {
+      useWorkerVersioning: true,
+      version: w2DeploymentVersion,
+      defaultVersioningBehavior: 'PINNED',
+    },
+    connection: nativeConnection,
+  });
+  const worker2Promise = worker2.run();
+  worker2Promise.catch((err) => {
+    t.fail('Worker 2.0 run error: ' + JSON.stringify(err));
```
mjameswh, line 64: Same

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,409 @@
...
+    workerDeploymentOptions: {
+      useWorkerVersioning: true,
+      version: w3DeploymentVersion,
+      defaultVersioningBehavior: 'PINNED',
+    },
+    connection: nativeConnection,
+  });
+  const worker3Promise = worker3.run();
+  worker3Promise.catch((err) => {
+    t.fail('Worker 3.0 run error: ' + JSON.stringify(err));
```
mjameswh, line 79: Same

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -0,0 +1,409 @@
...
+import { toCanonicalString, WorkerDeploymentVersion } from '@temporalio/common';
+import { temporal } from '@temporalio/proto';
+import { Worker } from './helpers';
+import { Context, makeTestFunction } from './helpers-integration';
+import { unblockSignal, versionQuery } from './workflows';
+
+const test = makeTestFunction({ workflowsPath: __filename });
+
+test('Worker deployment based versioning', async (t) => {
+  const taskQueue = 'worker-deployment-based-versioning-' + randomUUID();
```
mjameswh, line 20: nit: Next time, consider using the `const { startWorkflow, createWorker, client, taskQueue } = helpers(t)` function. That can save quite a few lines of boilerplate.

packages/testing/src/testing-workflow-environment.ts
```diff
@@ -25,6 +25,15 @@ export type TimeSkippingTestWorkflowEnvironmentOptions = {
   client?: ClientOptionsForTestEnv;
 };
 
+/**
+ * Options for {@link TestWorkflowEnvironment.createExistingServer}
+ */
+export type ExistingServerTestWorkflowEnvironmentOptions = {
+  /** If not set, defaults to localhost:7233 */
+  address?: string;
+  client?: ClientOptions;
```
mjameswh, line 34: Why `ClientOptions` rather than `ClientOptionsForTestEnv` like other variants? I'm particularly worried about the fact that this would make it possible for caller to pass in an existing `Connection`, which would override (line 96) the one we create at line 222.

packages/worker/src/worker-tuner.ts
```diff
@@ -231,6 +236,11 @@ export interface SlotReserveContext {
    */
   workerBuildId: string;
 
+  /**
+   * The deployment version of the worker that is requesting the reservation
+   */
+  workerDeploymentVersion?: WorkerDeploymentVersion;
```
mjameswh, line 242: `@experimental`

packages/worker/src/worker-tuner.ts
```diff
@@ -231,6 +236,11 @@ export interface SlotReserveContext {
    */
   workerBuildId: string;
```
mjameswh, line 237: Should this one be marked as deprecated?

packages/common/src/worker-deployments.ts
Sushisource: This isn't really either of those things, it can show up in the UI that way, etc. As of the latest API changes now that's less of a thing since this is actually well typed everywhere - but I don't know that we need to hide this really. We didn't in the other SDKs.

packages/core-bridge/src/worker.rs
Sushisource: No, you can't. I meant to message you about that - but, into and from don't appear to work together. Seems like something that needs to be fixed in the macros.

packages/test/src/test-worker-deployment-versioning.ts
Sushisource: I couldn't use those helpers for reasons I can't remember perfectly, now - I think it had to do with the fact that I need multiple workers and need them all to use the same TQ but different other options.

packages/core-bridge/ts/native.ts
Sushisource: They're still kinda needed as they're how you select the old versioning strategy. We could pass _that_ straight through to the bridge though, but, I figure we can just remove them once they're no longer available at all.

packages/testing/src/testing-workflow-environment.ts
Sushisource: Really just because you need namespace - but I can add that in to the type explicitly

packages/client/src/workflow-client.ts
```diff
@@ -1227,6 +1227,7 @@ export class WorkflowClient extends BaseClient {
       cronSchedule: options.cronSchedule,
       header: { fields: headers },
       priority: options.priority ? compilePriority(options.priority) : undefined,
+      versioningOverride: options.versioningOverride ? options.versioningOverride : undefined,
```
mjameswh, line 1230:
```suggestion
      versioningOverride: options.versioningOverride ?? undefined,
```

packages/client/src/workflow-client.ts
```diff
@@ -1296,6 +1297,7 @@ export class WorkflowClient extends BaseClient {
       cronSchedule: opts.cronSchedule,
       header: { fields: headers },
       priority: opts.priority ? compilePriority(opts.priority) : undefined,
+      versioningOverride: opts.versioningOverride ? opts.versioningOverride : undefined,
```
mjameswh, line 1300:
```suggestion
      versioningOverride: opts.versioningOverride ?? undefined,
```

packages/common/src/worker-deployments.ts
mjameswh: In that case, let's at least give it a more specific name. This will be exported from the `common` package, and can't be renamed in the future. Maybe `toCanonicalDeploymentName`?

packages/core-bridge/src/worker.rs
mjameswh: Oh, then I'll revisit the macro at a later time.

For now, though, can still use the `set_property_from` macro instead, just to keep the code a bit cleaner.

mjameswh (approved):
Still a few nits, but no need for another round of review after that. You can safely ignore the unrelated test failures on the feature repo due to protobufjs, as well as the doc publish error due to vercel token needing refresh.

SamuelMS:
@Sushisource @mjameswh `defaultVersioningBehavior` is a mandatory option in `workerDeploymentOptions`, but `undefined` and `UNSPECIFIED` aren't accepted as documented here: https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning

TypeScript sample snippet from the docs:

```typescript
const myWorker = await Worker.create({
  taskQueue,
  workerDeploymentOptions: {
    useWorkerVersioning: true,
    version: { buildId: '1.0', deploymentName: 'llm_srv' },
    // does not work
    defaultVersioningBehavior: 'UNSPECIFIED',
  },
  connection: nativeConnection,
});
```

In my use-case, I actually want to set `useWorkerVersioning` to `false` – to ignore versioning entirely, but to have online workers self-report in the workflow deployments section of the UI regardless.

TypeScript throws e.g. `TS2322: Type undefined is not assignable to type VersioningBehavior`, and at runtime you get e.g. `Error: Unknown versioning behavior: UNSPECIFIED`. Excluding `defaultVersioningBehavior` entirely throws `temporal_sdk_core::worker::workflow: Error while completing workflow activation error=status: InvalidArgument, message: "versioning behavior cannot be specified without deployment options being set with versioned mode", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }` when workflows run.

Let me know if I'm misinterpreting how this is intended to be used, or if you'd like a dedicated issue raised for this.

Thank you!

PR #1685, opened 2025-04-21
fix!: Use plain Error in FailureConverter, WorkflowFailedError, and WorkflowUpdateFailedError

bergundy:
💥 BREAKING CHANGE 💥 

Instead of expecting that the FailureConverter will always return a TemporalFailure.
This change is required for the FailureConverter to support Nexus HandlerError.

NOTE: a similar breaking change has been made in the Java SDK a little while ago: https://github.com/temporalio/sdk-java/pull/2365


THardy98 (approved):
lgtm, seems pretty straightforward but worth getting a look from James in case I'm not considering something

PR #1690, opened 2025-04-22
Add vitest support to `@temporalio/nyc-test-coverage`

TastyPi:
## What was changed

Add support for `vitest` coverage reports using istanbul

## Why?

More test frameworks supported = better!

## Checklist

1. Closes 

I did not create an issue since it was reasonably easy to add and I figure discussions on whether or not to support can be done in the PR

2. How was this tested:

Locally with my setup that uses vitest

3. Any docs updates needed?

packages/nyc-test-coverage/README.md


gg-vara:
Would be a great addition, thanks @TastyPi  !

PR #1691, opened 2025-04-23
Expose Poller Automation

Sushisource:
Needs https://github.com/temporalio/sdk-typescript/pull/1679 to merge first

## What was changed
Expose options to opt-in to new poller autoscaling behavior

## Why?
Poller autoscaling is cool!

## Checklist

1. Closes 

2. How was this tested:
 Added simple test. More meaningful tests are in Core, we just need to make sure option is passed properly here.

3. Any docs updates needed?


packages/test/src/test-worker-poller-autoscale.ts
```diff
@@ -0,0 +1,82 @@
...
+    t.true(
+      workflow_pollers[0].endsWith('2') || workflow_pollers[0].endsWith('1'),
+      'First workflow poller count should be 1 or 2'
+    );
+    t.true(
+      workflow_pollers[1].endsWith('2') || workflow_pollers[1].endsWith('1'),
+      'Second workflow poller count should be 1 or 2'
+    );
+
+    const workflowPromises = Array(20)
```
mjameswh, line 65: There's no test assertion pass this point. I suppose you only want to confirm it doesn't throw, but should at least do `t.pass()` or some other trivial check.

packages/test/src/test-worker-poller-autoscale.ts
```diff
@@ -0,0 +1,82 @@
...
+        const handle = await localEnv.client.workflow.start(workflows.waitOnSignalThenActivity, {
+          taskQueue,
+          workflowId: `resource-based-${uuid()}`,
+        });
+        await handle.signal('my-signal', 'finish');
+        return handle.result();
+      });
+
+    await Promise.all(workflowPromises);
+    worker.shutdown();
```
mjameswh, line 77: I'm pretty sure `localEnv.teardown()` below will hang the process if `worker.shutdown()` isn't called first (e.g. because of a thrown error above). Though that should be fixed by the refactor, so feel free to just ignore.

packages/worker/src/worker-options.ts
```diff
@@ -568,6 +582,46 @@ export interface WorkerOptions {
   };
 }
 
+export type PollerBehavior = PollerBehaviorSimpleMaximum | PollerBehaviorAutoscaling;
+
+/**
+ * A poller behavior that will automatically scale the number of pollers based on feedback
+ * from the server. A slot must be available before beginning polling.
+ */
+export interface PollerBehaviorAutoscaling {
```
mjameswh, line 591: Should this marked as experimental?

packages/worker/src/worker-options.ts
```diff
@@ -362,6 +362,20 @@ export interface WorkerOptions {
    */
   maxConcurrentActivityTaskPolls?: number;
 
+  /**
+   * Specify the behavior of workflow task polling.
+   *
+   * @default A fixed maximum whose value is min(10, maxConcurrentWorkflowTaskExecutions).
+   */
+  workflowTaskPollerBehavior?: PollerBehavior;
```
mjameswh, line 370: Experimental?

packages/worker/src/worker-options.ts
```diff
@@ -362,6 +362,20 @@ export interface WorkerOptions {
...
+
+  /**
+   * Specify the behavior of activity task polling.
+   *
+   * @default A fixed maximum whose value is min(10, maxConcurrentActivityTaskExecutions).
+   */
+  activityTaskPollerBehavior?: PollerBehavior;
```
mjameswh, line 377: Experimental?

packages/worker/src/worker-options.ts
```diff
@@ -793,15 +849,30 @@ function addDefaultWorkerOptions(options: WorkerOptions, logger: Logger): Worker
...
+      : behavior.type === 'simple-maximum'
+        ? { type: 'simple-maximum', maximum: behavior.maximum ?? defaultMax }
+        : {
+            type: 'autoscaling',
+            minimum: behavior.minimum ?? 1,
+            initial: behavior.initial ?? 5,
+            maximum: behavior.maximum ?? 100,
+          };
+
+  const wftPollerBehavior = createPollerBehavior(maxWFTPolls, workflowTaskPollerBehavior);
```
mjameswh, line 864: Is this per WFT poller types? Is there a risk that we might starve either regular or sticky task queue pollers?

mjameswh: Also, I assume invalid configuration will be reported by Core?

mjameswh:
Are we considering making autoscaling 1/5/100 be the default at some point? I'd expect that to give much better performance than our current default values and heuristics.

mjameswh (approved):

Sushisource:
> Are we considering making autoscaling 1/5/100 be the default at some point? I'd expect that to give much better performance than our current default values and heuristics.

Yes, it should become the default eventually

packages/test/src/test-worker-poller-autoscale.ts
Sushisource: I always forget that's a thing. Seems so nonsense to me that you'd have to call pass explicitly.

packages/worker/src/worker-options.ts
Sushisource: Will leave these unmarked since I think these fields themselves are very unlikely to change, just the details of the autoscale behavior.

Sushisource:
No longer needed

Issue #1443, opened 2024-06-14
[Bug] Failure to start ephemeral server prevents shutdown of the process


mjameswh:
### Describe the bug

We regularly see patterns like the following in our CI; I presume this also affects users. The cause of the failure to start an ephemeral server isn't important (in this case, trying to use the time skipping server on Linux ARM, which isn't supported).

```
integration-workflows › before hook
Rejected promise returned by test. Reason:
UnexpectedError {
  message: 'Failed to start ephemeral server: HTTP status client error (404 Not Found) for url (https://temporal.download/temporal-test-server/default?arch=arm64&platform=linux&sdk-name=sdk-typescript&sdk-version=1.10.1)',
}

integration-workflows › after.always hook
Rejected promise returned by test. Reason:
TypeError {
  message: 'Cannot read properties of undefined (reading \'teardown\')',
}
› <anonymous> (src/helpers-integration.ts:89:30)

...then process never completes shutdown
```

diogotorres97:
Do we have a fix to run this on ARM machines?

PR #1698, opened 2025-04-29
chore(bridge): Refactor TS-Core bridge

### LINKED ISSUE #1302: [Bug] Node discards promises created by calling into Core Bridge with callback
### Describe the bug

Our bridge-to-rust layer extensively use a pattern where native Rust code is called with a callback function, which gets called asynchronously on completion of some Rust-side operation; on the TS side, we use the `promisify` function from the `node:util` built-in module to create a `Promise` that will resolve when the native code eventually invoke the provided callback function.

It appears that in some cases (possibly related to ESM and/or use of worker threads), Node may fail to recognize that such promise may still eventually resolve, and may therefore discard the promise as non-resolved. When that happens, it may cause sudden termination of the Node process (with [exit code 13](https://stackoverflow.com/questions/64602114/why-is-node-exiting-with-exit-code-13-rather-than-hanging)) or of some worker thread (for example when running tests).

I believe this is due to the fact that callback functions are [not marked as "referenced"](https://nodejs.org/docs/latest-v21.x/api/n-api.html#deciding-whether-to-keep-the-process-running) by the native code when they get passed to non-node threads (eg. when passing them to the Tokio runtime thread). Unfortunately, Neon doesn't currently expose the Thread Safe Functions API, which would have been the proper way to fix this issue.

### Minimal Reproduction

> ⚠️ The reproduction code presented below explicitly calls the `Runtime.shutdown()` function. **That function is not meant for public usage**. Calling it explicitly should never be required, and the specific sequence demonstrated below is expected to actually fail by throwing an exception.
>
> This particular sequence was only kept because it is the simplest currently known sequence that predictably trigger the issue described above, ie. that execution of the code gets completely dropped in some situations, with the node process exiting with error 13, rather than properly waiting on the completion of the promise (which in this case, should result in an error being thrown). ⚠️

1. Make sure `package.json` contains the `"type": "module"` directive.

2. Save the following code to a file named example.ts

    ```
    import { Runtime } from '@temporalio/worker';
    import { TestWorkflowEnvironment } from '@temporalio/testing';
    import { setTimeout } from 'node:timers/promises';
    
    const main = async () => {
      const runtime = Runtime.install({});
    
      console.log('Start test env start');
      const testEnvironment = await TestWorkflowEnvironment.createLocal();
      console.log('shutdown test env start');
      await testEnvironment.teardown();
      console.log('shutdown completed');
    
      console.log('shutdown runtime - start');
      try {
        await runtime.shutdown();
      } catch (e) {
        console.error(e);
      }
    
      console.log('shutdown runtime - stop');
    };
    
    await main();
    console.log('after main');
    ```

3. Execute that code with `tsx example.ts ; echo $?`, and observe the following output:

    ```
    Start test env start
    shutdown test env start
    shutdown completed
    shutdown runtime - start
    13
    ```

4. Modify the code by replacing the `await runtime.shutdown();` line by `await Promise.all([runtime.shutdown(), setTimeout(2000)]);`.

5. Run again with the same command, and observe the following result:

    ```
    Start test env start
    shutdown test env start
    shutdown completed
    shutdown runtime - start
    [UnexpectedError: channel closed]
    shutdown runtime - stop
    after main
    0
    ```

Note that the `[UnexpectedError: channel closed]` error is expected, as the runtime has already been shutdown after shutting down the test environment (because there was no more native resources being tracked by the runtime). This is not a problem by itself, but was simply kept as an easy way to demonstrate the present issue.

Running the same code in non-ESM context completes as expected.

### LINKED ISSUE #1413: [Feature Request] Warn when runtime can't shutdown due left over `NativeConnection`

By API contract, the user is responsible to `close()` `NativeConnection` objects they created explicitly (implicitly created connections are closed when its owner `Worker` is closed).

It is however too easy to forget doing so, and be left wondering why the process isn't completing as expected. With little work, we could emit warn messages when a `NativeConnection` is left opened for a few seconds after completion of the last associated Worker.

### LINKED ISSUE #1495: [Bug] Reenable OTLP export tests
### Describe the bug

- [Two of our TS integration tests](https://github.com/temporalio/sdk-typescript/blob/56207d8bb19e813e405a9530eed1dc44b6d8466a/packages/test/src/test-otel.ts#L65-L229) for Core OTLP metrics exports have recently started being extremely noisy (warnings: `OpenTelemetry metrics error occurred. Metrics exporter otlp failed with the grpc server returns error (Internal error): , detailed error message: Missing response message.`). Those tests are using a bare HTTP2 server to grossly emulate a OTLP collector grpc-service. Due to a recent update to Core (maybe to one of its underlying dependencies), the OTLP emitter now reports failure to get a proper grpc response form the collector service.

- These error messages have coincided with one case where our integration tests completely hanged in CI (I stopped them after 3.5 hours), though I don't know at this point either there is an actual connection between the these two symptoms.

- Until we can rewrite those tests to use a more convincing mock collector, I have disabled those tests.

### LINKED ISSUE #1634: [Feature Request] Custom metric bucket sizes

See https://github.com/temporalio/features/issues/606. Need to expose histogram bucket overrides made available in https://github.com/temporalio/sdk-core/pull/844.

mjameswh:
## What changed

This is a major refactor of the Core Bridge layer. It strives to reduce complexity, and increase reliability and maintainability of the TS SDK bridge layer, as well as realign it with how more recent Core-based SDKs have been designed. More importantly, this refactor is a stepping stone to increase velocity of further development on the TS SDK, introducing the structure and abstractions required to efficiently implement several upcoming SDK features.

Here are some of the most significant changes introduced by this refactor:

- Bridge APIs (i.e. those define in `@temporalio/core-bridge`) now adhere to much stricter rules.

  - In short, types defined by the bridge are now designed as Data Transfer Objects.

  - Bridge types now align on Core SDK's APIs, rather than corresponding user-facing APIs). That moves concerns such as data conversion, validation, default value handling, type restructuring, etc, to the TS side, rather than relying on unclear and inconsistent expectations between TS and Rust.
	
  - All properties on bridge APIs are now non-optional (e.g. can't define a property as `someProp?: someType`); `null` (rather than `undefined`) is used to indicate a value that is _intentionally_ left unset (i.e. the equivalent of `None` in Rust).

  - These new rules ensure that most incoherencies in type definitions between TS and Rust will be caught early, avoiding some recurrent error patterns we have had previously in TS SDK. Several such bugs were in fact uncovered and fixed while working on this PR.
	
  - These rules also make the Rust code much simpler, moving complexity back to the TS side, where we have better context and better ability to implement more complex business logic and report back errors. That makes the bridge code more approachable by TS-first developpers, easier to maintain and keep in sync with evolution of the Core SDK, and much more reviewable by our engineers.

- Most Rust side APIs (including entrypoint functions and most DTO type definitions) are now defined using macros, which avoid explicit handling of type conversion. That previously resulted in extremely verbose code, drowning actual business logic in data access and conversion plumbing.
  
  - Bridge-level API definitions (i.e. function signatures and DTOs) align almost one-to-one between the TS side and Rust side, allowing side-by-side review. Macros automatically provide required type conversions in and out, including basic type safety checks and error reporting where appropriate.

- Redresses and documents ownership rules for Rust-to-JS objects. The "Opaque Handle" abstraction introduced by this PR provides a clear yet simple to use ownership model for Runtime, Client, Worker, Ephemeral Servers and other native resources that are handed down to the JS world. This model notably guarantees proper clean up of resources, even when they are mistakenly dropped and garbage collected from the JS side (e.g. omitting to call `close` or `shutdown` on a resource), or dropped on the Rust side due to unexpected errors.

  - This fixes multiple cases of lifecycle issues that were causing process hang on shutdown. Fixes #1413 and #1443.
	
  - The bridge now properly handle cases where multiple Runtime coexists. We do not encourage that practice at the moment, and the Runtime class do not provide any API to intentionally do so, but there were previously scenarios that could result in such situations, 

- Bridge APIs now schedule async tasks directly to the Tokio runtime. The bridge itself no longer need event dispatching threads. This considerably simplify the effort requried to add new features to the SDK, and makes the code more approachable for new developers.

  - API calls into Rust now return actual JS `Promise` objects, rather than expecting TS side use of `promissify` and callback functions. Those Promise are backed by Rust `Future`.
   
    This is a huge benefit from a reliability perspective, as we can now guarantee that a `Promise` will be properly rejected if it is dropped by the Rust without having been settled. This could previously result in various unexpected behaviors including process hangs and suddent process termination. Fixes #1302.
	
  - Thanks to the `BridgeResult`, `BridgeError` and `TryIntoJs` types, async implementation of bridge APIs can directly return appropriate values or errors, without the need for implementer to explicitly handle conversion back into JS values (which can only be done on the Node thread, making this kind of pattern quite messy and verbose to implement).
	
  - `BridgeResult` and `BridgeError` also ensure proper handling of Neon's `Throw` errors, indicating that the Node thread is currently in "throw state". This particular state makes it difficult to properly handle errors in Rust/Node code, as any attempt to use the JS Context once it has fall into "throw state" will result in a panic. `BridgeResult` also restore the functionality of the Rust's `?` suffix operator, by transparently encapsulating Neon's `Throw` into `BridgeError`, and back into `Throw` once the `BridgeResult` surfaces back in place from where it can be handled.

  - Introduced other well-defined abstractions for some recurring needs, such as sync and async callback functions into JS, and `AbortController`. Those were previously reimplemented at each call-site.

- Bridge APIs are now clearly marked as _internal_ only.

  - Made it very clear that any API defined in the `@temporalio/core-bridge` package is internal APIs, and offers absolutely no stability guarantee. APIs meant for public usage must be defined and exposed by user facing packages exclusively (i.e. `@temporalio/worker` and `@temporalio/testing`). Reexporting bridge APIs from public facing APIs is also not allowed.

  - Public APIs that were previously exported by `@temporalio/core-bridge` have all been moved to user facing packages. Documentation has been added where missing on public types. Bridge types no longer have typedoc themselves as that would be redundant and unhelpful.

- Some other note worthy points:

  - Logs forwarding from Rust to JS is now implemented using a push model rather than pull model. All other Core-based SDKs were already using push-based log forwarding.
	
  - The runtime now ensure proper cleanup of the Otel metrics exporter; this could previously cause tons of error messages on runtime shutdown. Fixes #1495.

  - Adds support of custom metric bucket sizes. Fixes #1634.


packages/core-bridge/ai-rules.md
```diff
@@ -0,0 +1,418 @@
+# Bridge Layer Coding Norms, Best Practices and Pertinent Knowledge
```
mjameswh, line 1: WIP - I'm leaving this doc here even though it is incomplete (I'll complete at a later time, post-merge). I think it still adds some value in its present form. Feel free to read and comment if curious, but don't feel obliged.

packages/core-bridge/src/helpers/abort_controller.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+struct AbortControllerJsCounterpart {
+    controller: Root<JsObject>,
+    abort: JsCallback<(String,), ()>,
+}
+
+const STATE_UNINITIALIZED: u8 = 0;
+const STATE_ARMED: u8 = 1;
+const STATE_ABORTED: u8 = 2;
```
Sushisource, line 63: I don't see where this state is ever set?

packages/core-bridge/src/helpers/errors.rs
```diff
@@ -0,0 +1,291 @@
...
+            }
+
+            Self::UnexpectedError(..) => throw_custom_error(cx, UNEXPECTED_ERROR, self.to_string()),
+
+            Self::WorkerShutdown => throw_custom_error(cx, SHUTDOWN_ERROR, self.to_string()),
+
+            Self::Other(e) => {
+                if let Some(thrown) = underlying_js_throw_error(&e) {
+                    // FIXME: Send to logger
+                    eprintln!("Error thrown from JavaScript side: {e:?}");
```
Sushisource, line 199: Needs fixing

packages/core-bridge/src/helpers/handles.rs
```diff
@@ -0,0 +1,149 @@
...
+/// - `OpaqueOutboundHandle` - for sending an opaque handle that was just created by the Rust
+///                            side to the JS side (e.g. as the return value of an API call).
+/// - `OpaqueInboundHandle` - for accepting an opaque handle sent from the JS side to the Rust
+///                           side (e.g. as an argument to an API call).
+pub struct OpaqueOutboundHandle<T: MutableFinalize + 'static> {
+    inner: T,
+}
+
+impl<T: MutableFinalize + 'static> OpaqueOutboundHandle<T> {
+    // We late-wrap into Arc<Box<RefCell<>>> so that T can be send across threads without mutex.
```
Sushisource, line 29: Where does the `Box` come from here? I don't actually see that in the code? An `Arc<Box<...>>` pretty much never makes sense really, since that's just 2x pointer indirection when 1 would do

Sushisource: Additionally an `Arc<RefCell<>>` seems like something that doesn't make sense. `Arc` is only `Send + Sync` if its inner type is, but `RefCell` isn't, so this outer type can't be either. Seemingly this should just be an `Rc`?

packages/core-bridge/src/helpers/try_from_js.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// impl TryFromJs for u128 {
+//     fn try_from_js<'cx, 'b>(
+//         cx: &mut impl Context<'cx>,
+//         js_value: Handle<'b, JsValue>,
+//     ) -> BridgeResult<Self> {
+//         Ok(js_value.downcast::<JsNumber, _>(cx)?.value(cx) as Self)
+//     }
+// }
```
Sushisource, line 112: Delete me

packages/core-bridge/src/logs.rs
```diff
@@ -0,0 +1,114 @@
...
+
+#[derive(Debug, Clone, Serialize)]
+pub struct LogEntry {
+    pub level: String,
+    pub timestamp: String, // u128 as a string - JSON doesn't support u128 numbers
+    pub message: String,
+
+    #[serde(serialize_with = "serialize_map_as_camel_case")]
+    pub fields: HashMap<String, serde_json::Value>,
+    //
```
Sushisource, line 39: Extra comment line

packages/core-bridge/src/runtime.rs
```diff
@@ -1,557 +1,446 @@
...
-    if let Ok(Err(e)) = result_receiver.blocking_recv() {
-        Err(cx.throw_error::<_, String>(e).unwrap_err())?;
+impl RuntimeExt for CoreRuntime {
+    fn future_to_promise<F, R>(&self, future: F) -> BridgeResult<BridgeFuture<R>>
+    where
+        F: Future<Output = Result<R, BridgeError>> + Send + 'static,
+        R: TryIntoJs + Send + 'static,
+    {
+        enter_sync!(self);
+        Ok(BridgeFuture::new(Box::pin(future)))
```
Sushisource, line 175: I think this would work better with `.instrument()`? I'm surprised this works since the subscriber guard would be dropped immediately after this function exits

packages/core-bridge/src/runtime.rs
```diff
@@ -1,557 +1,446 @@
...
-/// Client will be returned in the supplied `callback`.
-pub fn client_new(mut cx: FunctionContext) -> JsResult<JsUndefined> {
-    let runtime = cx.argument::<BoxedRuntime>(0)?;
-    let opts = cx.argument::<JsObject>(1)?;
-    let callback = cx.argument::<JsFunction>(2)?;
+////////////////////////////////////////////////////////////////////////////////////////////////////
 
-    let client_options = opts.as_client_options(&mut cx)?;
+// IMORTANT: Any struct or enum below this point must be kept in sync with the type of the same name
+//           in native.ts. Similarly,
```
Sushisource, line 208: Unfinished comment

packages/core-bridge/ts/native.ts
```diff
@@ -0,0 +1,350 @@
...
+ * a property that is set to `undefined` and a property that is missing;
+ * i.e. `obj.get_value(cx, "prop")` will return `undefined` in both cases.
+ *
+ * We therefore follow the following principles for our TypeScript/Rust interfaces:
+ *
+ * - Always use `null` to indicate an intentionally unspecified optional value
+ *   in TypeScript interfaces. This will be converted to `None` on the Rust side.
+ * - Explicitly set _every_ properties on objects sent to the native code,
+ *   including optional properties (e.g. `{ prop: input.prop ?? null }`).
+ * - Never use the "optional property" syntax in TypeScript (i.e. `prop?: T`).
```
Sushisource, line 27: :+1:

packages/test/src/test-bridge.ts
```diff
@@ -0,0 +1,297 @@
...
+});
+
+test("Stopping Worker after creating another runtime doesn't fail", async (t) => {
+  const runtime0 = native.newRuntime(GenericConfigs.runtime.basic);
+  const runtime1 = native.newRuntime(GenericConfigs.runtime.basic);
+
+  const client0 = await native.newClient(runtime0, GenericConfigs.client.basic);
+  const worker0 = native.newWorker(client0, GenericConfigs.worker.basic);
+  await native.workerValidate(worker0);
+  const wftPromise0 = native.workerPollWorkflowActivation(worker0);
```
Sushisource, line 169: Some warnings about unused bindings in this file

packages/test/src/test-otel.ts
```diff
@@ -106,106 +124,102 @@ test.serial('Runtime.install() accepts metrics.otel.url without headers', async
...
-        telemetryOptions: {
-          metrics: {
-            otel: {
-              url: `http://127.0.0.1:${port}/v1/metrics`,
-              http: true,
-              headers: {
-                'x-test-header': 'test-value',
-              },
-              metricsExportInterval: 10,
+  // try {
```
Sushisource, line 177: Leftover comment

Sushisource (commented):
Overall looking good. This is absolutely massive so I couldn't give it super detailed scrutiny but tried to focus on the important looking bits.

packages/core-bridge/bridge-macros/src/derive_tryfromjs.rs
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+    };
+
+    // Uncomment this line to print the expanded macro
+    // eprintln!(
+    //     "=== struct {} ===\n{}\n======",
+    //     struct_ident.to_string(),
+    //     expanded
+    // );
```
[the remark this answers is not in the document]
mjameswh: Ah! Thanks for the tip. I was assuming there would be something like that, yet couldn't find it anywhere.

packages/core-bridge/src/runtime.rs
mjameswh: The future will execute on one of Tokio’s thread, which are bound with specific tracing contexts immediately after being created.

AFAICT the node thread would not need to access the tracing subscriber once the implementation of the “js_function” returns.

packages/core-bridge/src/helpers/abort_controller.rs
mjameswh: They are used in all operations on the atomic `state` variable, below.

packages/core-bridge/src/helpers/handles.rs
mjameswh: > Where does the Box come from here? I don't actually see that in the code? An Arc<Box<...>> pretty much never makes sense really, since that's just 2x pointer indirection when 1 would do

Indeed. Stale comment.

> `Arc<RefCell<>>` seems like something that doesn't make sense

Indeed. It is now `JsBox<Rc<Wrapper<T>>>`, where `Wrapper<T> = RefCell<Option<T>>`.

packages/core-bridge/src/helpers/abort_controller.rs
Sushisource: Sorry, that wasn't clear. I meant to highlight specifically ABORTED - I don't see where it ever gets set?

packages/core-bridge/src/helpers/errors.rs
```diff
@@ -70,6 +72,10 @@ pub enum BridgeError {
     #[error("Worker has been shutdown")]
     WorkerShutdown,
 
+    /// A custom error that can be used to wrap errors that are not otherwise handled.
```
bergundy, line 75: Docstring seems wrong here.

packages/core-bridge/src/helpers/properties.rs
```diff
@@ -23,9 +23,35 @@ pub trait ObjectExt: Object {
         }
         <T>::try_from_js(cx, value).field(key)
     }
+
+    /// Set a property on a JS Object, converting it from a Rust value using `TryIntoJs`.
+    fn set_property_from<'cx, C: Context<'cx>, T: TryIntoJs>(
```
bergundy, line 28: This is handy!

packages/core-bridge/ts/errors.ts
```diff
@@ -27,6 +28,35 @@ export class UnexpectedError extends Error {
...
+  name: 'ServiceError';
+  message: string;
+  code: number;
+  details: string;
+  metadata: Record<string, Buffer | string>;
+  stack?: string;
+}
+
+@SymbolBasedInstanceOfError('ServiceError')
+export class ServiceError extends Error implements grpc.ServiceError {
```
bergundy, line 41: I would ideally merge TransportError and ServiceError, they're different representations of the tonic error.

You can remove this new error type and add the new properties to `TransportError`.

bergundy: I was too lazy to do this as part of my PR but with your `from` and `into` helpers, it should be fairly straightforward.

bergundy, line 41: This class is missing a docstring BTW (if you plan on keeping it).

packages/core-bridge/src/helpers/abort_controller.rs
mjameswh: Lines 157 and 185...
And ABORTED is at line 202.

packages/core-bridge/src/worker.rs
```diff
@@ -1,454 +1,871 @@
...
-    let callback = cx.argument::<JsFunction>(1)?;
-    match worker.borrow().as_ref() {
-        None => {
-            callback_with_unexpected_error(&mut cx, callback, "Tried to use closed Worker")?;
+/// Let drop handle the cleanup.
+impl MutableFinalize for HistoryForReplayTunnelHandle {}
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+
+mod config {
```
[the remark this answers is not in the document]
mjameswh: Yeah, been hesitant, but I think it helps having that in the same file the feature they relate too. If anything, I would move them out of their `::config` modules, into their respective parents. I'll leave it that way for now, that's something that can easily be tuned later.

packages/core-bridge/src/helpers/errors.rs
mjameswh: Totally. Good catch.

packages/worker/src/connection.ts
```diff
@@ -9,21 +13,141 @@ import { Runtime } from './runtime';
...
+   * await conn.withAbortSignal(ctrl.signal, () => client.workflow.execute(myWorkflow, options));
+   * ```
+   *
+   * @returns value returned from `fn`
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal
+   */
+  async withAbortSignal<ReturnType>(abortSignal: AbortSignal, fn: () => Promise<ReturnType>): Promise<ReturnType> {
+    const cc = this.callContextStorage.getStore();
+    return await this.callContextStorage.run({ ...cc, abortSignal }, fn);
```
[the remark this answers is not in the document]
mjameswh: Oh, we completely missed that indeed. And probably most devs using abort signals did too.

I'll note for the moment, but won't change the semantic in this PR.

Sushisource (approved):

PR #1699, opened 2025-04-30
NativeConnection powered Client

bergundy:
This work enables both Nexus handlers and activities to get a `Client` powered by the `NativeConnection` of the `Worker` they are run in.

I left a TODO to support request cancelation via `withAbortSignal`, I may wait for @mjameswh's refactor before adding that.

### DIFF
--- a/packages/core-bridge/src/runtime.rs
+++ b/packages/core-bridge/src/runtime.rs
@@ -1,4 +1,10 @@
-use crate::{conversions::*, errors::*, helpers::*, worker::*};
+use crate::{
+    client::{BoxedClient, Client, CoreClient, RpcCall, RpcError, client_invoke},
+    conversions::*,
+    errors::*,
+    helpers::*,
+    worker::*,
+};
 use neon::{context::Context, prelude::*};
 use std::{
     cell::{Cell, RefCell},
@@ -8,9 +14,9 @@ use std::{
     sync::Arc,
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
-use temporal_client::{ClientInitError, ConfiguredClient, TemporalServiceClientWithMetrics};
+use temporal_client::ClientInitError;
 use temporal_sdk_core::{
-    ClientOptions, CoreRuntime, RetryClient, TokioRuntimeBuilder, WorkerConfig,
+    ClientOptions, CoreRuntime, TokioRuntimeBuilder, WorkerConfig,
     api::telemetry::CoreTelemetry,
     ephemeral_server::EphemeralServer as CoreEphemeralServer,
     init_replay_worker, init_worker,
@@ -22,8 +28,6 @@ use tokio::sync::{
     oneshot,
 };
 use tokio_stream::wrappers::ReceiverStream;
-
-pub type CoreClient = RetryClient<ConfiguredClient<TemporalServiceClientWithMetrics>>;
 
 #[derive(Clone)]
 pub struct EphemeralServer {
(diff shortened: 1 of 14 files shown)


bergundy:
I'm adding tests for using the client including the `TransportError` to `ServiceError` conversion.

packages/core-bridge/src/runtime.rs
```diff
@@ -219,6 +219,46 @@ pub fn start_bridge_loop(
...
+                                JsArrayBuffer::from_slice(cx, &bytes)
+                            }),
+                            Err(err) => send_error(channel.clone(), callback, |cx| {
+                                match err {
+                                    RpcError::Tonic(err) => {
+                                        let code = cx.number(err.code() as u32);
+                                        let metadata = JsObject::new(cx);
+                                        let details = err.details();
+                                        if !details.is_empty() {
+                                            let k = cx.string("grpc-status-details-bin");
```
mjameswh, line 235: Have you confirmed that these `grpc-status-details-bin` ends up being decoded properly by the Client, i.e. same as those coming from a JS Connection?

bergundy: Yup, there's a test that verifies that.

packages/worker/src/connection.ts
```diff
@@ -1,29 +1,208 @@
...
 
+  public readonly workflowService: WorkflowService;
+  readonly callContextStorage = new AsyncLocalStorage<CallContext>();
+
   /**
    * nativeClient is intentionally left private, framework code can access it with `extractNativeClient` (below)
    */
-  protected constructor(private nativeClient: Client) {}
+  protected constructor(private nativeClient: Client) {
+    this.workflowService = WorkflowService.create(this.sendRequest.bind(this), false, false);
```
mjameswh, line 78: Definitely not required for this PR, but any reason we should _not_ also implement other services over `NativeConnection` at some later point?

packages/worker/src/connection.ts
```diff
@@ -1,29 +1,208 @@
 import util from 'node:util';
+import { AsyncLocalStorage } from 'node:async_hooks';
+import * as grpc from '@grpc/grpc-js';
+import { getRelativeTimeout } from '@grpc/grpc-js/build/src/deadline';
```
mjameswh, line 4: That function is pretty simple. I'd prefer we inline it here rather than have an import to their internal project structure.

packages/worker/src/connection.ts
```diff
@@ -1,29 +1,208 @@
...
+import { ConnectionLike, Metadata, CallContext, WorkflowService } from '@temporalio/client';
+import { SymbolBasedInstanceOfError } from '@temporalio/common/lib/type-helpers';
 import { NativeConnectionOptions } from './connection-options';
 import { Runtime } from './runtime';
 
 const updateHeaders = util.promisify(clientUpdateHeaders);
 const updateApiKey = util.promisify(clientUpdateApiKey);
 
+@SymbolBasedInstanceOfError('ServiceError')
+export class ServiceError extends Error {
```
mjameswh, line 25: I don't think we should export this class, nor use the `@SymbolBasedInstanceOfError` decorator on it, as it is likely to be confused with the `ServiceError` class exported by the `client` package (yet, isn't exactly the same thing, as the later corresponds to errors rethrown by the Client, not those coming from the `Connection`).

In the `client` package, we simply use the `isGrpcServiceError()` type guard. So I suggest you keep this class here, just make it private and remove the decorator, and possibly add an `implements GrpcServiceError` if you feel like it.

mjameswh (approved):
Left a few important comments, but approving.

packages/worker/src/connection.ts
bergundy: We can definitely do that later, I just wanted to limit the scope for this MVP. I will open an issue.

packages/worker/src/connection.ts
bergundy: Will do.

packages/worker/src/connection.ts
bergundy: Hmm... I feel like any error that we throw should be exported but I'm not highly opinionated here.
FYI grpc-js `ServiceError` is not a concrete implementation, which is why I created this class.

WDYT?

bergundy: I ended up not exporting this since it's easy to export later while we cannot remove an already exported class.

PR #1708, opened 2025-05-08
feat(nexus): Initial Nexus support

bergundy:
## What changed

This adds support for Nexus in the Temporal TS SDK.

All Nexus-related APIs are experimental, and some changes are expected before this reaches GA.

## Review notes

- The team is still discussing expectations regarding `HandlerError` and `OperationError` serialization. In the mean time, this PR mostly adheres to the original spec (i.e. a Nexus should have either a `message` _or_ a `cause`), but little effort has been put into compliance testing, as we know the present model is broken in certain aspects.
- Tests use the HTTP API directly in lieu of a workflow caller or strongly typed client, we can refactor those later.
- More tests will continue to be added in the near future, after this PR.


packages/common/src/internal-non-workflow/codec-helpers.ts
```diff
@@ -91,6 +91,17 @@ export async function decodeArrayFromPayloads(
   return arrayFromPayloads(payloadConverter, await decodeOptional(payloadCodecs, payloads));
 }
 
+/**
+ * Decode `payloads` and then return {@link arrayFromPayloads}`.
```
mjameswh, line 95: Docstring is wrong

packages/worker/src/worker-options.ts
```diff
@@ -916,11 +968,43 @@ export function compileWorkerOptions(rawOpts: WorkerOptions, logger: Logger): Co
...
+    Map<string, nexus.OperationHandler<any, any> | nexus.SyncOperationHandler<any, any>>
+  >();
+  for (const s of opts.nexusServices ?? []) {
+    if (!s.name) {
+      throw new TypeError('Tried to register a Nexus service with no name');
+    }
+    if (nexusServices.has(s.name)) {
+      throw new TypeError(`Duplicate registration of nexus service ${s.name}`);
+    }
+    const ops = new Map<string, nexus.OperationHandler<any, any> | nexus.SyncOperationHandler<any, any>>();
```
mjameswh, line 989: Should that be part of the Nexus SDK? Logic is trivial at present, but there I expect there will be some normalization work to be done at a later point (e.g. once we add decorators) which we wouldn't to diverge from one Nexus implementation to another. This also seems to be what we did in Java, and kind-of what we did in Go.

bergundy: Yeah, we discussed moving this to the Nexus SDK with Chad yesterday. I will do that.

packages/test/src/test-nexus-handler.ts
```diff
@@ -0,0 +1,524 @@
+import { randomUUID } from 'node:crypto';
+import anyTest, { TestFn } from 'ava';
+import getPort from 'get-port';
+import * as nexus from 'nexus-rpc';
+import * as protoJsonSerializer from 'proto3-json-serializer';
+import * as temporalnexus from '@temporalio/nexus';
```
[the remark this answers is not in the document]
bergundy: I will fix.

packages/worker/src/worker-options.ts
mjameswh: Oh, I missed that part of the conversation.

> I will do that.

I only wanted to confirm that we agree on direction. I can do it myself if you don't have time/prefer not to do it (same apply to all suggestions I make on this PR).

packages/test/src/test-nexus-handler.ts
```diff
@@ -0,0 +1,524 @@
...
+
+test.before(async (t) => {
+  const logEntries = new Array<LogEntry>();
+
+  const logger = new DefaultLogger('INFO', (entry) => {
+    logEntries.push(entry);
+  });
+
+  Runtime.install({ logger });
+  t.context.httpPort = await getPort();
```
mjameswh, line 31: TODO(JWH): This will be flaky in CI due to port collisions, and users will often need to do that themselves. Move this to Core.

packages/test/src/test-nexus-handler.ts
```diff
@@ -0,0 +1,524 @@
...
+});
+
+test.after.always(async (t) => {
+  await t.context.env.teardown();
+});
+
+test.beforeEach(async (t) => {
+  const taskQueue = t.title + randomUUID();
+  const { env } = t.context;
+  const response = await env.connection.operatorService.createNexusEndpoint({
```
mjameswh, line 45: I understand _this_ is the way for now, but what's the intent regarding this? Do we expect users to have to do the same in their own Temporal+Nexus tests?

packages/worker/src/nexus.ts
```diff
@@ -0,0 +1,433 @@
...
+
+  protected async startOperation(
+    payload: Payload | undefined,
+    options: nexus.StartOperationOptions
+  ): Promise<coresdk.nexus.INexusTaskCompletion> {
+    try {
+      const input = await decodeFromPayload(this.dataConverter, payload);
+      if (typeof this.handler === 'function') {
+        const handler = this.handler as nexus.SyncOperationHandler<unknown, unknown>;
+        const output = await this.invokeUserCode('startOperation', handler.bind(undefined, input, options));
```
mjameswh, line 124: Would be preferable to add some defensive checks here to prevent a user from returning a `HandlerStartOperationResultSync` or `HandlerStartOperationResultAsync` from their `SyncOperationHandler`. Sounds like an easy error to make. But I don't think there's currently a safe way of doing this.

packages/test/src/test-nexus-handler.ts
bergundy: That's a good question. in Java we have a shortcut, we should probably make it easier in the test environment for Core based SDKs. @dandavison, wondering if you considered this in Python.

dandavison: No, I have not considered APIs for user Temporal Nexus testing yet; I'll add a TODO to the code. In my tests I have a helper function like this for creating Nexus endpoints.

packages/worker/src/nexus.ts
```diff
@@ -0,0 +1,433 @@
...
+      failure: await errorToNexusFailure(this.dataConverter, cause),
+    };
+  }
+
+  protected async startOperation(
+    payload: Payload | undefined,
+    options: nexus.StartOperationOptions
+  ): Promise<coresdk.nexus.INexusTaskCompletion> {
+    try {
+      const input = await decodeFromPayload(this.dataConverter, payload);
```
dandavison, line 121: If the data converter fails here, I assume it should result in a `BAD_REQUEST` 400 Nexus HTTP response, right? If that's correct can you confirm that it does?

bergundy: It's a test worth adding. Thanks!

packages/test/src/test-nexus-handler.ts
bergundy: I think we need an issue to track this.

dandavison: 👍 Created one for Python SDK-3837 "Python Nexus user testing utilities and docs"

bergundy: I would have made this a features issue with checkboxes for each SDK but 🤷, thanks!

packages/test/src/test-nexus-link-converter.ts
```diff
@@ -0,0 +1,163 @@
+import test from 'ava';
+import Long from 'long';
+import { temporal } from '@temporalio/proto';
+import {
+  convertWorkflowEventLinkToNexusLink,
+  convertNexusLinkToWorkflowEventLink,
+} from '@temporalio/worker/lib/nexus/link-converter';
```
dandavison, line 7:
```suggestion
} from '@temporalio/nexus/lib/link-converter';
```

packages/nexus/src/context.ts
```diff
@@ -0,0 +1,131 @@
...
+  return { workflowId: handle.workflowId, runId: handle.firstExecutionRunId };
+}
+
+export type WorkflowRunOperationHandler<I, O> = (
+  input: I,
+  options: nexus.StartOperationOptions
+) => Promise<WorkflowHandle<O>>;
+
+export class WorkflowRunOperation<I, O> implements nexus.OperationHandler<I, O> {
+  constructor(readonly handler: WorkflowRunOperationHandler<I, O>) {}
```
dandavison, line 112: Given that `WorkflowRunOperation` implements `OperationHandler`, isn't there an argument that `WorkflowRunOperation` should be called `WorkflowRunOperationHandler` and that the function passed into the constructor (type currently called `WorkflowRunOperationHandler`) should be called `startHandler: WorkflowRunOperationStartHandler`?

packages/nexus/src/token.ts
```diff
@@ -0,0 +1,78 @@
+const OPERATION_TOKEN_TYPE_WORKFLOW_RUN = 1;
+
+// OperationTokenType is used to identify the type of operation token.
+// Currently, we only have one type of operation token: WorkflowRun.
+type OperationTokenType = typeof OPERATION_TOKEN_TYPE_WORKFLOW_RUN;
+
+interface WorkflowRunOperationToken {
```
dandavison, line 7: Will you still be happy with this name when we have other operations such as `WorkflowSignalWithStartOperation` and `WorkflowUpdateWithStartOperation` that start workflows? Or is it better to give it a slightly more general name that indicates that it represents a workflow but is agnostic regarding how that workflow was started?

packages/test/src/test-nexus-handler.ts
```diff
@@ -0,0 +1,716 @@
...
+        nexus.service('testService', {
+          testOp: nexus.operation<void, void>(),
+        }),
+        {
+          testOp: new temporalnexus.WorkflowRunOperation<void, void>(async (ctx) => {
+            return await temporalnexus.startWorkflow(ctx, 'some-workflow', {
+              workflowId,
+              // To test attaching multiple callers to the same operation.
+              workflowIdConflictPolicy: 'USE_EXISTING',
+            });
```
dandavison, line 610: The design proposed here is that, when a user wants to start a workflow that is the backing execution for their Nexus operation, they must call a special `temporalnexus.startWorkflow` instead of the usual `client.startWorkflow`. The "special" version takes care of
- propagating nexus request ID to StartWorkflow request ID
- establishing bidirectional linkage


I'm going to note some downsides with this design; no implication intended that other designs are free of downsides.

1. The user may not know to do this, and obtain a normal Temporal client and start their workflow in the usual way via `client.startWorkflow`.
2. The user may write a Nexus operation that starts multiple workflows, intending only the final one to be the "backing" workflow. But if they use `temporalnexus.startWorkflow` for more than one workflow, they will get unexpected behavior.

dandavison: I'm partly mentioning this because I had been using a similar design for Python (in Python I was hoping not to introduce a second `start_workflow`). But there are similar challenges.

bergundy: So far we've been saying that it's better for the user to be explicit about the workflow that they intend to use to back the operation. Sure, there are cases where users may try to call this function twice in a handler and we can decide whether it's undefined behavior or a runtime error.

I don't think implicitly doing all of the things the "special" function does in `client.startWorkflow` is the right approach. And note that the context here is passed explicitly as we've discussed in the long explicit vs. implicit context discussion, let's not go back on that.

I like this version the best as it is consistent with Go, and allows a user to intercept failures.

FTR, this helper also propagates the callbacks and populates onConflictOptions when USE_EXISTING is used to attach a callback to an already running workflow.

dandavison: I agree we don't want to do the "special" things implicitly in `client.startWorkflow`.

I think what we want to do is almost exactly what you're doing, but instead of returning `WorkflowHandle<OutputT>`, return a type like `WorkflowOperationToken<OutputT>`.

For Python's solution to this, see https://github.com/temporalio/sdk-python/pull/813/files#r2160553793.

With what you've got currently, the user could easily make a mistake and use the normal `client.startWorkflow`, thinking that they are creating a nexusified workflow. And that would be very bad: there would be no type-check-time error and no request-time error; they will not be alerted to their mistake until they see that their operation result is never delivered and there is no bidirectional linking.

But having `temporalnexus.startWorkflow` return `WorkflowOperationToken<OutputT>` solves that: they still have the familiar `startWorkflow` interface, but it's impossible for the user to make the mistake in Typescript due to the type error.

These are the requirements that I believe we should be designing for here:

1. A Nexus operation handler must be able to start a nexusified workflow:
  a. delivers the final Nexus operation result
  b. has bidirectional linking
  c. Nexus request ID propagated into the StartWorkflow request
2. A Nexus operation handler must also be able to start “normal” workflows.
3. It should not be possible for a user who is trying to do (1) to get confused and accidentally do (2) instead.
4. The above must all hold whether a user is using the “shorthand” form (`WorkflowRunOperation`) or manually implementing an operation handler.

packages/nexus/src/context.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Starts a workflow run for a {@link WorkflowRunOperationHandler}, linking the execution chain to a Nexus Operation
+ * (subsequent runs started from continue-as-new and retries). Automatically propagates the callback, request ID, and
+ * back and forward links from the Nexus options to the Workflow.
+ */
+export async function startWorkflow<T extends Workflow>(
+  ctx: nexus.StartOperationContext,
+  workflowTypeOrFunc: string | T,
+  workflowOptions: WorkflowStartOptions<T>
+): Promise<WorkflowHandle<T>> {
+  const { client, taskQueue } = getHandlerContext();
```
dandavison, line 99: Users will need to be able to override with a client / task queue that is not the same as that inuse by the Nexus worker.

packages/test/src/test-nexus-handler.ts
bergundy: As discussed offline, I think we're good if we expect a `WorkflowHandle` that is specific to `@temporalio/nexus`.

packages/nexus/src/context.ts
mjameswh: Makes sense. Fixed.

packages/nexus/README.md
```diff
@@ -0,0 +1,8 @@
+# `@temporalio/nexus`
+
+[![NPM](https://img.shields.io/npm/v/@temporalio/nexus?style=for-the-badge)](https://www.npmjs.com/package/@temporalio/nexus)
+
+Part of [Temporal](https://temporal.io)'s [TypeScript SDK](https://docs.temporal.io/typescript/introduction/).
+
+- [API reference](https://typescript.temporal.io/api/namespaces/nexus)
```
THardy98, line 7: This link 404s atm

packages/nexus/src/index.ts
```diff
@@ -0,0 +1,14 @@
+export {
+  //
```
THardy98, line 2: nit: remove?

packages/test/src/test-nexus-handler.ts
mjameswh: Replaced by our existing `getRandomPort` helper.

packages/worker/src/nexus.ts
mjameswh: Dismissing. That is guaranteed by the Nexus RPC SDK itself.

packages/workflow/src/nexus.ts
```diff
@@ -0,0 +1,208 @@
...
+  ): Promise<nexus.OperationOutput<O>>;
+
+  /**
+   * Start a Nexus Operation and wait for its completion taking an operation name.
+   */
+  executeOperation<K extends nexus.OperationKey<T['operations']>>(
+    op: K,
+    input: nexus.OperationInput<T['operations'][K]>,
+    options?: Partial<StartNexusOperationOptions>
+  ): Promise<nexus.OperationOutput<T['operations'][K]>>;
```
THardy98: Is there reasonable benefit to have an overload for this where the user provides the key instead of the direct operation? (same for the `start` calls)

The benefit of this is not clear to me, but I'm not as familiar with the nexus-rpc semantics

packages/workflow/src/internals.ts
```diff
@@ -657,12 +667,55 @@ export class Activator implements ActivationHandler {
...
-  public resolveNexusOperation(_: coresdk.workflow_activation.IResolveNexusOperation): void {
-    throw new Error('TODO');
+  public resolveNexusOperation(activation: coresdk.workflow_activation.IResolveNexusOperation): void {
+    const seq = getSeq(activation);
+
+    if (activation.result?.completed) {
+      const result = this.payloadConverter.fromPayload(activation.result.completed);
+
+      // It is possible for ResolveNexusOperation to be received without a prior ResolveNexusOperationStart,
+      // e.g. either because the handler completed the operation synchronously.
```
THardy98: Can you clarify this case?
If the handler completes the operation synchronously, is no start activation ever emit?

packages/workflow/src/interceptors.ts
```diff
@@ -245,6 +260,64 @@ export interface LocalActivityInput {
   readonly attempt: number;
 }
 
+/**
+ * Input for {@link WorkflowOutboundCallsInterceptor.startNexusOperation}.
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export interface StartNexusOperationInput {
```
THardy98, line 268: Is this ever user-facing? Do the fields need docstrings?

packages/workflow/src/nexus.ts
```diff
@@ -0,0 +1,208 @@
...
+      untrackPromise(
+        scope.cancelRequested.catch(() => {
+          const completed =
+            !activator.completions.nexusOperationStart.has(seq) &&
+            !activator.completions.nexusOperationComplete.has(seq);
+
+          if (!completed) {
+            activator.pushCommand({
+              requestCancelNexusOperation: { seq },
+            });
```
THardy98, line 201: Why not just check `nexusOperationComplete` ?

THardy98 (commented):
leaving this for now, but haven't finished my review

packages/nexus/README.md
mjameswh: API docs will get published once this PR is merged to main. The link will remain a 404 till then.

packages/nexus/src/index.ts
mjameswh: That's intentional, to prevent Prettier from collapsing the list of reexported members into a single line. That makes it easier to review code diffs that affect the public API of our packages.

There are surely better ways to achieve that, but that one works fine just now.

packages/workflow/src/nexus.ts
mjameswh: As it is now, a user may do either of these:

```
  client.executeOperation(service.someOperation, ...)
```

or

```
  client.executeOperation('someOperation', ...)
```

Support for both forms is a requirement.

packages/workflow/src/internals.ts
mjameswh: The Nexus spec allows a remote service to complete either synchronously or asynchronously (for the purpose of this discussion, synchronicity is from the perspective of the Temporal server, as it is the one making a network request to that remote service).

If the service responds asynchronously, the Temporal server will record a `NexusOperationStartedEvent` in the workflow history, and later append a `NexusOperationCompletedEvent` to indicate actual completion. If the service responds synchronously, the Temporal server will only write a `NexusOperationCompletedEvent`, thus saving one event in the workflow history.

Core SDK could possibly fill the gap for the missing `NexusOperationStartedEvent`, but it was decided not to do so, so we have to handle both of those scenarios here.

packages/workflow/src/nexus.ts
mjameswh: Because cancellation could also be requested for an operation that is still in the "Start" state.

packages/common/src/converter/failure-converter.ts
```diff
@@ -10,16 +13,42 @@ import {
...
+import { makeProtoEnumConverters } from '../internal-workflow';
 import { isError } from '../type-helpers';
 import { msOptionalToTs } from '../time';
 import { arrayFromPayloads, fromPayloadsAtIndex, PayloadConverter, toPayloads } from './payload-converter';
 
+// Can't import enums into the workflow sandbox, use this helper type and enum converter instead.
+const NexusHandlerErrorRetryBehavior = {
+  RETRYABLE: 'RETRYABLE',
+  NON_RETRYABLE: 'NON_RETRYABLE',
+} as const;
```
THardy98: Is this more generally, can't import **proto** enums in to the workflow sandbox?

packages/workflow/src/interceptors.ts
```diff
@@ -245,6 +260,64 @@ export interface LocalActivityInput {
...
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export interface StartNexusOperationOutput {
+  /**
+   * The token for the Nexus Operation.
+   *
+   * May be undefined if the Nexus Operation completed synchronously, in which case the {@link result}
+   * will be immediately resolved.
+   */
```
dandavison:
```suggestion
   * Undefined if the Nexus Operation completed synchronously, in which case the {@link result}
   * will be immediately resolved.
   */
```

packages/workflow/src/interceptors.ts
```diff
@@ -245,6 +260,64 @@ export interface LocalActivityInput {
...
+  token?: string;
+
+  /**
+   * A promise that will resolve when the Nexus operation completes, either with the result of the
+   * operation if it completed successfully, or a failure if the Nexus Operation failed.
+   */
```
dandavison, line 317: FWIW - when it comes to cancellation types this comment will be technically incomplete/inaccurate: under `WaitRequested` the promise will resolve with a cancellation error if the cancel handler succeeds, and it will resolve with a non-cancellation error if the cancel handler fails, in both cases regardless of the operation outcome.

nit: decide on capitalization of [Oo]peration and [N]nexus and make docstrings consistent.

packages/workflow/src/interceptors.ts
```diff
@@ -245,6 +260,64 @@ export interface LocalActivityInput {
...
+   *
+   * Optional: defaults to the maximum allowed by the Temporal server.
+   */
+  scheduleToCloseTimeout?: Duration;
+
+  /**
+   * A fixed, single-line summary for this Nexus Operation that may appear in the UI/CLI.
+   * This can be in single-line Temporal markdown format.
+   *
+   * @experimental User metadata is a new API and suspectible to change.
```
dandavison: typo

packages/workflow/src/interceptors.ts
```diff
@@ -245,6 +260,64 @@ export interface LocalActivityInput {
...
+  /**
+   * The end to end timeout for the Nexus Operation.
+   *
+   * Optional: defaults to the maximum allowed by the Temporal server.
+   */
+  scheduleToCloseTimeout?: Duration;
+
+  /**
+   * A fixed, single-line summary for this Nexus Operation that may appear in the UI/CLI.
+   * This can be in single-line Temporal markdown format.
```
dandavison, line 293: Am I supposed to know what "Temporal markdown format" is?

packages/common/src/converter/failure-converter.ts
```diff
@@ -220,7 +286,9 @@ export class DefaultFailureConverter implements FailureConverter {
     }
     const err = this.failureToErrorInner(failure, payloadConverter);
     err.stack = failure.stackTrace ?? '';
-    err.failure = failure;
+    if (err instanceof TemporalFailure) {
+      err.failure = failure;
+    }
```
THardy98, line 291: Has the blind `err.failure = failure` caused issues before?

packages/workflow/src/internals.ts
dandavison:
```suggestion
      // e.g. because the handler completed the operation synchronously.
```

packages/workflow/src/nexus.ts
```diff
@@ -0,0 +1,208 @@
...
+      //         and the result promise is contained in that Output object. As a consequence of this,
+      //         the result promise/completion does not exist until the StartNexusOperation event is received.
+      //         That's totally different from what we did in ChildWorkflow, but I think that's cleaner from
+      //         interceptors point of view, and will make it easier to extend the API in the future.
+      const { token, result: resultPromise } = await execute({
+        endpoint: options.endpoint,
+        service: options.service.name,
+        operation: opName,
+        options: operationOptions ?? {},
+        nexusHeader: {},
```
dandavison: Do you want to make it possible for the user to pass headers now, or leave an open issue for it?

dandavison: Also I know Go uses the (weird) singular for "header"; does TS also? Python [uses](https://github.com/temporalio/sdk-python/blob/main/temporalio/workflow.py#L5212-L5213) "headers".

packages/workflow/src/nexus.ts
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * A Nexus client for invoking Nexus Operations for a specific service from a Workflow.
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export interface NexusClient<T extends nexus.ServiceDefinition> {
+  /**
+   * Start a Nexus Operation and wait for its completion taking a {@link nexus.operation}.
+   */
```
dandavison:
```suggestion
   * Start a Nexus Operation and wait for its completion taking a {@link nexus.operation}.
   * Returns the operation's result.   
   */
```

packages/workflow/src/nexus.ts
```diff
@@ -0,0 +1,208 @@
...
+  executeOperation<O extends T['operations'][keyof T['operations']]>(
+    op: O,
+    input: nexus.OperationInput<O>,
+    options?: Partial<StartNexusOperationOptions>
+  ): Promise<nexus.OperationOutput<O>>;
+
+  /**
+   * Start a Nexus Operation and wait for its completion taking an operation name.
+   */
```
dandavison:
```suggestion
  /**
   * Start a Nexus Operation and wait for its completion taking an operation name.
   * Returns the operation's result.   
   */
```

packages/workflow/src/nexus.ts
dandavison: Would it be worth giving users some guidance in the docstrings regarding which of the two overloads (operation vs operation name) they should use?

Maybe dumb question: doesn't the string overload always provide better type safety? The structural typing (operation) overload allows a user to refer to an operation that has a name that does not match a name on the service for which the client was created, giving rise to a request-time `HandlerError` I believe e.g.

```typescript
const service = nexus.service('test', {
  syncOp: nexus.operation<string, string>({ name: 'my-sync-op' }),
  asyncOp: nexus.operation<string, string>(),
});

const service2 = nexus.service('test', {
  syncOp2: nexus.operation<string, string>({ name: 'my-sync-op' }),
  asyncOp2: nexus.operation<string, string>(),
});

export async function caller(endpoint: string, op: keyof typeof service.operations, action: string): Promise<string> {
  const client = workflow.createNexusClient({
    endpoint,
    service,
  });
  return await workflow.CancellationScope.cancellable(async () => {
    let opRef = service2.operations['syncOp2'];
    const handle = await client.startOperation(opRef, action);
```

THardy98 (commented):
I have limited context on the nexus semantics, so I'll leave that to @dandavison and @bergundy . The integration lgtm to me though.

packages/workflow/src/interceptors.ts
mjameswh: Yes, that's user facing, and yes this could benefit from some doc strings.

But we don't have any on other interceptors input/output types, so I'll just skip for the moment.

packages/workflow/src/nexus.ts
THardy98: Duh 🤦 
What is the result of cancelling a completed operation then? It's already completed.

packages/workflow/src/interceptors.ts
THardy98: All the normal Markdown elements except we strip away images, hmtl, and scripts.

Maybe it's worth clarifying that in the docstring

mjameswh: Good question.

Unfortunately, we don't yet have a doc page to link to, which is ok at this point as User Metadata is still considered prerelease.

And FWIW, that's the same typedoc text as we have in other SDKs.

packages/workflow/src/interceptors.ts
mjameswh: Fixed everywhere, thanks.

packages/worker/src/interceptors.ts
```diff
@@ -85,6 +86,68 @@ export interface ActivityInboundCallsInterceptorFactory {
...
+/**
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export type NexusInterceptors = {
+  inbound?: NexusInboundCallsInterceptor;
+  outbound?: NexusOutboundCallsInterceptor;
+};
+
+/**
+ * A function that takes a Nexus Context and returns an interceptor
```
dandavison: I don't think I understand this comment

packages/worker/src/interceptors.ts
```diff
@@ -85,6 +86,68 @@ export interface ActivityInboundCallsInterceptorFactory {
...
+ * Output for NexusInboundCallsInterceptor.execute
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export interface NexusExecuteOutput {
+  readonly result: unknown;
+}
+
+/**
+ * A function that takes a Nexus Context and returns an interceptor
```
dandavison: As above: I don't understand this comment

packages/worker/src/interceptors.ts
THardy98: +1

packages/workflow/src/nexus.ts
mjameswh: - `nexusOperationStart` => waiting for the operation to be started on the remote nexus handler service.
- `nexusOperationComplete` => waiting for the operation to be completed asynchronously by the remote nexus handler.

So while the op is in `nexusOperationComplete`, it is not yet completed, it is _pending_ completion, and so it can still be cancelled.

packages/workflow/src/interceptors.ts
mjameswh: > nit: decide on capitalization of [Oo]peration and [N]nexus and make docstrings consistent.

Done

mjameswh: > FWIW - when it comes to cancellation types this comment will be technically incomplete/inaccurate: under WaitRequested the promise will resolve with a cancellation error if the cancel handler succeeds, and it will resolve with a non-cancellation error if the cancel handler fails, in both cases regardless of the operation outcome.

Thanks for the heads-up.

packages/workflow/src/nexus.ts
mjameswh: > Also I know Go uses the (weird) singular for "header"; does TS also? Python [uses](https://github.com/temporalio/sdk-python/blob/main/temporalio/workflow.py#L5212-L5213) "headers".

Good catch. Renamed to `headers` everywhere (no `nexus` prefix, plural form). That's coherent with both Python and other TS interceptors.

mjameswh: > Do you want to make it possible for the user to pass headers now, or leave an open issue for it?

Not sure what's the expectation here. For all other APIs, headers are not exposed directly in the "main" user facing call, only in interceptors. Are doing this differently for Nexus ops? If that's only in interceptors, then that should already be working now (though no test atm covering that).

In any case, I'm leaving that for later.

packages/common/src/converter/failure-converter.ts
mjameswh: Right, this is about proto enums. Clarified.

packages/common/src/converter/failure-converter.ts
mjameswh: That was simply not possible before. I mean, until now, `err` here would always be a `TemporalFailure` (look at the method signature). Now, it may be a Nexus `HandlerError`, which is part of the (Temporal agnostic) Nexus RPC.

I'm considering storing the failure on HandlerError anyway, possibly using a private symbol property. But we are revisiting nexus error serialization at the moment, so I'm deferring that decision until we've set our mind.

packages/worker/src/interceptors.ts
mjameswh: Right, that comment should have been on `NexusInterceptorsFactory`, above. I improved that one, as well as that of the equivalent `ActivityInterceptorsFactory` type, near the top of this file.

However, I'd point out that typedocs on everything interceptors-related (Client/Workflow/Activity/Nexus) is very disappointing. I'm taking note to revisit _all_ interceptors soon.

packages/worker/src/interceptors.ts
mjameswh: Removed description on that specific typedoc. It should have one, but so should all other interceptor-related types, but most don't. I'll revisit in a distinct PR.

packages/worker/src/nexus/index.ts
```diff
@@ -0,0 +1,329 @@
...
+    };
+    // In case some interceptors use the metric meter while initializing...
+    if (this.interceptors == null) return baseTags;
+    return composeInterceptors(this.interceptors.outbound, 'getMetricTags', (a) => a)(baseTags);
+  }
+
+  protected async startOperation(
+    ctx: nexus.StartOperationContext,
+    payload: Payload | undefined
+  ): Promise<coresdk.nexus.INexusTaskCompletion> {
```
dandavison, line 102: Just a comment: I note that in Python, all the classes that have a "Handler"-like interface are in the nexus SDK:

https://github.com/temporalio/sdk-python/blob/607641bf1e8250699b9c7ae0e87230d43b26c2ff/temporalio/worker/_nexus.py#L62

whereas here you have a class with a handler-like interface in the Tempral worker. Perhaps Python could/should be written more like yours; I hadn't considered it. Do you have any comments on that difference?

packages/test/src/test-nexus-handler.ts
mjameswh: > As discussed offline, I think we're good if we expect a WorkflowHandle that is specific to @temporalio/nexus.

I added a type brand symbol property to `temporalnexus`'s `WorkflowHandle`, so that it can't be assigned from a `client`'s `WorkflowHandle`. That results in a compile-time error if user is mistakenly using the later in lieu of the former. No runtime check (should we?).

Also added a test to confirm.

packages/worker/src/nexus/index.ts
mjameswh: Not sure what difference you are trying to point out.

That's an internal method of the worker to bridge between what Nexus Tasks provided by Core/the Temporal server, and the Nexus RPC's APIs. It really can't be moved to the Nexus RPC, as it is dealing with Core's `NexusTaskCompletion` and Temporal `Payload`.

It is equivalent to [this method](https://github.com/temporalio/sdk-python/blob/607641bf1e8250699b9c7ae0e87230d43b26c2ff/temporalio/worker/_nexus.py#L249-L317) in Python.

The signature of the methods are slightly different, I think that's mostly copying the respective style of existing Workflow/Activity Task handling in various SDKs. I really don't think we need to worry about cross-SDK alignment at that level.

packages/workflow/src/nexus.ts
mjameswh: > Maybe dumb question: doesn't the string overload always provide better type safety? The structural typing (operation) overload allows a user to refer to an operation that has a name that does not match a name on the service for which the client was created

No, both provide the same level of type safety. I [just added a test](https://github.com/temporalio/sdk-typescript/pull/1708/files#diff-2814fd44158a1ab11f8abc4cf9316b77925ea580ba3a7a88edb654dee27435c4R201-R313) to demonstrate this.

> Would it be worth giving users some guidance in the docstrings regarding which of the two overloads (operation vs operation name) they should use?

We have no opinion on that question. We intentionally chose to allow both.

I however expanded the typedoc to clarify that this expects the _operation's property's name_, rather than the _operation's name_. That may sound a bit confusing at first, but in practice, I'd expect most users will either not override operation names, or they will rely on their IDE's autocompletion, or TSC's compilation errors, to point out incorrect usage. Only pure-JS users (with no implicit TS IDE support) and Nexus framework implementers would really be exposed to this notion.

packages/test/src/test-nexus-handler.ts
mjameswh: This will be added post-initial-release.

THardy98 (approved):
Approving, but should get approval from someone who has more familiarity with the nexus semantics before merging.

Specifically, approval on the `nexus` package and the corresponding `test`s.

The changes integrating this with the TS SDK lgtm

mjameswh (approved):
I'm comfortable with the PR as it stands at this point.

However, given that I myself authored a significant part of this PR, I think it's pertinent to wait for some complementary signoffs.

packages/workflow/src/nexus.ts
dandavison: > both provide the same level of type safety.

Using an operation method reference, users can do this without a type error:

```typescript
await service1client.startOperation(service2.operations.syncOp2, action)
```

That would result in a request time error. I.e. it would send a request for `service='service1` and `operation='syncOp2'`, which is invalid.

But using string operation names they cannot represent that invalid nexus request. For example, `service1client.startOperation('syncOp2', action)` would be a type-check-time error.

My question is: given that operation method references are less type safe in this sense, what value do they add?

packages/nexus/src/workflow-helpers.ts
```diff
@@ -0,0 +1,153 @@
...
+    requestId: ctx.requestId,
+  };
+
+  if (workflowOptions.workflowIdConflictPolicy === 'USE_EXISTING') {
+    internalOptions.onConflictOptions = {
+      attachLinks: true,
+      attachCompletionCallbacks: true,
+      attachRequestId: true,
+    };
+  }
```
dandavison: Minor, but Go and Java set these all to `true` conditional only on it being a nexus-backing workflow, but not on the conflict policy value:

Go: https://github.com/temporalio/sdk-go/blob/9ca81c60/temporalnexus/operation.go#L332
Java: https://github.com/temporalio/sdk-java/blob/b6b42903/temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java#L153-L158

The server honors the options conditional on the conflict policy: https://github.com/temporalio/temporal/blob/fdb7f31a4/service/history/api/startworkflow/api.go#L423-L425

I suggest that all SDKs follow Go and Java and pass them unconditionally, and allow the server to decide what goes into effect. This will reduce the possibility of differences creeping in if we add/change conflict policies in the future.

mjameswh: I see. Good catch.

Yes, it indeed makes sense to be consistent across SDK, to avoid future surprises.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -0,0 +1,151 @@
...
+  if (ctx.callbackUrl) {
+    internalOptions.completionCallbacks = [
+      {
+        nexus: { url: ctx.callbackUrl, header: ctx.callbackHeaders },
+        links, // pass in links here as well for older servers, newer servers dedupe them.
+      },
+    ];
+  }
+
+  const { taskQueue: userSpeficiedTaskQueue, ...rest } = workflowOptions;
```
dandavison: typo in this variable name

dandavison (approved):
LGTM! Nice work @mjameswh and @bergundy.

PR #1711, opened 2025-05-08
Simplify the type of `MockActivityEnvironment.run`

### LINKED ISSUE #1710: [Feature Request] Better type inference for `MockActivityEnvironment.run`

This code:

```typescript
async function foo(): Promise<string> {
  return "foo";
}
const result = await new MockActivityEnvironment().run(foo);
result.startsWith("foo");
```

Does not compile with strict types because `result` is inferred to be `unknown` instead of `string`.

Change the signature of `MockActivityEnvironment.run` from:

```typescript
run<P extends any[], R, F extends ActivityFunction<P, R>>(fn: F, ...args: P): Promise<R>;
```

to

```typescript
run<F extends ActivityFunction>(fn: F, ...args: Parameters<F>): ReturnType<F>
```

I've checked this locally and it makes my example infer the type of `result` correctly.

TastyPi:
## What was changed

Simplified the type of `MockActivityEnvironment.run`

## Why?

Improves the type inference of the return type.

This example:

```typescript
async function foo(): Promise<string> {
  return "foo";
}
const result = await new MockActivityEnvironment().run(foo);
result.startsWith("foo");
```

Does not currently compile with strict types because `result` is inferred as `unknown` instead of `string`. This change makes the type inference work correctly.

## Checklist

1. Closes #1710

2. How was this tested:

Made the change on my local machine and saw the red squiggly lines disappear.

3. Any docs updates needed?

I don't know, this is technically a breaking change if anyone is currently setting the type parameters.

### DIFF
--- a/packages/testing/src/mocking-activity-environment.ts
+++ b/packages/testing/src/mocking-activity-environment.ts
@@ -54,7 +54,7 @@ export class MockActivityEnvironment extends events.EventEmitter {
   /**
    * Run a function in Activity Context
    */
-  public async run<P extends any[], R, F extends ActivityFunction<P, R>>(fn: F, ...args: P): Promise<R> {
+  public async run<F extends ActivityFunction>(fn: F, ...args: Parameters<F>): ReturnType<F> {
     return this.activity.runNoEncoding(fn as ActivityFunction<any, any>, { args, headers: {} }) as Promise<R>;
   }
 }


packages/testing/src/mocking-activity-environment.ts
```diff
@@ -54,7 +54,7 @@ export class MockActivityEnvironment extends events.EventEmitter {
   /**
    * Run a function in Activity Context
    */
-  public async run<P extends any[], R, F extends ActivityFunction<P, R>>(fn: F, ...args: P): Promise<R> {
+  public async run<F extends ActivityFunction>(fn: F, ...args: Parameters<F>): ReturnType<F> {
```
mjameswh, line 57: That's not compiling, because the compiler can't confirm that the return type of this async function is a `Promise`.

I think something like this should work:

```
public async run<F extends ActivityFunction>(fn: F, ...args: Parameters<F>): Promise<Awaited<ReturnType<F>>> {
```

packages/testing/src/mocking-activity-environment.ts
```diff
@@ -54,7 +54,7 @@ export class MockActivityEnvironment extends events.EventEmitter {
...
     return this.activity.runNoEncoding(fn as ActivityFunction<any, any>, { args, headers: {} }) as Promise<R>;
```
mjameswh, line 58: This line is still using the `R` type parameter. Again, you might have no choice but to cast the return value into `Promise<Awaited<ReturnType<F>>>`.

And while you are there, can you please remove the type cast on `fn`? I very much doubt this is required.

mjameswh:
Thanks!

At the moment, CI isn't passing. See my comments for hints on how to resolve these.

You may also add a test in `test-mockactivityenv.rs` using the exact code snippet you showed above. Basically:

```
test('MockActivityEnvironment return type is correctly inferred', async (t) => {
  async function foo(): Promise<string> {
    return 'foo';
  }
  const result = await new MockActivityEnvironment().run(foo);
  // There should be no compile time error on this line
  result.startsWith('foo');
  t.pass();
});
```

packages/testing/src/mocking-activity-environment.ts
TastyPi: I just removed `async` since it doesn't use `await`.

TastyPi:
Thanks for the review @mjameswh, admittedly I just did a quick edit in GitHub's integrated editor, sorry for not doing it properly :sweat_smile: 

I've made the changes you suggested, and I *think* it's all good now. I tried running `npm run test` in my repo, but a lot of seemingly unrelated tests failed, I'm not sure if I  just haven't got my local environment set up correctly or if that is expected.

PR #1719, opened 2025-05-30
Add application failure category

### LINKED ISSUE #1672: [Feature Request] Apply application failure logging and metrics behaviour according to ApplicationErrorCategory
See corresponding features issue:
https://github.com/temporalio/features/issues/614

Will need to wait for core:
https://github.com/temporalio/sdk-core/issues/897

Core-based SDKs only need to apply corresponding logging behaviour, core handles metrics.

THardy98:
## What was changed
Added category to `ApplicationFailure`, users can now specify a category to signify the severity of their application failure. Application failure categories also map to different observability behaviours, namely, a BENIGN application failure downgrades log emissions to `DEBUG` for activity failures.

1. Closes #1672

2. How was this tested:
small integration test

### DIFF
--- a/packages/common/src/failure.ts
+++ b/packages/common/src/failure.ts
@@ -101,6 +101,29 @@ export const [encodeRetryState, decodeRetryState] = makeProtoEnumConverters<
   'RETRY_STATE_'
 );
 
+export const ApplicationFailureCategory = {
+  /**
+   * BENIGN category errors emit DEBUG level logs and do not record metrics
+   */
+  BENIGN: 'BENIGN',
+} as const;
+
+export type ApplicationFailureCategory = (typeof ApplicationFailureCategory)[keyof typeof ApplicationFailureCategory];
+
+export const [encodeApplicationFailureCategory, decodeApplicationFailureCategory] = makeProtoEnumConverters<
+  temporal.api.enums.v1.ApplicationErrorCategory,
+  typeof temporal.api.enums.v1.ApplicationErrorCategory,
+  keyof typeof temporal.api.enums.v1.ApplicationErrorCategory,
+  typeof ApplicationFailureCategory,
+  'APPLICATION_ERROR_CATEGORY_'
+>(
+  {
+    [ApplicationFailureCategory.BENIGN]: 1,
+    UNSPECIFIED: 0,
+  } as const,
+  'APPLICATION_ERROR_CATEGORY_'
+);
+
 export type WorkflowExecution = temporal.api.common.v1.IWorkflowExecution;
 
 /**
@@ -172,7 +195,8 @@ export class ApplicationFailure extends TemporalFailure {
     public readonly nonRetryable?: boolean | undefined | null,
     public readonly details?: unknown[] | undefined | null,
     cause?: Error,
-    public readonly nextRetryDelay?: Duration | undefined | null
+    public readonly nextRetryDelay?: Duration | undefined | null,
+    public readonly category?: ApplicationFailureCategory | undefined | null
   ) {
     super(message, cause);
   }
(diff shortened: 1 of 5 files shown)


packages/common/src/failure.ts
```diff
@@ -261,6 +285,12 @@ export interface ApplicationFailureOptions {
    * Cause of the failure
    */
   cause?: Error;
+
+  /**
+   * Severity category of the application error.
+   * Maps to corresponding client-side logging/metrics behaviors.
```
mjameswh, line 291: Not sure what is meant by "corresponding **client-side** logging/metrics behaviors".

THardy98: As in not metrics reported by server

packages/core-bridge/src/worker.rs
```diff
@@ -771,10 +769,7 @@ mod custom_slot_supplier {
                 slot_type: SK::kind().into(),
                 task_queue: ctx.task_queue().to_string(),
                 worker_identity: ctx.worker_identity().to_string(),
-                worker_deployment_version: ctx
-                    .worker_deployment_version()
-                    .clone()
-                    .map(Into::into),
+                worker_deployment_version: ctx.worker_deployment_version().clone().map(Into::into),
```
mjameswh, line 772: Can you please check why the spurious reformat here? Could be that previous commit was not properly formatted, but that sometimes happen due to update in prettier. If the later, then we should make another PR at some point in the near future to bump our prettier dependency and reformat everything.

packages/common/src/failure.ts
```diff
@@ -172,7 +195,8 @@ export class ApplicationFailure extends TemporalFailure {
     public readonly nonRetryable?: boolean | undefined | null,
     public readonly details?: unknown[] | undefined | null,
     cause?: Error,
-    public readonly nextRetryDelay?: Duration | undefined | null
+    public readonly nextRetryDelay?: Duration | undefined | null,
+    public readonly category?: ApplicationFailureCategory | undefined | null
```
mjameswh, line 199: `ApplicationFailure` may also be thrown from Workflows. Does `category` being set has any impact anywhere? If not, do we plan to do so in the near future? Should we mention some limitation in doctypes?

packages/common/src/failure.ts
mjameswh: I see. What about this?

```suggest
Affects worker-side logging and metrics behaviors of this failure.
```

packages/common/src/failure.ts
THardy98: It's initially targeted just for activity failure logging, with the idea to expand on this based on user feedback. Aside from that, it's just part of the proto failure.

For the doctypes, do you mean to explicitly call out that the BENIGN category only impacts activity failure logging? I suppose we could, but not sure it's necessary.

mjameswh:
That's a pretty simple feature, but still, should probably add `@experimental` everywhere, just for safety.

packages/common/src/failure.ts
mjameswh: If the `category` property is only significant for `ApplicationFailure`s thrown from activities, it would be nice to users to indicate that fact. Otherwise, we can expect that some users will spend time trying to understand why their benign `ApplicationFailure` thrown from a workflow are still being logged, and then reporting that as a bug.

THardy98:
> That's a pretty simple feature, but still, should probably add `@experimental` everywhere, just for safety.

added to the category enum definition

packages/common/src/failure.ts
THardy98: Updated the doc string:

```
/** 
 * A category to describe the severity and change the observability behavior of an application failure.
 * 
 * Currently, observability behaviour changes are limited to:
 * - activities that fail due to a BENIGN application failure emit DEBUG level logs and do not record metrics
 * 
 * @experimental Category is a new feature and may be subject to change.
 */
 ```

packages/core-bridge/src/worker.rs
THardy98: my `package-lock.json` has prettier at `3.1.1` which leads me to think it's not a prettier update
(for reference, `package.json` has prettier at `^3.1.1"`

mjameswh (approved):
Great, thanks!

PR #1729, opened 2025-06-19
Activity pause/unpause

### LINKED ISSUE #1665: Heartbeating activities should be interrupted when the activities are paused.
See https://github.com/temporalio/features/issues/602

THardy98:
## What was changed
Added support to pause/unpause an activity. Cancellations due to pause are received through the heartbeat mechanism of the activity.

## Why?
Part of activity operator work.

1. Closes #1665

2. How was this tested:
Couple integration tests

3. Any docs updates needed?
Likely

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.12.1",
+  "version": "1.12.2",
   "npmClient": "npm",
   "command": {
     "publish": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,7 +56,7 @@
         "eslint-plugin-tsdoc": "^0.2.17",
         "fs-extra": "^11.2.0",
         "json5": "^2.2.3",
-        "lerna": "^7.4.2",
+        "lerna": "^8.2.3",
         "prettier": "^3.1.1",
         "ts-prune": "^0.10.3",
         "typescript": "^5.6.3",
@@ -390,9 +390,9 @@
       }
     },
     "node_modules/@cypress/request": {
-      "version": "3.0.7",
-      "resolved": "https://registry.npmjs.org/@cypress/request/-/request-3.0.7.tgz",
-      "integrity": "sha512-LzxlLEMbBOPYB85uXrDqvD4MgcenjRBLIns3zyhx7vTPj/0u2eQhzXvPiGcaJrV38Q9dbkExWp6cOHPJ+EtFYg==",
+      "version": "3.0.9",
+      "resolved": "https://registry.npmjs.org/@cypress/request/-/request-3.0.9.tgz",
+      "integrity": "sha512-I3l7FdGRXluAS44/0NguwWlO83J18p0vlr2FYHrJkWdNYhgVoiYo61IXPqaOsL+vNxU1ZqMACzItGK3/KKDsdw==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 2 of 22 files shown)


packages/testing/src/mocking-activity-environment.ts
```diff
@@ -30,7 +30,7 @@ export interface MockActivityEnvironmentOptions {
  * will immediately be in a cancelled state.
  */
 export class MockActivityEnvironment extends events.EventEmitter {
-  public cancel: (reason?: any) => void = () => undefined;
+  public cancel: (reason?: any, details?: any) => void = () => undefined;
```
mjameswh: What would a `MockActivityEnvironment` user be expected to provide as input for the `details` argument? The signature says `any`, and the `ActivityCancellationDetails` class has a private constructor, so it looks like it is simply impossible for users to provide anything useful here.

One solution would be to have `ActivityCancellationDetails` be just an interface, rather than a class, but then we'd have to make all properties optional, as adding more fields in the future would be a backward incompatible change.

Can you please look at what we do with this in other SDKs, and try to figure out a proper API for this?

And btw, `reason` being `any` also feels wrong.

packages/activity/src/index.ts
```diff
@@ -289,6 +298,11 @@ export class Context {
    */
   public readonly metricMeter: MetricMeter;
 
+  /**
+   * Holder object for activity cancellation details
+   */
+  public readonly cancellationDetails: ActivityCancellationDetailsHolder;
```
mjameswh: I understand why you introduced this holder. I'm not convinced this is the best solution. An alternative would have been to split `Context` into an interface AND an implementation class. We'd obviously need to keep to export a `Context` object, to keep the `current()` static method, but that wouldn't be hard to do. But whatever, there's nothing wrong with the holder approach.

My real concern is that we're making that indirection be part of the public `Context` API. That's not desirable from an API perspective, and totally avoidable by keeping that field private, and exposing a `cancellationDetails` property getter instead.

mjameswh:
I haven't compared with what we did in other SDKs, but it looks like there was quite a few discussions while you were working on the corresponding Python PR.

Are you sure this PR correctly mirrors the final state of your Python one?

packages/activity/src/index.ts
THardy98: I've made this private and provided a getter. 
I'm in favour of this simple change to begin with, if the introduction of shared reference mutable fields grows, we can opt to split Context into interface + class impl.

THardy98:
> I haven't compared with what we did in other SDKs, but it looks like there was quite a few discussions while you were working on the corresponding Python PR.
> 
> Are you sure this PR correctly mirrors the final state of your Python one?

Yes, this mirrors the Python activity pause change

packages/activity/src/index.ts
```diff
@@ -347,6 +363,10 @@ export class Context {
     });
     return Promise.race([this.cancelled.finally(() => clearTimeout(handle)), timer]);
   };
+
+  public cancellationDetails(): ActivityCancellationDetails | undefined {
```
mjameswh: Make this a getter. No readon to expose this as a function on Context. Note that this only applies here, you will need to keep the function as the top level, as you can't have a getter directly on a module.

```suggestion
  public get cancellationDetails(): ActivityCancellationDetails | undefined {
```

mjameswh: Please add Typedoc here.

packages/activity/src/index.ts
```diff
@@ -427,6 +447,13 @@ export function cancelled(): Promise<never> {
   return Context.current().cancelled;
 }
 
+/**
+ * Returns the cancellation details for this activity, if any.
```
mjameswh: nit: prefer the imperative form: `Return`, no third person `'s`.

packages/activity/src/index.ts
mjameswh: Should this be marked as `@experimental`? Same for other public facing APIs introduced in this PR.

packages/common/src/index.ts
```diff
@@ -8,6 +8,7 @@ import * as encoding from './encoding';
 import * as helpers from './type-helpers';
 
 export * from './activity-options';
+export * from './activity-cancellation-details';
```
mjameswh: The `ActivityCancellationDetailsHolder` type should not be exposed publicly. It is presently reexported due to this line. Do an `export { specificType1, specificType2 } ...` instead.

mjameswh (commented):
A few moere details before we can merge.

packages/activity/src/index.ts
THardy98: changed to getter, added Typedoc with experimental to public facing APIs

PR #1730, opened 2025-06-19
Activity reset

### LINKED ISSUE #1693: [typescript] SDK support for activity reset
See https://github.com/temporalio/features/issues/620

THardy98:
## What was changed
Added support to reset an activity. Cancellations due to reset are received through the heartbeat mechanism of the activity.

## Why?
Part of activity operator work.

1. Closes #1693

2. How was this tested:
Integration test

3. Any docs updates needed?
Likely

### DIFF
--- a/packages/client/src/async-completion-client.ts
+++ b/packages/client/src/async-completion-client.ts
@@ -252,13 +252,13 @@ export class AsyncCompletionClient extends BaseClient {
     } catch (err) {
       this.handleError(err);
     }
+    // Note that it is possible for a heartbeat response to have multiple fields
+    // set as true (i.e. cancelled and pause).
     if (cancelRequested) {
       throw new ActivityCancelledError('cancelled');
-    }
-    if (reset) {
+    } else if (reset) {
       throw new ActivityResetError('reset');
-    }
-    if (paused) {
+    } else if (paused) {
       throw new ActivityPausedError('paused');
     }
   }
--- a/packages/client/src/async-completion-client.ts
+++ b/packages/client/src/async-completion-client.ts
@@ -43,6 +43,13 @@ export class ActivityCancelledError extends Error {}
 export class ActivityPausedError extends Error {}
 
 /**
+ * Thrown by {@link AsyncCompletionClient.heartbeat} when the reporting Activity
+ * has been reset.
+ */
+@SymbolBasedInstanceOfError('ActivityResetError')
+export class ActivityResetError extends Error {}
+
+/**
  * Options used to configure {@link AsyncCompletionClient}
  */
 export type AsyncCompletionClientOptions = BaseClientOptions;
@@ -219,6 +226,7 @@ export class AsyncCompletionClient extends BaseClient {
     const payloads = await encodeToPayloads(this.dataConverter, details);
     let cancelRequested = false;
     let paused = false;
+    let reset = false;
     try {
       if (taskTokenOrFullActivityId instanceof Uint8Array) {
         const response = await this.workflowService.recordActivityTaskHeartbeat({
@@ -229,6 +237,7 @@ export class AsyncCompletionClient extends BaseClient {
         });
         cancelRequested = !!response.cancelRequested;
         paused = !!response.activityPaused;
+        reset = !!response.activityReset;
       } else {
         const response = await this.workflowService.recordActivityTaskHeartbeatById({
           identity: this.options.identity,
@@ -238,12 +247,16 @@ export class AsyncCompletionClient extends BaseClient {
         });
         cancelRequested = !!response.cancelRequested;
         paused = !!response.activityPaused;
+        reset = !!response.activityReset;
       }
     } catch (err) {
       this.handleError(err);
     }
     if (cancelRequested) {
       throw new ActivityCancelledError('cancelled');
+    }
+    if (reset) {
+      throw new ActivityResetError('reset');
     }
     if (paused) {
       throw new ActivityPausedError('paused');
--- a/packages/test/src/helpers-integration.ts
+++ b/packages/test/src/helpers-integration.ts
@@ -299,7 +299,7 @@ export function configurableHelpers<T>(
 export async function setActivityState(
   handle: WorkflowHandle,
   activityId: string,
-  state: 'pause' | 'unpause' | 'reset'
+  state: 'pause' | 'unpause' | 'reset' | 'pause & reset'
 ): Promise<void> {
   const desc = await handle.describe();
   const req = {
@@ -314,9 +314,11 @@ export async function setActivityState(
     await handle.client.workflowService.pauseActivity(req);
   } else if (state === 'unpause') {
     await handle.client.workflowService.unpauseActivity(req);
+  } else if (state === 'reset') {
+    await handle.client.workflowService.resetActivity({ ...req, resetHeartbeat: true });
   } else {
-    const resetReq = { ...req, resetHeartbeat: true };
-    await handle.client.workflowService.resetActivity(resetReq);
+    await handle.client.workflowService.pauseActivity(req);
+    await handle.client.workflowService.resetActivity({ ...req, resetHeartbeat: true });
   }
   await waitUntil(async () => {
     const { raw } = await handle.describe();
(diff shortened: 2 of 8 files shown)


packages/test/src/helpers-integration.ts
```diff
@@ -295,22 +299,33 @@ export async function setActivityPauseState(handle: WorkflowHandle, activityId:
...
   };
-  if (pause) {
+  if (state === 'pause') {
     await handle.client.workflowService.pauseActivity(req);
-  } else {
+  } else if (state === 'unpause') {
     await handle.client.workflowService.unpauseActivity(req);
+  } else {
+    const resetReq = { ...req, resetHeartbeat: true };
+    await handle.client.workflowService.resetActivity(resetReq);
```
cretz: I am looking into doing this in other langs too, can you confirm the current dev server supports this and/or do we need to upgrade to an RC one for tests?

THardy98: As in supports the `resetActivity` call? Yes should be fine. Latest CLI version is 1.4.1, which has temporal server version 1.28.0. `ResetActivity` was included in 1.27.1 with some improvements in 1.28.0

Edit: will need to set `frontend.activityAPIsEnabled` to true to enable this API

packages/client/src/async-completion-client.ts
```diff
@@ -238,13 +247,17 @@ export class AsyncCompletionClient extends BaseClient {
...
         paused = !!response.activityPaused;
+        reset = !!response.activityReset;
       }
     } catch (err) {
       this.handleError(err);
     }
     if (cancelRequested) {
       throw new ActivityCancelledError('cancelled');
     }
+    if (reset) {
```
mjameswh: Should we be worried about this cascade of `if` statements? Can multiple of these conditions be true at the same time? If so, are we 100% sure about the order?

packages/test/src/helpers-integration.ts
```diff
@@ -295,22 +299,33 @@ export async function setActivityPauseState(handle: WorkflowHandle, activityId:
...
     // If we are pausing: success when either
     //  • paused flag is true  OR
     //  • the activity vanished (it completed / retried)
-    if (pause) return activityInfo ? activityInfo.paused ?? false : true;
-    // If we are unpausing: success when either
-    //  • paused flag is false  OR
-    //  • the activity vanished (already completed)
-    return activityInfo ? !activityInfo.paused : true;
+    if (state === 'pause') {
+      return activityInfo ? activityInfo.paused ?? false : true;
```
mjameswh: Not new code, but pretty hard on reader.

packages/worker/src/activity.ts
```diff
@@ -186,8 +188,17 @@ export class Activity {
         } else if (this.cancelReason) {
           // Either a CancelledFailure that we threw or AbortError from AbortController
           if (err instanceof CancelledFailure) {
-            // If cancel due to activity pause, emit an application failure for the pause.
-            if (this.context.cancellationDetails?.paused) {
+            // If cancel due to activity pause or reset, emit an application failure.
+            if (this.context.cancellationDetails?.reset) {
```
mjameswh, line 213: Are we confident on if clause ordering? reset vs pause?

packages/worker/src/activity.ts
```diff
@@ -186,8 +188,17 @@ export class Activity {
...
+              return {
+                failed: {
+                  failure: await encodeErrorToFailure(
+                    this.dataConverter,
+                    new ApplicationFailure('Activity reset', 'ActivityReset')
```
mjameswh, line 218: - Retryable or not? Prefer making this explicit, for the benefit of readers.
- Is there any official semantic on this error? Does the server expect specific type?
- Should we avoid attaching stack trace on that specific error?

I'm a bit concerned here that `ApplicationFailure` are meant for user usage, so exposing activity reset as that error type might be "misinterpreted" as something they need to investigate, so they would look at the stack trace and try to figure it out. "Activity Reset" is a pretty generic name, an error message that users could have written themselves...

mjameswh (commented):
Had a very quick look at it, pointing out anything that popped to my mind without any real analysis, so be careful.

Leaving more complete review to @Sushisource.

packages/client/src/async-completion-client.ts
THardy98: I believe it's possible for multiple of these conditions can be true at the same time - user issues an activity cancel and a reset command between heartbeats (not sure if server has some logic that guards against this).

In any case, the ordering of these statements follows that of core:
https://github.com/temporalio/sdk-core/blob/d34c1d6d393462a816baf2469c256a21ffbaf196/core/src/worker/activities/activity_heartbeat_manager.rs#L148-L159

packages/worker/src/activity.ts
THardy98: Between reset and pause yes. Added a conditional for explicit cancellations, ordered _before_ the `reset` branch (to keep the cancellation -> reset -> pause ordering, though this is just for logging)

packages/worker/src/activity.ts
THardy98: 1. I've make non-retryable explicitly false.
2. The server does not expect a specific error type (this error returned is agnostic to server)
3. I don't really see a reason why

I'm less concerned about using `ApplicationFailure` for this, partly because it's a user-triggered error, and because it's what we chose to do for Python and TS for activity pause (though other SDKs diverge - Java introduced `ActivityPausedException` as a subtype of `ActivityCompletionException`, Go cancels the context with a `ErrActivityPaused` cause).

The alternative is we don't wrap and we emit a `CancelledFailure` with message `RESET`, see:
https://github.com/temporalio/sdk-typescript/blob/9089bc682b4334287e52cd0bdb4547836f63e335/packages/worker/src/activity.ts#L77-84
and
https://github.com/temporalio/sdk-typescript/blob/9089bc682b4334287e52cd0bdb4547836f63e335/packages/worker/src/worker.ts#L1072-1092
(the `else` conditional)

Whatever choice we make, we should also apply to the `pause` case, and potentially to Python as well.

THardy98: @Sushisource wdyt

Sushisource: IMO what Java (wrap with more specific types) is probably the most sensible. But I agree we'd need to apply it to Python and pause as well.

You should get @dandavison 's take and coordinate Python change with him

THardy98: @tconley1428 is working on activity reset right now

Sushisource: Then him, hehe :sweat_smile:

THardy98: After some discussion - we'll leave this as is.

packages/test/src/helpers-integration.ts
Sushisource: +1

packages/client/src/async-completion-client.ts
Sushisource: Well, though more than one may be true, the way this is written where each branch throws, that's kind of irrelevant. I would make them elses for that reason I think.

packages/test/src/test-integration-workflows.ts
```diff
@@ -1494,15 +1494,47 @@ test('Activity can be cancelled via pause and retry after unpause', async (t) =>
...
+      const activityInfo = raw.pendingActivities?.find((act) => act.activityId === testActivityId);
+      return !!(activityInfo && (await hasActivityHeartbeat(handle, testActivityId, 'heartbeated')));
+    }, 10000);
+
+    await setActivityState(handle, testActivityId, 'reset');
+    const result = await handle.result();
+    t.deepEqual(result, {
+      cancelRequested: false,
+      notFound: false,
+      paused: false,
```
Sushisource, line 1530: Do we have any tests where multiple things are true at once? EX: Paused and reset?

Sushisource (approved):
This makes sense to me, approving, but I have couple minor things to address, and possibly an extra test to add.

packages/test/src/test-integration-workflows.ts
THardy98: added

PR #1741, opened 2025-07-14
[HOLD] chore: support opentelemetry v2

### LINKED ISSUE #1658: [Feature Request] Add support for opentelemetry v2

JackTreble:
Closed: Being delivered by https://github.com/temporalio/sdk-typescript/pull/1951

## What was changed
Support OpenTelemetry v2

## Why?

So the latest versions of OpenTelemetry are supported going forward
This was a fairly naive update libraries job using the upgrade docs as a guide and the otel tests to check behaviour.

If you're reading this months in the future feel free to take on this PR, I may forget to come back to it!

## Checklist

1. Closes #1658 

2. How was this tested:
- Using the test suite, although some unrelated tests failed locally. might be my setup?

3. Any docs updates needed?

- Probably need to extract changes from https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md


mjameswh:
@JackTreble Thanks for your work on this, that looks awesome.

I'm a bit worried that this might be a breaking change for our users. Anything you can say about that? I suppose, for example, that users will need to simultaneously change their own code to also use OTel 2, and possibly also upgrade other OTel integration modules they have, right? Does OTel provide guidance for a smooth upgrade path?

In any case, we'll need to cross that bridge at one point or another, so I'd consider merging this as part of our next minor release. In the mean time, I'll leave this PR pending.

sdedovic:
Two notes:
- please see https://github.com/temporalio/sdk-typescript/issues/1779 and https://github.com/open-telemetry/opentelemetry-js/issues/5828#event-19364472301
- in my own codebase I have made similar changes. By importing the `NodeTracerProvider` from the node sdk, it adds the `async_hooks` and `events` modules which cause workers to fail unless you pass these to `bundlerOptions.ignoreModules` due to being sources of non-determinism. Have you tested this in your code?

sepehr500:
Just a heads up. LangFuse (and other LLMOps frameworks) require v2 of otel for compatibility. So given Temporal's push to be an llm agent framework, this will hold it back. Also, v1 otel will only be supported until February 2026 (I think) looking at this https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md

JackTreble:
Closed: Being delivered by https://github.com/temporalio/sdk-typescript/pull/1951

PR #1757, opened 2025-08-08
Eager Workflow Start

### LINKED ISSUE #1348: [Feature Request] Eager Workflow Start
Support for Eager Workflow Start (EWS) was added in the server in 1.20, and is currently supported in Java, Python, Core, and Go SDKs. It reduces the latency to start a workflow by allowing the client to directly schedule the first workflow task to a local worker. See [https://github.com/temporalio/features/issues/242](https://github.com/temporalio/features/issues/242) for details.

The goal is to enable support for EWS in the TS SDK.

This is dependent on [https://github.com/temporalio/sdk-typescript/issues/3776](https://github.com/temporalio/sdk-typescript/issues/3776)

THardy98:
## What was changed
The main change to accommodate eager workflow start was adding `startWithDetails` to `WorkflowClientInterceptor` to support a (new) return type of `WorkflowStartOutput`. This allows us to return an object, giving us the ability to add new fields without issue - we add `eagerlyStarted` in this PR. This new interceptor method is injected if not defined at the internal `_start` call, and adapts the existing `start` interceptor method to use `startWithDetails`. We also define a new handle called `WorkflowHandleWithStartDetails` which extends the existing `WorkflowHandleWithFirstExecutionRunId`, which now includes whether the workflow was started eagerly or not. Adding fields to the handle is an adopted pattern in the TS SDK.

## Why?
Support eager workflow start.

1. Closes #1348

2. How was this tested:
Existing integration tests and 2 new integration tests.

3. Any docs updates needed?
Probably

### DIFF
--- a/packages/client/src/interceptor-adapters.ts
+++ b/packages/client/src/interceptor-adapters.ts
@@ -1,0 +1,42 @@
+import { WorkflowClientInterceptor, WorkflowStartInput, WorkflowStartOutput } from './interceptors';
+
+// Apply adapters to workflow client interceptor.
+export function adaptWorkflowClientInterceptor(i: WorkflowClientInterceptor): WorkflowClientInterceptor {
+  return adaptLegacyStartInterceptor(i);
+}
+
+// Adapt legacy `start` interceptors to the new `startWithDetails` interceptor.
+function adaptLegacyStartInterceptor(i: WorkflowClientInterceptor): WorkflowClientInterceptor {
+  // If it already has the new method, or doesn't have the legacy one, no adaptation is needed.
+  // eslint-disable-next-line deprecation/deprecation
+  if (i.startWithDetails || !i.start) {
+    return i;
+  }
+
+  // This interceptor has a legacy `start` but not `startWithDetails`. We'll adapt it.
+  return {
+    ...i,
+    startWithDetails: async (input, next): Promise<WorkflowStartOutput> => {
+      let downstreamOut: WorkflowStartOutput | undefined;
+
+      // Patched `next` for legacy `start` interceptors.
+      // Captures the full `WorkflowStartOutput` while returning `runId` as a string.
+      const patchedNext = async (patchedInput: WorkflowStartInput): Promise<string> => {
+        downstreamOut = await next(patchedInput);
+        return downstreamOut.runId;
+      };
+
+      const runIdFromLegacyInterceptor = await i.start!(input, patchedNext); // eslint-disable-line deprecation/deprecation
+
+      // If the interceptor short-circuited (didn't call `next`), `downstreamOut` will be undefined.
+      // In that case, we can't have an eager start.
+      if (downstreamOut === undefined) {
+        return { runId: runIdFromLegacyInterceptor, eagerlyStarted: false };
+      }
+
+      // If `next` was called, honor the `runId` from the legacy interceptor but preserve
+      // the `eagerlyStarted` status from the actual downstream call.
+      return { ...downstreamOut, runId: runIdFromLegacyInterceptor };
+    },
+  };
+}
--- a/packages/client/src/interceptor-adapters.ts
+++ b/packages/client/src/interceptor-adapters.ts
@@ -1,6 +1,5 @@
 import { WorkflowClientInterceptor, WorkflowStartInput, WorkflowStartOutput } from './interceptors';
 
-// Apply adapters to workflow client interceptor.
 export function adaptWorkflowClientInterceptor(i: WorkflowClientInterceptor): WorkflowClientInterceptor {
   return adaptLegacyStartInterceptor(i);
 }
--- a/packages/client/src/interceptors.ts
+++ b/packages/client/src/interceptors.ts
@@ -124,6 +124,8 @@ export interface WorkflowClientInterceptor {
    *
    * If you implement this method,
    * {@link signalWithStart} most likely needs to be implemented too
+   *
+   * @deprecated in favour of {@link startWithDetails}
    */
   start?: (input: WorkflowStartInput, next: Next<this, 'start'>) => Promise<string /* runId */>;
 
--- a/packages/client/src/interceptors.ts
+++ b/packages/client/src/interceptors.ts
@@ -132,8 +132,7 @@ export interface WorkflowClientInterceptor {
   /**
    * Intercept a service call to startWorkflowExecution
    *
-   * Successor to {@link start}. Unlike {@link start}, this method returns
-   * start details via {@link WorkflowStartOutput}.
+   * This method returns start details via {@link WorkflowStartOutput}.
    *
    * If you implement this method,
    * {@link signalWithStart} most likely needs to be implemented too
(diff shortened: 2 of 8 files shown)


packages/client/src/types.ts
```diff
@@ -163,6 +163,12 @@ export interface ConnectionLike {
    * @see https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal
    */
   withAbortSignal<R>(abortSignal: AbortSignal, fn: () => Promise<R>): Promise<R>;
+
+  /**
+   * Capability flag that determines whether the connection supports eager workflow start.
+   * This will only be true if the underlying connection is a {@link NativeConnection}.
+   */
+  readonly supportsEagerStart?: boolean;
```
mjameswh: @THardy98 I did [a similar thing](https://github.com/temporalio/sdk-typescript/pull/1769/files#diff-aebd6517d8da25ba56bc895d268847cf7cc26a16275652f20ce3667438d4de61R371-R373) as part of #1769.

Not sure if there's value in exposing this one publicly.

packages/client/src/workflow-options.ts
```diff
@@ -54,6 +54,12 @@ export interface WorkflowOptions extends CommonWorkflowOptions {
    * @experimental Deployment based versioning is experimental and may change in the future.
    */
   versioningOverride?: VersioningOverride;
+
+  /**
+   * Potentially reduce the latency to start this workflow by encouraging the server to
+   * start it on a local worker running with this same client.
+   */
+  requestEagerStart?: boolean;
```
mjameswh, line 62: Should it really need to be requested explicitly? Not sure what we did in other languages. If there's really value to this, should consider that as a tri-state? Like `false = absolutely not`, `undefined = if you can, but dont complain`, `true = yes, that's a requirement`.

packages/client/src/interceptors.ts
```diff
@@ -120,6 +126,18 @@ export interface WorkflowClientInterceptor {
    * {@link signalWithStart} most likely needs to be implemented too
    */
   start?: (input: WorkflowStartInput, next: Next<this, 'start'>) => Promise<string /* runId */>;
+
+  /**
```
mjameswh, line 132: Nice, I'm really happy to finally have a path forward.

- Deprecate old method, and explicitly encourage users to prefer implementing the new one instead.
- Not sure about the `WithDetails` suffix for new hooks, but 🤷🏻.

packages/client/src/workflow-client.ts
```diff
@@ -513,14 +523,53 @@ export class WorkflowClient extends BaseClient {
...
     interceptors: WorkflowClientInterceptor[]
-  ): Promise<string> {
+  ): Promise<WorkflowStartOutput> {
     const workflowType = extractWorkflowType(workflowTypeOrFunc);
     assertRequiredWorkflowOptions(options);
     const compiledOptions = compileWorkflowOptions(ensureArgs(options));
 
-    const start = composeInterceptors(interceptors, 'start', this._startWorkflowHandler.bind(this));
+    // Adapt legacy `start` interceptors to the new `startWithDetails` interface.
+    const adaptedInterceptors: WorkflowClientInterceptor[] = interceptors.map((i) => {
```
mjameswh: Nits:
- Should probably move these adaptors to a distinct file. The `workflow-client.ts` file is already quite heavy.
- I think we should adapt interceptors on Client creation, so that we don't have to do this repeatedly here.

packages/client/src/workflow-options.ts
THardy98: The Python SDK defaults this parameter to `false` for all calls, it doesn't look like it's enabled by default.

packages/client/src/types.ts
THardy98: changed to use similar symbol pattern (`defineProperty` + symbol)

packages/client/src/interceptors.ts
THardy98: deprecated 👍

packages/client/src/interceptor-adapters.ts
```diff
@@ -0,0 +1,42 @@
+import { WorkflowClientInterceptor, WorkflowStartInput, WorkflowStartOutput } from './interceptors';
+
+// Apply adapters to workflow client interceptor.
+export function adaptWorkflowClientInterceptor(i: WorkflowClientInterceptor): WorkflowClientInterceptor {
```
Sushisource: Comment mostly just repeats function name. The comment for the "real" function seems sufficient.

packages/client/src/interceptors.ts
```diff
@@ -118,8 +124,22 @@ export interface WorkflowClientInterceptor {
...
+   *
+   * @deprecated in favour of {@link startWithDetails}
    */
   start?: (input: WorkflowStartInput, next: Next<this, 'start'>) => Promise<string /* runId */>;
+
+  /**
+   * Intercept a service call to startWorkflowExecution
+   *
+   * Successor to {@link start}. Unlike {@link start}, this method returns
+   * start details via {@link WorkflowStartOutput}.
```
Sushisource: This comment will need to change when the deprecated start is removed. I'd get rid of the reference to the old thing and just talk about what it does.

packages/client/src/workflow-client.ts
```diff
@@ -1610,7 +1644,11 @@ export class WorkflowClient extends BaseClient {
       const factories = (this.options.interceptors as WorkflowClientInterceptors).calls ?? [];
       return factories.map((ctor) => ctor({ workflowId, runId }));
     }
-    return Array.isArray(this.options.interceptors) ? (this.options.interceptors as WorkflowClientInterceptor[]) : [];
+    const interceptors = Array.isArray(this.options.interceptors)
+      ? (this.options.interceptors as WorkflowClientInterceptor[])
+      : [];
+    // Apply adapters to workflow client interceptors.
```
Sushisource: Obvious comment

packages/client/src/workflow-options.ts
```diff
@@ -54,6 +54,12 @@ export interface WorkflowOptions extends CommonWorkflowOptions {
    * @experimental Deployment based versioning is experimental and may change in the future.
    */
   versioningOverride?: VersioningOverride;
+
+  /**
+   * Potentially reduce the latency to start this workflow by encouraging the server to
+   * start it on a local worker running with this same client.
```
Sushisource:
```suggestion
   * Potentially reduce the latency to start this workflow by requesting that the server
   * start it on a local worker running with this same client.
```

packages/client/src/workflow-client.ts
THardy98: moved to `interceptor-adapters.ts`

I tried moving the adapt interceptors logic to client creation, but it causes an otel interceptor test to fail (`Otel interceptor spans are connected and complete`). I spent probably too much time trying to debug this (I believe it has to do with losing `this` context when we try to transform the interceptor), but wasn't able to solve it without resorting back to does this transformation directly in `_start`

PR #1758, opened 2025-08-11
[Feature] Environment Configuration

### LINKED ISSUE #1727: [Feature Request] Environment Configuration

See https://github.com/temporalio/features/issues/441 and https://github.com/temporalio/proposals/blob/master/all-sdk/external-client-configuration.md

THardy98:
## What was changed
Environment configuration for Typescript SDK, using core base envconfig module

## Checklist

1. Closes #1727

2. How was this tested:
`test-envconfig.ts`

3. Any docs updates needed?
No

### DIFF
--- a/packages/client/src/envconfig.ts
+++ b/packages/client/src/envconfig.ts
@@ -414,11 +414,6 @@ export interface LoadClientConfigOptions {
    * contents of the file.
    */
   configSource?: DataSource;
-  /**
-   * If true, file loading is disabled. This is only used when `configSource`
-   * is not present.
-   */
-  disableFile?: boolean;
   /** If true, will error on unrecognized keys in the TOML file. */
   configFileStrict?: boolean;
   /**
@@ -492,7 +487,6 @@ export class ClientConfig {
     const bridgeConfig = native.loadClientConfig(
       path ?? null,
       data ?? null,
-      options.disableFile ?? false,
       options.configFileStrict ?? false,
       options.overrideEnvVars ?? null
     );
--- a/packages/core-bridge/src/envconfig.rs
+++ b/packages/core-bridge/src/envconfig.rs
@@ -135,7 +135,6 @@ impl From<CoreDataSource> for DataSource {
 fn load_client_config(
     path: Option<String>,
     data: Option<Vec<u8>>,
-    disable_file: bool,
     config_file_strict: bool,
     env_vars: Option<HashMap<String, String>>,
 ) -> BridgeResult<ClientConfig> {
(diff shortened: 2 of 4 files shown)


packages/client/src/envconfig.ts
```diff
@@ -0,0 +1,525 @@
+import * as fs from 'fs';
```
mjameswh, line 1: We can't have the `@temporalio/client` package depends on Core. User-side env config stuff will need to go either in the `worker` package, or we'll make a new package for this (e.g. `@temporalio/config`).

packages/client/src/envconfig.ts
```diff
@@ -0,0 +1,525 @@
...
+
+/**
+ * A data source for configuration, which can be a path to a file,
+ * the string contents of a file, or raw bytes.
+ *
+ * It is defined as a tagged union, to be consumed by other parts of the SDK.
+ */
+export type DataSource = { path: string } | { data: string | Buffer };
+
+interface ClientConfigTLSObject {
```
mjameswh, line 14: Seems like that type is meant to be exposed publicly, no? Same for `ClientConfigProfileObject` and some others.

packages/client/src/envconfig.ts
```diff
@@ -0,0 +1,525 @@
...
+  if (source === undefined) {
+    return undefined;
+  }
+
+  if ('path' in source) {
+    try {
+      return fs.readFileSync(source.path);
+    } catch (error) {
+      throw new Error(
+        `Failed to read file at path "${source.path}": ${error instanceof Error ? error.message : String(error)}`
```
mjameswh, line 88:
```suggestion
        `Failed to read file at path "${source.path}": ${(error as Error).message ?? String(error)}`
```

packages/client/src/envconfig.ts
```diff
@@ -0,0 +1,525 @@
...
+      data ?? null,
+      options.disableFile ?? false,
+      options.disableEnv ?? false,
+      options.configFileStrict ?? false,
+      options.overrideEnvVars ?? null
+    );
+    return clientConfigProfileFromBridge(bridgeProfile);
+  }
+
+  public toObject(): ClientConfigProfileObject {
```
mjameswh, line 348: Why do you need these `toObject` and `fromObject` methods?

packages/client/src/envconfig.ts
```diff
@@ -0,0 +1,525 @@
...
+   * @example
+   * ```ts
+   * const profile = ClientConfigProfile.load({ profile: 'production' });
+   * const { connectionOptions, namespace } = profile.toClientConnectConfig();
+   *
+   * const connection = await Connection.connect(connectionOptions);
+   * const client = new Client({ connection, namespace });
+   * ```
+   */
+  public toClientConnectConfig(): ClientConnectConfig {
```
mjameswh, line 386: I don't think that will work to instantiate a `NativeConnection`/`Worker`.

packages/core-bridge/src/envconfig.rs
```diff
@@ -0,0 +1,196 @@
...
+
+impl From<CoreClientConfig> for ClientConfig {
+    fn from(c: CoreClientConfig) -> Self {
+        Self {
+            profiles: c.profiles.into_iter().map(|(k, v)| (k, v.into())).collect(),
+        }
+    }
+}
+
+#[derive(TryIntoJs, Serialize)]
```
mjameswh, line 43: Why `Serialize`/`serde`? Seems like this is all being transfered using `TryIntoJs`.

THardy98:
Standalone `envconfig` package here: https://github.com/temporalio/sdk-typescript/pull/1795

PR #1762, opened 2025-08-13
Add Node 24 to CI test matrix

mjameswh:
## What was changed

- Add Node 24 to our CI test matrix.
  - Node 24 is out and will become the Active TLS in October 2025.
  - Note that we recommend use of Active TLS for production-grade Temporal applications. That is Node 22 at this time.


packages/test/src/test-workflows.ts
```diff
@@ -4,6 +4,7 @@ import vm from 'node:vm';
 import anyTest, { ExecutionContext, TestFn } from 'ava';
 import dedent from 'dedent';
 import Long from 'long'; // eslint-disable-line import/no-named-as-default
+import escapeStringRexep from 'escape-string-regexp';
```
chris-olszewski: Eventually this can be replaced by [`RegExp.escape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape). I chose this over a polyfill since the package was already pulled as a transitive dependency.

package-lock.json
```diff
@@ -12424,7 +12424,8 @@
     },
     "node_modules/nexus-rpc": {
       "version": "0.0.1",
-      "resolved": "git+ssh://git@github.com/nexus-rpc/sdk-typescript.git#f594a7fd9e33bd14e5ce1ed04c5225fc708e7866",
+      "resolved": "https://registry.npmjs.org/nexus-rpc/-/nexus-rpc-0.0.1.tgz",
```
chris-olszewski, line 12427: Switching from the git reference to the published package over the git one as NPM 11 has a [breaking change](https://docs.npmjs.com/cli/v11/using-npm/changelog#%EF%B8%8F-breaking-changes-1) which skips `prepare` scripts from running when `--ignore-scripts` is used. `nexus-rpc` needs `prepare` to produce `lib/` files which are not checked in, but are present in the published package.

packages/test/src/test-workflows.ts
```diff
@@ -172,6 +172,17 @@ function compareCompletion(
   );
 }
 
+// Compare stack traces while handling $CLASS keyword to match any identifier that isn't consistent across Node versions.
+// As of Node 24.6.0 type names are now present on source mapped stack traces
+// See [f33e0fcc83954f728fcfd2ef6ae59435bc4af059](https://github.com/nodejs/node/commit/f33e0fcc83954f728fcfd2ef6ae59435bc4af059)
+function compareFailureStackTrace(t: ExecutionContext<Context>, actual: string, expected: string) {
+  const escapedTrace = expected
+    .replace(/[|\\{}()[\]^$+*?.]/g, '\\$&')
+    .replace(/-/g, '\\x2d')
```
chris-olszewski: This can eventually be replaced with `RegExp.escape`, but that is only available as of Node 24.

packages/test/src/helpers.ts
```diff
@@ -98,6 +98,21 @@ export function cleanStackTrace(ostack: string): string {
...
 
+/**
+ * Compare stack traces using $CLASS keyword to match any inconsistent identifiers
+ *
+ * As of Node 24.6.0 type names are now present on source mapped stack traces which leads
+ * to different stack traces depending on Node version.
+ * See [f33e0fcc83954f728fcfd2ef6ae59435bc4af059](https://github.com/nodejs/node/commit/f33e0fcc83954f728fcfd2ef6ae59435bc4af059)
+ */
+export function compareStackTraceIdentifiers<T>(t: ExecutionContext<T>, actual: string, expected: string): void {
+  const escapedTrace = expected
```
chris-olszewski, line 109: Can be replaced with `RegExp.escape` added in Node 24 eventually

package-lock.json
mjameswh: Uhh? Actually the git-based dependency should have never been commited to main, not sure how come it slipped through. Replacing it with the published release is definitely the right thing to do.

packages/test/src/test-workflows.ts
mjameswh: Good to know, but that will be years. We're supporting Node 18 at the moment.

packages/test/src/test-integration-split-one.ts
```diff
@@ -204,11 +204,12 @@ test.serial('activity-failure with ApplicationFailure', configMacro, async (t, c
...
   t.deepEqual(err.cause.cause.details, ['details', 123, false]);
-  t.is(
-    cleanOptionalStackTrace(err.cause.cause.stack),
+  compareStackTraceIdentifiers(
+    t,
+    cleanOptionalStackTrace(err.cause.cause.stack)!,
     dedent`
   ApplicationFailure: Fail me
-      at Function.nonRetryable (common/src/failure.ts)
+      at $CLASS.nonRetryable (common/src/failure.ts)
```
mjameswh, line 212: That comes out very neat. Great!

packages/test/src/helpers.ts
```diff
@@ -98,6 +98,21 @@ export function cleanStackTrace(ostack: string): string {
...
 }
 
+/**
+ * Compare stack traces using $CLASS keyword to match any inconsistent identifiers
+ *
+ * As of Node 24.6.0 type names are now present on source mapped stack traces which leads
+ * to different stack traces depending on Node version.
+ * See [f33e0fcc83954f728fcfd2ef6ae59435bc4af059](https://github.com/nodejs/node/commit/f33e0fcc83954f728fcfd2ef6ae59435bc4af059)
+ */
+export function compareStackTraceIdentifiers<T>(t: ExecutionContext<T>, actual: string, expected: string): void {
```
mjameswh: Not sure about this function's name. Why `...Identifiers`?

Also, can probably avoid the `T` generic parameter, this function never depends on that type. Try accepting `ExcutionContext` instead, or `ExcutionContext<unknown>` if the parameter is required (it is not AFAIR).

packages/test/src/test-workflows.ts
```diff
@@ -800,24 +804,24 @@ test('failSignalWorkflow', async (t) => {
...
-            'Signal failed',
-            dedent`
-          ApplicationFailure: Signal failed
-              at Function.nonRetryable (common/src/failure.ts)
-              at test/src/workflows/fail-signal.ts
-          `,
-            'Error'
-          ),
-        ],
+        [makeFailWorkflowExecution('Signal failed', 'Error')],
```
mjameswh: What if we kept the reference stack trace as argument to `makeFailWorkflowExecution()`, and instead did the special handling of stack traces as part of `compareCompletion()`? We'd simply iterate over history events, looking for `WorkflowExecutionFailure` events (if any). WDYT?

packages/test/src/helpers.ts
chris-olszewski: I am not especially happy with the name. I wasn't sure if `compareStackTrace` was too generic.

mjameswh: `compareStackTrace` seems fine to me. And that's an internal, test-only function anyway, so we can reconsider in the future if we ever feel it needs to be more specific.

packages/test/src/test-workflows.ts
chris-olszewski: This is far cleaner. Thank you for the suggestion!

PR #1769, opened 2025-08-20
Expose Client on Activity Context

mjameswh:
## What was changed

- Added `Context.client` and `getClient()` in `@temporalio/activity`.
  It is backed by the Worker's NativeConnection.
- Now accept Client Interceptors to be specified on a Worker, through `WorkerOptions.interceptors.client`; those interceptors apply to the `Client` passed through activities context.
- Added support for the Operator Service, Test Service and Health Service over `NativeConnection`.


THardy98 (approved):
Generally looks good to me. Small comment on workflow service contract test.

Big PR, the connection injector cleanup is sweet. I think we could have reduced the diff by doing the workflow interceptor changes in a separate PR, but nbd.

packages/test/src/test-native-connection.ts
```diff
@@ -235,7 +260,147 @@ test('all WorkflowService methods are implemented', async (t) => {
...
+  methods.sort();
+  for (const method of methods) {
+    await (connection.testService as any)[method]({});
+    t.true(calledMethods.has(method), `method ${method} not called`);
+  }
+
+  await connection.close();
+  server.forceShutdown();
+});
+
```
THardy98, line 402: Do we have a corresponding test for workflow service?

mjameswh: Yes, it already existed.

PR #1771, opened 2025-08-21
Lint: Cleanup interceptors types and add type checking on wf completion promises

mjameswh:
## What was changed

- Cleanup interceptors types
- Add type checking on workflow completion Promises


packages/workflow/src/interceptors.ts
```diff
@@ -20,68 +18,212 @@ import { ChildWorkflowOptionsWithDefaults, ContinueAsNewOptions } from './interf
...
+/**
+ * A mapping from interceptor type to an optional list of interceptor implementations
+ */
+export interface WorkflowInterceptors {
+  inbound?: WorkflowInboundCallsInterceptor[];
+  outbound?: WorkflowOutboundCallsInterceptor[];
+  internals?: WorkflowInternalsInterceptor[];
+}
+
+// Workflow Inbound Calls Interceptors /////////////////////////////////////////////////////////////////////////////////
```
THardy98, line 49: nit: comment format is not consistent with the Activity/Worker interceptor formats

packages/workflow/src/interceptors.ts
```diff
@@ -20,68 +18,212 @@ import { ChildWorkflowOptionsWithDefaults, ContinueAsNewOptions } from './interf
...
+ * Input for {@link WorkflowInboundCallsInterceptor.handleQuery}.
+ */
 export interface QueryInput {
   readonly queryId: string;
   readonly queryName: string;
   readonly args: unknown[];
   readonly headers: Headers;
 }
 
+// Workflow Outbound Calls Interceptors /////////////////////////////////////////////////////////////////////////////////
```
THardy98, line 124: ditto

packages/workflow/src/interceptors.ts
```diff
@@ -151,90 +274,31 @@ export interface SignalWorkflowInput {
...
-}
+/**
+ * Input for {@link WorkflowOutboundCallsInterceptor.getMetricTags}.
+ */
+export type GetMetricTagsInput = MetricTags;
 
-/** Input for WorkflowInternalsInterceptor.dispose */
-// eslint-disable-next-line @typescript-eslint/no-empty-object-type
-export interface DisposeInput {}
+// Workflow Internals Interceptors /////////////////////////////////////////////////////////////////////////////////////
```
THardy98, line 301: ditto

THardy98 (approved):
Skimmed through - assuming it's the same change from the activity client + interceptors PR.

Left a couple nits but feel free to ignore

packages/workflow/src/interceptors.ts
mjameswh: Well observed, but that's actually intentional. They are not the same level of separation 😉

But using section separators is a practice I just recently introduced, so it might still not be obvious yet, and it's totally possible that I've done it inconsistently elsewhere.

I'll leave as-is for now.

PR #1777, opened 2025-08-27
Fix worker logs not getting fliushed in Core log level set to ERROR

### LINKED ISSUE #1750: [Bug] Workflow logs not propagated unless telemetryOptions.logging.filter is DEBUG or TRACE
### Describe the bug

Using sample https://github.com/temporalio/samples-typescript/tree/main/custom-logger unless the worker 
configures core logging filter to DEBUG or TRACE, workflow application logging, i.e.,  using `log.info` in workflow code, will not show up in the logs.

This is confusing because it would be expected if telemetryOptions.logging.filter is set to INFO, a `log.info` in workflow code will be reflected in the logs

### Minimal Reproduction
Change line https://github.com/temporalio/samples-typescript/blob/main/custom-logger/src/worker.ts#L33 to

`filter: makeTelemetryFilterString({ core: 'INFO' }),`

and the log.info in 
https://github.com/temporalio/samples-typescript/blob/main/custom-logger/src/workflows/index.ts#L10

does not show up in the logs

### Environment/Versions

Latest TS SDK 1.12.1

mjameswh:
## What was changed

Considerably improved the `Runtime`'s `NativeLogCollector` logic.

- `NativeLogCollector` now starts its own interval timer to trigger log flushing.
- `NativeLogCollector` now ensures that the number of buffered log entries can't exceed 2000.
- `NativeLogCollector` now ensure that each message gets throttled by at least 100 ms. This is required to ensure correct global ordering of messages.

## Why?

- When `Runtime` is configured to forward logs from Core to the lang side, we use a lang-side buffer (`NativeLogCollector`) to reorder log entries coming from different sources (notably Workflow log APIs through Sinks, and Core) based on their absolute timestamp. This helps reduce incoherencies in log order.
- Since 1.12.0, `NativeLogCollector` was relying on Core periodic flushes to perform its own flushing to the downstream logger. However, it turns out that Core won't flush unless there has been at least one log message in the mean time, which can be a rare occurrence if Core's log level is set to WARN or ERROR.
- This has been causing considerable delays in propagation of workflow logs (fixes #1750), as well as memory leak as the log buffer may reach disproportionate size.

### DIFF
--- a/packages/worker/src/runtime-logger.ts
+++ b/packages/worker/src/runtime-logger.ts
@@ -1,7 +1,7 @@
 import { Heap } from 'heap-js';
 import { SdkComponent } from '@temporalio/common';
 import { native } from '@temporalio/core-bridge';
-import { DefaultLogger, LogEntry, Logger, LogTimestamp } from './logger';
+import { DefaultLogger, FlushableLogger, LogEntry, Logger, LogTimestamp } from './logger';
 
 /**
  * A log collector that accepts log entries either through the TS `Logger` interface (e.g. used by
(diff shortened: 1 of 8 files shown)


packages/worker/src/runtime-logger.ts
```diff
@@ -25,10 +25,60 @@ export class NativeLogCollector {
...
+    this.logger = new DefaultLogger('TRACE', this.appendOne.bind(this));
+    (this.logger as FlushableLogger).flush = this.flush.bind(this);
+    (this.logger as FlushableLogger).close = this.close.bind(this);
+
     this.downstream = downstream;
     this.receive = this.receive.bind(this);
+
+    // Flush the buffer every so often.
+    // Unref'ed so that it doesn't prevent the process from exiting.
+    this.flushIntervalTimer = setInterval(this.flushExpired.bind(this), this.flushPassIntervalMs).unref();
```
THardy98: `// Flush the buffer every so often. -> // Flush expired messages at an interval.`

Might be worth also specifying that an "expired" message is simply a message that exceeds the `minBufferTimeMs`.

packages/worker/src/runtime-logger.ts
```diff
@@ -44,14 +94,20 @@ export class NativeLogCollector {
...
+      this.flushUnconditionally();
+      this.flushExpired();
     } catch (_e) {
       // We're not allowed to throw from here, and conversion errors have already been handled in
       // convertFromNativeLogEntry(), so an error at this point almost certainly indicates a problem
       // with the downstream logger. Just swallow it, there's really nothing else we can do.
     }
   }
 
+  private appendOne(entry: LogEntry): void {
```
THardy98, line 107: bit of a nit, but would prefer a better name here (i.e. `appendOneAndFlush`)
it's not clear unless you read the method that you are also flushing

packages/worker/src/runtime-logger.ts
```diff
@@ -78,15 +134,49 @@ export class NativeLogCollector {
...
     }
-    this.buffer.clear();
+  }
+
+  /**
+   * Flush messages without regard to the time threshold, up to a given number of messages.
+   *
+   * If no limit is provided, flushes messages in excess to the maximum buffer size.
+   */
+  private flushUnconditionally(maxFlushCount?: number): void {
```
THardy98: Might be worth mentioning that this usage of this method can cause incorrect ordering.

THardy98 (commented):
I see that this is an improvement over an unbounded buffer and has more forceful flushing mechanisms with the threshold flush and the unconditional flush.

I'm a bit concerned with hardcoded values that configure the buffer size and flush timers (`minBufferTimeMs`, `flushPassIntervalMs`, and `maxBufferSize`). The only potential issue I can see (aside from incorrectly ordered logs) is if the log emission rate is greater than the flushing rate. Not sure if that's a realistic concern.

In any case, it would be nice to benchmark the flushing functions. It would give a clear indicator at what log emission threshold these buffer/flush config values no longer work.

I generally think this is an improvement, depending on how strenuous the general logging case is.

THardy98 (approved):
Approving, given that we suspect the unconditional flush to handle when the timer flush is overwhelmed.

mjameswh:
Split code, renamed methods, and rephrased comments to clarify the points raised by @THardy98.

mjameswh:
> I'm a bit concerned with hardcoded values that configure the buffer size and flush timers (minBufferTimeMs, flushPassIntervalMs, and maxBufferSize). The only potential issue I can see (aside from incorrectly ordered logs) is if the log emission rate is greater than the flushing rate. Not sure if that's a realistic concern.

It is not. That's what `flushUnconditionally()` is for. Renamed to `flushExcess()` for clarity.

PR #1786, opened 2025-09-19
Don't set error status on otel spans for benign exceptions

### LINKED ISSUE #1776: [Feature Request] Reclassify Benign Application errors in OpenTelemetry
See issue description here:

https://github.com/temporalio/sdk-python/issues/1041

THardy98:
## What was changed
DWISOTT

## Why?
We don't want benign exceptions to alert failures

1. Closes #1776

2. How was this tested:
Integration test

3. Any docs updates needed?
Don't think so

### DIFF
--- a/packages/interceptors-opentelemetry/src/instrumentation.ts
+++ b/packages/interceptors-opentelemetry/src/instrumentation.ts
@@ -3,7 +3,7 @@
  * @module
  */
 import * as otel from '@opentelemetry/api';
-import { Headers, defaultPayloadConverter } from '@temporalio/common';
+import { ApplicationFailure, ApplicationFailureCategory, Headers, defaultPayloadConverter } from '@temporalio/common';
 
 /** Default trace header for opentelemetry interceptors */
 export const TRACE_HEADER = '_tracer-data';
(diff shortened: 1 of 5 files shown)


packages/test/package.json
```diff
@@ -8,7 +8,8 @@
     "build:ts": "tsc --build",
     "build:protos": "node ./scripts/compile-proto.js",
     "test": "ava ./lib/test-*.js",
-    "test.watch": "ava --watch ./lib/test-*.js"
+    "test.watch": "ava --watch ./lib/test-*.js",
+    "test-otel": "ava ./lib/test-otel.js"
```
tconley1428: Does otel seem important enough to have this separately, or was this a during development thing?

packages/interceptors-opentelemetry/src/instrumentation.ts
```diff
@@ -43,8 +43,10 @@ async function wrapWithSpan<T>(
     span.setStatus({ code: otel.SpanStatusCode.OK });
     return ret;
   } catch (err: any) {
+    const isBenignErr = err instanceof ApplicationFailure && err.category === ApplicationFailureCategory.BENIGN;
     if (acceptableErrors === undefined || !acceptableErrors(err)) {
-      span.setStatus({ code: otel.SpanStatusCode.ERROR, message: err instanceof Error ? err.message : String(err) });
+      const statusCode = isBenignErr ? otel.SpanStatusCode.UNSET : otel.SpanStatusCode.ERROR;
+      span.setStatus({ code: statusCode, message: err instanceof Error ? err.message : String(err) });
```
mjameswh: Being an `Error` is neither necessary nor sufficient to deduce that `message` will be present. Prefer this instead:

```suggestion
      span.setStatus({ code: statusCode, message: (err as Error).message ?? String(err) });
```

packages/test/package.json
mjameswh: > Does otel seem important enough to have this separately,

No. I'd prefer we don't add such things.

THardy98: development thing - i'll remove

packages/interceptors-opentelemetry/src/instrumentation.ts
THardy98: changed

mjameswh (approved):

PR #1789, opened 2025-09-24
Bump uuid to 13.0.0

TastyPi:
## What was changed

Bumped `uuid` dependency from 9.0.1 to 13.0.0 and removed `@types/uuid` dependency

## Why?

`uuid` started being packaged with its own types in 10.0.0. We've been having weird issues with a tool that is trying to use `@types/uuid` to check types for `uuid@13.0.0`. Rather than trying to fix the tool, it seems reasonable to upgrade `uuid` so we can remove the dependency on `@types/uuid`. Temporal is our last dependency that depends on an old version of `uuid`.

## Checklist

1. Closes N/A (I didn't create an issue, can create one if needed)

2. How was this tested:

I tried to run `npm run build` and `npm run test`, but both failed for seemingly unrelated reasons.

Every import of this package is `import { v4 as uuid4 } from 'uuid'` which is still the same in 13.0.0, so it shouldn't change anything.

3. Any docs updates needed?

No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3143,12 +3143,6 @@
       "integrity": "sha512-/Ad8+nIOV7Rl++6f1BdKxFSMgmoqEoYbHRpPcx3JEfv8VRsQe9Z4mCXeJBzxs7mbHY/XOZZuXlRNfhpVPbs6ZA==",
       "dev": true
     },
-    "node_modules/@types/uuid": {
-      "version": "9.0.7",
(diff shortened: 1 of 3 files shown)


mjameswh:
`uuid` 12 and over is ESM only, which makes it unusable for us at the moment.

I think upgrading to `uuid` 11 would be possible, but we'll need to confirm this would have no operational impact.

Issue #1778, opened 2025-08-27
[Feature Request] Ensure gRPC binary metadata headers are supported


cretz:
### Describe the solution you'd like

See https://github.com/temporalio/features/issues/671. This is likely only a problem for native connection. Probably just need to accept `Uint8Array` as metadata value and send as binary to Tonic for -bin keys.

PR #1791, opened 2025-10-02
Add support for gRPC binary metadata values

chris-olszewski:
## What was changed
Added ability to send binary gRPC headers via `NativeConnection`.

Also added a test to verify that binary headers were accepted by `Connection`.

## Why?
Make progress on https://github.com/temporalio/features/issues/671

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1778

2. How was this tested:
Added setting and sending of binary headers at both the client and call level for both native and the standard client connection.

3. Any docs updates needed?
I believe the type expansion from `string` to `string | Buffer` is the only user facing change and that change should appear in the API reference.

Each commit is reviewable on its own.

### DIFF
--- a/packages/core-bridge/src/client.rs
+++ b/packages/core-bridge/src/client.rs
@@ -88,13 +88,23 @@ pub fn client_new(
 #[js_function]
 pub fn client_update_headers(
     client: OpaqueInboundHandle<Client>,
-    headers: HashMap<String, String>,
+    headers: HashMap<String, MetadataValue>,
 ) -> BridgeResult<()> {
+    let (ascii_headers, bin_headers) = config::partition_headers(Some(headers));
     client
         .borrow()?
         .core_client
         .get_client()
-        .set_headers(headers)
+        .set_headers(ascii_headers.unwrap_or_default())
+        .map_err(|err| BridgeError::TypeError {
+            message: format!("Invalid metadata key: {err}"),
+            field: None,
+        })?;
+    client
+        .borrow()?
+        .core_client
+        .get_client()
+        .set_binary_headers(bin_headers.unwrap_or_default())
         .map_err(|err| BridgeError::TypeError {
             message: format!("Invalid metadata key: {err}"),
             field: None,
@@ -742,7 +752,7 @@ mod config {
         }
     }
 
-    fn partition_headers(
+    pub(super) fn partition_headers(
         headers: Option<HashMap<String, MetadataValue>>,
     ) -> (
         Option<HashMap<String, String>>,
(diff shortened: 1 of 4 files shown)


packages/core-bridge/src/helpers/try_from_js.rs
```diff
@@ -175,8 +175,9 @@ impl<T: TryFromJs> TryFromJs for HashMap<String, T> {
         let mut map = Self::new();
         for key_handle in props {
             let key = key_handle.to_string(cx)?.value(cx);
-            let value = obj.get_value(cx, key_handle)?;
-            map.insert(key, T::try_from_js(cx, value)?);
+            let js_value = obj.get_value(cx, key_handle)?;
+            let value = T::try_from_js(cx, js_value).field(&key)?;
```
chris-olszewski, line 179: Improves the error message slightly by indicating which field on the object had the type issue

packages/core-bridge/src/client.rs
```diff
@@ -711,4 +741,32 @@ mod config {
...
+    fn partition_headers(
+        headers: Option<HashMap<String, MetadataValue>>,
+    ) -> (
+        Option<HashMap<String, String>>,
+        Option<HashMap<String, Vec<u8>>>,
+    ) {
+        let Some(headers) = headers else {
+            return (None, None);
+        };
+        // Maybe with_capacity this assuming primarily ascii headers?
```
chris-olszewski: Unsure if making the assumption that all headers are text would matter. It would just avoid the need to grow the hashmap if the number of text headers exceeded the initial hashmap allocation.

packages/core-bridge/src/client.rs
```diff
@@ -711,4 +741,31 @@ mod config {
...
+    fn partition_headers(
+        headers: Option<HashMap<String, MetadataValue>>,
+    ) -> (
+        Option<HashMap<String, String>>,
+        Option<HashMap<String, Vec<u8>>>,
+    ) {
+        let Some(headers) = headers else {
+            return (None, None);
+        };
+        let mut ascii_headers = HashMap::default();
```
chris-olszewski, line 764: Possible optimization is to create the hashmap with the same capacity as the initial headers map. Only downside I could see would be if the headers were mostly binary.

mjameswh: This is once per native client + on `clientUpdateHeader`. Neither of them should be called frequently enough to justify efforts on performance optimization IMO.

But if we'd indeed want to go there, I'd initiate _both_ hashmaps at the capacity of the initial headers map, up to a certain upper limit (something like 128 or 256). That should optimize performance in all cases while providing a reasonable upper limit on how much memory we may allocate uselessly.

But honestly, I wouldn't mind that one.

PR #1792, opened 2025-10-02
Integration with AI SDK

tconley1428:
## What was changed
Added a new `AiSDKPlugin` to support running AI SDK in workflows. The plugin registers new activities for model invocation and optionally `MCP` tools. Workflows can then use `temporalProvider` to create models in workflows.

## Checklist

1. Closes 

2. How was this tested:
Added a new test suite with local model coverage. `AI_SDK_REMOTE_TESTS` env variable is available to run against a real endpoint.

3. Any docs updates needed?

### DIFF
--- a/packages/ai-sdk/src/activities.ts
+++ b/packages/ai-sdk/src/activities.ts
@@ -76,26 +76,34 @@ export function createActivities(provider: ProviderV2, mcpClientFactories?: McpC
 function activitiesForName(name: string, mcpClientFactory: McpClientFactory): object {
   async function listToolsActivity(args: ListToolArgs): Promise<Record<string, ListToolResult>> {
     const mcpClient = await mcpClientFactory(args.clientArgs);
-    const tools = await mcpClient.tools();
+    try {
+      const tools = await mcpClient.tools();
 
-    return Object.fromEntries(
-      Object.entries(tools).map(([k, v]) => [
-        k,
-        {
-          description: v.description,
-          inputSchema: v.inputSchema,
-        },
-      ])
-    );
+      return Object.fromEntries(
+        Object.entries(tools).map(([k, v]) => [
+          k,
+          {
+            description: v.description,
+            inputSchema: v.inputSchema,
+          },
+        ])
+      );
+    } finally {
+      await mcpClient.close();
+    }
   }
   async function callToolActivity(args: CallToolArgs): Promise<any> {
     const mcpClient = await mcpClientFactory(args.clientArgs);
-    const tools = await mcpClient.tools();
-    const tool = tools[args.name];
-    if (tool === undefined) {
-      throw ApplicationFailure.retryable(`Tool ${args.name} not found.`);
+    try {
+      const tools = await mcpClient.tools();
+      const tool = tools[args.name];
+      if (tool === undefined) {
+        throw ApplicationFailure.retryable(`Tool ${args.name} not found.`);
+      }
+      return tool.execute(args.args, args.options);
+    } finally {
+      await mcpClient.close();
     }
-    return tool.execute(args.args, args.options);
   }
   return {
     [name + '-listTools']: listToolsActivity,
--- a/packages/ai-sdk/src/mcp.ts
+++ b/packages/ai-sdk/src/mcp.ts
@@ -23,6 +23,9 @@ export interface TemporalMCPClientOptions {
  * A Temporal MCP Client which uses activities to execute list tools and call tools.
  * This should match by name an MCPClientFactory registered with the AI SDK plugin to function.
  *
+ * This is intentionally similar to the AI SDK's ``MCPClient``, but is lacking functions not used by the framework.
+ * It doesn't close either since the underlying clients are created in activities.
+ *
  * @experimental The AI SDK plugin is an experimental feature; APIs may change without notice.
  */
 export class TemporalMCPClient {


packages/interceptors-opentelemetry/package.json
```diff
@@ -14,10 +14,10 @@
...
   "license": "MIT",
   "dependencies": {
-    "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/core": "^1.19.0",
-    "@opentelemetry/resources": "^1.19.0",
-    "@opentelemetry/sdk-trace-base": "^1.19.0"
+    "@opentelemetry/api": "1.9.0",
+    "@opentelemetry/core": "1.25.1",
+    "@opentelemetry/resources": "1.25.1",
+    "@opentelemetry/sdk-trace-base": "1.25.1"
```
chris-olszewski: Unless there's an issue with a newer version of these packages, we should probably be a good package use a range here
```suggestion
    "@opentelemetry/api": "^1.9.0",
    "@opentelemetry/core": "^1.25.1",
    "@opentelemetry/resources": "^1.25.1",
    "@opentelemetry/sdk-trace-base": "^1.25.1"
```

.github/workflows/ci.yml
```diff
@@ -253,6 +253,12 @@ jobs:
         continue-on-error: true
 
       # Sample 3: fetch-esm to local server
+      #
```
tconley1428: Merge artifact, will remove

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,85 @@
...
+  LanguageModelV2Content,
+  LanguageModelV2FinishReason,
+  LanguageModelV2ResponseMetadata,
+  LanguageModelV2Usage,
+  ProviderV2,
+  SharedV2Headers,
+  SharedV2ProviderMetadata,
+} from '@ai-sdk/provider';
+import type { experimental_MCPClient as MCPClient } from '@ai-sdk/mcp';
+import { ToolCallOptions } from 'ai';
```
chris-olszewski:
```suggestion
import type { ToolCallOptions } from 'ai';
```

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,85 @@
...
+            description: v.description,
+            inputSchema: v.inputSchema,
+          },
+        ])
+      );
+    },
+    async callTool(args: CallToolArgs): Promise<any> {
+      const mcpClient = await mcpClientFactory(args.clientArgs);
+      const tools = await mcpClient.tools();
+      return tools[args.name].execute(args.args, args.options);
```
chris-olszewski: Might be worthwhile making a special error when the tool isn't found instead of a `Cannot read properties of undefined`

packages/ai-sdk/src/index.ts
```diff
@@ -0,0 +1,9 @@
+// eslint-disable-next-line import/no-unassigned-import
+import 'ai';
```
chris-olszewski: A quick comment about what side effects we need from this import would be helpful

packages/ai-sdk/src/mcp.ts
```diff
@@ -0,0 +1,40 @@
+import { ToolSet } from 'ai';
+import * as workflow from '@temporalio/workflow';
+import { ActivityOptions } from '@temporalio/workflow';
```
chris-olszewski:
```suggestion
import type { ActivityOptions } from '@temporalio/workflow';
```

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,85 @@
...
+ * These activities allow workflows to call AI models while maintaining Temporal's
+ * execution guarantees and replay safety.
+ *
+ * @param provider The AI SDK provider to use for model invocations
+ * @param mcpClientFactory
+ * @returns An object containing the activity functions
+ *
+ * @experimental The AI SDK integration is an experimental feature; APIs may change without notice.
+ */
+export const createActivities = (provider: ProviderV2, mcpClientFactory?: (_: any) => Promise<MCPClient>): object => {
```
chris-olszewski: Unless we want to intentionally obscure what functions this returns
```suggestion
export const createActivities = (provider: ProviderV2, mcpClientFactory?: (_: any) => Promise<MCPClient>) => {
```

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,85 @@
...
+      return await model.doGenerate(options);
+    },
+  };
+  if (mcpClientFactory === undefined) {
+    return activities;
+  }
+  return {
+    ...activities,
+    async listTools(args: ListToolArgs): Promise<Record<string, ListToolResult>> {
+      const mcpClient = await mcpClientFactory(args.clientArgs);
```
chris-olszewski: Is there something preventing us from explicitly restricting the type of `clientArgs` to match what `mcpClientFactory` takes? I'm a little concerned with the amount of `any` in these types

packages/ai-sdk/src/plugin.ts
```diff
@@ -0,0 +1,24 @@
+import { ProviderV2 } from '@ai-sdk/provider';
```
chris-olszewski:
```suggestion
import type { ProviderV2 } from '@ai-sdk/provider';
```

packages/ai-sdk/src/mcp.ts
```diff
@@ -0,0 +1,40 @@
+import { ToolSet } from 'ai';
```
chris-olszewski:
```suggestion
import type { ToolSet } from 'ai';
```

chris-olszewski (commented):
Apologies for slow review. My biggest concern is usage of `any` for some of the user facing types where ideally we would want to rely on the type system to prevent mismatching args.

packages/ai-sdk/src/activities.ts
tconley1428: If we can find something more robust to do, good. The arguments to the factory are intended to be user provided types, we can't really put any particular bound on them.

packages/ai-sdk/src/index.ts
tconley1428: Yep not needed.

package.json
```diff
@@ -56,9 +57,9 @@
     "temporalio": "file:packages/meta"
   },
   "devDependencies": {
-    "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/core": "^1.19.0",
-    "@opentelemetry/sdk-node": "^0.46.0",
+    "@opentelemetry/api": "1.9.0",
```
mjameswh: Any reason we need these to have pinned versions instead of version ranges?

packages/worker/src/workflow/bundler.ts
```diff
@@ -255,7 +255,7 @@ exports.importInterceptors = function importInterceptors() {
       ignoreWarnings: [/Failed to parse source map/],
     };
 
-    const compiler = webpack(this.webpackConfigHook(options));
+    const compiler = webpack(this.webpackConfigHook(options))!;
```
mjameswh: I don't see any reason for the non-null assertion operator on this line.

packages/test/pnpm-lock.yaml
```diff
@@ -0,0 +1,5648 @@
+lockfileVersion: '9.0'
```
mjameswh: There should be no lockfile in this directory. It's probably the result of a mistake when you rebased your branch over Chris' works on switching the repo to PNPM.

Just delete the file, and the sibling `node_modules` directory if you have one on your local machine, then rerun `pnpm i` from the root of the project.

package.json
```diff
@@ -38,6 +38,7 @@
     "docs": "cd packages/docs && pnpm run maybe-install-deps-and-build-docs"
   },
   "dependencies": {
+    "@temporalio/ai-sdk": "workspace:*",
```
mjameswh, line 42: You also need to add to the `meta` package's pacakge.json, tsconfig.json and index.ts files. These control what's included in generated API docs.

packages/test/package.json
```diff
@@ -24,15 +24,20 @@
...
     "@grpc/grpc-js": "^1.12.4",
     "@grpc/proto-loader": "^0.7.10",
-    "@opentelemetry/api": "^1.7.0",
-    "@opentelemetry/core": "^1.19.0",
-    "@opentelemetry/exporter-trace-otlp-grpc": "^0.46.0",
-    "@opentelemetry/sdk-node": "^0.46.0",
-    "@opentelemetry/sdk-trace-base": "^1.19.0",
-    "@opentelemetry/semantic-conventions": "^1.19.0",
+    "@modelcontextprotocol/sdk": "^1.10.2",
+    "@opentelemetry/api": "1.9.0",
```
mjameswh: Can we avoid pinned dependencies here?

packages/test/src/test-ai-sdk.ts
```diff
@@ -0,0 +1,322 @@
...
+  middlewareWorkflow,
+  telemetryWorkflow,
+  toolsWorkflow,
+} from './workflows/ai-sdk';
+import { helpers, makeTestFunction } from './helpers-integration';
+import { getWeather } from './activities/ai-sdk';
+import { Worker } from './helpers';
+import EventType = temporal.api.enums.v1.EventType;
+
+const remoteTests = !!process.env.AI_SDK_REMOTE_TESTS;
```
mjameswh: Do you know that `!!"false"` is true?

Please use the following idiom instead (or look at `isSet` in `worker-options.ts`).
That will ensure that this variable follows the same thruthness conventions as our other env variables.

```suggestion
const remoteTests = ['1', 't', 'true'].includes((process.env.AI_SDK_REMOTE_TESTS ?? 'false').toLowerCase()),
```

packages/test/src/test-ai-sdk.ts
```diff
@@ -0,0 +1,322 @@
...
+function* helloWorkflowGenerator(): Generator<ModelResponse> {
+  yield textResponse('Test Haiku');
+}
+
+const test = makeTestFunction({
+  workflowsPath: require.resolve('./workflows/ai-sdk'),
+});
+
+test('Hello world agent responds in haikus', async (t) => {
+  t.timeout(120 * 1000);
```
mjameswh: I'm a bit worried about setting individual test timeout to 2 minutes. That means that eventual bugs that would prevent a worker from shutting down will cause the CI job to time out while running this test (which runs very early because `ai` starts with an `a`), with almost no context on what failed.

I suppose that's only pertinent when `remoteTests` is true, right?
And `remoteTests` would be false in CI, right?
If so, can we only bump up the test timeout to 2 minutes when remote tests are enabled?

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,88 @@
...
+ * @returns An object containing the activity functions
+ *
+ * @experimental The AI SDK integration is an experimental feature; APIs may change without notice.
+ */
+export const createActivities = (provider: ProviderV2, mcpClientFactory?: (_: any) => Promise<MCPClient>): object => {
+  const activities = {
+    async invokeModel(
+      modelId: string,
+      options: LanguageModelV2CallOptions
+    ): Promise<{
```
mjameswh: Seems like this inline type is duplicated in `doGenerate`, and possibly other places. Can we extract that to a single name ddefinition?

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,88 @@
...
+ * These activities allow workflows to call AI models while maintaining Temporal's
+ * execution guarantees and replay safety.
+ *
+ * @param provider The AI SDK provider to use for model invocations
+ * @param mcpClientFactory
+ * @returns An object containing the activity functions
+ *
+ * @experimental The AI SDK integration is an experimental feature; APIs may change without notice.
+ */
+export const createActivities = (provider: ProviderV2, mcpClientFactory?: (_: any) => Promise<MCPClient>): object => {
```
mjameswh: A few things:

- Prefer the function syntax rather than the const lambda form unless there's a reason for the later;
- Make a named type definition for `McpClientFactory`.
- Make the arg's type `unknown` rather than `any`.

```suggestion
// In mcp.ts
type McpClientFactory = (args: unknown) => Promise<MCPClient>;

export function createActivities(provider: ProviderV2, mcpClientFactory?: McpClientFactory): object {
```

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,88 @@
...
+  const activities = {
+    async invokeModel(
```
mjameswh: Should we not namespace those activities somehow, to avoid eventual conflicts with user defined activities, or those from other plugins? Could be something like e.g. `__ai_temporal_invokeModel`?

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,88 @@
...
+      providerMetadata?: SharedV2ProviderMetadata;
+      request?: { body?: unknown };
+      response?: LanguageModelV2ResponseMetadata & { headers?: SharedV2Headers; body?: unknown };
+      warnings: Array<LanguageModelV2CallWarning>;
+    }> {
+      const model = provider.languageModel(modelId);
+      return await model.doGenerate(options);
+    },
+  };
+  if (mcpClientFactory === undefined) {
```
mjameswh: I think we should still register the `listTools` and `callTool` activities even if there is no actual tool provided. The workflow side has no way to figure this out. `listTools` could return an empty array, and `callTool` could would throw the `Tool xxcxc not found` error, or even `No tool registered`.

packages/ai-sdk/src/mcp.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class TemporalMCPClient {
+  constructor(
+    // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+    readonly clientArgs?: any,
+    readonly options?: ActivityOptions
+  ) {}
+
+  async tools(): Promise<ToolSet> {
+    const tools: Record<string, ListToolResult> = await workflow
```
mjameswh: Is it the way you did this in Python? I mean, `listTools` being an activity call, which is call everytime `tools()` is called?

A few things that can be considered to make this more efficient:
- Cache `tools` after the first call
- Make `listTools` a local activity call instead of a regular activity;
- Extract the list of tools at bundle time and inject it as a static value into the workflow bundle (so no activity call needed to list tools).

packages/ai-sdk/src/plugin.ts
```diff
@@ -0,0 +1,24 @@
+import type { ProviderV2 } from '@ai-sdk/provider';
+import type { experimental_MCPClient as MCPClient } from '@ai-sdk/mcp';
+import { SimplePlugin } from '@temporalio/plugin';
+import { createActivities } from './activities';
+
+export interface AiSDKPluginOptions {
```
mjameswh: Need an experimental annotation here. And also some docs.

packages/ai-sdk/src/plugin.ts
```diff
@@ -0,0 +1,24 @@
...
+  mcpClientFactory?: (args?: any) => Promise<MCPClient>;
+}
+
+/**
+ * A Temporal plugin that integrates AI SDK providers for use in workflows.
+ * This plugin creates activities that allow workflows to invoke AI models.
+ *
+ * @experimental The AI SDK plugin is an experimental feature; APIs may change without notice.
+ */
+export class AiSDKPlugin extends SimplePlugin {
```
mjameswh: nil: This is inconsistent. AI and SDK are both acronyms, and so both shoujld either be written with only the first letter in capital, or both entirely in uppercase.

mjameswh: Same for options.

packages/ai-sdk/src/provider.ts
```diff
@@ -0,0 +1,92 @@
...
+    content: Array<LanguageModelV2Content>;
+    finishReason: LanguageModelV2FinishReason;
+    usage: LanguageModelV2Usage;
+    providerMetadata?: SharedV2ProviderMetadata;
+    request?: { body?: unknown };
+    response?: LanguageModelV2ResponseMetadata & { headers?: SharedV2Headers; body?: unknown };
+    warnings: Array<LanguageModelV2CallWarning>;
+  }> {
+    const result = await workflow
+      .proxyActivities(this.options ?? { startToCloseTimeout: '10 minutes' })
```
mjameswh: The `??` here will not guarantee that `startToCloseTimeout` is set, as `this.options` could be an empty object or similar.

Might be better to do `{ startToCloseTimeout: '10 minutes' , ...this.options }` instead.

I'd also suggest doing that in the class constructor instead of here, but no strong strong opinion on that one.

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,88 @@
+import {
```
mjameswh:
```suggestion
import type {
```

packages/ai-sdk/src/provider.ts
```diff
@@ -0,0 +1,92 @@
+import {
```
mjameswh:
```suggestion
import type {
```

packages/ai-sdk/src/testing.ts
```diff
@@ -0,0 +1,99 @@
+import {
```
mjameswh:
```suggestion
import type {
```

packages/ai-sdk/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "temporal",
+    "workflow",
+    "ai",
+    "ai-sdk",
+    "llm"
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT",
+  "dependencies": {
+    "@ai-sdk/provider": "^2.0.0",
```
mjameswh: I was able to build the project with only _type imports_ to the `@ai/*` packages, so it should be possible to lower these two dependencies to dev + peer dependencies instead of regular ones.

packages/ai-sdk/src/testing.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+/**
+ * A test provider implementation that creates TestModel instances for testing purposes.
+ * This provider allows testing AI-enabled workflows without external API dependencies.
+ *
+ * @experimental The AI SDK integration is an experimental feature; APIs may change without notice.
+ */
+export class TestProvider implements ProviderV2 {
```
mjameswh: I'm not clear why it is pertinent for this mock to be part of our public API. That would be used in a user's tests on the worker side, not in workflow, so there's no concern regarding eventual import of non-workflow-safe code.

Doesn't the `ai` package itself provide this kind of test provider? Or some other external package that we could recommend instead of taking responsibility for this?

mjameswh (changes requested):
See my previous comments.

package.json
tconley1428: Chris said the same, I thought I updated it. Will check again

packages/test/pnpm-lock.yaml
tconley1428: Sounds good

packages/test/src/test-ai-sdk.ts
tconley1428: Yeah, we could do that. I don't currently have it set to run the remoteTests at all in CI.

packages/ai-sdk/src/activities.ts
tconley1428: We didn't in Python, and it does come with the downside of being UI clutter since the name is shown there.

packages/ai-sdk/src/mcp.ts
tconley1428: It is one of the implementations we provided. Caching is a potential addition, but `tools()` is generally called once. 
> Extract the list of tools at bundle time and inject it as a static value into the workflow bundle (so no activity call needed to list tools).

I don't believe this is valid as the tool list is not known to be static.

packages/ai-sdk/src/provider.ts
tconley1428: Yeah, I think that's what I was thinking but not what I did, good call.

packages/ai-sdk/src/plugin.ts
tconley1428: Fair point

packages/ai-sdk/src/testing.ts
tconley1428: They have something kind of similar, although it doesn't do exactly what I needed. In Python we did expose it because OpenAI had basically no unit test support, but I could see an argument here for keeping it in test at least until there's a strong need.

packages/worker/src/workflow/bundler.ts
tconley1428: I have local issues with this for some reason, meant to remove it for PR.

packages/ai-sdk/src/activities.ts
tconley1428: That brings some reminders up for me. I need to reevaluate this to allow multiple MCP servers, that will change some things.

packages/ai-sdk/src/mcp.ts
tconley1428: It can be added later and is easy to cache for the user if needed, so I'll leave that off for now.

packages/ai-sdk/src/activities.ts
mjameswh: This comment was marked as resolved, but the code still shows a regular import.

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,103 @@
...
+ * These activities allow workflows to call AI models while maintaining Temporal's
+ * execution guarantees and replay safety.
+ *
+ * @param provider The AI SDK provider to use for model invocations
+ * @param mcpClientFactories A mapping of server names to functions to create mcp clients
+ * @returns An object containing the activity functions
+ *
+ * @experimental The AI SDK integration is an experimental feature; APIs may change without notice.
+ */
+export const createActivities = (provider: ProviderV2, mcpClientFactories?: [string, McpClientFactory][]): object => {
```
mjameswh: nit: Please use the function syntax for top level definitions, rather than const+lambda form.

```suggestion
export function createActivities(provider: ProviderV2, mcpClientFactories?: [string, McpClientFactory][]): object {
```

mjameswh: Same for other functions in this file.

packages/ai-sdk/src/plugin.ts
```diff
@@ -0,0 +1,29 @@
...
+import type { McpClientFactory } from './mcp';
+
+/**
+ * Options for the AI SDK plugin
+ *
+ * @experimental The AI SDK plugin is an experimental feature; APIs may change without notice.
+ */
+export interface AiSdkPluginOptions {
+  modelProvider: ProviderV2;
+  mcpClientFactories?: [string, McpClientFactory][];
```
mjameswh: I think that specific line deserves non-trivial documentation.

mjameswh: Also, why is this an array of string+factory, rather than an object (i.e. `Record<string, McpClientFactory> or `{ [serverName: string]: McpClientFactory }`)? The object form would guarantee that there can't be two factories with the same server name, and that server name can't be an empty string.

mjameswh: Also, should we make this a type union that would accept a single-server form, with no server name? Or does that pose a risk of NDE if users later add other servers (and therefore now need to specify a server name for the first server)?

packages/ai-sdk/src/mcp.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export class TemporalMCPClient {
+  constructor(readonly options: TemporalMCPClientOptions) {}
+
+  async tools(): Promise<ToolSet> {
+    workflow.log.info(`Options: ${this.options.activityOptions}`);
+    const activities = workflow.proxyActivities({ startToCloseTimeout: '10 minutes', ...this.options.activityOptions });
+
+    const listActivity = activities[this.options.name + '-listTools'];
+    if (listActivity === undefined) {
```
mjameswh: This can't happen.

The proxy object returned by `proxyActivities` has no knowledge of which activities have been registered on the worker. What ever property you ask from it, it will _always_ return a proxy function. It's only when the activity is scheduled and the activity task gets handled by a worker that it will figure out that there's no registered activity function under that name, which will be considered as a retryable failure, causing that workflow execution stuck retrying the activity _forever_ (the 10 minutes `startToCloseTimeout` won't help in that case, because that timeout is per attempt, not global).

packages/ai-sdk/src/plugin.ts
tconley1428: Yeah I don't love the single server form. Object is good, I think I had some problem with it but it makes more sense.

packages/ai-sdk/src/mcp.ts
tconley1428: Okay

tconley1428: The compiler doesn't appear to know that however, without it I get `TS2722: Cannot invoke an object which is possibly undefined`. Would you recommend just enforcing with `!`?

packages/ai-sdk/src/provider.ts
```diff
@@ -0,0 +1,81 @@
...
+  readonly supportedUrls = {};
+
+  constructor(
+    readonly modelId: string,
+    readonly options?: ActivityOptions
+  ) {}
+
+  async doGenerate(options: LanguageModelV2CallOptions): Promise<InvokeModelResult> {
+    const activities = workflow.proxyActivities({ startToCloseTimeout: '10 minutes', ...this.options });
+    if (activities.invokeModel === undefined) {
```
mjameswh: Same as before, the proxy object returned by `proxyActivities` will always give back a proxy function for whatever activity name that is requested. The type checker doesn't know about that. Just add a `!` non-null assertion operator, and remove this condition.

packages/ai-sdk/src/provider.ts
```diff
@@ -0,0 +1,81 @@
...
+      throw new Error('Unable to find model activity. Is the plugin registered?');
+    }
+    const result = await activities.invokeModel({ modelId: this.modelId, options });
+    if (result === undefined) {
+      throw new Error('Received undefined response from model activity.');
```
mjameswh: You need to throw an `ApplicationFailure` instead, which will cause the Workflow Execution to fail.

Throwing a plain `Error` will cause a WFT failure, and therefore, WFT retries. Which, in this case, is guaranteed to fail again, and again forever, as the activity call already completed and won't get called again on WFT retries.

By the way, this will also be true in all other languages, except for Java. Had a quick look at your Python implementation, and it seems ok, but you may want to double check.

packages/ai-sdk/src/activities.ts
```diff
@@ -0,0 +1,103 @@
...
+        },
+      ])
+    );
+  }
+  async function callToolActivity(args: CallToolArgs): Promise<any> {
+    const mcpClient = await mcpClientFactory(args.clientArgs);
+    const tools = await mcpClient.tools();
+    const tool = tools[args.name];
+    if (tool === undefined) {
+      throw new Error(`Tool ${args.name} not found.`);
```
mjameswh:
```suggestion
      throw ApplicationFailure.retryable(`Tool ${args.name} not found.`);
```

packages/ai-sdk/src/mcp.ts
```diff
@@ -0,0 +1,67 @@
...
+        toolName,
+        {
+          execute: async (args: any, options) => {
+            const activities = workflow.proxyActivities({
+              summary: toolName,
+              startToCloseTimeout: '10 minutes',
+              ...this.options,
+            });
+            const callActivity = activities[this.options.name + '-callTool'];
+            if (callActivity === undefined) {
```
mjameswh: Same as before, this can't be undefined.

mjameswh (approved):
Awesome! Thanks again.

PR #1793, opened 2025-10-06
chore: migrate to PNPM

chris-olszewski:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


.github/workflows/ci.yml
```diff
@@ -282,7 +283,7 @@ jobs:
   # FIXME: Update this job to reuse native build artifacts from compile-native-binaries
   features-tests:
     name: Features Tests
-    uses: temporalio/features/.github/workflows/typescript.yaml@main
+    uses: temporalio/features/.github/workflows/typescript.yaml@olszewski/pnpm
```
chris-olszewski: Used to verify CI setup. Once both PRs are approved will switch to `main` before landing

.github/actions/setup-node-deps/action.yml
```diff
@@ -25,33 +25,40 @@ outputs:
...
+        key: pnpm-main-${{ inputs.platform }}-${{ hashFiles('./pnpm-lock.yaml') }}
         restore-keys: |
-          npm-main-${{ inputs.platform }}-
+          pnpm-main-${{ inputs.platform }}-
 
     - name: Download dependencies
       shell: bash
       run: |
-        npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose
+        pnpm install --frozen-lockfile || pnpm install --frozen-lockfile || pnpm install --frozen-lockfile
```
mjameswh: Not 100% sure about this, but it seems reasonable to hope that `pnpm` will be more stable and not require this "retry 3 times" hack.

.github/workflows/ci.yml
```diff
@@ -260,6 +260,7 @@ jobs:
 
       # End samples
 
+      # TODO: PNPM doesn't write logs to file by default
```
mjameswh: I wouldn't mind just removing that step.

lerna.json
```diff
@@ -1,6 +1,7 @@
 {
   "version": "1.13.2",
   "npmClient": "npm",
+  "npmClient": "pnpm",
```
mjameswh, line 3: I'd expect the diff to show the previous line being deleted. Could it be that you duplicated the line rather than modify it?

package.json
mjameswh, line 1: There are still some scripts in this file that use `npm`.

packages/core-bridge/package.json
mjameswh, line 1: Some scripts here are still using npm.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,28 @@
+packages:
+  - 'packages/activity'
+  - 'packages/client'
+  - 'packages/cloud'
+  - 'packages/common'
+  - 'packages/core-bridge'
+  - 'packages/create-project'
+  - 'packages/docs'
```
mjameswh: I'm not sure what to do with the docs package. It was previously handled in isolation both to avoid conflicting dependencies and to speed up loading dependencies from an empty cache. PNPM may certainly help on both of those aspects, but I'm not convinced that that will be sufficient.

Could we consider removing docs from this workspace, and add a distinct pnpm-workspace.yaml file inside of docs?

lerna.json
chris-olszewski: Bad merge on my part 🤦

PR #1794, opened 2025-10-06
Plugins/initial

tconley1428:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


packages/client/src/connection.ts
```diff
@@ -441,6 +446,10 @@ export class Connection {
    * This method does not verify connectivity with the server. We recommend using {@link connect} instead.
    */
   static lazy(options?: ConnectionOptions): Connection {
+    options = options || {};
+    for (const plugin of options.plugins || []) {
```
tconley1428: ?? instead.

packages/client/src/connection.ts
```diff
@@ -685,3 +695,12 @@ export class Connection {
     return wrapper as WorkflowService;
   }
 }
+
+export interface Plugin {
```
tconley1428: Just give the full name here.

packages/worker/src/connection.ts
```diff
@@ -231,13 +233,24 @@ export class NativeConnection implements ConnectionLike {
...
     const enableTestService = internalOptions.supportsTestService ?? false;
 
     try {
       const runtime = Runtime.instance();
-      const client = await runtime.createNativeClient(options);
-      return new this(runtime, client, enableTestService);
+
+      let connectNative = async () => await runtime.createNativeClient(options)
+      for (const plugin of options.plugins || []) {
+        const cn = connectNative
```
tconley1428: ;

packages/worker/src/worker.ts
```diff
@@ -1956,6 +1970,16 @@ export class Worker {
    * To stop polling, call {@link shutdown} or send one of {@link Runtime.options.shutdownSignals}.
    */
   async run(): Promise<void> {
+    let nextFunction = (w: Worker) => w.runInternal();
+    for (const plugin of this.plugins) {
```
tconley1428: reverse list

packages/worker/src/connection.ts
```diff
@@ -231,13 +233,24 @@ export class NativeConnection implements ConnectionLike {
...
     const internalOptions = (options as InternalConnectionOptions)?.[InternalConnectionOptionsSymbol] ?? {};
     const enableTestService = internalOptions.supportsTestService ?? false;
 
     try {
       const runtime = Runtime.instance();
-      const client = await runtime.createNativeClient(options);
-      return new this(runtime, client, enableTestService);
+
+      let connectNative = async () => await runtime.createNativeClient(options)
+      for (const plugin of options.plugins || []) {
```
tconley1428: Reverse list

tconley1428: Use composeInterceptors instead?

packages/worker/src/plugin.ts
```diff
@@ -0,0 +1,190 @@
...
+   *
+   * Args:
+   *   config: The replay worker configuration to potentially modify.
+   *
+   * Returns:
+   *   The modified worker configuration.
+   */
+  configureReplayWorker(config: ReplayWorkerOptions): ReplayWorkerOptions;
+
+  runWorker(worker: Worker, next: (w: Worker) => Promise<void>): Promise<void>;
```
tconley1428: Document that it won't happen on replay worker

packages/worker/src/connection.ts
```diff
@@ -349,3 +363,14 @@ function tagMetadata(metadata: Metadata): Record<string, native.MetadataValue> {
     ])
   );
 }
+
+export interface Plugin {
+  configureNativeConnection(options: NativeConnectionOptions): NativeConnectionOptions;
+
+  connectNative(next: () => Promise<native.Client>): Promise<native.Client>;
```
tconley1428: Remove connectNative until we have both.

packages/test/src/test-plugins.ts
```diff
@@ -0,0 +1,260 @@
...
+      workflowExecutionTimeout: '30 seconds',
+      workflowId: randomUUID(),
+    });
+
+    t.is(result, 'Hello');
+  });
+});
+
+test('Worker plugins are passed from native connection', async (t) => {
+  const env = await TestWorkflowEnvironment.createLocal({ connectionPlugins: [new ExamplePlugin()] });
```
tconley1428: Add client plugins with plugin filtering.

packages/client/src/connection.ts
```diff
@@ -130,6 +130,8 @@ export interface ConnectionOptions {
    * @default 10 seconds
    */
   connectTimeout?: Duration;
+
+  plugins?: ConnectionPlugin[];
```
tconley1428, line 142: Make sure to note this is passed to things which take the connection.

packages/test/src/workflows/plugins.ts
```diff
@@ -0,0 +1,15 @@
+import { proxyActivities } from '@temporalio/workflow';
+import type * as activities from '../activities';
+
+const { echo } = proxyActivities<typeof activities>({
+  startToCloseTimeout: '20s',
+  retry: { initialInterval: 5, maximumAttempts: 1, nonRetryableErrorTypes: ['NonRetryableError'] },
+});
+
+export async function hello_workflow(): Promise<string> {
```
tconley1428: Underscores!

packages/worker/src/connection.ts
```diff
@@ -349,3 +363,14 @@ function tagMetadata(metadata: Metadata): Record<string, native.MetadataValue> {
     ])
   );
 }
+
+export interface Plugin {
```
tconley1428: NativeConnectionPlugin

tconley1428: Add `name` fields to interfaces.

packages/worker/src/plugin.ts
```diff
@@ -0,0 +1,46 @@
...
+  get name(): string;
+
+  /**
+   * Hook called when creating a worker to allow modification of configuration.
+   *
+   * This method is called during worker creation and allows plugins to modify
+   * the worker configuration before the worker is fully initialized. Plugins
+   * can add activities, workflows, interceptors, or change other settings.
+   */
+  configureWorker(options: WorkerOptions): WorkerOptions;
```
tconley1428: Make things other than name optional.

packages/worker/src/plugin.ts
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * This method is not called when running a replay worker, as activities will not be
+   * executed, and global state can't affect the workflow.
+   */
+  runWorker(worker: Worker, next: (w: Worker) => Promise<void>): Promise<void>;
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function isWorkerPlugin(p: any): p is WorkerPlugin {
+  return 'configureWorker' in p && 'configureReplayWorker' in p && 'runWorker' in p;
```
tconley1428: Pass all plugins to child objects in case they are used by grandchildren.

packages/plugin/src/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+  readonly tls?: PluginParameter<TLSConfig | boolean | null>;
+  readonly apiKey?: PluginParameter<string | (() => string)>;
+  readonly dataConverter?: PluginParameter<DataConverter>;
+  readonly clientInterceptors?: PluginParameter<ClientInterceptors>;
+  readonly activities?: PluginParameter<object>;
+  readonly nexusServices?: PluginParameter<nexus.ServiceHandler<any>[]>;
+  readonly workflowsPath?: PluginParameter<string>;
+  readonly workflowBundle?: PluginParameter<WorkflowBundleOption>;
+  readonly workerInterceptors?: PluginParameter<WorkerInterceptors>;
+  readonly runContext?: {
```
tconley1428: Rework into a single function

packages/plugin/src/plugin.ts
```diff
@@ -0,0 +1,212 @@
...
+  NativeConnectionOptions,
+  TLSConfig,
+} from '@temporalio/worker';
+
+type PluginParameter<T> = T | ((p: T | undefined) => T);
+
+export interface SimplePluginOptions {
+  readonly name: string;
+  readonly tls?: PluginParameter<TLSConfig | boolean | null>;
+  readonly apiKey?: PluginParameter<string | (() => string)>;
```
tconley1428: Remove function option.

packages/client/src/client.ts
```diff
@@ -52,9 +62,21 @@ export class Client extends BaseClient {
   public readonly taskQueue: TaskQueueClient;
 
   constructor(options?: ClientOptions) {
+    options = options ?? {};
+
+    // Add client plugins from the connection
+    options.plugins = (options.plugins ?? []).concat(options.connection?.plugins ?? []);
+
+    // Process plugins first to allow them to modify connect configuration
+    for (const plugin of options?.plugins ?? []) {
```
chris-olszewski: No need to handle undefined cases since they were set right before this
```suggestion
    for (const plugin of options.plugins) {
```

packages/client/src/client.ts
```diff
@@ -95,6 +117,7 @@ export class Client extends BaseClient {
       workflow: {
         queryRejectCondition: this.workflow.options.queryRejectCondition,
       },
+      plugins: plugins ?? [],
```
chris-olszewski, line 121: `plugins` will always be present
```suggestion
      plugins,
```

packages/client/src/client.ts
```diff
@@ -52,9 +62,21 @@ export class Client extends BaseClient {
...
   constructor(options?: ClientOptions) {
+    options = options ?? {};
+
+    // Add client plugins from the connection
+    options.plugins = (options.plugins ?? []).concat(options.connection?.plugins ?? []);
+
+    // Process plugins first to allow them to modify connect configuration
+    for (const plugin of options?.plugins ?? []) {
+      if (plugin.configureClient !== undefined) {
+        options = plugin.configureClient(options);
```
chris-olszewski, line 74: Would we ever expect/support a plugin altering the plugin list?

I'm not sure I can think of a reason for doing this, but the current implementation would allow altering the list, but the iteration wouldn't change. e.g. a removed plugin can still alter the options/added plugin will not get a chance to alter the options

packages/client/src/types.ts
```diff
@@ -122,6 +122,7 @@ export interface CallContext {
  */
 export interface ConnectionLike {
   workflowService: WorkflowService;
+  plugins: any[];
```
chris-olszewski: Can we restrict the type to just `NativeConnectionPlugin` or `ConnectionPlugin`?

packages/client/src/client.ts
tconley1428: We don't intend to support that. If you have an idea of how to enforce that, we could.

packages/client/src/types.ts
tconley1428: I _think_ so now that methods are optional. Really it could be any kind of plugin to get passed down, but they all can act like each other now.

packages/client/src/client.ts
tconley1428: Technically true but the compiler doesn't know it. For the lest typescripty of us, would this be recommended, silencing the compiler, or `!`?

packages/client/src/types.ts
tconley1428: Actually I'm not sure it can reference Native since it is in the client package, but `ConnectionPlugin` should be sufficient now that all plugins are sorta every plugin.

packages/client/src/client.ts
chris-olszewski: Ah apologies, I somehow missed the red squiggle when I suggested the change. If the compiler fails to infer a type narrowing, then I generally prefer to be overly cautious (keep the `?? []`). Ignore my comment

packages/client/src/client.ts
chris-olszewski: You could leverage `Omit` to make it clear that `plugins` won't be respected e.g. `configureClient?(options: ClientOptions): Omit<ClientOptions, 'plugins'>`

Running the configurations becomes a little more awkward as you have to destructure `options` each iteration e.g.
```
    // Add client plugins from the connection
    options.plugins = (options.plugins ?? []).concat(options.connection?.plugins ?? []);
    let { plugins, ...optionsWithoutPlugins } = options;

    // Process plugins first to allow them to modify connect configuration
    for (const plugin of options.plugins) {
      if (plugin.configureClient !== undefined) {
        optionsWithoutPlugins = plugin.configureClient({...optionsWithoutPlugins, plugins });
      }
    }
    options = { ...optionsWithoutPlugins, plugins };

    super(options);
```

It doesn't warn/error if the user tries to alter the plugin list, but any changes they make won't be respected.

tconley1428: `Omit` is a good idea. It seems like I can just change the type in the plugin to `Omit<ClientOptions, "plugins">` without any other change. They can return something with changed plugins, but expresses the intent

package.json
```diff
@@ -45,6 +45,7 @@
     "@temporalio/interceptors-opentelemetry": "file:packages/interceptors-opentelemetry",
     "@temporalio/nexus": "file:packages/nexus",
     "@temporalio/nyc-test-coverage": "file:packages/nyc-test-coverage",
+    "@temporalio/plugin": "file:packages/plugin",
```
mjameswh, line 49: You also need to add that package in the following places:
- `tsconfig.prune.json`
- `packages/meta/package.json`
- `packages/meta/tsconfig.json`
- `packages/meta/src/index.ts`

packages/client/src/client.ts
```diff
@@ -15,6 +16,14 @@ export interface ClientOptions extends BaseClientOptions {
    */
   interceptors?: ClientInterceptors;
 
+  /**
+   * List of plugins to register with the client.
+   *
+   * Plugins allow you to extend and customize the behavior of Temporal clients through a chain of
+   * responsibility pattern. They can intercept and modify client creation.
+   */
```
mjameswh, line 25: Please add this on all publicly exposed types, functons and other APIs introduced in this PR.

```
@experimental Plugins is an experimental feature; APIs may change without notice.
```

packages/plugin/src/plugin.ts
```diff
@@ -0,0 +1,198 @@
+import type * as nexus from 'nexus-rpc';
+import type { DataConverter } from '@temporalio/common';
+import {
```
mjameswh:
```suggestion
import type {
```

packages/plugin/src/plugin.ts
```diff
@@ -0,0 +1,198 @@
...
+): WorkerInterceptors | undefined {
+  return resolveParameterWithResolution(existing, parameter, (existing, parameter) => ({
+    client: resolveClientInterceptors(existing.client, parameter.client),
+    activity: resolveAppendParameter(existing.activity, parameter.activity),
+    nexus: resolveAppendParameter(existing.nexus, parameter.nexus),
+    workflowModules: resolveAppendParameter(existing.workflowModules, parameter.workflowModules),
+  }));
+}
+
+// eslint-disable-next-line deprecation/deprecation
```
mjameswh: The code got reformated, so this next-line comment is now misplaced.

packages/plugin/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "types": "./lib/index.d.ts",
+  "keywords": [
+    "temporal",
+    "workflow",
+    "worker",
+    "plugin"
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT",
+  "dependencies": {},
```
mjameswh, line 15: Need to add dev dependencies to sibling packages that are used for type import

packages/plugin/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./lib",
+    "rootDir": "./src"
+  },
+  "include": ["./src/**/*.ts"]
```
mjameswh, line 8: Needs to add "references" to other packages.

packages/plugin/src/plugin.ts
```diff
@@ -0,0 +1,198 @@
...
+  TLSConfig,
+} from '@temporalio/worker';
+
+type PluginParameter<T> = T | ((p: T | undefined) => T);
+
+export interface SimplePluginOptions {
+  readonly name: string;
+  readonly tls?: PluginParameter<TLSConfig | boolean | null>;
+  readonly apiKey?: PluginParameter<string>;
+  readonly dataConverter?: PluginParameter<DataConverter>;
```
mjameswh, line 49: We should have a custom resolve handler for `DataConverters` that merge each of the three subproperties independently. Under its present form, it would be essentially impossible to set this property as an immediate value.

packages/plugin/src/plugin.ts
```diff
@@ -0,0 +1,198 @@
...
+    return {
+      ...options,
+      dataConverter: resolveParameter(options.dataConverter, this.options.dataConverter),
+      workflowsPath: resolveParameter(options.workflowsPath, this.options.workflowsPath),
+      workflowBundle: resolveParameter(options.workflowBundle, this.options.workflowBundle),
+      interceptors: resolveWorkerInterceptors(options.interceptors, this.options.workerInterceptors),
+    };
+  }
+
+  async runWorker(worker: Worker, next: (w: Worker) => Promise<void>): Promise<void> {
```
mjameswh: No need to mark it as `async` if the function isn't using the `await` keyword.

packages/client/src/plugin.ts
```diff
@@ -0,0 +1,19 @@
+import type { ClientOptions } from './client';
```
mjameswh: nit: After a second pass over the whole PR, I think I'd prefer the `XxxPlugin` interface to be defined in the same type that defines `Xxx`, like you did for `Connection`.

packages/worker/src/plugin.ts
```diff
@@ -0,0 +1,41 @@
+import type { ReplayWorkerOptions, WorkerOptions } from './worker-options';
+import type { Worker } from './worker';
+
+/**
+ * Plugin interface for worker functionality.
+ *
+ * Plugins provide a way to extend and customize the behavior of Temporal workers.
+ * They allow you to intercept and modify worker configuration and worker execution.
+ */
+export interface WorkerPlugin {
```
mjameswh: Move this one to `worker-options.ts`, like at the end of the file.

packages/worker/src/worker.ts
```diff
@@ -555,6 +562,7 @@ export class Worker {
       compiledOptionsWithBuildId,
       logger,
       metricMeter,
+      options.plugins || [],
```
mjameswh:
```suggestion
      options.plugins ?? [],
```

packages/plugin/src/plugin.ts
tconley1428: I'm not 100% sure I agree. There is enough ambiguity in how one might resolve conflicts there that it might be better to require the user to define it. Python currently allows the fixed value to override it, I'm not sure the behavioral difference there is worth it.

packages/worker/src/worker-options.ts
```diff
@@ -1129,6 +1145,47 @@ function toNativeDeploymentOptions(options?: WorkerDeploymentOptions): native.Wo
...
+   * can add activities, workflows, interceptors, or change other settings.
+   */
+  configureWorker?(options: WorkerOptions): WorkerOptions;
+
+  /**
+   * Hook called when creating a replay worker to allow modification of configuration.
+   *
+   * This method is called during worker creation and allows plugins to modify
+   * the worker configuration before the worker is fully initialized. Plugins
+   * can add activities, workflows, interceptors, or change other settings.
```
mjameswh: nit: should not mention activities in the case of a replay worker.

```suggestion
   * can add workflows, interceptors, or change other settings.
```

packages/worker/src/worker.ts
```diff
@@ -1956,6 +1982,19 @@ export class Worker {
...
    */
   async run(): Promise<void> {
+    if (this.isReplayWorker) {
+      return this.runInternal();
+    }
+    let runWorker = (w: Worker) => w.runInternal();
+    for (let i = this.plugins.length - 1; i >= 0; --i) {
+      const rw = runWorker;
+      const plugin = this.plugins[i];
+      runWorker = (w: Worker) => plugin.runWorker?.(w, rw) ?? rw(w);
```
mjameswh: Please use an `if` block here.

packages/worker/src/workflow/bundler.ts
```diff
@@ -307,6 +315,21 @@ exports.importInterceptors = function importInterceptors() {
   }
 }
 
+export interface BundlerPlugin {
```
mjameswh, line 325: Missing typedoc + experimental tag.

PR #1795, opened 2025-10-07
Introduce `envconfig` package

### LINKED ISSUE #1727: [Feature Request] Environment Configuration

THardy98:
## What was changed
Created `envconfig` package, functionality that allows users to configure `Connection` or `NativeConnection` backed clients (or just the connection itself for that matter).

## Why?
Allows users to configure connections/clients from config files instead of source code.

1. Closes #1727 

2. How was this tested:
Integration test suite at `test-envconfig.ts`

3. Any docs updates needed?
Yes

### DIFF
--- a/packages/worker/src/connection-options.ts
+++ b/packages/worker/src/connection-options.ts
@@ -74,8 +74,8 @@ export function toNativeClientOptions(options: NativeConnectionOptions): native.
         serverRootCaCert: tlsInput.serverRootCACertificate ? Buffer.from(tlsInput.serverRootCACertificate) : null,
         clientTlsConfig: tlsInput.clientCertPair
           ? {
-              clientCert: Buffer.from(tlsInput.clientCertPair.crt),
-              clientPrivateKey: Buffer.from(tlsInput.clientCertPair.key),
+              clientCert: tlsInput.clientCertPair.crt && Buffer.from(tlsInput.clientCertPair.crt),
+              clientPrivateKey: tlsInput.clientCertPair.key && Buffer.from(tlsInput.clientCertPair.key),
             }
           : null,
       }


packages/envconfig/src/envconfig.ts
```diff
@@ -0,0 +1,77 @@
...
+  };
+}
+
+export function toTLSConfig(config?: ClientConfigTLS): TLSConfig | boolean | undefined {
+  if (config === undefined) {
+    return undefined;
+  }
+  if (config.disabled === true) {
+    return false;
+  }
```
chris-olszewski, line 60: More of a question than a suggestion, but if upstream `undefined` and `false` are treated the same, is there a reason for distinguishing between the two here?
```suggestion
  if (config === undefined || config?.disabled === true) {
    return undefined;
  }
```

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,137 @@
...
+  server_ca_cert_path?: string;
+  server_ca_cert_data?: string;
+  server_name?: string;
+  disable_host_verification?: boolean;
+}
+
+export interface tomlClientConfigCodec {
+  endpoint?: string;
+  auth?: string;
+}
```
chris-olszewski: Is there a reason for the lower cased type names?

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,155 @@
...
+  return Buffer.isBuffer(source.data) ? source.data : Buffer.from(source.data);
+}
+
+export function readPathSync(path?: string): Buffer | undefined {
+  if (path === undefined) {
+    return undefined;
+  }
+
+  return fs.readFileSync(path);
+}
```
chris-olszewski: I'm not sure I understand the usefulness of this function. Shouldn't the caller just check the presence of `path`? It adds some confusion about control flow as opposed to just using `fs.readFileSync` directly

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,333 @@
...
+    // If env var also not provided, fallback to default profile name.
+    if (profileName === undefined) {
+      profileName = DEFAULT_CONFIG_FILE_PROFILE;
+      profileSet = false;
+    }
+    const tomlProfile = tomlClientConfig.profile[profileName];
+    // If toml profile does not exist and an explicit profile was requested, error.
+    if (tomlProfile === undefined && profileSet) {
+      throw new Error(`Profile '${profileName}' not found in config data`);
+    }
```
chris-olszewski:
```suggestion
    // If profile name not provided, fallback to env variable.
    const profileName = options.profile ?? getEnvVar(envProvider, 'TEMPORAL_PROFILE');
     // If env var also not provided, fallback to default profile name.
    const tomlProfile = tomlClientConfig.profile[profileName ?? DEFAULT_CONFIG_FILE_PROFILE];
    // If toml profile does not exist and an explicit profile was requested, error.
    if (tomlProfile === undefined && profileName) {
      throw new Error(`Profile '${profileName}' not found in config data`);
    }
```

chris-olszewski (commented):
Mostly questions. Not sure I have enough context to approve

packages/envconfig/src/envconfig.ts
THardy98: They aren't necessarily treated the same upstream. The usage here:
```
  // TLS is enabled if we have an explicit TLS config, or if an api key is provided.
  const tls = toTLSConfig(profile.tls) ?? (profile.apiKey !== undefined ? true : undefined);
```
basically says:
`toTLSConfig = undefined -> check if apiKey was provided`
`toTLSConfig = false -> tls is false`
(^ or more generally:`toTLSConfig = non-nullish value -> tls is non-nullish value`)

packages/envconfig/src/types.ts
THardy98: No, I initially had these types as `TOMLClientConfigTLS` etc, but it was hard to read, so I search-and-replaced to `toml` and must've missed these types. I'll change these to be capitalized.

packages/envconfig/src/utils.ts
THardy98: Yeah fair - could probably be replaced with a ternary expression

edit: removed & replaced with ternary

packages/envconfig/src/envconfig-toml.ts
THardy98: Nice - much clearer.

package.json
```diff
@@ -42,6 +42,7 @@
     "@temporalio/cloud": "file:packages/cloud",
     "@temporalio/common": "file:packages/common",
     "@temporalio/create": "file:packages/create-project",
+    "@temporalio/envconfig": "file:packages/envconfig",
```
mjameswh, line 45: You will also need to add this package to the meta package (`packages/meta`), both in `package.json` and `src/index.ts`. The meta package is a legacy thing, but still controls what gets included in generated api docs.

packages/envconfig/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@temporalio/envconfig",
+  "version": "1.0.0",
```
mjameswh: Let's set the initial version to `0.0.1` to avoid confusion with something that would have actually existed in SDK 1.0.0. It will automatically get bumped to `1.13.2` (or whatever, same as other packages) when we publish a first release that includes it.

packages/envconfig/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "temporal",
+    "environment",
+    "configuration",
+    "client"
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT",
+  "dependencies": {
+    "@temporalio/common": "file:../common",
+    "@temporalio/worker": "file:../worker",
```
mjameswh: We specifically want to avoid a regular dependency on the `worker` package, as that would force `Client`-only users to still install that package and everything that comes with it.

From a glance at the code, it seems like the only import on `worker` is for an interface, so it should really be an `import type` statement, which can be satisfied with a dev dependency rather than a regular one.

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,129 @@
+import { NativeConnectionOptions } from '@temporalio/worker';
+
+/**
+ * A data source for configuration, which can be a path to a file,
+ * the string contents of a file, or raw bytes.
+ */
+export type ConfigDataSource = { path: string } | { data: string | Buffer };
```
mjameswh: Experimental

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+export interface ClientConfigTLS {
+  disabled?: boolean;
+  serverName?: string;
+  clientCert?: ConfigDataSource;
+  clientKey?: ConfigDataSource;
+  serverCACert?: ConfigDataSource;
+}
+
+export interface ClientConnectConfig {
```
mjameswh, line 29: Experimental

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,129 @@
...
+  /** If true, will error on unrecognized keys in the TOML file. */
+  configFileStrict?: boolean;
+  /**
+   * A dictionary of environment variables to use for loading and overrides.
+   * If not provided, the current process's environment is used.
+   */
+  overrideEnvVars?: Record<string, string>;
+}
+
+export interface ClientConfigProfile {
```
mjameswh, line 69: Experimental + doc

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Client configuration represents a client config file.
+ *
+ * @experimental Environment configuration is new feature and subject to change.
+ */
+export interface ClientConfig {
+  /** Map of profile name to its corresponding ClientConfigProfile. */
+  profiles: Record<string, ClientConfigProfile>;
+}
+
+export interface ClientConfigFromTomlOptions {
```
mjameswh, line 113: Experimental + doc

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,129 @@
...
+  // If true, will error if there are unrecognized keys.
+  strict: boolean;
+}
+
+export interface TomlClientConfig {
```
mjameswh: nit: Let's move this interface and other `Toml*` types to the `envconfig-toml.ts` file. Having them here, next to other exported types, makes it non-obvious that these are not meant to be publicly exposed, and why they use a different naming scheme. I'd even consider adding an `@internal` annotation comment on these, for extra clarity.

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,326 @@
...
+
+function getEnvVar(provider: EnvProvider, key: string): string | undefined {
+  if (provider.kind === 'Map') {
+    return provider.map[key];
+  }
+  return process.env[key];
+}
+
+export function tomlLoadClientConfig(options: LoadClientConfigOptions): TomlClientConfig {
+  const envProvider: EnvProvider = options.overrideEnvVars
```
mjameswh: Why not:

```
const env = options.overrideEnvVars ?? process.env;
```

And then you can simply do `env[key]` everywhere; no need for the `EnvProvider` type, nor the `getEnvVar` function.

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,326 @@
...
+  TomlClientConfigTLS,
+} from './types';
+import { normalizeGrpcMetaKey } from './utils';
+
+function sourceToStringData(source?: ConfigDataSource): string | undefined {
+  if (source === undefined) {
+    return undefined;
+  }
+  if ('path' in source) {
+    return readFileSync(source.path)?.toString();
```
mjameswh: Pass `{ encoding: 'utf-8' }` as the second argument to `readFileSync`, and its return value will be a string. No need for the extra `toString()` call.

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,326 @@
...
+export function tomlLoadClientConfig(options: LoadClientConfigOptions): TomlClientConfig {
+  const envProvider: EnvProvider = options.overrideEnvVars
+    ? { kind: 'Map', map: options.overrideEnvVars }
+    : { kind: 'System' };
+
+  let configData = undefined;
+  try {
+    configData = sourceToStringData(options.configSource) ?? getFallbackConfigData(envProvider);
+  } catch (error) {
+    const isFileNotFound = error instanceof Error && 'code' in error && error.code === 'ENOENT';
```
mjameswh: The `instanceof` check isn't adding any value here. Nor is the `'code' in error` part.

```suggestion
    const isFileNotFound = (error as Error).code === 'ENOENT';
```

packages/envconfig/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  ClientConfig,
+  ClientConfigProfile,
+  ClientConfigTLS,
+  LoadClientConfigOptions,
+  LoadClientProfileOptions,
+  ClientConfigFromTomlOptions,
+  ConfigDataSource,
+} from './types';
+
+export { fromTomlConfig, fromTomlProfile, toTomlConfig, toTomlProfile, loadConfigData } from './utils';
```
mjameswh, line 20: Do we really want to expose all of those functions publicly? If so, we'll also need to expose the `Toml*` types.

But I'd preach to avoid conflating the public API surface at this point.

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,326 @@
...
+}
+
+export function tomlLoadClientConfig(options: LoadClientConfigOptions): TomlClientConfig {
+  const envProvider: EnvProvider = options.overrideEnvVars
+    ? { kind: 'Map', map: options.overrideEnvVars }
+    : { kind: 'System' };
+
+  let configData = undefined;
+  try {
+    configData = sourceToStringData(options.configSource) ?? getFallbackConfigData(envProvider);
```
mjameswh, line 88: Are you sure about having the fallback in the try block? You say below that File Not Found is ok… Shouldn't we default to fallback config in that case?

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,147 @@
...
+export function normalizeGrpcMetaKey(key: string): string {
+  return key.toLocaleLowerCase().replace('_', '-');
+}
+
+export function fromTomlProfile(tomlProfile: TomlClientConfigProfile): ClientConfigProfile {
+  let grpcMeta: Record<string, string> | undefined = undefined;
+  if (tomlProfile.grpc_meta !== undefined) {
+    grpcMeta = {};
+    // Normalize GRPC meta keys.
+    for (const key in tomlProfile.grpc_meta) {
```
mjameswh: There are a few gotchas to iterating with the `in` operator, which may or may not affect you in a specific use case but may not be immediately obvious. Prefer the `of` operator instead.

You may also get both the key and value at once, so you don't need to fetch for it form the loop body.

```suggestion
    for (const [key, value] of Object.entries(tomlProfile.grpc_meta)) {
```

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,147 @@
...
+  };
+  return filterNullAndUndefined(profile);
+}
+
+export function toTomlProfile(profile: ClientConfigProfile): TomlClientConfigProfile {
+  let grpc_meta: Record<string, string> | undefined = undefined;
+  if (profile.grpcMeta !== undefined) {
+    grpc_meta = {};
+    // Normalize GRPC meta keys.
+    for (const key in profile.grpcMeta) {
```
mjameswh: Use `for .. of` instead of `for ... in`.

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,326 @@
...
+}
+
+function getTLSFromEnvVars(envProvider: EnvProvider): TomlClientConfigTLS | undefined {
+  const tlsConfig: TomlClientConfigTLS = {};
+  const tlsEnvVar = getEnvVar(envProvider, 'TEMPORAL_TLS');
+  if (tlsEnvVar !== undefined) {
+    tlsConfig.disabled = envVarToBool(tlsEnvVar);
+  }
+  const clientCertPath = getEnvVar(envProvider, 'TEMPORAL_TLS_CLIENT_CERT_PATH');
+  if (clientCertPath !== undefined) {
```
mjameswh: To make the code a bit simpler, why not instead unconditionally assign the properties, then use `filterNullAndUndefined` to get only keep fields that have values.

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,326 @@
...
+  const codecFromEnv = getCodecFromEnvVars(envProvider);
+  profile.codec = profile.codec || codecFromEnv ? { ...profile.codec, ...codecFromEnv } : undefined;
+  applyGrpcMetaFromEnvVars(profile, envProvider);
+}
+
+function getTLSFromEnvVars(envProvider: EnvProvider): TomlClientConfigTLS | undefined {
+  const tlsConfig: TomlClientConfigTLS = {};
+  const tlsEnvVar = getEnvVar(envProvider, 'TEMPORAL_TLS');
+  if (tlsEnvVar !== undefined) {
+    tlsConfig.disabled = envVarToBool(tlsEnvVar);
```
mjameswh: I'd suggest making `envVarToBool()` undefined-aware, i.e. make it accept `undefined` as input, returning the same in that case.

mjameswh: Combining this and the above suggestion, this could become:

```
const tlsConfig: TomlClientConfigTLS = filterNullAndUndefined({
  disabled: envVarToBool(env['TEMPORAL_TLS']),
  client_cert_path: env['TEMPORAL_TLS_CLIENT_CERT_PATH'],
  client_cert_data: env['TEMPORAL_TLS_CLIENT_CERT_DATA'],
  // etc 
});
return Object.keys(tlsConfig).length > 0 ? tlsConfig : undefined;
```

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,326 @@
...
+  }
+  return profile;
+}
+
+function applyProfileEnvVars(profile: TomlClientConfigProfile, envProvider: EnvProvider) {
+  profile.address = getEnvVar(envProvider, 'TEMPORAL_ADDRESS') ?? profile.address;
+  profile.namespace = getEnvVar(envProvider, 'TEMPORAL_NAMESPACE') ?? profile.namespace;
+  profile.api_key = getEnvVar(envProvider, 'TEMPORAL_API_KEY') ?? profile.api_key;
+  const tlsFromEnv = getTLSFromEnvVars(envProvider);
+  profile.tls = profile.tls || tlsFromEnv ? { ...profile.tls, ...tlsFromEnv } : undefined;
```
mjameswh, line 232: Have a look at the `mergeObjects` and `deepMerge`, from `object-helpers.ts`. That could be much simpler.

packages/envconfig/src/index.ts
THardy98: Yes. We have followed a pattern across SDKs to allow users to convert to/from TOML representations on their own.

packages/envconfig/src/envconfig-toml.ts
THardy98: Hmm, valid but we haven't done this in other SDKs. The core/go/java approach falls back to empty profile objects if the file isn't found:

Ex (core)
```
    if let Some(data) = toml_data {
        ClientConfig::from_toml(
            &data,
            ClientConfigFromTOMLOptions {
                strict: options.config_file_strict,
            },
        )
    } else {
        Ok(ClientConfig::default())
    }
```

packages/envconfig/src/envconfig-toml.ts
THardy98: `mergeObjects` would be nice here, but Typescript is having issues taking in `undefined` parameters despite the overload. That being said, I don't think this is _that_ unreadable or complex.

packages/envconfig/src/types.ts
THardy98: Moved, added @internal annotation

packages/envconfig/src/envconfig-toml.ts
THardy98: changed to `const isFileNotFound = (error as NodeJS.ErrnoException)?.code === 'ENOENT'` regular `Error` does not have a `code` property

packages/envconfig/src/envconfig-toml.ts
THardy98: much nicer

packages/envconfig/src/index.ts
mjameswh: In that case, we'll need to also export the `Toml*` interfaces and remove the `@internal` annotation on those.

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,116 @@
...
+  namespace?: string;
+}
+
+/**
+ * Options for loading a client configuration profile.
+ *
+ * @experimental Environment configuration is new feature and subject to change.
+ */
+export interface LoadClientProfileOptions {
+  /** The profile to load from the config. Defaults to "default". */
```
mjameswh:
```suggestion
  /** The name of the profile to load from the config. Defaults to "default". */
```

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * @internal
+ * @experimental Environment configuration is new feature and subject to change.
+ */
+export interface TomlClientConfigCodec {
+  endpoint?: string;
+  auth?: string;
+}
+
+function sourceToStringData(source?: ConfigDataSource): string | undefined {
```
mjameswh: Surely you didn't mean that it is valid to call this function without an argument…

```suggestion
function sourceToStringData(source: ConfigDataSource | undefined): string | undefined {
```

packages/envconfig/src/types.ts
```diff
@@ -0,0 +1,116 @@
+import type { NativeConnectionOptions } from '@temporalio/worker';
+
+/**
+ * A data source for configuration, which can be a path to a file,
+ * the string contents of a file, or raw bytes.
+ *
+ * @experimental Environment configuration is new feature and subject to change.
+ */
+export type ConfigDataSource = { path: string } | { data: string | Buffer };
```
mjameswh: I'm hesitant here.

For one thing, `Buffer` is a Node-specific type, which will pose issues at some point in the future when we'll add official support for other JS runtimes. We can use that class as much as we want internally, because replacing it at a later time won't be a breaking change, but it was a mistake to ever use that type in our public APIs. That should be `Uint8Array` instead, which is part of the EcmaScript standard.

Second, I get the feeling that `data` being `string | Uint8Array` is adding complexity without any real benefit. Any reason not to make this `Uint8Array` only?

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
+import { readFileSync } from 'fs';
+import { filterNullAndUndefined } from '@temporalio/common/lib/internal-workflow';
+import { ClientConfigProfile, ClientConfigTLS, ClientConfig, ConfigDataSource } from './types';
+import { normalizeGrpcMetaKey, TomlClientConfig, TomlClientConfigProfile, TomlClientConfigTLS } from './envconfig-toml';
+
+export function loadConfigData(source?: ConfigDataSource): Buffer | undefined {
```
mjameswh: This is exposed publicly, so it needs experimental + doc

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+  if (!source) return undefined;
+
+  if ('path' in source) {
+    return readFileSync(source.path);
+  }
+
+  return Buffer.isBuffer(source.data) ? source.data : Buffer.from(source.data);
+}
+
+export function fromTomlProfile(tomlProfile: TomlClientConfigProfile): ClientConfigProfile {
```
mjameswh, line 27: experimental + doc

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    address: tomlProfile.address,
+    namespace: tomlProfile.namespace,
+    apiKey: tomlProfile.api_key,
+    tls: fromTomlTLS(tomlProfile.tls),
+    grpcMeta,
+  };
+  return filterNullAndUndefined(profile);
+}
+
+export function toTomlProfile(profile: ClientConfigProfile): TomlClientConfigProfile {
```
mjameswh, line 51: experimental + doc

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    address: profile.address,
+    namespace: profile.namespace,
+    api_key: profile.apiKey,
+    tls: toTomlTLS(profile.tls),
+    grpc_meta,
+  };
+  return filterNullAndUndefined(tomlProfile);
+}
+
+export function fromTomlTLS(tomlTLS?: TomlClientConfigTLS): ClientConfigTLS | undefined {
```
mjameswh, line 75: experimental + doc

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    disabled: tomlTLS.disabled,
+    serverName: tomlTLS.server_name,
+    clientCert: toConfigDataSource(tomlTLS.client_cert_path, tomlTLS.client_cert_data, 'client_cert'),
+    clientKey: toConfigDataSource(tomlTLS.client_key_path, tomlTLS.client_key_data, 'client_key'),
+    serverCACert: toConfigDataSource(tomlTLS.server_ca_cert_path, tomlTLS.server_ca_cert_data, 'server_ca_cert'),
+  };
+  return filterNullAndUndefined(clientConfigTLS);
+}
+
+export function toTomlTLS(tlsConfig?: ClientConfigTLS): TomlClientConfigTLS | undefined {
```
mjameswh, line 94: experimental + doc

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    client_cert_data: clientCert?.data?.toString() ?? undefined,
+    client_key_path: clientKey?.path,
+    client_key_data: clientKey?.data?.toString(),
+    server_ca_cert_path: serverCACert?.path,
+    server_ca_cert_data: serverCACert?.data?.toString(),
+  };
+  return filterNullAndUndefined(tomlConfigTLS);
+}
+
+export function fromTomlConfig(tomlConfig: TomlClientConfig): ClientConfig {
```
mjameswh, line 119: experimental + doc

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+  const profiles: Record<string, ClientConfigProfile> = {};
+
+  for (const [profileName, profile] of Object.entries(tomlConfig.profile)) {
+    profiles[profileName] = fromTomlProfile(profile);
+  }
+
+  return { profiles };
+}
+
+export function toTomlConfig(config: ClientConfig): TomlClientConfig {
```
mjameswh, line 134: experimental + doc

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+  const profile: Record<string, TomlClientConfigProfile> = {};
+
+  for (const [profileName, configProfile] of Object.entries(config.profiles)) {
+    profile[profileName] = toTomlProfile(configProfile);
+  }
+
+  return { profile };
+}
+
+export function toPathAndData(source?: ConfigDataSource): { path?: string; data?: Buffer } | undefined {
```
mjameswh: Does this really need to be exported?

packages/envconfig/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+  if ('path' in source) {
+    return { path: source.path };
+  }
+  if (Buffer.isBuffer(source.data)) {
+    return { data: source.data };
+  }
+  return { data: Buffer.from(source.data, 'utf8') };
+}
+
+export function toConfigDataSource(
```
mjameswh: Does it need to be `export`?

packages/envconfig/src/types.ts
THardy98: I had switched it to use `Uint8Array` while making the change to make this standalone without core for this exact reason. But I noticed usage of `Buffer` in `tls-config.ts` and in the worker, so I figured that this was already a path walked and thought it'd be more consistent.

No issues changing to `Uint8Array`.

packages/envconfig/src/envconfig-toml.ts
THardy98: Uh yes - lol thank you

packages/envconfig/src/utils.ts
THardy98: Nope - unexported

packages/envconfig/src/utils.ts
THardy98: It's used as a helper by tests

packages/envconfig/src/types.ts
THardy98: Changed to `Uint8Array` along with the TLS fields that were formerly using Buffer

PR #1798, opened 2025-10-15
Add nightly throughput stress

THardy98:
## What was changed
Added a nightly workflow to run Omes throughput stress scenario.

### DIFF
--- a/.github/workflows/nightly-throughput-stress.yml
+++ b/.github/workflows/nightly-throughput-stress.yml
@@ -190,14 +190,3 @@ jobs:
             }
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_SDK_ALERTS_WEBHOOK }}
-
-      - name: Report test results
-        if: always()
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-throughput-stress.yml
```diff
@@ -0,0 +1,192 @@
+name: Nightly Throughput Stress
+
+on:
+  schedule:
+    # Run at 3 AM PST (11:00 UTC) - offset from existing nightly
+    - cron: '00 11 * * *'
+  push:
+    branches:
+      - nightly_tps
```
chris-olszewski, line 9: Once done testing the workflow
```suggestion
      - main
```

THardy98: TYFR :)
I don't think we intend to run this on every push, I added the "push to nightly_tps triggers the workflow" just for testing. Just running nightly I think is sufficient

PR #1803, opened 2025-10-21
fix: remove await from OTEL handleSignal interceptor

### LINKED ISSUE #1677: [Bug] Upgrading OTel Interceptor to 1.11.5+ breaks signal-with-start and may cause NDE
### Describe the bug

Assuming a workflow is launched with `signalWithStart`, the following workflow code in TS SDK 1.11.3 or earlier would result in sequence A-B-C. Same in later releases if not using the OTel workflow interceptor. However, in 1.11.5+ with OTel, we’d get A-C-B (i.e. equivalent of signal coming in a later WFT).

```
export async function myWorkflow(name: string): Promise<void> {
  log('A');
  setHandler(defineSignal('mySignal'), async () => {
    log('B');
  });
  log('C');
}
```

This change is due to addition of tracing propagation on inbound signals in 1.11.5 (#1449). Order of promise completion on outbound `scheduleLocalActivity` has similarly been modified in 1.11.6 (#1577). Both of these changes may result in Non-Determinism errors when older workflows are replayed on newer release of the SDK with the OTel interceptor installed. 

For context, the changes mentioned previously added a certain number of `await` statements to the corresponding code paths, so the actual code is now executed as a microtask rather than synchronously, hence the change in execution order. As a general rule, interceptors provided by the SDK itself should as much as possible avoid adding yield points. We should also document this risk on the Workflow Interceptor types, as user-implemented interceptors may face similar issues.

chris-olszewski:
## What was changed
Remove yield points from the `@temporalio/interceptors-opentelemetry` workflow interceptors. Some yield points were there since the initial version of the package and others were added in https://github.com/temporalio/sdk-typescript/pull/1449 and https://github.com/temporalio/sdk-typescript/pull/1577

## Why?
These yield points have the ability to cause NDE due to the inclusion of these interceptors. In addition, it broke `startWithSignal` operations as signal handlers were not immediately processed upon registration if the workflow was started with the given signal.

Depends on https://github.com/temporalio/sdk-core/pull/1040

## Checklist

1. Closes #1677 

2. How was this tested:
Created histories from the affected versions and added tests that replay these histories. These are red/green tests where they initially failed without the changes, but with the flagging logic they passed.

The only interceptors that are not covered by these replay tests are:
 - inbound `execute`
 - outbound `scheduleLocalActivity`
 - outbound `continueAsNew`

As I was unable to trigger NDE from removing/adding yield points to these interceptors.

3. Any docs updates needed?
Possibly a warning on https://docs.temporal.io/develop/typescript/interceptors#workflow-interceptors-registration about adding yield points to workflow interceptors.

### DIFF
--- a/.github/actions/setup-node-deps/action.yml
+++ b/.github/actions/setup-node-deps/action.yml
@@ -1,0 +1,57 @@
+name: 'Setup Node and Install Dependencies'
+description: 'Sets up Node.js, configures NPM cache, and installs dependencies with retry logic'
+
+inputs:
+  node-version:
+    description: 'Version of Node.js to install'
+    required: false
+    default: '22'
+  platform:
+    description: 'Platform identifier for cache key (e.g., linux-x64, macos-arm, windows-x64)'
+    required: true
+  save-cache:
+    description: 'Whether to save the NPM cache at the end'
+    required: false
+    default: 'false'
+
+outputs:
+  cache-dir:
+    description: 'Path to the NPM cache directory'
+    value: ${{ steps.npm-cache-dir.outputs.dir }}
+  cache-hit:
+    description: 'Whether the cache was restored successfully'
+    value: ${{ steps.restore-cache.outputs.cache-hit }}
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Install Node
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ inputs.node-version }}
+
+    - name: Get NPM cache directory
+      id: npm-cache-dir
+      shell: bash
+      run: echo "dir=$(npm config get cache)" >> ${GITHUB_OUTPUT}
+
+    - name: Restore NPM cache
+      id: restore-cache
+      uses: actions/cache/restore@v4
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: npm-main-${{ inputs.platform }}-${{ hashFiles('./package-lock.json') }}
+        restore-keys: |
+          npm-main-${{ inputs.platform }}-
+
+    - name: Download dependencies
+      shell: bash
+      run: |
+        npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose
+
+    - name: Save NPM cache
+      if: inputs.save-cache == 'true'
+      uses: actions/cache/save@v4
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: npm-main-${{ inputs.platform }}-${{ hashFiles('./package-lock.json') }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,10 +159,11 @@ jobs:
           name: corebridge-native-debug-${{ matrix.platform }}
           path: ./packages/core-bridge/releases
 
-      - name: Install Node
-        uses: actions/setup-node@v4
+      - name: Setup Node and Install Dependencies
+        uses: ./.github/actions/setup-node-deps
         with:
           node-version: ${{ matrix.node-release-override || matrix.node }}
+          platform: ${{ matrix.platform }}
 
       # On Windows, the 'runner.temp' variable uses backslashes as path separators, but
       # that may pose problems in later steps when we try to join that with subpaths;
@@ -175,23 +176,6 @@ jobs:
         id: tmp-dir
         run: echo "dir=$(pwd)" >> ${GITHUB_OUTPUT}
         working-directory: ${{ runner.temp }}
-
-      - name: Get NPM cache directory
-        id: npm-cache-dir
-        run: echo "dir=$(npm config get cache)" >> ${GITHUB_OUTPUT}
-
-      - name: Restore NPM cache
-        uses: actions/cache/restore@v4
-        with:
-          path: ${{ steps.npm-cache-dir.outputs.dir }}
-          key: npm-main-${{ matrix.platform }}-${{ hashFiles('./package-lock.json') }}
-          restore-keys: |
-            npm-main-${{ matrix.platform }}-
-
-      - name: Download dependencies
-        # Make up to 3 attempts to install NPM dependencies, to work around transient NPM errors :(
-        run: |
-          npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose
 
       - name: Compile code
         run: npm run build -- --ignore @temporalio/core-bridge
--- a/.github/workflows/conventions.yml
+++ b/.github/workflows/conventions.yml
@@ -20,23 +20,11 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Install Node
-        uses: actions/setup-node@v4
+      - name: Setup Node and Install Dependencies
+        uses: ./.github/actions/setup-node-deps
         with:
-          node-version: 22
-
-      - name: Get NPM cache directory
-        id: npm-cache-dir
-        shell: bash
-        run: echo "dir=$(npm config get cache)" >> ${GITHUB_OUTPUT}
-
-      - name: Restore NPM cache
-        uses: actions/cache/restore@v4
-        with:
-          path: ${{ steps.npm-cache-dir.outputs.dir }}
-          key: npm-main-linux-x64-${{ hashFiles('./package-lock.json') }}
-          restore-keys: |
-            npm-main-linux-x64-
+          node-version: '22'
+          platform: 'linux-x64'
 
       - name: Install protoc
         uses: arduino/setup-protoc@v3
@@ -57,10 +45,6 @@ jobs:
           shared-key: linux-x64
           env-vars: ''
           save-if: false
-
-      - name: Download dependencies
-        run: |
-          npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose
 
       # eslint-import-resolver-typescript requires packages to be built
       - name: Compile all non-rust code
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -39,29 +39,11 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Install Node
-        uses: actions/setup-node@v4
+      - name: Setup Node and Install Dependencies
+        uses: ./.github/actions/setup-node-deps
         with:
-          node-version: 22
-
-      - name: Get NPM cache directory
-        id: npm-cache-dir
-        shell: bash
-        run: echo "dir=$(npm config get cache)" >> ${GITHUB_OUTPUT}
-
-      - name: Restore NPM cache
-        uses: actions/cache/restore@v4
-        with:
-          path: ${{ steps.npm-cache-dir.outputs.dir }}
-          key: npm-main-linux-x64-${{ hashFiles('./package-lock.json') }}
-          restore-keys: |
-            npm-main-linux-x64-
-
-      - name: Download dependencies
-        # Make up to 3 attempts to install NPM dependencies, to work around transient NPM errors
-        # Don't build during install phase since we're going to explicitly build anyway
-        run: |
-          npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose || npm ci --ignore-scripts --verbose
+          node-version: '22'
+          platform: 'linux-x64'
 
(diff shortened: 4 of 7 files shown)


packages/interceptors-opentelemetry/src/worker/index.ts
```diff
@@ -34,7 +34,7 @@ export class OpenTelemetryActivityInboundInterceptor implements ActivityInboundC
   }
 
   async execute(input: ActivityExecuteInput, next: Next<ActivityInboundCallsInterceptor, 'execute'>): Promise<unknown> {
-    const context = await extractContextFromHeaders(input.headers);
+    const context = await Promise.resolve(extractContextFromHeaders(input.headers));
```
chris-olszewski: I kept the await here since we can't use the sdk flags to gate it. I don't think this promise can affect replaying, but I left it since I don't believe we can use SDK flags in this context. Happy to remove promise if you think that would be a better solution.

packages/interceptors-opentelemetry/src/worker/index.ts
```diff
@@ -34,7 +34,7 @@ export class OpenTelemetryActivityInboundInterceptor implements ActivityInboundC
   }
 
   async execute(input: ActivityExecuteInput, next: Next<ActivityInboundCallsInterceptor, 'execute'>): Promise<unknown> {
-    const context = await extractContextFromHeaders(input.headers);
+    const context = extractContextFromHeaders(input.headers);
```
chris-olszewski, line 37: Unconditionally removed this await since we cannot check sdk flags in this context. I was not able to trigger any NDE with this `await` being present/not present. I believe this removal is safe, but can add a `Promise.resolve` to keep the yield point if we think that would be safer.

packages/workflow/src/flags.ts
```diff
@@ -44,6 +44,44 @@ export const SdkFlags = {
...
   ProcessWorkflowActivationJobsAsSingleBatch: defineFlag(2, true, [buildIdSdkVersionMatches(/1\.11\.[01]/)]),
+  /**
+   * In 1.11.3 and previous versions, the interceptor for `handleSignal` provided
+   * by @temporalio/interceptors-opentelemetry did not have a yield point in it.
+   * A yield point was accidentally added in later versions. This added yield point
+   * can cause NDE if there was a signal handler and the workflow was started with a signal.
+   *
+   * This yield point was removed in 1.13.2, but in order to prevent workflows from the
+   * affected versions resulting in NDE, we have to inject the yield point on replay.
+   * This flag should be enabled for SDK versions newer than 1.11.3 or older than 1.13.2.
```
THardy98: `1.11.3 or older than 1.13.2` -> `1.11.3 and older than 1.13.2`
no?

packages/workflow/src/flags.ts
```diff
@@ -44,6 +44,44 @@ export const SdkFlags = {
...
+  /**
+   * The interceptors provided by @temporalio/interceptors-opentelemetry initially had unnecessary yield points.
+   * If replaying a workflow created from these versions a yield point is injected to prevent any NDE.
+   *
+   * If the history does not include the SDK version, default to enabled since the yields were present since the OTEL
+   * package was created.
+   *
+   * @since Introduced in 1.13.2
+   */
+  OpenTelemetryInterceptorInsertYield: defineFlag(3, false, [isBefore({ major: 1, minor: 13, patch: 2 }, true)]),
```
THardy98: Shares the same flag ID as the previous flag (you have 3, 3, 4 heh)

Wonder why this wasn't caught, in principle, assertions relying on `OpenTelemetryHandleSignalInterceptorInsertYield` (previous flag) to be true because of a version _greater_ than 1.11.3 (but before 1.13.2) should be false. Maybe it's the `true` default on this flag?

packages/test/src/workflows/signal-workflow.ts
```diff
@@ -0,0 +1,21 @@
+import { defineSignal, startChild, setHandler, sleep, condition } from '@temporalio/workflow';
+
+const approveTopSecret = defineSignal('approve');
+
+// A workflow that simply calls an activity
+export async function topSecretGreeting(name: string): Promise<string> {
+  const handle = await startChild(topSecretGreetingChild, {
+    args: [name],
+  });
+  await Promise.all([handle.signal(approveTopSecret), sleep('1ms')]);
```
THardy98, line 10: Why do we need `sleep` ?
Is it for task ordering (we want to assert/expect signal to come prior to the sleep timer) ?

packages/interceptors-opentelemetry/src/workflow/workflow-module-loader.ts
```diff
@@ -3,6 +3,9 @@
  * @module
  */
 import type * as WorkflowModule from '@temporalio/workflow';
+import type { SdkFlags as SdkFlagsT } from '@temporalio/workflow/lib/flags';
+
+type SdkFlags = typeof SdkFlagsT;
```
THardy98, line 8: Can't we just `import type { SdkFlags } from '@temporalio/workflow/lib/flags';` above?

packages/interceptors-opentelemetry/src/workflow/workflow-module-loader.ts
```diff
@@ -30,6 +33,21 @@ export function getWorkflowModule(): typeof WorkflowModule {
...
+ * Returns if an SDK flag was set
+ *
+ * Expects to be called only after `ensureWorkflowModuleLoaded`.
+ * Will throw if `@temporalio/workflow` is not available
+ */
+export function hasSdkFlag(flag: keyof SdkFlags): boolean {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { SdkFlags } = require('@temporalio/workflow/lib/flags');
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { getActivator } = require('@temporalio/workflow/lib/global-attributes');
```
THardy98: Why do we need to import in the function?
Generally - maybe this indicates the function belongs in `@temporalio/workflow/lib` and we just need a single internal import to use this (instead of 3). This file could just re-export it, or import directly in the `index.ts` usage, not sure.

packages/interceptors-opentelemetry/src/workflow/index.ts
```diff
@@ -52,6 +52,7 @@ function getTracer(): otel.Tracer {
  */
 export class OpenTelemetryInboundInterceptor implements WorkflowInboundCallsInterceptor {
   protected readonly tracer = getTracer();
+  protected readonly maybeInjectYield = true;
```
THardy98: Don't think we use this

packages/interceptors-opentelemetry/src/worker/index.ts
THardy98: I don't really know the implications of this change, why do you think this is safe?
My immediate intuition is that it'll causing different microtask queuing.

packages/workflow/src/flags.ts
chris-olszewski: 🤦 

The ids are only used for checking if the flag is one of the known flags, but a flag is only marked as a known flag if the default for the flag is `true`. Since these are `false`, the alternative condition will always get evaluated.

packages/test/src/workflows/signal-workflow.ts
chris-olszewski: Yes, just using the `sleep` to create a `TimerCreated` history event. You can see in the history that with the old OTEL interceptors the timer ended up getting created before the signal was sent. [link](https://github.com/temporalio/sdk-typescript/pull/1803/files#diff-557c5fded7eb629e0c682ae7a7fc62bf9338a828be0e8f700189b0a1d0122c16R214)

packages/interceptors-opentelemetry/src/workflow/workflow-module-loader.ts
chris-olszewski: This is an outcome of the recent PR to make `@temporalio/workflow` an optional peer dep for the interceptors: https://github.com/temporalio/sdk-typescript/pull/1812, so we have to do some song and dance around importing from that package now that it might not be present.

From my understanding talking to @mjameswh we don't want to advertise these APIs for general consumption. With that in mind, I think keeping these in the nested modules is okay.

packages/interceptors-opentelemetry/src/workflow/workflow-module-loader.ts
chris-olszewski: For some reason TS does resolution in a weird way so if we do that and then change the signature to `keyof typeof SdkFlags`, TS will complain of exporting the private type `SdkFlags` even though we're only requiring the keys of the type. Doing the little type alias get TS to correctly fully resolve the `keyof` operator to the type.

packages/interceptors-opentelemetry/src/worker/index.ts
chris-olszewski: It will cause different microtask queueing, but I suppose the question is if that change will be impactful. I can add a synthesized yield point back here to be extra cautious

packages/test/src/workflows/signal-start-otel.ts
```diff
@@ -0,0 +1,29 @@
...
+  OpenTelemetryInboundInterceptor,
+  OpenTelemetryOutboundInterceptor,
+  OpenTelemetryInternalsInterceptor,
+} from '@temporalio/interceptors-opentelemetry/lib/workflow';
+
+export const startSignal = workflow.defineSignal('startSignal');
+
+const { a, b, c } = workflow.proxyLocalActivities({
+  scheduleToCloseTimeout: '1m',
+});
```
THardy98, line 12: where are these activities coming from? I don't see any activities registered on the test workers, am i missing something?

packages/interceptors-opentelemetry/src/workflow/workflow-module-loader.ts
THardy98: Ah ok. But I feel as though if we're importing `import type { SdkFlags as SdkFlagsT } from '@temporalio/workflow/lib/flags';` at the top anyway, it's a bit odd.

packages/interceptors-opentelemetry/src/worker/index.ts
THardy98: Yeah I'm not sure - my feeling is that less disruption would be better, but @mjameswh would be better to ask

packages/test/src/workflows/signal-workflow.ts
THardy98: Nice!

packages/test/src/workflows/signal-start-otel.ts
chris-olszewski: I did remove the code to create the histories since the tests only replay history to check for verify there isn't NDE. You can view the code used at https://github.com/temporalio/sdk-typescript/pull/1803/commits/abf1134f3ba0b2ba1828e42395af90c83e6cadb5

packages/interceptors-opentelemetry/src/workflow/workflow-module-loader.ts
chris-olszewski: Is it the type aliasing that is odd or the fact that we have a top level import and an import in a function?

packages/interceptors-opentelemetry/src/worker/index.ts
mjameswh: Determinism and microtask ordering concerns only apply to the Workflow interceptor. This is the activity interceptor, so removal is safe.

+1 to removing the `await` keyword here given that it is useless anyway.

packages/interceptors-opentelemetry/src/workflow/workflow-module-loader.ts
mjameswh: Yeah, that's kind of ugly, but that can't easily be avoided due to the optional dependency.

packages/workflow/src/flags.ts
```diff
@@ -68,9 +106,77 @@ export function assertValidFlag(id: number): void {
...
+function isAfter(compare: SemVer, missingDefault?: boolean): AltConditionFn {
+  return isCompared(compare, -1, missingDefault);
+}
+
+/**
+ * Creates an `AltConditionFn` that checks if the SDK version is between the provided versions.
+ * The range check is exclusive.
+ * An optional default can be provided in case the SDK version is not available.
+ */
+function isBetween(lowEnd: SemVer, highEnd: SemVer, missingDefault?: boolean): AltConditionFn {
```
mjameswh: Let's make `lowEnd` be inclusive, and `highEnd` to be exclusive. That's more common and arguably more practical.

mjameswh (approved):

PR #1805, opened 2025-10-22
[BREAKING CHANGE] Change RawValue API - create RawValue directly from Payload

THardy98:
## What was changed
Changes the `RawValue` API. Formerly, you could create a `RawValue` from any given value via the constructor. However, it was not possible to create a `RawValue` directly with a payload.

This PR changes the API to instead:
- allow `RawValue` to be created directly from a payload, via its constructor (changing the constructor signature is the breaking change)
- all `RawValue` to still be created with any given value, via a static method `fromValue`

## Why?
Imo, this API makes more sense. If we wanted to keep the old API, but expand it to include a static for `fromPayload` it would probably look something like this:
```
  static fromPayload<T = unknown>(payload: Payload): RawValue<T> {
      const instance = Object.create(RawValue.prototype) as RawValue<T>;
      (instance as any)._payload = payload;
      return instance;
  }
```
which isn't very nice, despite being a basic case.

alternatively, we could expand the type of the value in the constructor to `T | Payload` and add some type guard checks. Also not very nice, imo.

2. How was this tested:
Updated existing tests.

3. Any docs updates needed?
Not sure

### DIFF
--- a/packages/common/src/converter/payload-converter.ts
+++ b/packages/common/src/converter/payload-converter.ts
@@ -122,8 +122,12 @@ export class RawValue<T = unknown> {
   private readonly _payload: Payload;
   private readonly [rawPayloadTypeBrand]: T = undefined as T;
 
-  constructor(value: T, payloadConverter: PayloadConverter = defaultPayloadConverter) {
-    this._payload = payloadConverter.toPayload(value);
+  private constructor(payload: Payload) {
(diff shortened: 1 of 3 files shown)


THardy98:
Not really expecting this to be merged or anything (at least, not without discussion). Mostly want to put this up for awareness.

Issue #1787, opened 2025-09-19
[Feature Request] Allow querying the port/address via TestWorkflowEnvironment


hanikesn:
### Is your feature request related to a problem? Please describe.

We're currently spawning a global TestWorkflowEnvironment for our integration tests. As we can't pass the connection/nativeConnection directly onwards we have to pass the address to the application itself. Unfortunately it's not possible to query the actual address string including the port, when spawning the TestWorkflowEnvironment via [createLocal](https://typescript.temporal.io/api/classes/testing.TestWorkflowEnvironment#createlocal) when using a [random free port](https://typescript.temporal.io/api/interfaces/testing.DevServerConfig#port).

### Describe the solution you'd like

Allow querying the port/address so it's possible to use a random free port when creating clients manually.

### Workaround
```typescript
const env = await TestWorkflowEnvironment.createLocal()
const address = native.ephemeralServerGetTarget(env["server"])
```

PR #1807, opened 2025-10-23
chore: update core

chris-olszewski:
## What was changed
TSIA

Other small changes: 
- renamed task queues for plugin tests so we don't hit the new namespace+client+task_queue uniqueness requirement
- disabled heartbeat by default. Will enable things in a separate PR, once configuring this is plumbed through TS.

## Why?
With the code moving, updating core is more involved than usual. Doing this in a separate PR to make it easier to review.

## Checklist

1. Closes N/A

2. How was this tested:
Existing tests

3. Any docs updates needed?
I do not believe so


packages/worker/src/worker-options.ts
```diff
@@ -968,6 +974,7 @@ function addDefaultWorkerOptions(
     stickyQueueScheduleToStartTimeout: '10s',
     maxHeartbeatThrottleInterval: '60s',
     defaultHeartbeatThrottleInterval: '30s',
+    skipClientWorkerSetCheck: false,
```
mjameswh: Should we be worried about defaulting this to `false`? I see we had to explicitly set this to `true` in many of our tests. Is there any risk that this might break users deployments or CIs?

packages/worker/src/worker-options.ts
```diff
@@ -464,6 +464,11 @@ export interface WorkerOptions {
    */
   defaultHeartbeatThrottleInterval?: Duration;
 
+  /**
+   * TODO: write
```
mjameswh: Just a reminder 😉

chris-olszewski: 🙈

packages/worker/src/worker-options.ts
yuandrew: CIs may break if their tests are running in parallel and using the same default namespace + task queue + same client, Would that be considered a breaking change? This path is now an error, whereas previously it just ignored duplicate registrations. Tests don't care about this req, but now they'll hit errors. Not sure if TS runs their tests in parallel, but rust hit this error in tests

chris-olszewski: I took this from the PR that bumped this in Python: https://github.com/temporalio/sdk-python/pull/1157/files#diff-71bb79f8f6f7277331d9df6e0c1de02a9b08a91977289ddf876398e8ad5a8d53R151

@yuandrew is this a concern?

packages/testing/src/testing-workflow-environment.ts
```diff
@@ -90,7 +90,8 @@ export class TestWorkflowEnvironment {
     protected readonly server: native.EphemeralServer | 'existing',
     connection: Connection,
     nativeConnection: NativeConnection,
-    namespace: string | undefined
+    namespace: string | undefined,
+    public readonly address: string
```
mjameswh, line 107: Can you confirm this addition resolves #1787?

chris-olszewski: Yes, added a test to verify that when used with `.createLocal` the field contains the correct address.

I initially missed that this was the public facing test environment class and not the wrapper we use for our own tests. I needed some way to create an additional connection and making this public seemed sensible, but open to other options.

PR #1808, opened 2025-10-24
deflake activity pause/reset tests

THardy98:
## What was changed
DWISOTT

## Why?
These tests having been failing most CI runs.

2. How was this tested:
Ran the same tests locally.

3. Any docs updates needed?
No


THardy98:
> LGTM
> 
> Just confirming that the deflake strategy is moving the logic into the activity instead of altering the activity from the outside?

Exactly - removes a lot of checking/polling of the activity's state from the outside, which can fail/are flaky

PR #1809, opened 2025-10-27
chore(prettier): ignore bridge macros target dir

chris-olszewski:
## What was changed
TSIA

## Why?
Before
```
Benchmark 1: npm exec prettier -- --end-of-line auto --check .
  Time (mean ± σ):     11.859 s ±  0.458 s    [User: 14.062 s, System: 2.327 s]
  Range (min … max):   11.107 s … 12.511 s    10 runs
```
After
```
Benchmark 1: npm exec prettier -- --end-of-line auto --check .
  Time (mean ± σ):      2.839 s ±  0.028 s    [User: 5.754 s, System: 0.668 s]
  Range (min … max):    2.806 s …  2.898 s    10 runs
```

## Checklist

1. Closes N/A

2. How was this tested:
N/A

3. Any docs updates needed?
N/A

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,7 @@
 node_modules
 build
 packages/core-bridge/target
+packages/core-bridge/bridge-macros/target
 sdk-core
 lib
 es2020


.prettierignore
```diff
@@ -1,6 +1,7 @@
 node_modules
 build
 packages/core-bridge/target
+packages/core-bridge/bridge-macros/target
```
mjameswh, line 4: Nice catch!

mjameswh (approved):

PR #1818, opened 2025-10-29
Enable Worker heartbeating, plumb plugin names to core, update Core to 44a6576

### LINKED ISSUE #1810: [Feature Request] Enable Worker Heartbeating

Worker heartbeating is implemented in Core, need to enable worker heartbeating by allowing users to specify the worker heartbeat interval when initializing Core Runtime

yuandrew:
## What was changed

- Plumb the ability to enable/disable worker heartbeating to TS lang layer
- plumb plugin names to Core, to show in worker heartbeats
- specify `WorkerTaskTypes` based on if workflows/activities/nexus handlers are registered to the worker
- bumped `sdk-core` to `44a6576`

## Why?

New feature!

## Checklist

1. Closes #1810

2. How was this tested:

Manually tested that heartbeats with a simple plugin from the worker and client both show up in plugins section of the worker heartbeat.

Also tested setting `workerHeartbeatInterval: null` disables heartbeating for the worker

3. Any docs updates needed?

---

> [!NOTE]
> Adds runtime-configurable worker heartbeats, forwards plugin names in worker heartbeats, introduces task-type gating, and migrates client TLS/options to new Core APIs.
> 
> - **Runtime/Core-bridge**:
>   - Add `worker_heartbeat_interval_millis` runtime option; plumb through to Core `heartbeat_interval`.
> - **Worker**:
>   - Introduce `WorkerTaskTypes` and wire to Core (`task_types`), replacing `enableNonLocalActivities` for capability gating.
>   - Forward `plugins` to Core as `PluginInfo` set for inclusion in heartbeats.
>   - Map “not enabled” completion errors for workflows/activities/nexus to `UnexpectedError`.
> - **Client**:
>   - Migrate TLS config/types (`TlsOptions`, `client_tls_options`) and switch to new Core builders (`maybe_*` setters).
> - **TypeScript bindings**:
>   - Update `native.ts` interfaces to new TLS, runtime, and worker options.
> - **Tests**:
>   - Adjust TLS error expectations and add test ensuring local activities work with remote activities disabled.
>   - Include runtime option `workerHeartbeatIntervalMillis` in bridge tests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fabb505258e145630d1b7ab0ca4a5884acff0fb7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/test/src/test-native-connection.ts
+++ b/packages/test/src/test-native-connection.ts
@@ -77,7 +77,7 @@ test('NativeConnection.connect() throws meaningful error when passed invalid add
 test('NativeConnection.connect() throws meaningful error when passed invalid clientCertPair', async (t) => {
   await t.throwsAsync(NativeConnection.connect({ tls: { clientCertPair: {} as any } }), {
     instanceOf: TypeError,
-    message: /tls\.clientTlsConfig\.clientCert: Missing property 'clientCert'/,
+    message: /tls\.clientTlsOptions\.clientCert: Missing property 'clientCert'/,
   });
 });
 


packages/core-bridge/src/worker.rs
```diff
@@ -166,6 +166,9 @@ pub fn worker_complete_workflow_activation(
                         ),
                     }
                 }
+                CompleteWfError::WorkflowNotEnabled => {
+                    BridgeError::UnexpectedError(format!("{err}"))
```
chris-olszewski: No strong opinions on this 🤷 
```suggestion
                    BridgeError::UnexpectedError(err.to_string())
```

packages/core-bridge/src/worker.rs
```diff
@@ -225,6 +228,9 @@ pub fn worker_complete_activity_task(
                     field: None,
                     message: format!("Malformed Activity Completion: {reason:?}"),
                 },
+                CompleteActivityError::ActivityNotEnabled => {
+                    BridgeError::UnexpectedError(format!("{err}"))
+                }
```
chris-olszewski, line 233: Same as above
```suggestion
                CompleteActivityError::ActivityNotEnabled => {
                    BridgeError::UnexpectedError(err.to_string())
                }
```

packages/worker/src/worker-options.ts
```diff
@@ -1090,14 +1090,20 @@ export function toNativeWorkerOptions(opts: CompiledWorkerOptionsWithBuildId): n
     workflowTaskPollerBehavior: toNativeTaskPollerBehavior(opts.workflowTaskPollerBehavior),
     activityTaskPollerBehavior: toNativeTaskPollerBehavior(opts.activityTaskPollerBehavior),
     nexusTaskPollerBehavior: toNativeTaskPollerBehavior(opts.nexusTaskPollerBehavior),
-    enableNonLocalActivities: opts.enableNonLocalActivities,
+    taskTypes: {
+      enableWorkflows: opts.workflowBundle !== undefined || opts.workflowsPath !== undefined,
+      enableLocalActivities: opts.maxCachedWorkflows > 0 && opts.activities.size > 0,
```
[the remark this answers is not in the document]
chris-olszewski: Looks like the bot is correct:
```
Error {
@temporalio/test:     message: 'Failed to convert WorkerOptions to CoreWorkerConfig: `task_types` cannot enable local activities without workflows',
@temporalio/test:   }
```
https://github.com/temporalio/sdk-typescript/actions/runs/19722768407/job/56508360895?pr=1818#step:11:13443

packages/core-bridge/src/worker.rs
```diff
@@ -566,14 +593,23 @@ mod config {
...
                 .default_heartbeat_throttle_interval(self.default_heartbeat_throttle_interval)
                 .max_task_queue_activities_per_second(self.max_task_queue_activities_per_second)
                 .max_worker_activities_per_second(self.max_activities_per_second)
                 .graceful_shutdown_period(self.shutdown_grace_time)
+                .plugins(
+                    self.plugins
+                        .into_iter()
+                        .map(|name| PluginInfo {
+                            name,
+                            version: String::new(),
```
chris-olszewski, line 610: Just for my own knowledge, is `version` from the past or not yet implemented?

yuandrew: Good question, I don't know. haha. The API was originally implemented with version, but not sure what the intent was. Will follow up on this.

packages/worker/src/worker-options.ts
```diff
@@ -1090,14 +1092,20 @@ export function toNativeWorkerOptions(opts: CompiledWorkerOptionsWithBuildId): n
     workflowTaskPollerBehavior: toNativeTaskPollerBehavior(opts.workflowTaskPollerBehavior),
     activityTaskPollerBehavior: toNativeTaskPollerBehavior(opts.activityTaskPollerBehavior),
     nexusTaskPollerBehavior: toNativeTaskPollerBehavior(opts.nexusTaskPollerBehavior),
-    enableNonLocalActivities: opts.enableNonLocalActivities,
+    taskTypes: {
+      enableWorkflows,
+      enableLocalActivities,
+      enableRemoteActivities: opts.enableNonLocalActivities && opts.activities.size > 0,
```
mjameswh, line 1098: That's basically saying:

```
enableRemoteActivities: (enableWorkflows && opts.activities.size > 0) && opts.activities.size > 0
```

Which is incorrect. We don't need workflows to enable remote activities. I'm surprised tests passed with this.

packages/worker/src/runtime-options.ts
```diff
@@ -32,6 +32,14 @@ export interface RuntimeOptions {
    */
   telemetryOptions?: TelemetryOptions;
 
+  /**
+   * Interval for worker heartbeats. `null` disables heartbeating.
```
mjameswh: I'm not totally comfortable using `null` as a user-facing marker to disable heartbeating. We do that internally at the bridge layer, to work around limitations of Node/NAPI/Neon, but I don't think we should do that in public APIs.

Instead, I suggest making this `Duration | boolean`, so `false` disables heartbeating, `true` or `== null` (which includes `undefined`) defaults to 60s, and anything else is parsed as a Duration.

yuandrew: ahh ok, let's do 0 to disable heartbeating. This is what I was planning on doing in dotnet

packages/worker/src/worker-options.ts
yuandrew: I don't think that's right,  you're referring to `enableLocalActivities`, but I'm using `opts.enableNonLocalActivities` here

packages/core-bridge/src/worker.rs
```diff
@@ -499,14 +506,15 @@ mod config {
...
-        enable_non_local_activities: bool,
+        task_types: WorkerTaskTypes,
         sticky_queue_schedule_to_start_timeout: Duration,
         max_cached_workflows: usize,
         max_heartbeat_throttle_interval: Duration,
         default_heartbeat_throttle_interval: Duration,
         max_activities_per_second: Option<f64>,
         max_task_queue_activities_per_second: Option<f64>,
         shutdown_grace_time: Option<Duration>,
+        plugins: Vec<String>,
```
cretz, line 517: I think we should consider deduping plugins if we don't already. Depending on plugin implementation, it may be very normal to configure multiple of the same plugin on the same worker (maybe each has different settings).

This should be done in every SDK IMO too if we don't already (could consider making that Vec a Set on the Core side).

yuandrew: good point, a set would ensure deduping. Created https://github.com/temporalio/sdk-core/pull/1072

packages/worker/src/runtime-options.ts
```diff
@@ -359,18 +367,20 @@ export interface PrometheusMetricsExporter {
...
-  telemetryOptions: native.RuntimeOptions;
+  runtimeOptions: native.RuntimeOptions;
   logger: Logger;
 }
 
 export function compileOptions(options: RuntimeOptions): CompiledRuntimeOptions {
   const { metrics, noTemporalPrefixForMetrics } = options.telemetryOptions ?? {}; // eslint-disable-line deprecation/deprecation
   const [logger, logExporter] = compileLoggerOptions(options);
 
+  const heartbeatMillis = msToNumber(options.workerHeartbeatInterval ?? '60s');
```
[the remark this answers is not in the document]
yuandrew: enforced in core

PR #1820, opened 2025-11-03
chore: make clippy lints blocking

chris-olszewski:
## What was changed
Made clippy lints errors instead of warnings so they will be blocking in CI

## Why?
Warnings are easy to miss/ignore. Raising them to errors ensures that they are always seen.

## Checklist

1. Closes N/A

2. How was this tested:
`npm run lint.check` will now have a non-zero exit code if there are clippy lints. Fixed/silenced all existing warnings

3. Any docs updates needed?
N/A

### DIFF
--- a/packages/core-bridge/src/worker.rs
+++ b/packages/core-bridge/src/worker.rs
@@ -940,18 +940,18 @@ mod custom_slot_supplier {
         fn from(info: &'a CoreSlotInfo<'a>) -> Self {
             match info {
                 CoreSlotInfo::Workflow(info) => Self::Workflow {
-                    workflow_type: info.workflow_type.to_string(),
+                    workflow_type: info.workflow_type.clone(),
                     is_sticky: info.is_sticky,
                 },
                 CoreSlotInfo::Activity(info) => Self::Activity {
-                    activity_type: info.activity_type.to_string(),
+                    activity_type: info.activity_type.clone(),
                 },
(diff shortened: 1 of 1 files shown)


packages/core-bridge/bridge-macros/src/derive_js_function.rs
```diff
@@ -18,6 +18,7 @@ pub fn js_function_impl(
     let args = &input.sig.inputs;
     let return_type = &input.sig.output; // Can we avoid in some cases?
     let fn_block = &input.block;
+    let attrs = &input.attrs;
```
chris-olszewski, line 21: Need to explicitly forward these since we're generating a new function that might contain code that requires a lint suppression.

packages/core-bridge/src/helpers/handles.rs
```diff
@@ -9,6 +9,7 @@ use parking_lot::Mutex;
 
 use super::{BridgeError, BridgeResult, IntoThrow, TryFromJs, TryIntoJs};
 
+#[allow(clippy::doc_overindented_list_items)]
```
chris-olszewski, line 12: Checked the formatting of `cargo doc` and everything looked fine so silenced the lint as it looks good in source.

mjameswh: Good to know, but we don't generate docs for the bridge anyway. None of it is public API or even pertinent to users.

mjameswh (approved):
Thanks for that.

Issue #1678, opened 2025-04-15
[Feature Request] Add missing hooks on OTel interceptors


mjameswh:
### Description

The OTel interceptor currently does not support tracing propagation on inbound update requests. We need to implement it.

We should also review the interceptor for any other missing hooks.

PR #1835, opened 2025-11-21
feat(@temporalio/interceptors-opentelemetry): implement all interceptors

chris-olszewski:
## What was changed
Add OTEL spans for all interceptors provided by the interface.

## Why?
As discovered in #1677 these can end up causing NDE errors if replaying an old history without these interceptors. Adding these all at once makes gating their usage far easier. 

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/1678

2. How was this tested:
- Existing replay tests
- Added a replay test of smorgasbord history from 1.13.2
- Updated span test to check for new spans added
- Added additional test to verify update with start behavior stays intact when used with OTEL interceptors

3. Any docs updates needed?
Doc comments should suffice

### DIFF
--- a/packages/interceptors-opentelemetry/src/instrumentation.ts
+++ b/packages/interceptors-opentelemetry/src/instrumentation.ts
@@ -62,7 +62,7 @@ async function wrapWithSpan<T>(
     span.setStatus({ code: otel.SpanStatusCode.OK });
     return ret;
   } catch (err: any) {
-    handleError(err, span, acceptableErrors);
+    maybeAddErrorToSpan(err, span, acceptableErrors);
     throw err;
   } finally {
     span.end();
@@ -79,14 +79,14 @@ function wrapWithSpanSync<T>(
     span.setStatus({ code: otel.SpanStatusCode.OK });
     return ret;
   } catch (err: any) {
-    handleError(err, span, acceptableErrors);
+    maybeAddErrorToSpan(err, span, acceptableErrors);
     throw err;
   } finally {
     span.end();
   }
 }
 
-function handleError(err: any, span: otel.Span, acceptableErrors?: (err: unknown) => boolean): void {
+function maybeAddErrorToSpan(err: any, span: otel.Span, acceptableErrors?: (err: unknown) => boolean): void {
   const isBenignErr = err instanceof ApplicationFailure && err.category === ApplicationFailureCategory.BENIGN;
   if (acceptableErrors === undefined || !acceptableErrors(err)) {
     const statusCode = isBenignErr ? otel.SpanStatusCode.UNSET : otel.SpanStatusCode.ERROR;


packages/interceptors-opentelemetry/src/client/index.ts
```diff
@@ -12,7 +36,7 @@ export interface InterceptorOptions {
  *
  * Wraps the operation in an opentelemetry Span and passes it to the Workflow via headers.
  */
-export class OpenTelemetryWorkflowClientInterceptor implements WorkflowClientInterceptor {
+export class OpenTelemetryWorkflowClientInterceptor implements Required<WorkflowClientInterceptor> {
```
chris-olszewski: Not sure if desired, but this will make adding interceptors build breaking for the OTEL package if they aren't implemented.

packages/interceptors-opentelemetry/src/client/index.ts
```diff
@@ -42,4 +66,138 @@ export class OpenTelemetryWorkflowClientInterceptor implements WorkflowClientInt
...
+    return await instrument({
+      tracer: this.tracer,
+      spanName: SpanName.WORKFLOW_TERMINATE,
+      fn: async (span) => {
+        span.setAttribute(WORKFLOW_ID_ATTR_KEY, input.workflowExecution.workflowId);
+        if (input.workflowExecution.runId) {
+          span.setAttribute(RUN_ID_ATTR_KEY, input.workflowExecution.runId);
+        }
+        if (input.reason) {
+          span.setAttribute(TERMINATE_REASON_ATTR_KEY, input.reason);
```
chris-olszewski, line 178: Unsure if this is useful information to stuff in the span by default

packages/interceptors-opentelemetry/src/workflow/index.ts
```diff
@@ -149,6 +226,24 @@ export class OpenTelemetryOutboundInterceptor implements WorkflowOutboundCallsIn
...
+  ): Promise<StartNexusOperationOutput> {
+    if (!hasSdkFlag('OpenTelemetryInterceptorsInstrumentsAllMethods')) return next(input);
+
+    return await instrument({
+      tracer: this.tracer,
+      spanName: `${SpanName.NEXUS_OPERATION_START}${SPAN_DELIMITER}${input.service}${SPAN_DELIMITER}${input.operation}`,
+      fn: async (span) => {
+        span.setAttribute(NEXUS_SERVICE_ATTR_KEY, input.service);
+        span.setAttribute(NEXUS_OPERATION_ATTR_KEY, input.operation);
+        span.setAttribute(NEXUS_ENDPOINT_ATTR_KEY, input.endpoint);
```
chris-olszewski, line 228: Not sure if this is too much info to add by default

packages/interceptors-opentelemetry/src/instrumentation.ts
```diff
@@ -14,6 +14,22 @@ import {
 export const TRACE_HEADER = '_tracer-data';
 /** As in workflow run id */
 export const RUN_ID_ATTR_KEY = 'run_id';
+/** As in workflow id */
+export const WORKFLOW_ID_ATTR_KEY = 'temporal_workflow_id';
+/** As in activity id */
+export const ACTIVITY_ID_ATTR_KEY = 'temporal_activity_id';
+/** As in update id */
+export const UPDATE_ID_ATTR_KEY = 'temporal_update_id';
```
chris-olszewski: These differ from Ruby/.Net/Python attributes which are camel cased.
Java is also camel cased without `temporal` prefix

packages/interceptors-opentelemetry/src/instrumentation.ts
```diff
@@ -14,6 +14,22 @@ import {
 export const TRACE_HEADER = '_tracer-data';
 /** As in workflow run id */
 export const RUN_ID_ATTR_KEY = 'run_id';
```
chris-olszewski, line 16: Ruby OTEL uses `temporalRunId`

packages/interceptors-opentelemetry/src/client/index.ts
mjameswh: Makes sense.

packages/interceptors-opentelemetry/src/client/index.ts
mjameswh: Arguably, tracing value is very limited on `terminate`, as that call will not be relayed to a workflow worker, and server is not emitting OTel spans. But still, trace will show the outbound grpc call, which can certainly turn out to be helpful in some cases, and then the termination reason could maybe be pertinent. I'm ok with that.

packages/interceptors-opentelemetry/src/instrumentation.ts
mjameswh: Oh? Have you compared with other SDKs, beside Ruby? We'd ideally want cross-SDKs compatibility of OTel tracing, as a customer could be operating different languages within a single Temporal application.

Not saying we'll prioritize, of course, but at the very least we should settle on what names we want to normalize to across the board, so that we eventually converge to something consistent.

packages/interceptors-opentelemetry/src/instrumentation.ts
```diff
@@ -48,20 +64,41 @@ async function wrapWithSpan<T>(
...
+    return ret;
+  } catch (err: any) {
+    handleError(err, span, acceptableErrors);
+    throw err;
+  } finally {
+    span.end();
+  }
+}
+
+function handleError(err: any, span: otel.Span, acceptableErrors?: (err: unknown) => boolean): void {
```
mjameswh: nit: can we think of another verb than `handle` here? Something that better communicates the fact that the function will be "encoding/attaching error details to the span", rather than "handling the error situation itself".

packages/interceptors-opentelemetry/src/workflow/index.ts
```diff
@@ -101,13 +167,31 @@ export class OpenTelemetryInboundInterceptor implements WorkflowInboundCallsInte
...
  */
-export class OpenTelemetryOutboundInterceptor implements WorkflowOutboundCallsInterceptor {
+export class OpenTelemetryOutboundInterceptor implements Required<WorkflowOutboundCallsInterceptor> {
   protected readonly tracer = getTracer();
 
   public constructor() {
     ensureWorkflowModuleLoaded();
   }
 
+  public async startTimer(
```
mjameswh: Are we tracing Timers in any other SDK?

packages/workflow/src/flags.ts
```diff
@@ -76,6 +76,16 @@ export const SdkFlags = {
...
+
+  /**
+   * In 1.13.3, all remaining interceptor methods were implemented in @temporalio/interceptors-opentelemetry
+   * including `handleUpdate`, `validateUpdate`, `handleQuery`, `startTimer`, and `startNexusOperation`.
+   * These add instrumentation and yield points that were not present in earlier versions.
+   * This flag gates these new interceptor methods to prevent NDE on replay of workflows from earlier versions.
+   *
+   * @since Introduced in 1.13.3
+   */
+  OpenTelemetryInterceptorsInstrumentsAllMethods: defineFlag(6, true, [isAtLeast({ major: 1, minor: 13, patch: 3 })]),
```
mjameswh: In this case, there's no need for the sdk-version-based alternate condition, as the numeric flag will be properly emited anyway. Arguably not a big deal to have the alt condition anyway, but if we always do, then we'll fall into the situation where the "alternative" is actually the "primary"...

packages/interceptors-opentelemetry/src/instrumentation.ts
chris-olszewski: - Java: `runId`
- Python `temporalRunId`
- Go: Doesn't have these
- PHP: N/A
- .NET: `temporalRunId`

packages/interceptors-opentelemetry/src/workflow/index.ts
chris-olszewski: No, I remove this.

packages/test/src/test-otel.ts
```diff
@@ -604,3 +679,43 @@ test('Can replay signal workflow from 1.13.1', async (t) => {
...
+          ],
+        },
+      },
+      hist
+    );
+  });
+});
+
+// Skipped as we only care that it compiles
+test.skip('otel interceptors are complete', async (t) => {
```
chris-olszewski, line 708: I built and checked at the docs and we end up leaking the `Required` type. Switched to a compile time check that these types satisfy the narrower type, but we only expose implementing the wider interface to users.

packages/interceptors-opentelemetry/src/instrumentation.ts
chris-olszewski: I kept `run_id` as I wasn't sure if moving to `temporalRunId` could be considered a breaking change. I could see the case that users have dashboard/queries using `run_id` that would break if we changed this.

packages/interceptors-opentelemetry/src/instrumentation.ts
chris-olszewski: Switched to `temporal` prefixed camel case to match others

packages/interceptors-opentelemetry/src/instrumentation.ts
mjameswh: Fair. We may however consider changing these while transitioning to OTel 2. I would mind being breaking at that point.

packages/test/src/test-otel.ts
mjameswh: Good thinking.

mjameswh (approved):
LGTM

PR #1842, opened 2025-11-27
chore: Convert all scripts to TypeScript

mjameswh:
## What changed

- Convert all internal scripts to TypeScript (instead of pure JS).
- Use `tsx` to run TS scripts.
- Removed a few scripts and doc files that are no longer pertinent.
- Do not build native code in Core Bridge on NPM's `install`. `install` can't be relied on, as it is disabled by some package managers due to security risks for users.

### DIFF
--- a/scripts/package.json
+++ b/scripts/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "scripts",
+  "private": true,
+  "version": "0.0.1",
+  "description": "Internal scripts for TS SDK repository",
+  "scripts": {
+    "build": "tsc --noEmit"
+  },
+  "keywords": [
+    "temporal",
+    "workflow",
+    "core",
+    "worker"
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT"
+}
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,7 @@
   "root": true,
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
-    "project": ["./packages/*/tsconfig.json", "./tsconfig.scripts.json"]
+    "project": ["./packages/*/tsconfig.json"]
   },
   "plugins": ["@typescript-eslint", "eslint-plugin-tsdoc", "deprecation", "import"],
   "extends": [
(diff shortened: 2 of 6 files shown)


mjameswh:
Need to replace the use of npm's `--ignore` argument in new `pnpm` commands. It is unknown and ignored by pnpm, causing Core to be rebuilt in CI during "Run integration tests" jobs; it shouldn't.

https://github.com/temporalio/sdk-typescript/actions/runs/20283391495/job/58251126305?pr=1842#step:7:1

packages/core-bridge/scripts/build.ts
```diff
@@ -83,16 +82,10 @@ function compile(requestedTarget) {
...
-  const { status, error } = spawnSync(cmd, argv, {
-    stdio: 'inherit',
-    shell: process.platform === 'win32',
-  });
-  if (status !== 0 || error) {
-    throw new Error(`Failed to build${target ? ' for ' + target : ''}: status code ${status}`, error);
-  }
+  console.log('Running cargo-cp-artifact', argv);
+
+  cargoCpArtifact(argv, process.env);
```
chris-olszewski, line 88: Checked and the function exported by `cargo-cp-artifact` takes the env object itself and not `{ env }`

packages/core-bridge/common.d.ts
```diff
@@ -0,0 +1,7 @@
+export const targets: string[];
```
chris-olszewski, line 1: We could move the `common.js` and `index.js` to TS, but decided to defer to another PR unless there's opposition

scripts/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "scripts",
+  "private": true,
+  "version": "0.0.1",
+  "description": "Internal scripts for TS SDK repository",
+  "scripts": {
+    "build": "tsc --noEmit"
```
chris-olszewski, line 7: Not really "building", but we need a new script to type check the scripts. For other packages we rely on `test` pulling them in for typechecking, but it doesn't make sense for scripts.

Decided to make a new `scripts` private package so we can keep our existing lerna based build strategy. No dependencies as these scripts's dependencies are hoisted.

packages/core-bridge/common.d.ts
mjameswh: I agree, that would be better in a different PR.

PR #1846, opened 2025-12-01
chore: enable no unchecked index option

chris-olszewski:
## What was changed
Enable `noUncheckedIndexedAccess` for our packages

## Why?
Joys of JS mean we should have some extra caution around indexing. I mostly took an approach of keeping existing behavior, but being more explicit about what happens if we encounter an `undefined`.

This is primarily intended to prevent any future bugs from creeping into the codebase.

## Checklist

1. Closes N/A

2. How was this tested:
`tsc`

3. Any docs updates needed?
N/A

### DIFF
--- a/packages/client/src/schedule-helpers.ts
+++ b/packages/client/src/schedule-helpers.ts
@@ -91,7 +91,7 @@ const [encodeMonth, decodeMonth] = makeCalendarSpecFieldCoders(
     const index = MONTHS.indexOf(month);
     return index >= 0 ? index + 1 : undefined;
   },
-  (month: number) => MONTHS[month - 1] ?? MONTHS[11],
+  (month: number) => MONTHS[month - 1],
   [{ start: 1, end: 12, step: 1 }], // default to *
   [{ start: 1, end: 12, step: 1 }]
 );
@@ -110,7 +110,7 @@ const [encodeDayOfWeek, decodeDayOfWeek] = makeCalendarSpecFieldCoders(
     const index = DAYS_OF_WEEK.indexOf(day);
     return index >= 0 ? index : undefined;
   },
-  (day: number) => DAYS_OF_WEEK[day] ?? DAYS_OF_WEEK[6],
+  (day: number) => DAYS_OF_WEEK[day],
   [{ start: 0, end: 6, step: 1 }], // default to *
   [{ start: 0, end: 6, step: 1 }]
 );
@@ -118,7 +118,7 @@ const [encodeDayOfWeek, decodeDayOfWeek] = makeCalendarSpecFieldCoders(
 function makeCalendarSpecFieldCoders<Unit>(
   fieldName: string,
   encodeValueFn: (x: Unit) => number | undefined,
-  decodeValueFn: (x: number) => Unit,
+  decodeValueFn: (x: number) => Unit | undefined,
   defaultValue: temporal.api.schedule.v1.IRange[],
   matchAllValue: temporal.api.schedule.v1.IRange[]
 ) {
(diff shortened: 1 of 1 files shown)


packages/client/src/schedule-helpers.ts
```diff
@@ -110,7 +110,7 @@ const [encodeDayOfWeek, decodeDayOfWeek] = makeCalendarSpecFieldCoders(
     const index = DAYS_OF_WEEK.indexOf(day);
     return index >= 0 ? index : undefined;
   },
-  (day: number) => DAYS_OF_WEEK[day],
+  (day: number) => DAYS_OF_WEEK[day] ?? DAYS_OF_WEEK[6],
```
chris-olszewski: Clamped this, but could see throwing

packages/client/src/schedule-helpers.ts
```diff
@@ -91,7 +91,7 @@ const [encodeMonth, decodeMonth] = makeCalendarSpecFieldCoders(
     const index = MONTHS.indexOf(month);
     return index >= 0 ? index + 1 : undefined;
   },
-  (month: number) => MONTHS[month - 1],
+  (month: number) => MONTHS[month - 1] ?? MONTHS[11],
```
chris-olszewski: Clamped this, but could see throwing

packages/common/src/encoding.ts
```diff
@@ -25,9 +25,11 @@ export class TextDecoder {
     for (; index < len; ) {
       nextEnd = index <= lenMinus32 ? 32 : (len - index) | 0;
       for (; pos < nextEnd; index = (index + 1) | 0, pos = (pos + 1) | 0) {
+        // @ts-ignore
```
chris-olszewski: Ignored these errors as I didn't want to add any additional checks here

packages/test/tsconfig.json
```diff
@@ -4,22 +4,51 @@
     "outDir": "./lib",
     "rootDir": "./src",
     "resolveJsonModule": true,
-    "experimentalDecorators": true
+    "experimentalDecorators": true,
+    "noUncheckedIndexedAccess": false,
```
chris-olszewski: Unchecked array access is not especially important in our tests

packages/workflow/src/nexus.ts
```diff
@@ -130,7 +130,12 @@ export function createNexusClient<T extends nexus.ServiceDefinition>(options: Ne
       input: nexus.OperationInput<T['operations'][nexus.OperationKey<T['operations']>]>,
       operationOptions?: StartNexusOperationOptions
     ) {
-      const opName = typeof operation === 'string' ? options.service.operations[operation]?.name : operation.name;
+      const opName =
+        typeof operation === 'string'
+          ? // Casting as string to cover up the fact that `opName` might be undefined.
+            // If this happens, then `execute` will produce a `NexusOperationFailure.NOT_FOUND`.
```
chris-olszewski, line 136: We have a test covering this case, but it only compiles with `@ts-expect-error` so I believe it is okay to leave this codepath as is.

packages/client/src/schedule-helpers.ts
mjameswh: That's the decode function, meaning the input value comes from the protobuf message sent by server, so an out of bound value here is not really possible.

But still, if we want to be extra careful, I'd suggest to let this function and the other one return `undefined`, and instead have the _caller_ (at line 154) throw in that case.

packages/common/src/converter/payload-search-attributes.ts
```diff
@@ -69,7 +69,7 @@ export class SearchAttributePayloadConverter implements PayloadConverter {
 
     const value = this.jsonConverter.fromPayload(payload);
     let arrayWrappedValue = Array.isArray(value) ? value : [value];
-    const searchAttributeType = decode(payload.metadata.type);
+    const searchAttributeType = payload.metadata.type ? decode(payload.metadata.type) : '';
```
mjameswh: I don't think we need to default to an empty string in this case. The only thing we do with `searchAttributeType` is to compare it for equality with `'Datetime'`. There's no problem for that purpose for `searchAttributeType` to be `string | undefined`.

chris-olszewski: Sounds good, the default of `''` was just matching the existing behavior of `decode(undefined)`

mjameswh (approved):

PR #1847, opened 2025-12-01
💥 TLS enabled if API key is provided

THardy98:
## What was changed
DWISOTT

1. Part of https://github.com/temporalio/features/issues/687

2. How was this tested:
Unit tests

3. Any docs updates needed?
Maybe?

### DIFF
--- a/packages/common/src/internal-non-workflow/tls-config.ts
+++ b/packages/common/src/internal-non-workflow/tls-config.ts
@@ -34,7 +34,7 @@ export type TLSConfigOption = TLSConfig | boolean | undefined | null;
  */
 export function normalizeTlsConfig(
   tls: TLSConfigOption,
-  apiKey: string | (() => string) | undefined
+  apiKey?: string | (() => string) | undefined
 ): TLSConfig | undefined {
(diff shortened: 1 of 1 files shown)


packages/common/src/internal-non-workflow/tls-config.ts
```diff
@@ -35,3 +35,10 @@ export type TLSConfigOption = TLSConfig | boolean | undefined | null;
 export function normalizeTlsConfig(tls: TLSConfigOption): TLSConfig | undefined {
   return typeof tls === 'object' ? (tls === null ? undefined : tls) : tls ? {} : undefined;
 }
+
+export function defaultTLSFromApiKey(
```
mjameswh: Why introduce a new function for this? Can't we simply pass a second argument to `normalizeTlsConfig`?

mjameswh (commented):
You missed a call to `normalizeTlsConfig()` in cloud-operations-client.ts.

packages/common/src/internal-non-workflow/tls-config.ts
THardy98: done

PR #1855, opened 2025-12-04
Corrected docstring for OtelCollectorExporter

### LINKED ISSUE #1854: [Feature Request] Correct docs for OtelCollectorExporter

The docstring for the Otel `url` is incorrect/confusing. It states that `OTEL_EXPORTER_OTLP_ENDPOINT` will override the value provided. This appears to be incorrect given that there is no code that references the environment variable. Additionally, if such code existed, it would break metrics export since `OTEL_EXPORTER_OTLP_ENDPOINT` for HTTP is typically `http://localhost:4318` but this package expects a URL with the path (e.g., `http://localhost:4318/v1/metrics`).

https://github.com/temporalio/sdk-typescript/blob/6b89dfc95ebb8554e439a0f2b2c26d5783bfa1d7/packages/worker/src/runtime-options.ts#L225-L236

Remove this note to avoid confusion.

### Additional context

N/A

clintonb:
## What was changed

Removed the incorrect reference to the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable.

## Why?

The documentation is wrong.

## Checklist

1. Closes #1854

2. How was this tested:

I manually tested with my own deployment.

3. Any docs updates needed?

This is a docs update.


mjameswh:
Ah, that's possible. There have been many recent changes to the OTel support in Core SDK. That may have got broken at some point. 

Out of curiosity, what led you to figure out this statement is correct? What was your use case for `OTEL_EXPORTER_OTLP_ENDPOINT`?

clintonb:
@mjameswh Every other TS/JS library expects `OTEL_EXPORTER_OTLP_ENDPOINT` to be the root of HTTP Otel endpoint (e.g., `http://localhost:4318`). This is the only library I've come across that expects a path: `http://localhost:4318/v1/metrics`.

All of my deployed workloads have `OTEL_EXPORTER_OTLP_ENDPOINT` defined because the rest of my code relies on it. While testing the use of HTTP metrics for Temporal, I realized that the docstring is wrong. In fact, I use `OTEL_EXPORTER_OTLP_ENDPOINT` to create the metrics URL:
```ts
const otelMetricsEndpoint = process.env.OTEL_EXPORTER_OTLP_ENDPOINT
  ? `${process.env.OTEL_EXPORTER_OTLP_ENDPOINT}/v1/metrics`
  : undefined;
```

chris-olszewski (approved):

PR #1868, opened 2026-01-02
fix: avoid disabling `CancellationScope` storage one workflow disposal

### LINKED ISSUE #1866: [Bug] Signal caused `condition` to fail with `CancelledFailure` on `1.14.0`

We bumped our Temporal dependencies to `1.14.0` today rom `1.13.1` and upon deploying started seeing workflows fail due to the following error:
```ts
CancelledFailure: Workflow cancelled
    at RootCancellationScope.cancel (/tmp/build_d69c691a/node_modules/.pnpm/@temporalio+workflow@1.14.0/node_modules/@temporalio/workflow/src/cancellation-scope.ts:264:16)
    at /tmp/build_d69c691a/node_modules/.pnpm/@temporalio+workflow@1.14.0/node_modules/@temporalio/workflow/src/workflow.ts:1171:36
```

Rolling back to `1.13.1` resolves the issue. We've reset the effected workflows to the signal event and they now run correctly. A race condition seems to have been introduced since we do observe some of our workflows succeeding with a signal and `condition`.

### Describe the bug
We started seeing phantom `CancelledFailure` errors surface in our workflows after updating to `1.14.0` that seem to be coming from `condition`. Nothing else about our code has changed. An example of what this looks like on our end:

Image

### Minimal Reproduction

Unable to reproduce locally so far. We have over 30 workflows in production that have experienced this.

### Environment/Versions

- OS and processor: Linux
- Temporal Version: 1.14.0
- Are you using Docker or Kubernetes or building Temporal from source? From source

chris-olszewski:
## What was changed
No longer call `AsyncLocalStorage.disable` for the stores used to track cancellation and update scopes.

## Why?
* `condition` with timeout creates a `CancellationScope` to perform the race between the sleep and the blocking condition [src](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/workflow.ts#L1167-L1172)
* On completion, the current `CancellationScope` is canceled
* Current `CancellationScope` is stored in a `AsyncLocalStorage` which falls back to the workflow cancellation scope if not in a context where there is an active store. [src](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/cancellation-scope.ts#L230)
* [1834](https://github.com/temporalio/sdk-typescript/pull/1834) added a `workflowModule.dispose()` call to the `dispose` method for reusable VM workflow
* This in turn disables the `AsyncLocalStorage` that stores the cancellation tokens ([src](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/worker-interface.ts#L260)). Any subsequent `getStore` calls that happen before a new `run` will return `undefined`.
* If a workflow ends up getting disposed after the `condition` scope is created, but before the scope is cancelled, then the workflow ends up get cancelled instead of the `condition` scope.
* This lines up from the trace provided by the user on [1866](https://github.com/temporalio/sdk-typescript/issues/1866) which is a `CancelledFailure: Workflow cancelled`, but coming from the cancel inside of the `condition` scope on [this line](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/workflow.ts#L1171)

## Checklist

1. Closes #1866

2. How was this tested:
Added failing integration test in first commit, second commit fixes the test.

3. Any docs updates needed?
N/A

### DIFF
--- a/packages/worker/src/workflow/reusable-vm.ts
+++ b/packages/worker/src/workflow/reusable-vm.ts
@@ -210,7 +210,15 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
    * Cleanup the pre-compiled script
    */
   public async destroy(): Promise<void> {
-    // TODO: disable AsyncLocalStorage used for cancellation/update scopes
+    if (this._context) {
+      vm.runInContext(
+        `{
+          __TEMPORAL__.api.destroy();
+        }`,
+        this._context,
+        { timeout: this.isolateExecutionTimeoutMs, displayErrors: true }
+      );
+    }
     globalHandlers.removeWorkflowBundle(this.workflowBundle);
     delete this._context;
   }
--- a/packages/worker/src/workflow/vm.ts
+++ b/packages/worker/src/workflow/vm.ts
@@ -126,6 +126,7 @@ export class VMWorkflowCreator implements WorkflowCreator {
 export class VMWorkflow extends BaseVMWorkflow {
   public async dispose(): Promise<void> {
     this.workflowModule.dispose();
+    this.workflowModule.destroy();
     VMWorkflowCreator.workflowByRunId.delete(this.runId);
     delete this.context;
   }
--- a/packages/workflow/src/worker-interface.ts
+++ b/packages/workflow/src/worker-interface.ts
@@ -10,6 +10,8 @@ import { WorkflowInterceptorsFactory } from './interceptors';
 import { WorkflowCreateOptionsInternal } from './interfaces';
 import { Activator } from './internals';
 import { setActivatorUntyped, getActivator } from './global-attributes';
+import { disableStorage } from './cancellation-scope';
+import { disableUpdateStorage } from './update-scope';
 
 // Export the type for use on the "worker" side
 export { PromiseStackStore } from './internals';
@@ -259,6 +261,11 @@ export function dispose(): void {
   } finally {
     activator.rethrowSynchronously = false;
   }
+}
+
+export function destroy(): void {
+  disableStorage();
+  disableUpdateStorage();
 }
 
 function isWorkflowFunctionWithOptions(obj: any): obj is WorkflowFunctionWithOptions<any[], any> {


packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -210,6 +210,16 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
    * Cleanup the pre-compiled script
    */
   public async destroy(): Promise<void> {
+    if (this._context) {
+      vm.runInContext(
+        `{
+          __TEMPORAL__.api.disableStorage();
+          __TEMPORAL__.api.disableUpdateStorage();
```
chris-olszewski: I think we still need to have these calls since it appears GCing the stores doesn't happen without the `dispose` call. Removing these calls would only affect users starting/stopping workers multiple times in the same process.

packages/worker/src/workflow/vm.ts
```diff
@@ -126,6 +126,7 @@ export class VMWorkflowCreator implements WorkflowCreator {
 export class VMWorkflow extends BaseVMWorkflow {
   public async dispose(): Promise<void> {
     this.workflowModule.dispose();
+    this.workflowModule.destroy();
```
chris-olszewski, line 129: Adding back the `disableStore`/`disableUpdateStore` calls that were present in `dispose` before this PR.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -210,6 +210,15 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
    * Cleanup the pre-compiled script
    */
   public async destroy(): Promise<void> {
+    if (this._context) {
+      vm.runInContext(
+        `{
+          __TEMPORAL__.api.destroy();
```
chris-olszewski, line 216: Not sure if there's a better way to call this

mjameswh:
Superseded by #1871.

PR #1871, opened 2026-01-07
feat(workflow): Automatically cleanup AsyncLocalStorage on workflow context disposal

### LINKED ISSUE #1432: [Feature Request] Simply proper usage of `AsyncLocalStorage` in Workflow context

### LINKED ISSUE #1866: [Bug] Signal caused `condition` to fail with `CancelledFailure` on `1.14.0`

mjameswh:
## What was changed and why

- Track instances of `AsyncLocalStorage` created inside the workflow sandbox and automatically clean them up on workflow context disposal, _after_ execution of the `dispose` hook on internal interceptors.

  - Proper use of `AsyncLocalStorage` was difficult due to the necessity of cleaning them up from an internal interceptor's `dispose` hook (fixes #1432).

- The `AsyncLocalStorage` instances created by the SDK to track cancellation and update scopes are now automatically cleaned up at the sandbox vm destruction time.

  - [PR 1834](https://github.com/temporalio/sdk-typescript/pull/1834) added a `workflowModule.dispose()` call during cleanup of the reusable VM workflow, but that resulted in a regression in 1.14.0 where a workflow might end up getting unexpectedly cancelled if evicted from cache while blocked on a `condition(fn, timeout)` (see investigation details below). Fixes #1866.

    - `condition` with timeout creates a `CancellationScope` to perform the race between the sleep and the blocking condition [src](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/workflow.ts#L1167-L1172)
    - On completion, the current `CancellationScope` is canceled
    - Current `CancellationScope` is stored in a `AsyncLocalStorage` which falls back to the workflow cancellation scope if not in a context where there is an active store. [src](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/cancellation-scope.ts#L230)
    - [1834](https://github.com/temporalio/sdk-typescript/pull/1834) added a `workflowModule.dispose()` call to the `dispose` method for reusable VM workflow
    - This in turn disables the `AsyncLocalStorage` that stores the cancellation tokens ([src](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/worker-interface.ts#L260)). Any subsequent `getStore` calls that happen before a new `run` will return `undefined`.
    - If a workflow ends up getting disposed after the `condition` scope is created, but before the scope is cancelled, then the workflow ends up get cancelled instead of the `condition` scope.
    - This lines up from the trace provided by the user on [1866](https://github.com/temporalio/sdk-typescript/issues/1866) which is a `CancelledFailure: Workflow cancelled`, but coming from the cancel inside of the `condition` scope on [this line](https://github.com/temporalio/sdk-typescript/blob/981d3f11eb47e491c19d4f7a27b460bef332ba29/packages/workflow/src/workflow.ts#L1171)

- Properly `dispose` a workflow execution context on early failure (e.g. if requested workflow type doesn't resolve to an exported workflow function).

  - We do not add the workflow execution context to the lang-side workflow cache if it failed during early initialization. It is therefore not possible to clean it up at a later time, i.e. on cache eviction. This could previously cause a memory leak if some some non-garbage-collectible resources (e.g. `AsyncLocalStorage`) had been allocated during early initialization. We now do ensure that `dispose` is properly called in that scenario.

## Checklist

1. How was this tested:

- Added failing integration test in first commit, second commit fixes the test.
- Actual disposal of `AsyncLocalStorage` lifecycle can't really be asserted through automated testing. We temporarily added `console.log` calls inside our mocked ALS class to confirm that `dispose()` is called at the correct moment in various scenarios.

2. Any docs updates needed?

- We should consider documenting the use of `AsyncLocalStorage` in workflow context.

### DIFF
--- a/packages/test/src/test-workflow-async-local-storage.ts
+++ b/packages/test/src/test-workflow-async-local-storage.ts
@@ -2,7 +2,6 @@ import type { AsyncLocalStorage } from 'async_hooks';
 import * as workflow from '@temporalio/workflow';
 import { helpers, makeTestFunction } from './helpers-integration';
 import { unblockSignal } from './workflows/testenv-test-workflows';
-import { startWorkflow } from '@temporalio/nexus';
 
 const test = makeTestFunction({
   workflowsPath: __filename,


packages/test/src/test-workflow-async-local-storage.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { createWorker, startWorkflow } = helpers(t);
+
+  const worker = await createWorker();
+
+  const wf1 = await startWorkflow(asyncLocalStorageWorkflow);
+  const wf2 = await startWorkflow(asyncLocalStorageWorkflow);
+  const wf3 = await startWorkflow(asyncLocalStorageWorkflow);
+
+  await worker.runUntil(Promise.all([wf1.result(), wf2.result(), wf3.result()]));
+  t.pass();
```
chris-olszewski, line 60: Are there any additional assertions that would be valuable here aside from "it doesn't throw"?

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -99,25 +102,59 @@ export function injectGlobals(context: vm.Context): void {
...
+      }
+    }
+
+    disable(): void {
+      super.disable();
+    }
+
+    __temporal_disableForReal(): void {
+      super.disable();
+    }
```
chris-olszewski, line 153: If both of these actually disable the ALS, do we need both?

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -18,6 +18,8 @@ import { convertDeploymentVersion } from '../utils';
 import { Workflow } from './interface';
 import { WorkflowBundleWithSourceMapAndFilename } from './workflow-worker-thread/input';
 
+import '@temporalio/workflow/lib/global-attributes'; // eslint-disable-line import/no-unassigned-import
```
chris-olszewski, line 22: Is there a side effect we need from this import?

mjameswh: We need that import because of the ambient global extensions that are declared by that file. I added a comment.

packages/worker/src/workflow/vm-shared.ts
mjameswh: Yeah, I was hesitant about whether we want the "normal" `disable()` function to work properly, or if it should be a noop. I'm now leading for the former, so I'll delete the `__temporal` one.

chris-olszewski: 👍 on having it function properly

packages/test/src/test-workflow-async-local-storage.ts
mjameswh: I played around with the possibility of observing the destruction of ALS by escaping the sandbox to create a `FinalizationRegistry`, which worked. The problem, though, is that I couldn't get that test to fail with the previous code, so that's not proving anything.

With that in mind, and given the complexity and fragility of that test, I decided not to commit it.

And I really can't think of any other side effect to look for.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -99,25 +103,62 @@ export function injectGlobals(context: vm.Context): void {
...
+            writable: false,
+            enumerable: false,
+            configurable: false,
+          });
+        sandboxGlobalThis.__temporal_globalSandboxDestructors!.push(this.disable.bind(this));
+      }
+    }
+
+    disable(): void {
+      console.log('AsyncLocalStorage.disable', this.name);
```
chris-olszewski:
```suggestion
```

chris-olszewski (approved):

PR #1874, opened 2026-01-08
chore: update deps

chris-olszewski:
## What was changed
Bump `@modelcontextprovider/sdk` and `qs`

## Why?
Clean up dependabot alerts

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/security/dependabot/290, https://github.com/temporalio/sdk-typescript/security/dependabot/291, https://github.com/temporalio/sdk-typescript/security/dependabot/293

2. How was this tested:
CI

3. Any docs updates needed?
N/A

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -38,4 +38,4 @@ peerDependencyRules:
 
 overrides:
   # Can be removed once a release of `express` with an updated qs is cut
-  "qs": "^6.14.1"
+  'qs': '^6.14.1'
--- a/packages/docs/pnpm-workspace.yaml
+++ b/packages/docs/pnpm-workspace.yaml
@@ -3,4 +3,4 @@ packages:
 
 overrides:
   # Can be removed once a release of `express` with an updated qs is cut
-  "qs": "^6.14.1"
+  'qs': '^6.14.1'


pnpm-workspace.yaml
```diff
@@ -27,9 +27,15 @@ minimumReleaseAge: 1440
 minimumReleaseAgeExclude:
   - '@temporalio/*'
   - nexus-rpc
+  # TODO: remove once this ages past min release age
+  - '@modelcontextprotocol/sdk@1.25.2'
```
chris-olszewski, line 31: Temporarily excluding this version as it is too young.

package.json
```diff
@@ -86,7 +86,13 @@
   },
   "engines": {
     "node": ">= 18.0.0",
-    "pnpm": ">= 10.16.0",
+    "pnpm": ">= 10.27.0",
```
chris-olszewski, line 89: Need to bump `pnpm` version to give us the ability to spot exclude specific versions of a package. Useful for cases such as `@modelcontextprovider/sdk` where the latest version with the security patch is too young, but we don't want to completely allowlist all future package versions.

package.json
```diff
@@ -86,7 +86,13 @@
...
     "rustc": ">= 1.53.0"
+  },
+  "devEngines": {
+    "packageManager": {
```
chris-olszewski, line 93: This field is apparently now what `corepack` prefers to read from

PR #1884, opened 2026-01-12
feat(opentelemetry): expose functionality as plugin

### LINKED ISSUE #1850: [Feature Request] Implement OTel interceptors v1 as Plugin

OTEL interceptors are awkward to use

Setting up OTEL should be as easy as
```
const plugin = new OpenTelemetryPlugin({resource: staticResource, traceExporter});
    activities,
    taskQueue: 'test-otel',
    plugins: [plugin],
});

const client = new Client({
    plugins: [plugin],
});
```

chris-olszewski:
## What was changed
Expose the OpenTelemetry interceptors as 2 plugins to reduce possible misuse of them.

## Why?
Correctly setting up OTEL interceptors requires correctly setting multiple configuration options in order for it to work correctly.

## Checklist

1. Closes #1850

2. How was this tested:
OTEL sample updated to use plugin. Applicable OTEL tests updated to use plugin instead of setting up interceptors directly. Not all tests applicable as some tests don't require fully configured OTEL setups.

3. Any docs updates needed?
No. We could advertise the plugin variant in the top level docs, but unsure if we want to wait until plugins are stable.

### DIFF
--- a/packages/interceptors-opentelemetry/src/plugin.ts
+++ b/packages/interceptors-opentelemetry/src/plugin.ts
@@ -39,7 +39,6 @@ export class OpenTelemetryPlugin extends SimplePlugin {
       clientInterceptors: {
         workflow: [new OpenTelemetryWorkflowClientInterceptor(interceptorOptions)],
       },
-      workflowInterceptorModules: [workflowInterceptorsPath],
       workerInterceptors: {
         client: {
           workflow: [new OpenTelemetryWorkflowClientInterceptor(interceptorOptions)],
--- a/packages/plugin/src/plugin.ts
+++ b/packages/plugin/src/plugin.ts
@@ -59,14 +59,6 @@ export interface SimplePluginOptions {
   readonly workflowBundle?: PluginParameter<WorkflowBundleOption>;
   /** Worker-side interceptors. When a value is provided, interceptors will be appended  */
   readonly workerInterceptors?: PluginParameter<WorkerInterceptors>;
-  /**
-   * Workflow interceptor modules to include when bundling workflow code.
-   * These modules are automatically added to {@link BundleOptions.workflowInterceptorModules}
-   * when using this plugin with {@link bundleWorkflowCode}.
-   *
-   * When a value is provided, modules will be appended to existing modules.
-   */
-  readonly workflowInterceptorModules?: PluginParameter<string[]>;
   /** Context function to wrap worker execution */
   readonly runContext?: (next: () => Promise<void>) => Promise<void>;
 }
@@ -157,12 +149,13 @@ export class SimplePlugin
    * @returns Modified bundle options with plugin configuration applied
    */
   configureBundler(options: BundleOptions): BundleOptions {
+    const workerInterceptors = resolveWorkerInterceptors(undefined, this.options.workerInterceptors);
     return {
       ...options,
       workflowsPath: resolveRequiredParameter(options.workflowsPath, this.options.workflowsPath),
       workflowInterceptorModules: resolveAppendParameter(
         options.workflowInterceptorModules,
-        this.options.workflowInterceptorModules
+        workerInterceptors?.workflowModules
       ),
     };
   }


packages/plugin/src/plugin.ts
```diff
@@ -59,6 +59,14 @@ export interface SimplePluginOptions {
...
   /** Worker-side interceptors. When a value is provided, interceptors will be appended  */
   readonly workerInterceptors?: PluginParameter<WorkerInterceptors>;
+  /**
+   * Workflow interceptor modules to include when bundling workflow code.
+   * These modules are automatically added to {@link BundleOptions.workflowInterceptorModules}
+   * when using this plugin with {@link bundleWorkflowCode}.
+   *
+   * When a value is provided, modules will be appended to existing modules.
+   */
+  readonly workflowInterceptorModules?: PluginParameter<string[]>;
```
chris-olszewski: Necessary for the OTEL plugin to function properly when used when prebundling

packages/interceptors-opentelemetry/src/plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * A client-side plugin that adds OpenTelemetry tracing to Workflow Client operations.
+ *
+ * Wraps client operations (start, signal, query, etc.) in OpenTelemetry spans and propagates
+ * trace context to Workflows via headers.
+ *
+ * @experimental Plugins is an experimental feature; APIs may change without notice.
+ */
+export class OpenTelemetryClientPlugin extends SimplePlugin {
```
chris-olszewski: I split these into 2 plugins vs making the resource/traceExporter optional to make it easier to use these plugins correctly.

If we wanted to make a stronger type error we could brand the various plugin interfaces and make it clear this plugin isn't valid to be used when creating a worker. At the moment since all plugin methods are optional any object with a `name` property is a valid plugin for any context. I would be open to branding the interfaces, but I think that deserves its own PR.

packages/interceptors-opentelemetry/src/plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+ * A worker-side plugin that adds OpenTelemetry tracing to Worker operations.
+ *
+ * Configures Activity and Workflow interceptors for trace propagation, and injects
+ * a span exporter sink for Workflow spans. Handles both regular Workers and Replay Workers.
+ *
+ * @experimental Plugins is an experimental feature; APIs may change without notice.
+ */
+export class OpenTelemetryWorkerPlugin extends SimplePlugin {
+  constructor(readonly otelOptions: OpenTelemetryWorkerPluginOptions) {
+    const workflowInterceptorsPath = require.resolve('./workflow-interceptors');
```
THardy98, line 35: Remind me, is it always the case that we want to register inbound, outbound, and internal workflow interceptors?

packages/interceptors-opentelemetry/src/plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+import { InjectedSinks, ReplayWorkerOptions, WorkerOptions } from '@temporalio/worker';
+import { InterceptorOptions, OpenTelemetryWorkflowClientInterceptor } from './client';
+import {
+  makeWorkflowExporter,
+  OpenTelemetryActivityInboundInterceptor,
+  OpenTelemetryActivityOutboundInterceptor,
+} from './worker';
+import { OpenTelemetrySinks } from './workflow';
+
+export type OpenTelemetryClientPluginOptions = InterceptorOptions;
```
THardy98: A bit weird that we have this type to extract it back into InterceptorOptions, but I have no real qualms with it.
I like that it's namespaced. Do we expect this to grow beyond InterceptorOptions in the future?

packages/interceptors-opentelemetry/src/plugin.ts
THardy98: > I split these into 2 plugins vs making the resource/traceExporter optional to make it easier to use these plugins correctly.

I agree with this sentiment.

> If we wanted to make a stronger type error we could brand the various plugin interfaces and make it clear this plugin isn't valid to be used when creating a worker. At the moment since all plugin methods are optional any object with a name property is a valid plugin for any context. I would be open to branding the interfaces, but I think that deserves its own PR.

Imo, I don't think this is absolutely necessary atm, given that you've separated them.

The overlying concern is relevant to all plugins. I think it's reasonable to create a separate issue to introduce a helper enforcing stronger plugin typing.

packages/test/src/test-otel.ts
```diff
@@ -268,33 +272,19 @@ if (RUN_INTEGRATION_TESTS) {
...
-          workflowModules: [require.resolve('./workflows/otel-interceptors')],
-          activity: [
-            (ctx) => ({
-              inbound: new OpenTelemetryActivityInboundInterceptor(ctx),
-              outbound: new OpenTelemetryActivityOutboundInterceptor(ctx),
-            }),
-          ],
-        },
-        sinks,
+        plugins: [new OpenTelemetryWorkerPlugin({ resource: staticResource, traceExporter })],
```
THardy98: This is really nice

packages/plugin/src/plugin.ts
THardy98: Looks reasonable to me, not sure if there are greater implications here @mjameswh

packages/interceptors-opentelemetry/src/plugin.ts
chris-olszewski: Good catch, I can definitely see people wanting to omit some of those interceptors. I'll provide the functionality via plugin options.

packages/interceptors-opentelemetry/src/plugin.ts
chris-olszewski: Possibly which is why I went with `export interface OpenTelemetryClientPluginOptions extends InterceptorOptions {}` initially which a lint told me is equivalent to just a type alias. If we end up needing to add additional options here we will do it via `extends InterceptorOptions` and it will be non-breaking.

packages/interceptors-opentelemetry/src/plugin.ts
chris-olszewski: I added the ability to disable any individual interceptor on the Worker plugin. I skipped this ability for the client plugin as the way to do that is just not use the plugin 🤷

packages/interceptors-opentelemetry/src/plugin.ts
```diff
@@ -0,0 +1,180 @@
...
+  options?: OpenTelemetryWorkerPluginOptions | OpenTelemetryClientPluginOptions
+): InterceptorOptions {
+  return options?.tracer ? { tracer: options.tracer } : {};
+}
+
+function isInterceptorEnabled(toggle: boolean | undefined): boolean {
+  return toggle !== false;
+}
+
+type WorkflowInterceptorModule =
```
chris-olszewski: I went down the road of generating this union, but it was not easily understood and felt like overkill since we probably won't be adding any new workflow interceptor types.

packages/interceptors-opentelemetry/src/plugin.ts
```diff
@@ -0,0 +1,180 @@
...
+  | 'inbound-outbound'
+  | 'inbound-internals'
+  | 'outbound-internals'
+  | 'inbound-outbound-internals';
+
+/**
+ * Returns the workflow interceptor module name based on the enabled toggles.
+ * Returns `null` if all workflow interceptors are disabled.
+ */
+function getWorkflowInterceptorModule(toggles: OpenTelemetryInterceptorOptions): WorkflowInterceptorModule | null {
```
mjameswh: Do you have practical use cases in mind that would justify one to only use only some of these, but not all of them?

chris-olszewski: I don't think there is a clear use case enabling subsets of the interceptors, but wanted to surface the same configuration options in the plugin that are available if setting up interceptors manually.

Happy to rip this out and have always enable all interceptors. We can go down this path if users ask for it.

mjameswh: Yeah, I'd prefer we don't. Adding that later remains a possibility. Removing it later once we've added it is more tricky.

packages/plugin/src/plugin.ts
mjameswh: Oh, yeah. I thought we had covered that case when in the initial Plugin PR, but looks like that got lost at some point.

However, I think we should avoid duplicating "workflow interceptor modules", i.e. once here and once in `WorkerInterceptors.workflowModules`. There's absolutely no use case that would justify registering wf interceptors only in `bundleWorkflowCode` or only in `Worker.create()`, and it is almost certain that some plugins developers will eventually fail to configure that properties in both places.

mjameswh: I suggest we don't add this property here; instead, let's make `configureBundler` add interceptors from `workerInterceptors.workflowModules`.

Or, we could do the opposite (only have it `SimplePluginOptions.workflowInterceptorModules`, and use that from both `configureBundler` and `configureWorker`).

No strong opinion on either of these two. I feel the former is a bit more coherent with the rest of SimplePlugin's API. The latter is more explicit, but breaks the option's structure (I wouldn't mind that being a breaking change, as Plugin API is still experimental).

In both cases, some typedocs should be added here and there to clarify the behavior.

mjameswh: Just realizing now that we have a similar issue with `clientInterceptors` vs `workerInterceptors.client`. With that in mind, it would probably make more sense to instead group all interceptors under a single `interceptors` property (i.e. not `workerInterceptor`), like `SimplePluginOptions.interceptors: { client, workflowModules, activity, nexus }`.

mjameswh: But I don't want to hold this PR on that decision.

So, let's do this: for this PR, let's not introduce this extra property here; just make `configureBundler` use interceptors from `workerInterceptors.workflowModules`. That will make this PR work fine immediately, and we can then refine the Plugin API independently, without impact on users of the OTel plugin.

packages/test/src/test-otel.ts
```diff
@@ -664,6 +648,82 @@ if (RUN_INTEGRATION_TESTS) {
     });
   }
 
+  test.serial('OpenTelemetryWorkerPlugin works with prebundled workflow code', async (t) => {
```
mjameswh:
```suggestion
  test.serial('OpenTelemetryPlugin works with prebundled workflow code', async (t) => {
```

packages/interceptors-opentelemetry/src/plugin.ts
mjameswh: Discussed this offline, and we ended up reverting to a single plugin.

mjameswh (approved):
I left two minor comments, good to merge after that.

PR #1886, opened 2026-01-14
fix(opentelemetry): `makeWorkflowExporter` accepts SpanProcessor 

### LINKED ISSUE #1779: [Bug] `@temporalio/interceptors-opentelemetry` OTEL bypasses a required SDK component
When using the `@temporalio/interceptors-opentelemetry` library I am seeing TONS of these errors:

```
ERROR --- node_modules/@opentelemetry/api/build/src/api/diag.js : ["Accessing resource attributes before async attributes settled"]
```

which are caused by the Temporal SDKs improper use of the OTEL SDK.

### Describe the bug
See this comment on the issue in the OTEL SDK: https://github.com/open-telemetry/opentelemetry-js/issues/5828#issuecomment-3227403161. To quote the linked comment, from one of the OTEL maintainers

> Looking at the code it looks like the following is happening:\
> -    @temporalio/interceptors-opentelemetry bypasses the SpanProcessor which is supposed to await the async attributes and passes spans directly to SpanExporter
> -    the SpanExporter then assumes that the resource was already awaited by the SpanProcessor, but since @temporalio/interceptors-opentelemetry bypassed the SpanProcessor this never happened.
> -    during normal operation of an OTel SDK, this log is not written - this happens since @temporalio/interceptors-opentelemetry bypasses a required SDK component.
>
> **Recommendation**: @temporalio/interceptors-opentelemetry should accept a SpanProcessor instead of a SpanExporter in [makeWorkflowExporter()](https://github.com/temporalio/sdk-typescript/blob/7c88cc7ab0d02f0002178ba03a349503d5bc481c/packages/interceptors-opentelemetry/src/worker/index.ts#L73-L90), and pass the spans to SpanProcessor#onEnd(Span) instead. This will properly await the resource attributes before exporting them, log errors and and handle batching in a user-configurable way, which may be desirable.

### Minimal Reproduction
See last comment on this issue: https://github.com/open-telemetry/opentelemetry-js/issues/5828#event-19364472301

```ts
export function makeWorkflowExporter(
  spanProcessor: SpanProcessor, // used with a BatchSpanProcessor for instance, which forwards to a SpanExporter
  resource: Resource
): InjectedSink<OpenTelemetryWorkflowExporter> {
  // note: somehow, `SpanProcessor#shutdown()` needs to be wired up somewhere to avoid data loss on shutdown.
  return {
    export: {
      fn: (info, spanData) => {
        const spans = spanData.forEach((serialized) => {
          Object.assign(serialized.attributes, info);
          // Spans are copied over from the isolate and are converted to ReadableSpan instances
          spanProcessor.onEnd(extractReadableSpan(serialized, resource));
        });
      },
    },
  };
}
```

### Environment/Versions

- OS and processor: x86_64-linux
- Temporal from source

### Additional context
This will affect https://github.com/temporalio/sdk-typescript/pull/1741

chris-olszewski:
## What was changed
Updated `makeWorkflowExporter` to accept a `SpanProcessor` and deprecate using `SpanExporter`.

## Why?
As detailed in #1779 `SpanExporter` does not wait for resource attributes to settle before exporting the span. We should instead encourage users to pass a `SpanProcessor` which holds off on exporting spans until resource attributes have settled.

## Checklist

1. Closes #1779 

2. How was this tested:
Added unit tests to both verify previous behavior and confirm that using a span provider fixes the bug.

3. Any docs updates needed?
Added deprecation notice when using `makeWorkflowExporter` with a `SpanExporter`. Will update OTEL sample once this is released.

### DIFF
--- a/packages/test/src/test-otel.ts
+++ b/packages/test/src/test-otel.ts
@@ -269,7 +269,7 @@ if (RUN_INTEGRATION_TESTS) {
       otel.start();
 
       const sinks: InjectedSinks<OpenTelemetrySinks> = {
-        exporter: makeWorkflowExporter(traceExporter, staticResource),
+        exporter: makeWorkflowExporter(new SimpleSpanProcessor(traceExporter), staticResource),
       };
 
       const worker = await Worker.create({
@@ -429,7 +429,7 @@ if (RUN_INTEGRATION_TESTS) {
       await otel.start();
 
       const sinks: InjectedSinks<OpenTelemetrySinks> = {
-        exporter: makeWorkflowExporter(exporter, staticResource),
+        exporter: makeWorkflowExporter(new SimpleSpanProcessor(exporter), staticResource),
       };
       const worker = await Worker.create({
         workflowsPath: require.resolve('./workflows'),
@@ -545,7 +545,7 @@ if (RUN_INTEGRATION_TESTS) {
     };
 
     const sinks: InjectedSinks<OpenTelemetrySinks> = {
-      exporter: makeWorkflowExporter(traceExporter, staticResource),
+      exporter: makeWorkflowExporter(new SimpleSpanProcessor(traceExporter), staticResource),
     };
 
     const worker = await Worker.create({


packages/interceptors-opentelemetry/src/worker/index.ts
```diff
@@ -122,13 +139,22 @@ export function makeWorkflowExporter(
...
           return extractReadableSpan(serialized, resource);
         });
-        // Ignore the export result for simplicity
-        exporter.export(spans, () => undefined);
+
+        if (givenSpanProcessor) {
+          spans.forEach((span) => processorOrExporter.onEnd(span));
+        } else {
+          // Ignore the export result for simplicity
+          processorOrExporter.export(spans, () => undefined);
```
chris-olszewski, line 147: Note, we could make the `export` function async and explicitly wait on resource attributes to settle instead of leaving the behavior as is. I don't think this is desired since we effectively would be choosing the `SpanProcessor` behavior for them.

packages/test/src/test-otel.ts
```diff
@@ -587,6 +587,81 @@ if (RUN_INTEGRATION_TESTS) {
     t.is(updateResult, true);
     t.is(workflowResult, true);
   });
+
+  // These tests verify makeWorkflowExporter's handling of async resource attributes:
+  // https://github.com/temporalio/sdk-typescript/issues/1779
+  // - Using SpanExporter directly: async attributes are not awaited
```
chris-olszewski, line 593: Not sure if we really need the `SpanExporter` case checked in. I mostly wrote it just to verify using `SpanProcessor` fixed the issue.

packages/test/src/test-otel.ts
```diff
@@ -587,6 +587,81 @@ if (RUN_INTEGRATION_TESTS) {
...
+  // - Using SpanProcessor: async attributes are awaited before export
+  for (const useSpanProcessor of [false, true]) {
+    test(`makeWorkflowExporter with ${
```
mjameswh, line 596: nit: Next time, prefer the use of ava's macros for these kind of parameterized tests. You can have a look at `testWorkflowCancellation` for an example of this. But I don't not fixing this one, I plan some rework of our tests soon enough anyway.

mjameswh (approved):

PR #1893, opened 2026-01-15
refactor(test): Minor tests refactorings

mjameswh:
## What was changed

- Add Nexus Endpoint registration to internal testing framework
- Refactor Workflow Cancellation tests to our modern internal testing framework
- Decrease default verbosity of internal test suites

### DIFF
--- a/packages/test/src/workflows/unblock-or-cancel.ts
+++ b/packages/test/src/workflows/unblock-or-cancel.ts
@@ -14,12 +14,15 @@ export async function unblockOrCancel(): Promise<void> {
   setHandler(unblockSignal, () => void (isBlocked = false));
   setHandler(isBlockedQuery, () => isBlocked);
   try {
+    console.log('Blocked');
     await condition(() => !isBlocked);
     isBlocked = false;
+    console.log('Unblocked');
   } catch (err) {
     if (!(err instanceof CancelledFailure)) {
       throw err;
     }
+    console.log('Cancelled');
   }
 }
 // @@@SNIPEND


packages/test/src/test-workflow-cancellation.ts
```diff
@@ -1,76 +1,77 @@
...
+        await workflow.result();
+        t.pass();
+      } else {
+        const err = await t.throwsAsync(workflow.result(), {
+          instanceOf: WorkflowFailedError,
+        });
+        if (!(err instanceof WorkflowFailedError)) {
+          throw new Error('Unreachable');
+        }
+        t.true(err.cause instanceof expected);
```
chris-olszewski:
```suggestion
        t.true(err!.cause instanceof expected);
```

packages/test/src/helpers-integration.ts
```diff
@@ -201,14 +202,18 @@ export interface Helpers {
...
+  registerNexusEndpoint(suffix?: string): Promise<string>;
 }
 
 export function configurableHelpers<T>(
   t: ExecutionContext<T>,
   workflowBundle: WorkflowBundle,
   testEnv: TestWorkflowEnvironment
 ): Helpers {
-  const taskQueue = t.title.replace(/ /g, '_');
+  const taskQueue = `${t.title}-${new Date().toISOString().replaceAll(/[:.zt]/g, '')}`
```
chris-olszewski: What is the motivation behind adding a timestring to the task queue?

mjameswh: The immediate reason is that, for simplicity, I reused the `taskQueue` name as the Nexus endpoint name. But the latter must be unique. Because of that constraint, I have seen occurrences of nexus endpoint registration errors in situations where I rerun the test multiple times while connected to a non-ephemeral server (i.e. by running tests with the `TEMPORAL_SERVICE_ADDRESS` env variable). Tests should obviously deregister their endpoints, but appending timestamp as an extra safety measure isn't an unreasonable thing to do.

More broadly, I've faced other occasional flakes (local, not CI) in non-nexus tests when running tests against a non-ephemeral server, due to a worker picking up tasks from a previous test run for the same taskqueue name (e.g. if some workflows didn't close at the end of the test). Appending timestamps to taskqueue names is also resolving that.

mjameswh: Well, I ended up removing the timestamp suffix.

I still think that was helpful and pertinent, but this was causing high-frequency flakes on tests that call `configurableHelpers` twice from the same test (mostly in `test-integration-{one,two,three}.ts`). Maybe I'll restore at a later point.

PR #1894, opened 2026-01-15
feat(workflow): Nexus Cancellation Types

mjameswh:
## What was changed

- Add Cancellation Types support for Nexus Operations scheduled
  using the workflow API. Based on temporalio/sdk-dotnet#589.

### DIFF
--- a/packages/workflow/src/nexus.ts
+++ b/packages/workflow/src/nexus.ts
@@ -292,7 +292,7 @@ export const NexusOperationCancellationType = {
 export type NexusOperationCancellationType =
   (typeof NexusOperationCancellationType)[keyof typeof NexusOperationCancellationType];
 
-export const [encodeNexusOperationCancellationType, decodeNexusOperationCancellationType] = makeProtoEnumConverters<
+const [encodeNexusOperationCancellationType, _] = makeProtoEnumConverters<
   coresdk.nexus.NexusOperationCancellationType,
   typeof coresdk.nexus.NexusOperationCancellationType,
   keyof typeof coresdk.nexus.NexusOperationCancellationType,


packages/test/src/test-workflow-nexus-cancellation.ts
```diff
@@ -0,0 +1,231 @@
...
+  );
+
+  return nexus.serviceHandler(service, {
+    startWorkflow,
+    startWorkflowFailCancel: {
+      start: startWorkflow.start.bind(startWorkflow),
+      cancel: async (_ctx, _token): Promise<void> => {
+        throw new nexus.HandlerError('NOT_IMPLEMENTED', 'Intentional failure');
+      },
+      // FIXME: Update nexus-rpc dependency, then remove these two methods
```
mjameswh, line 77: Update of the nexus rpc dependency will be done in a distinct PR, at a later time.

packages/workflow/src/interceptors.ts
```diff
@@ -288,6 +289,21 @@ export interface StartNexusOperationOptions {
    */
   readonly scheduleToCloseTimeout?: Duration;
 
+  /**
+   * Determines:
+   * - whether cancellation requests should be propagated from the Workflow to the Nexus Operation, and
+   * - whether and when should the Operation's cancellation be reported back to the Workflow
```
chris-olszewski: This mix of a bullet list and the conjoining `and` is a little hard to parse for me. I think the `and` can be dropped
```suggestion
   * - whether cancellation requests should be propagated from the Workflow to the Nexus Operation
   * - whether and when should the Operation's cancellation be reported back to the Workflow
```

packages/workflow/src/nexus.ts
```diff
@@ -231,3 +234,75 @@ function startNexusOperationNextHandler({
     });
   });
 }
+
+/**
+ * Determines:
+ * - whether cancellation requests should be propagated from the Workflow to the Nexus Operation, and
+ * - whether and when should the Operation's cancellation be reported back to the Workflow
```
chris-olszewski: Same note as the other doc comment. It might be worth putting a non-doc comment mentioning to keep these comments in sync.
```suggestion
 * - whether cancellation requests should be propagated from the Workflow to the Nexus Operation
 * - whether and when should the Operation's cancellation be reported back to the Workflow
```

PR #1897, opened 2026-01-16
Add test util to create Nexus endpoint.

Quinn-With-Two-Ns:
Add test utility to create Nexus endpoint. This matches what we have in [Java](https://www.javadoc.io/static/io.temporal/temporal-testing/1.30.1/io/temporal/testing/TestWorkflowEnvironment.html#createNexusEndpoint(java.lang.String,java.lang.String)) and [Dotnet](https://github.com/temporalio/sdk-dotnet/blob/a0a70a113c18547d12e41ab06e2fcd83fc8d6a57/src/Temporalio/Testing/WorkflowEnvironment.cs#L220C37-L220C61)

---

> [!NOTE]
> Adds convenient Nexus endpoint management for tests and refactors tests accordingly.
> 
> - Introduces `createNexusEndpoint(name, taskQueue)` and `deleteNexusEndpoint(endpoint)` on `TestWorkflowEnvironment`; exports `NexusEndpointIdentifier`
> - Refactors Nexus tests to use the new helpers instead of direct `operatorService` calls; adds a test covering create/delete lifecycle
> - CI/docs/release workflows: adjust build step to also `--filter '!typescript-sdk'` alongside `!@temporalio/core-bridge`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6604d40afb48b33ea6630e9672d44a6e2ac70286. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/testing/src/testing-workflow-environment.ts
```diff
@@ -357,6 +358,59 @@ export class TestWorkflowEnvironment {
...
+        name,
+        target: {
+          worker: {
+            namespace: this.namespace ?? 'default',
+            taskQueue,
+          },
+        },
+      },
+    });
+    return response.endpoint!;
```
chris-olszewski: Can we throw a descriptive response here if we do end up not getting back an endpoint on the response? With `!` we're just passing on the `undefined` to the end user.

packages/testing/src/testing-workflow-environment.ts
```diff
@@ -357,6 +358,59 @@ export class TestWorkflowEnvironment {
...
+   * ```ts
+   * const endpoint = await testEnv.createNexusEndpoint('my-endpoint', 'my-task-queue');
+   * // ... use the endpoint ...
+   * await testEnv.deleteNexusEndpoint(endpoint);
+   * ```
+   */
+  async deleteNexusEndpoint(endpoint: temporal.api.nexus.v1.IEndpoint): Promise<void> {
+    await this.connection.operatorService.deleteNexusEndpoint({
+      id: endpoint.id!,
+      version: endpoint.version!,
```
chris-olszewski:
```suggestion
      id: endpoint.id,
      version: endpoint.version,
```

packages/testing/src/testing-workflow-environment.ts
mjameswh: Agreed, though, to be fair, we're generally not paranoid about these kinds of checks. The grpc library will throw an error if server failed with some grpc error code, and if the service succeeded the request, then it is _expected_ to return a valid value.

mjameswh:
I'd propose this:

```diff
-  async createNexusEndpoint(name: string, taskQueue: string): Promise<temporal.api.nexus.v1.IEndpoint> {
+  async createNexusEndpoint(name: string, taskQueue: string): Promise<NexusEndpoint> {
     // ...

+    if (!response.endpoint?.id || !response.endpoint?.version) {
+      throw new TypeError('Unexpected response from createNexusEndpoint');
+    }

-    return response.endpoint;
+    return {
+      id: response.endpoint.id,
+      version: response.endpoint.version,
+      raw: response.endpoint,
+    };
   }

-  async deleteNexusEndpoint(endpoint: Pick<temporal.api.nexus.v1.IEndpoint>): Promise<void> {
+  async deleteNexusEndpoint(endpoint: Pick<NexusEndpoint, 'id' | 'version'>): Promise<void> {
     // ...
   }
```

with the following type definition at the bottom of `testing-workflow-environment.ts` (and reexported from that package's `index.ts`)

```ts
export type NexusEndpoint = {
  id: NonNullable<temporal.api.nexus.v1.IEndpoint['id']>;
  version: NonNullable<temporal.api.nexus.v1.IEndpoint['version']>;
  raw: temporal.api.nexus.v1.IEndpoint;
};
```

This is not intended to be a general-purpose Operator Service Client, so I would personally be ok reducing `NexusEndpoint` to only `{id, version}`, without anything else. But then, doesn't cost much keeping the original protobuf `IEndpoint` object as `raw`. This is something we did in a few other places in TS SDK.

PR #1906, opened 2026-01-21
feat(bun): add experimental bun support

chris-olszewski:
## What was changed
Adding experimental Bun support:
 - Work around bug in Bun where `postMessage` silently drops protobuf objects
 - Work around bug in Bun where object references get corrupted during `postMessage`
 - Work around bug in Bun where `Object.getOwnPropertyDescriptor` returns stale results for numerical properties
 - Work around https://github.com/oven-sh/bun/issues/20145
 - Work around `WorkerThread.terminate()` never resolving sometimes in Bun
 - Update test assertions to match output differences in Bun
 - Skip 3 tests: 2 regarding stack traces queries, 1 OOM test where Bun does not OOM where Node usually does
 - Adding Bun to our CI matrix

Current limitations:
- Stack traces do not have source maps applied to them
- Stack trace queries will return empty results
- No guarantee workflows between Node and Bun are valid

## Why?
TSIA

## Checklist

1. Partially addresses #1334

2. How was this tested:
Existing tests, [stress test](https://github.com/temporalio/sdk-typescript/actions/runs/21518197290/job/62001902844), ran through a dozen or so samples

3. Any docs updates needed?
Yes, we will need to top level documentation about the experimental status of Bun support as well as the current limitations.

### DIFF
--- a/packages/worker/src/workflow/vm-shared.ts
+++ b/packages/worker/src/workflow/vm-shared.ts
@@ -374,7 +374,6 @@ export abstract class BaseVMWorkflow implements Workflow {
   ): Promise<coresdk.workflow_completion.IWorkflowActivationCompletion> {
     try {
       if (this.context === undefined) throw new IllegalStateError('Workflow isolate context uninitialized');
-      activation = coresdk.workflow_activation.WorkflowActivation.fromObject(activation);
       if (!activation.jobs) throw new TypeError('Expected workflow activation jobs to be defined');
 
       // Queries are particular in many ways, and Core guarantees that a single activation will not
--- a/packages/worker/src/workflow/workflow-worker-thread.ts
+++ b/packages/worker/src/workflow/workflow-worker-thread.ts
@@ -69,7 +69,7 @@ async function handleRequest({ requestId, input }: WorkerThreadRequest): Promise
         // To work around this bug, we encode activations
         activation = coresdk.workflow_activation.WorkflowActivation.decode(input.activation);
       } else {
-        activation = input.activation;
+        activation = coresdk.workflow_activation.WorkflowActivation.fromObject(input.activation);
       }
       const completion = await workflow.activate(activation);
       const maybeEncodedCompletion = isBun


packages/test/src/helpers.ts
```diff
@@ -104,8 +105,23 @@ export function cleanStackTrace(ostack: string): string {
  * As of Node 24.6.0 type names are now present on source mapped stack traces which leads
  * to different stack traces depending on Node version.
  * See [f33e0fcc83954f728fcfd2ef6ae59435bc4af059](https://github.com/nodejs/node/commit/f33e0fcc83954f728fcfd2ef6ae59435bc4af059)
+ *
+ * In Bun, stack traces contain bundled paths instead of source-mapped paths.
+ * We only compare the first line (error message) which is consistent across runtimes.
```
chris-olszewski: TODO: avoid this and match exact stack traces across both runtimes

packages/test/src/test-interceptors.ts
```diff
@@ -245,17 +246,23 @@ if (RUN_INTEGRATION_TESTS) {
...
-          at $CLASS.nonRetryable (common/src/failure.ts)
-          at Object.continueAsNew (test/src/workflows/interceptor-example.ts)
-          at workflow/src/workflow.ts
-          at continueAsNewToDifferentWorkflow (test/src/workflows/continue-as-new-to-different-workflow.ts)
-    `
-    );
+
+    const cleanedStack = cleanOptionalStackTrace(err.cause.stack)!;
+    if (isBun) {
+      // Bun doesn't apply source maps to workflow bundle, so stack traces show bundle paths.
```
chris-olszewski: TODO: match exact stack trace bun produces

packages/test/src/test-workflows.ts
```diff
@@ -1724,6 +1738,10 @@ test('globalOverrides', async (t) => {
 test('logAndTimeout', async (t) => {
   const { workflowType, workflow } = t.context;
   const completion = await activate(t, makeStartWorkflow(workflowType));
+  // In Bun, strip stack trace to first line since bundled paths differ from source-mapped paths
```
chris-olszewski: TODO: match exact Bun stacktrace

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -255,12 +283,23 @@ export class VMWorkflowThreadProxy implements Workflow {
   ): Promise<coresdk.workflow_completion.IWorkflowActivationCompletion> {
     const output = await this.workerThreadClient.send({
       type: 'activate-workflow',
-      activation,
+      // TODO: It appears that some messages will never get sent with Bun's postMessage
+      // Specific example is test-payload-converter.ts 'Worker encodes/decodes a protobuf containing a binary array'
+      // I have narrowed it down to the `arguments` field of the init workflow job causing the no send, but not further
+      // Encoding activation results in the message being delivered
+      activation: isBun ? coresdk.workflow_activation.WorkflowActivation.encode(activation).finish() : activation,
```
mjameswh, line 287: Not urgent, but it would be interesting to see how performance compares on Node if we'd always encode activations and activation completions, then pass the buffer through `MessagePort.send`'s `transferList`.

Unintuitively, v8's serialization isn't particularly fast—it'd actually be slower than JSON according to a credible source, though I never confirmed that assertion myself. So I'm wondering here if it might turn out that encoding before transfer + passing ownership of the serialized buffer (that's the `transferList` thing) might always be a good thing anyway, rather than a bun-specific workaround...

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -423,7 +427,7 @@ export abstract class BaseVMWorkflow implements Workflow {
         this.workflowModule.activate(
           coresdk.workflow_activation.WorkflowActivation.fromObject({ ...activation, jobs: rest })
         );
-        this.tryUnblockConditionsAndMicrotasks();
+        await this.tryUnblockConditionsAndMicrotasks();
```
mjameswh: 🤔 Oh... I think a yield point here might hurt us.

Is it possible that we might have multiple workflow executions trying to concurrently take control of the shared vm? All calls going through `this.workflowModule` would properly restore the execution context, so that would technically not be a problem on those (though that might have a performance impact), but what about microtasks? There might be a wf exec context switch while there are pending microtasks that belongs to the first execution context.

That could possibly explain the isolation test failure you observed.

If that's indeed the case, a possible solution would be to add some form of locking mechanism around the whole `activate` function.

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -73,14 +94,15 @@ async function handleRequest({ requestId, input }: WorkerThreadRequest): Promise
...
-        // Do this on a copy of the object, as workflowInfo is the live object.
-        call.workflowInfo = {
-          ...call.workflowInfo,
-          unsafe: { ...call.workflowInfo.unsafe },
-        };
         delete (call.workflowInfo.unsafe as any).now;
+        // Use structuredClone for Bun to work around a postMessage bug where
+        // shared object references get corrupted during serialization.
+        // Node.js handles shared references correctly with a shallow copy.
+        call.workflowInfo = isBun
```
mjameswh: So Bun is giving access to `structuredClone` inside vms? Nice!

Using `structuredClone` is actually the right thing to do here. We used to have `("structuredClone" in globalThis) ? structuredClone(call.workflowInfo) : ..fallback..`.

But then I realized at some point that Node isn't making `structuredClone` available inside `vm`s, so we were always using the fallback, so I removed the ternary. But now, I'd just restore the original condition, rather than `isBun`.

packages/worker/src/workflow/vm-shared.ts
chris-olszewski: I did a refactor to avoid this yield point for Node so we don't regress.

I'm not sure the yield point is an issue for correctness. All of the pending microtasks should be executed from the queue before we exit the task. Since we only switch the bag when calling into the workflow module, the bag should be in the same state for the microtasks that get run before the task exits.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -5,13 +5,106 @@ import { native } from '@temporalio/core-bridge';
...
+    Object.defineProperty(globalThis, '__TEMPORAL_CALL_INTO_SCOPE', {
+      value: __TEMPORAL_CALL_INTO_SCOPE, writable: false, enumerable: false, configurable: false
+    });
+  }`;
+}
+
+// This is a workaround for a bug in Bun where Object.getOwnPropertyDescriptor returns
+// stale values for numeric properties after modification. We must read/write numeric
+// properties directly.
+function generateBunCallIntoScopeScript(): string {
```
chris-olszewski, line 54: I don't love the duplicated code between these two call into scope scripts, but this does allow us to not have any impact on the existing Node codepath while still being able to provide the Bun workaround.

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -255,12 +281,22 @@ export class VMWorkflowThreadProxy implements Workflow {
...
       runId: this.runId,
     });
     if (output?.type !== 'activation-completion') {
       throw new TypeError(`Got invalid response output from Workflow Worker thread ${output}`);
     }
+    if (output.completion instanceof Uint8Array) {
+      if (isBun) {
+        return coresdk.workflow_completion.WorkflowActivationCompletion.decode(output.completion);
+      } else {
+        throw new Error('got encoded message even when not using bun');
```
chris-olszewski: Not sure about if we actually want to error here. I could see it being useful as a debug assertion while running tests to make sure we aren't accidentally on this codepath in Node, but if a user hit this codepath, I think we would want to just decode.

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -360,6 +361,11 @@ export abstract class BaseVMWorkflow implements Workflow {
 
   /**
    * Send request to the Workflow runtime's worker-interface
+   *
+   * In order to properly work around Bun's lack of support for `microtaskMode: afterEvaluate` it is important to
+   * immediately schedule a task after each call into the `workflowModule`. This task ensures that any microtasks
+   * from a specific workflow will execute while the vm is still setup for said workflow.
+   * This is why there are `if (isBun) await new Promise(setImmediate)` scattered throughout this function.
```
chris-olszewski, line 370: I do not love how this looks at the moment, but unsure of a good approach without forking this method that avoids adding any yield points to the Node codepath.

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -73,14 +96,16 @@ async function handleRequest({ requestId, input }: WorkerThreadRequest): Promise
...
-        call.workflowInfo = {
-          ...call.workflowInfo,
-          unsafe: { ...call.workflowInfo.unsafe },
-        };
         delete (call.workflowInfo.unsafe as any).now;
+        // Use structuredClone when available
+        // This lets us work around a bug in Bun's postMessage where
+        // shared object references get corrupted during serialization.
+        call.workflowInfo =
+          'structuredClone' in globalThis
```
chris-olszewski, line 102: Note this is a change for the Node codepath as well as a workaround for Bun.

packages/test/src/test-integration-split-three.ts
```diff
@@ -29,7 +29,7 @@ test('cancel-http-request', configMacro, async (t, config) => {
   t.pass();
 });
 
-if ('promiseHooks' in v8) {
+if ('promiseHooks' in v8 && !isBun) {
```
chris-olszewski, line 33: At the moment stack trace queries just return empty results. I can add in a throw for these queries in Bun if desired.

packages/test/src/test-integration-split-one.ts
```diff
@@ -170,12 +170,34 @@ test.serial('activity-failure with Error', configMacro, async (t, config) => {
...
-      at throwAnError (test/src/activities/index.ts)
-  `
+  compareStackTrace(
+    t,
+    cleanOptionalStackTrace(err.cause.cause.stack)!,
+    isBun
+      ? dedent`
+    Error: Fail me
+        at throwAnError (test/lib/activities/index.js)
+        at execute (worker/lib/activity.js)
```
mjameswh: That stack trace should have been cut at this line when serializing the failure. We need to add a new pattern to the cutoff stack patterns. See failure-converter.ts:60

packages/test/src/test-integration-split-one.ts
```diff
@@ -263,11 +307,29 @@ test.serial('child-workflow-failure', configMacro, async (t, config) => {
...
-      ApplicationFailure: failure
-          at $CLASS.nonRetryable (common/src/failure.ts)
-          at throwAsync (test/src/workflows/throw-async.ts)
-    `
+      isBun
+        ? dedent`
+        ApplicationFailure: failure
+            at nonRetryable (test/workflow-bundle-$HASH.js)
+            at throwAsync (test/workflow-bundle-$HASH.js)
+            at startWorkflowNextHandler (test/workflow-bundle-$HASH.js)
```
mjameswh: Same thing, failure-converter.ts:64

packages/common/src/converter/failure-converter.ts
```diff
@@ -57,13 +57,13 @@ function combineRegExp(...regexps: RegExp[]): RegExp {
...
 const CUTOFF_STACK_PATTERNS = combineRegExp(
   /** Activity execution */
-  /\s+at Activity\.execute \(.*[\\/]worker[\\/](?:src|lib)[\\/]activity\.[jt]s:\d+:\d+\)/,
+  /\s+at (Activity\.)?execute \(.*[\\/]worker[\\/](?:src|lib)[\\/]activity\.[jt]s:\d+:\d+\)/,
   /** Nexus execution */
   /\s+at( async)? NexusHandler\.invokeUserCode \(.*[\\/]worker[\\/](?:src|lib)[\\/]nexus[\\/]index\.[jt]s:\d+:\d+\)/,
-  /** Workflow activation */
-  /\s+at Activator\.\S+NextHandler \(.*[\\/]workflow[\\/](?:src|lib)[\\/]internals\.[jt]s:\d+:\d+\)/,
+  /** Workflow activation (inbound handlers only) */
+  /\s+at( async)? (Activator\.)?(startWorkflow|queryWorkflow|signalWorkflow|update|validateUpdate)NextHandler \(.*\.[jt]s:\d+:\d+\)/,
```
chris-olszewski, line 64: Don't love enumerating these here, but the lack of `Activator` prefixes on the stack trace for Bun meant we were matching too broadly such that we changed Node behavior e.g. [timerNextHandler](https://github.com/temporalio/sdk-typescript/actions/runs/21881886145/job/63166436502#step:12:6913) calls becoming cutoff points.

We could gate the more permissive regexp behind `isBun`, but we would need to pass that into the workflow context which I would like to avoid if possible.

packages/test/src/test-integration-split-three.ts
mjameswh: We should throw a failure, similar to what .Net does:

https://github.com/temporalio/sdk-dotnet/blob/main/src/Temporalio/Worker/WorkflowInstance.cs#L1727-L1728

packages/test/src/helpers.ts
```diff
@@ -99,17 +100,30 @@ export function cleanStackTrace(ostack: string): string {
...
 
 /**
- * Compare stack traces using $CLASS keyword to match any inconsistent identifiers
+ * Compare stack traces using keywords to match any inconsistent parts of the stack trace
  *
  * As of Node 24.6.0 type names are now present on source mapped stack traces which leads
  * to different stack traces depending on Node version.
  * See [f33e0fcc83954f728fcfd2ef6ae59435bc4af059](https://github.com/nodejs/node/commit/f33e0fcc83954f728fcfd2ef6ae59435bc4af059)
+ *
+ * Bun does not support promist hooks meaning we are currently unable to apply the source map on stack traces and workflow bundles
```
mjameswh:
```suggestion
 * Bun does not support promise hooks meaning we are currently unable to apply the source map on stack traces and workflow bundles
```

packages/test/src/test-nexus-handler.ts
```diff
@@ -465,7 +490,17 @@ test('cancel Operation Handler errors', async (t) => {
       compareStackTrace(
         t,
         cleanStackTrace(err.stack!),
-        `ApplicationFailure: deliberate failure
+        isBun
+          ? `ApplicationFailure: deliberate failure
+    at create (common/lib/failure.js)
+    at cancel (test/lib/test-nexus-handler.js)
+    at cancel (nexus-rpc/lib/handler/service-registry.js)
+    at invokeUserCode (worker/lib/nexus/index.js)
```
mjameswh: failure-converter.ts:62

packages/worker/src/workflow/workflow-worker-thread.ts
```diff
@@ -61,10 +63,27 @@ async function handleRequest({ requestId, input }: WorkerThreadRequest): Promise
...
         throw new IllegalStateError(`Tried to activate non running workflow with runId: ${input.runId}`);
       }
-      const completion = await workflow.activate(input.activation);
+      let activation;
+      if (input.activation instanceof Uint8Array) {
+        // Some activation messages get silently dropped by Bun's postMessage.
+        // To work around this bug, we encode activations
+        activation = coresdk.workflow_activation.WorkflowActivation.decode(input.activation);
+      } else {
+        activation = input.activation;
```
mjameswh: Given that we need to have these two branches here, I'd suggest moving `vm-shared.ts#activate`'s second line here in the Node branch, as calling `fromObject` on the object we just decoded adds useless processing.

```
      activation = coresdk.workflow_activation.WorkflowActivation.fromObject(activation);
```

packages/worker/src/workflow/workflow-worker-thread.ts
mjameswh: No problem with that.

packages/worker/src/workflow/vm-shared.ts
mjameswh: I agree. Not elegant, but no better option.

PR #1910, opened 2026-01-22
chore: break up test package

chris-olszewski:
## What was changed
Break up the `test` package so tests for specific packages reside in the package.

As part of this, we also changed how each package gets built. Previously `pnpm run build` worked as building `test` would end up building every package it depended upon. As we no longer expect this to always hold, we now have each package build itself.

In order to keep this PR more reviewable, I avoided breaking up any test files and only moved entire files where they were applicable. There's certainly more work that can be done for breaking up `test`, but this suits our most purposes for the moment.

## Why?
As we introduce support for OTEL v2, we will need to test both v1 and v2 of OTEL. The easiest way to achieve this is by having separate packages for each. In order to support this each package needs to own it's own tests.

## Checklist

1. Closes N/A

2. How was this tested:
`pnpm run test`

3. Any docs updates needed?
N/A


.github/workflows/ci.yml
```diff
@@ -178,7 +178,8 @@ jobs:
         working-directory: ${{ runner.temp }}
 
       - name: Compile code
-        run: pnpm --recursive --stream --filter '!@temporalio/core-bridge' --filter '!typescript-sdk' run build
+        # Temporary measure to work around PNPM installs of local packages
```
chris-olszewski: I don't love this, but this is to avoid making the PR even more invasive.

Crux of the issue is with how `pnpm install` works for `workspace` protocol. If the `lib` for our internal packages is not present at install time it doesn't get symlinked. This results in tests failing as they can't find the `lib/index.js` entry point. The second install after build fixes this since pnpm will properly link the `lib` dir now that it exists.

A proper fix would be leveraging something like [publishConfig](https://pnpm.io/package_json#publishconfig) to use a different entrypoint for dev vs what we ship.

packages/nyc-test-coverage/package.json
```diff
@@ -39,6 +42,7 @@
     "@types/istanbul-lib-coverage": "^2.0.6",
     "@types/istanbul-lib-instrument": "^1.7.7",
     "@types/webpack": "^5.28.5",
+    "typescript": "^5.6.3",
```
chris-olszewski, line 45: Need to specify TS version as otherwise `ts-loader` grabs a different TS version than the rest of the repo and the latest version has some [type signatures changed](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#libdts-changes) resulting in failed compilation.

packages/test/src/helpers.ts
```diff
@@ -1,247 +1,47 @@
...
-export const test: TestFn<unknown> = inWorkflowContext() ? (noopTest as any) : ava;
+  createBaseBundlerOptions,
+  loadHistory as loadHistoryBase,
+  saveHistory as saveHistoryBase,
+  RUN_TIME_SKIPPING_TESTS,
+  test,
+  noopTest,
+} from '@temporalio/test-helpers';
+
+// Re-export from test-helpers
```
chris-olszewski, line 17: Re-exporting here just so reviewing this PR has less noise from a bunch of import changes

packages/test/src/test-local-activities.ts
```diff
@@ -1,93 +1,17 @@
...
-  WorkflowBundle,
-  WorkerOptions,
-} from '@temporalio/worker';
+import { bundleWorkflowCode, DefaultLogger, LogLevel, Runtime } from '@temporalio/worker';
 import * as workflow from '@temporalio/workflow';
-import { test as anyTest, bundlerOptions, Worker, TestWorkflowEnvironment } from './helpers';
+import { test as anyTest, Worker, TestWorkflowEnvironment, BaseContext, helpers } from '@temporalio/test-helpers';
+import { bundlerOptions } from './helpers';
 
-// FIXME MOVE THIS SECTION SOMEWHERE IT CAN BE SHARED //
```
chris-olszewski, line 16: Seemed like as good a time as any to fix this

.github/workflows/conventions.yml
```diff
@@ -51,4 +51,6 @@ jobs:
         run: pnpm --recursive --stream --filter '!@temporalio/core-bridge' --filter '!typescript-sdk' run build
 
       - run: pnpm run lint:check
+        env:
+          NODE_OPTIONS: '--max-old-space-size=8192'
```
chris-olszewski, line 55: OOM workaround. Totally unrelated, I am looking at migrating us to `oxlint`

mjameswh (approved):

PR #1922, opened 2026-02-04
feat(runtime): add support for buffered metrics

### LINKED ISSUE #1177: [Feature Request] Forward Core's metrics to TypeScript-land

Core is adding support for exporting metrics to lang (see [Core PR](https://github.com/temporalio/sdk-core/pull/544)). This will allow various customizations to metrics handling, including:

- Exporting additional/custom metrics;
- Filtering out some metrics (to reduce cost, since commercial metrics collectors generally charge per metric);
- Export to metrics collectors that are not supported by Core itself.

This ticket is tracking effort at adding support in TS SDK for accepting and handling metrics from Core SDK. The scope of this ticket does not extend to the development of a metric exporter for any specific metrics collector.

mjameswh:
## What was changed

- Add experimental support for Buffered Metrics. This feature allows exporting metrics emitted by the Core SDK through a lang-side exporter. Resolves #1177.
- Also:
    - Updated Core SDK to latest

### DIFF
--- a/packages/test/src/test-otel.ts
+++ b/packages/test/src/test-otel.ts
@@ -2,6 +2,7 @@
 /**
  * Manual tests to inspect tracing output
  */
+import * as http from 'http';
 import * as http2 from 'http2';
 import * as otelApi from '@opentelemetry/api';
 import { SpanStatusCode, createTraceState } from '@opentelemetry/api';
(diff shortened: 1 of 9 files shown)


packages/test/src/test-otel.ts
```diff
@@ -64,7 +63,7 @@ async function withFakeGrpcServer(
         });
         res.write(
           // This is a raw gRPC response, of length 0
-          Buffer.from([
+          new Uint8Array([
```
chris-olszewski: Reason for this change other than general Buffer -> Uint8Array migration?

packages/worker/src/runtime-options.ts
```diff
@@ -359,6 +361,33 @@ export interface PrometheusMetricsExporter {
...
+     * @default 10000
+     */
+    maxBufferSize?: number;
+
+    /**
+     * If set to true, the exporter will use seconds for durations instead of milliseconds.
+     *
+     * @default false
+     */
+    useSecondsForDurations?: boolean;
```
chris-olszewski: For my own knowledge, what's the reason for offering this configuration?

packages/test/src/test-otel.ts
mjameswh: For some reason, that usage of `Buffer` was showing as an error in my editor, even though that was building fine.

packages/worker/src/runtime-options.ts
mjameswh: I don't know why we introduced that initially.

In any case, we do now in every SDKs, and already have it in TS on both OTel and Prometheus metric exporter, so at this point, that's feature parity.

packages/test/src/test-otel.ts
chris-olszewski: I wonder if your editor is using TS 5.9: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#libdts-changes

PR #1925, opened 2026-02-05
chore: Prune Dev Server dynamic configs

mjameswh:
## What was changed

- Prune out the list of dynamic configs used by our SDK internal test suite. Many of those dynamic configs are either now default or have been dropped in the server.

### DIFF
--- a/packages/test/src/helpers-integration.ts
+++ b/packages/test/src/helpers-integration.ts
@@ -47,10 +47,10 @@ const defaultDynamicConfigOptions = [
 function setupRuntime(recordedLogs?: { [workflowId: string]: LogEntry[] }, runtimeOpts?: Partial<RuntimeOptions>) {
   const logger = recordedLogs
     ? new DefaultLogger('DEBUG', (entry) => {
-      const workflowId = (entry.meta as any)?.workflowInfo?.workflowId ?? (entry.meta as any)?.workflowId;
-      recordedLogs![workflowId] ??= [];
-      recordedLogs![workflowId].push(entry);
-    })
+        const workflowId = (entry.meta as any)?.workflowInfo?.workflowId ?? (entry.meta as any)?.workflowId;
+        recordedLogs![workflowId] ??= [];
+        recordedLogs![workflowId].push(entry);
+      })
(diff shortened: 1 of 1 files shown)


packages/test/src/helpers-integration.ts
```diff
@@ -40,17 +40,8 @@ export interface Context {
...
-  'frontend.workerVersioningDataAPIs=true',
-  'frontend.workerVersioningWorkflowAPIs=true',
   'system.enableActivityEagerExecution=true',
-  'system.enableDeploymentVersions=true',
-  'system.enableEagerWorkflowStart=true',
-  'system.forceSearchAttributesCacheRefreshOnRead=true',
-  'worker.buildIdScavengerEnabled=true',
-  'worker.removableBuildIdDurationSinceDefault=1',
-  'component.nexusoperations.recordCancelRequestCompletionEvents=true',
+  'history.enableRequestIdRefLinks=true',
```
chris-olszewski, line 44: What is the best way to find an explanation of what a dynamic config does? I followed some of the links at https://docs.temporal.io/references/dynamic-configuration, but wasn't able to find this one

mjameswh: Yeah... That's difficult.

For context, there are basically two distinct categories of dynamic configs.

1. Some dynamic configs are user-tuneable feature flags, providing server-level or namespace-level control on exposed features and some performance-related options. Those are relatively stable over time, and meant to be used by Temporal server administrators, so we try to document those properly, e.g. on that page you referred to, but the reference is definitely incomplete.
2. Other dynamic configs are introduced as gatekeepers for not-yet-stable server-side features. Those are short-lived (just a few server releases, then they get removed), and not something we want to broadly expose to users. Now, _we_ (Temporalites) obviously see a lot of those, even though the exact signification of a given dynamic config is often a bit obscure to anyone outside that specific feature crew.

Most dynamic configs (all that start with `frontend`, `history`, `matching` or `worker`) should be defined in [this file](https://github.com/temporalio/temporal/blob/release/v1.29.x/common/dynamicconfig/constants.go). So you can walk that file at various tags and branches to see when a specific flag was added, what's its default value, and whether it is still pertinent on latest releases.

From there, you may also get the constant used by the server to identify that feature, and search the code base to see what is gated by that constant.

Issue #1887, opened 2026-01-14
Allow logging in queries and update validators


cretz:
Currently logging is suppressed during queries and update validators because `isReplaying()` returns true. We should allow logging during these live read operations.

Related: temporalio/features#718

Reference implementations:
- Python: https://github.com/temporalio/sdk-python/pull/1274
- .NET: https://github.com/temporalio/sdk-dotnet/pull/594
- Ruby: https://github.com/temporalio/sdk-ruby/pull/378

PR #1927, opened 2026-02-05
Allow logging queries validators during replay

THardy98:
## What was changed
Added `isReplayingHistoryEvents` to `unsafe` workflow info as state that tracks when we are replaying inside queries and update validators. We use this state to allow logs to still be emit even when replaying (we want to allow logs to emit during these read-only operations, but not when replaying history events).

## Why?
Allow logs to be emit from queries and update validators during replay.

1. Closes https://github.com/temporalio/sdk-typescript/issues/1887

2. How was this tested:
Added integration test

3. Any docs updates needed?
No?


packages/test/src/test-sinks.ts
```diff
@@ -554,4 +555,72 @@ if (RUN_INTEGRATION_TESTS) {
...
+      taskQueue,
+      sinks,
+      // Avoid waiting for sticky execution timeout on worker transition
+      stickyQueueScheduleToStartTimeout: '1s',
+    });
+
+    // Empty recorded messages
+    recordedMessages = [];
+
+    // Run the entire workflow through workflow 2 (will replay).
```
chris-olszewski:
```suggestion
    // Run the entire workflow through worker 2 (will replay).
```

packages/test/src/test-sinks.ts
```diff
@@ -554,4 +555,72 @@ if (RUN_INTEGRATION_TESTS) {
       },
     ]);
   });
+
+  test('Logging is allowed in query handlers and update validators', async (t) => {
+    const taskQueue = `${__filename}-${t.title}`;
+
+    let recordedMessages = Array<{ message: string; isReplaying: boolean }>();
```
chris-olszewski: Not a strong opinion, but I generally prefere `[]` over `Array`
```suggestion
    let recordedMessages: { message: string; isReplaying: boolean }[] = [];
```

chris-olszewski (approved):
LGTM

packages/workflow/src/interfaces.ts
```diff
@@ -234,6 +234,18 @@ export interface UnsafeWorkflowInfo {
   readonly now: () => number;
 
   readonly isReplaying: boolean;
+
+  /**
+   * Whether the workflow is currently replaying history events.
+   *
+   * This is similar to {@link isReplaying}, but returns `false` during query handlers and update
```
mjameswh, line 244: Please also add typedoc on `isReplaying`.

packages/workflow/src/interfaces.ts
```diff
@@ -234,6 +234,18 @@ export interface UnsafeWorkflowInfo {
...
+   * validators, which are live read-only operations that should not be considered as replaying
+   * history events.
+   *
+   * Use this for log filtering: log messages should be emitted during queries and update validators
```
mjameswh: I don't feel the prescriptive tone is appropriate here, for a few reasons; notably, this flag is not only about logging, and the SDK already internally takes that decision internally, so users don't have to "use this property for log filtering".

I'd suggest instead phrasing this as a matter of fact. Basically, "this property is true => workflow log messages are suppressed and sinks defined with callDuringReplay=false wont get processed"

mjameswh (approved):
Left some minor review comments. Good to merge after that.

PR #1950, opened 2026-02-24
feat: add serialization context to data conversion interfaces

THardy98:
## Summary

Closes: https://github.com/temporalio/sdk-typescript/issues/1661

Add serialization context to the data conversion interfaces (`PayloadConverter`, `FailureConverter`, `PayloadCodec`), enabling converters to vary behavior based on which workflow or activity owns the data being serialized. Brings the TypeScript SDK to parity with the Python SDK's serialization context support.

## How it works

Converters can optionally implement `withContext(context: SerializationContext)` to receive a context-bound copy. If not implemented, the converter is used as-is.

Context is threaded at every serialization boundary: 
- client → server (workflow start, queries, signals, updates, schedule actions, async completion)
- worker → activity (args, results, failures, heartbeats)
- workflow isolate (activity commands, child workflows, external signals/cancels).

**Seq-based context tracking:**

Both the workflow activator (`internals.ts`) and the `WorkflowCodecRunner` maintain parallel maps from command sequence number → serialization context. When a command is issued (e.g. `scheduleActivity`), the context is stored keyed by seq. When the corresponding resolve job arrives (e.g. `resolveActivity`), the stored context is retrieved, used for deserialization, and deleted. This is necessary because scheduling and resolution happen in separate activation cycles.

## What to pay attention to

- **Two parallel seq maps** — the activator tracks contexts for the `PayloadConverter`/`FailureConverter` (workflow isolate side), while the codec runner tracks contexts for `PayloadCodec` (worker side). They follow the same lifecycle but operate independently.
- **`AsyncCompletionClient.withContext()`** — the only way to bind context. By-ID and by-token paths without it use the bare converter.

## Test plan

- `test-serialization-context.ts` — Integration tests using a `Tracer` type that accumulates context tags at each serialization hop. Full trace arrays are asserted with `t.deepEqual` to catch false positives. Covers: workflow start, activity, child workflow, query, update, signal, external signal/cancel, schedule actions, async completion (by-id, bound-token, unbound-token).
- `test-workflow-codec-runner.ts` — 9 unit tests verifying seq map contents and codec context binding for: activity, local activity, child workflow round-trips, workflow-level commands, signal/cancel independence, context-free fallback, forgetRun.

```bash
pnpm run build
pnpm -C packages/test exec ava ./lib/test-workflow-codec-runner.js
pnpm -C packages/test exec ava ./lib/test-serialization-context.js
RUN_INTEGRATION_TESTS=1 pnpm -C packages/test exec ava
```

### DIFF
--- a/packages/test/src/test-serialization-context.ts
+++ b/packages/test/src/test-serialization-context.ts
@@ -88,7 +88,13 @@ if (RUN_INTEGRATION_TESTS) {
     });
 
     // Workflow activities default activityId to seq, and this workflow schedules a single activity first.
-    const activityTag = activityContextTag({ type: 'activity', namespace, workflowId, activityId: '1', isLocal: false });
+    const activityTag = activityContextTag({
+      type: 'activity',
+      namespace,
+      workflowId,
+      activityId: '1',
+      isLocal: false,
+    });
     // Local activities share the activity seq counter, so the local activity (scheduled second) gets activityId '2'.
     const localActivityTag = activityContextTag({
       type: 'activity',
@@ -312,7 +318,11 @@ if (RUN_INTEGRATION_TESTS) {
         activityId: unboundInfo.activityId,
         isLocal: false,
       });
-      const unboundWorkflowTag = workflowContextTag({ type: 'workflow', namespace, workflowId: workflowIdUnboundToken });
+      const unboundWorkflowTag = workflowContextTag({
+        type: 'workflow',
+        namespace,
+        workflowId: workflowIdUnboundToken,
+      });
       // Unbound token without withContext(): same as by-ID — no client context
       t.deepEqual(unboundResult.trace, [
         `to:none`,
--- a/packages/workflow/src/workflow.ts
+++ b/packages/workflow/src/workflow.ts
@@ -403,7 +403,7 @@ function startChildWorkflowExecutionNextHandler({
     type: 'workflow',
     namespace: activator.info.namespace,
     workflowId,
-    workflowType: workflowType,
+    workflowType,
   };
   const payloadConverter = withPayloadConverterContext(activator.payloadConverter, context);
   const startPromise = new Promise<string>((resolve, reject) => {
--- a/packages/client/src/schedule-helpers.ts
+++ b/packages/client/src/schedule-helpers.ts
@@ -253,7 +253,12 @@ export async function encodeScheduleAction(
   action: CompiledScheduleAction,
   headers: Headers
 ): Promise<temporal.api.schedule.v1.IScheduleAction> {
-  const context: WorkflowSerializationContext = { type: 'workflow', namespace, workflowId: action.workflowId, workflowType: action.workflowType };
+  const context: WorkflowSerializationContext = {
+    type: 'workflow',
+    namespace,
+    workflowId: action.workflowId,
+    workflowType: action.workflowType,
+  };
   const contextDataConverter = withSerializationContext(dataConverter, context);
   return {
     startWorkflow: {
@@ -332,7 +337,12 @@ export async function decodeScheduleAction(
   if (pb.startWorkflow) {
     const workflowId = pb.startWorkflow.workflowId!;
     const workflowType = pb.startWorkflow.workflowType?.name ?? undefined;
-    const contextDataConverter = withSerializationContext(dataConverter, { type: 'workflow', namespace, workflowId, workflowType });
+    const contextDataConverter = withSerializationContext(dataConverter, {
+      type: 'workflow',
+      namespace,
+      workflowId,
+      workflowType,
+    });
     const { staticSummary, staticDetails } = await decodeUserMetadata(
       contextDataConverter,
       pb.startWorkflow?.userMetadata
--- a/packages/worker/src/workflow-codec-runner.ts
+++ b/packages/worker/src/workflow-codec-runner.ts
@@ -198,7 +198,12 @@ export class WorkflowCodecRunner {
                   : undefined;
 
               const initializeContext = job.initializeWorkflow?.workflowId
-                ? { type: 'workflow' as const, namespace: this.namespace, workflowId: job.initializeWorkflow.workflowId, workflowType: job.initializeWorkflow.workflowType ?? undefined }
+                ? {
+                    type: 'workflow' as const,
+                    namespace: this.namespace,
+                    workflowId: job.initializeWorkflow.workflowId,
+                    workflowType: job.initializeWorkflow.workflowType ?? undefined,
+                  }
                 : workflowContext;
               const initializeCodecs = this.codecsForContext(initializeContext);
               const workflowCodecs = this.codecsForContext(workflowContext);
--- a/packages/workflow/src/internals.ts
+++ b/packages/workflow/src/internals.ts
@@ -588,7 +588,12 @@ export class Activator implements ActivationHandler {
 
   public initializeWorkflow(activation: coresdk.workflow_activation.IInitializeWorkflow): void {
     const { continuedFailure, lastCompletionResult, memo, searchAttributes } = activation;
-    const workflowContext = { type: 'workflow' as const, namespace: this.info.namespace, workflowId: this.info.workflowId, workflowType: this.info.workflowType };
+    const workflowContext = {
+      type: 'workflow' as const,
+      namespace: this.info.namespace,
+      workflowId: this.info.workflowId,
+      workflowType: this.info.workflowType,
+    };
     const payloadConverter = withPayloadConverterContext(this.payloadConverter, workflowContext);
     const failureConverter = withFailureConverterContext(this.failureConverter, workflowContext);
 
@@ -1390,14 +1395,24 @@ export class Activator implements ActivationHandler {
   }
 
   errorToFailure(err: unknown): ProtoFailure {
-    const workflowContext = { type: 'workflow' as const, namespace: this.info.namespace, workflowId: this.info.workflowId, workflowType: this.info.workflowType };
+    const workflowContext = {
+      type: 'workflow' as const,
+      namespace: this.info.namespace,
+      workflowId: this.info.workflowId,
+      workflowType: this.info.workflowType,
+    };
     const payloadConverter = withPayloadConverterContext(this.payloadConverter, workflowContext);
     const failureConverter = withFailureConverterContext(this.failureConverter, workflowContext);
     return failureConverter.errorToFailure(err, payloadConverter);
   }
 
   failureToError(failure: ProtoFailure): Error {
-    const workflowContext = { type: 'workflow' as const, namespace: this.info.namespace, workflowId: this.info.workflowId, workflowType: this.info.workflowType };
+    const workflowContext = {
+      type: 'workflow' as const,
+      namespace: this.info.namespace,
+      workflowId: this.info.workflowId,
+      workflowType: this.info.workflowType,
+    };
     const payloadConverter = withPayloadConverterContext(this.payloadConverter, workflowContext);
     const failureConverter = withFailureConverterContext(this.failureConverter, workflowContext);
     return failureConverter.failureToError(failure, payloadConverter);


THardy98:
Sorry I don't know how to lessen the diff for `workflow-codec-runner.ts`.

Open to suggestions as the diff looks worse than it actually is

packages/common/src/index.ts
```diff
@@ -13,6 +13,12 @@ export * from './converter/data-converter';
 export * from './converter/failure-converter';
 export * from './converter/payload-codec';
 export * from './converter/payload-converter';
+export { withSerializationContext } from './converter/serialization-context';
```
chris-olszewski: This module has many exported identifiers, why only re-export one? Are the other ones only intended to be used by our internals?

packages/test/src/test-serialization-context.ts
```diff
@@ -0,0 +1,429 @@
...
+        args: [makeTracer('start-input'), childWorkflowId, signalTargetWorkflowId, cancelTargetWorkflowId],
+      });
+
+      const queryResult = await handle.query(serializationContextQuery, makeTracer('query-input'));
+      const updateResult = await handle.executeUpdate(serializationContextUpdate, {
+        args: [makeTracer('update-input')],
+      });
+      await handle.signal(serializationContextFinishSignal, makeTracer('finish-signal-input'));
+
+      const result = await handle.result();
```
chris-olszewski: Small suggestion, but I prefer keeping assertions outside of the worker function.

```suggestion
    const { result, queryResult, updateResult } = await worker.runUntil(async () => {
      const handle = await client.workflow.start(serializationContextWorkflow, {
        workflowId,
        taskQueue,
        args: [makeTracer('start-input'), childWorkflowId, signalTargetWorkflowId, cancelTargetWorkflowId],
      });

      const queryResult = await handle.query(serializationContextQuery, makeTracer('query-input'));
      const updateResult = await handle.executeUpdate(serializationContextUpdate, {
        args: [makeTracer('update-input')],
      });
      await handle.signal(serializationContextFinishSignal, makeTracer('finish-signal-input'));

      const result = await handle.result();
      return { result, queryResult, updateResult };
    });
```

packages/common/src/converter/serialization-context.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+export interface WorkflowSerializationContext {
+  namespace: string;
+  workflowId: string;
+}
+
+/**
+ * Context for activity-level serialization operations.
+ */
+export interface ActivitySerializationContext extends WorkflowSerializationContext {
```
chris-olszewski: Discussed on Slack, but I believe this shouldn't be extends due to SAA

packages/common/src/converter/serialization-context.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (
+    payloadConverter === converter.payloadConverter &&
+    failureConverter === converter.failureConverter &&
+    !codecsChanged
+  ) {
+    return converter;
+  }
+  return { payloadConverter, failureConverter, payloadCodecs };
+}
```
chris-olszewski, line 115: Given our use `===` here, we might want to inform users via doc comment about the implications of returning `this` from `withContext` in their implementations. An alternative would be to just check if `withContext` is defined on any of the converters or codecs and assume change if one is present.

packages/client/src/async-completion-client.ts
```diff
@@ -127,7 +142,10 @@ export class AsyncCompletionClient extends BaseClient {
   async complete(fullActivityId: FullActivityId, result: unknown): Promise<void>;
 
   async complete(taskTokenOrFullActivityId: Uint8Array | FullActivityId, result: unknown): Promise<void> {
-    const payloads = await encodeToPayloads(this.dataConverter, result);
+    const dataConverter = this.context
+      ? withSerializationContext(this.dataConverter, this.context)
+      : this.dataConverter;
```
chris-olszewski: Might be worth making this snippet a getter to use in `complete`/`fail`/`reportCancellation`/`heartbeat`.

packages/worker/src/worker.ts
```diff
@@ -1078,6 +1084,14 @@ export class Worker {
                       this.options.namespace,
                       this.options.taskQueue
                     );
+                    contextDataConverter = withSerializationContext(this.options.loadedDataConverter, {
```
chris-olszewski: Do we need to be concerned about not reaching this line and using the data converter without context in the `catch`?

packages/workflow/src/workflow.ts
```diff
@@ -36,6 +36,10 @@ import { composeInterceptors } from '@temporalio/common/lib/interceptors';
 import { temporal } from '@temporalio/proto';
 import { deepMerge } from '@temporalio/common/lib/internal-workflow';
 import { throwIfReservedName } from '@temporalio/common/lib/reserved';
+import {
+  ActivitySerializationContext,
```
chris-olszewski:
```suggestion
  type ActivitySerializationContext,
```

packages/common/src/index.ts
THardy98: this actually shouldn't be exported, should be private

packages/common/src/converter/serialization-context.ts
THardy98: yup 👍

packages/common/src/converter/serialization-context.ts
```diff
@@ -0,0 +1,100 @@
+import type { LoadedDataConverter } from './data-converter';
+import type { FailureConverter } from './failure-converter';
+import type { PayloadCodec } from './payload-codec';
+import type { PayloadConverter } from './payload-converter';
+
+/**
+ * Context for workflow-level serialization operations.
+ */
+export interface WorkflowSerializationContext {
```
mjameswh, line 9: We have `workflowType?` in Activity context, but not here. Of course, that information is not always available, but when it is, shouldn't we also include it here?

Or put another way, how can that information be useful to serializers on an activity, but not needed on a workflow?

mjameswh, line 9: We should consider adding some form of discriminant on serialization contexts. Maybe a `type: 'workflow'`.

Two reasons:
- As it is now, the only way to determine whether the current serialization belongs to a workflow or an activity, and therefore which fields can be accessed, is to first do `"isLocal" in context` (because `isLocal` is the only field which is guaranteed to be set in one context but not he other).
- There will likely be other types of serialization context in the future, which may further complicate proper use.

packages/common/src/converter/serialization-context.ts
```diff
@@ -0,0 +1,100 @@
...
+  /** Namespace of the activity. */
+  namespace: string;
+  /** Activity ID for this execution when provided by the activity context source. */
+  activityId?: string;
+  /** Parent workflow ID when this activity is associated with a workflow. */
+  workflowId?: string;
+  /** Parent workflow type when this activity is associated with a workflow. */
+  workflowType?: string;
+  /** Whether the activity is a local activity started from a workflow. */
+  isLocal: boolean;
```
mjameswh, line 40: Thought: with that information on hand, that means we (i.e. the sdk, not the user), could now skip codecs on local activities's input args, as those are never transferred through the network/persisted to history anyway. That optimization was not possible until now.

packages/common/src/converter/serialization-context.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+export function withPayloadCodecContext(codec: PayloadCodec, context: SerializationContext): PayloadCodec {
+  return codec.withContext?.(context) ?? codec;
+}
+
+/**
+ * Return a loaded data converter where all components are context-bound when supported.
+ */
+// ts-prune-ignore-next (imported via lib/converter/serialization-context)
+export function withSerializationContext(
```
mjameswh, line 88: Can you please confirm that this function is not intended to be ever be used in workflow context?

packages/common/src/converter/serialization-context.ts
mjameswh: IIUC, this is only a memory optimization, right? If so, the fact that `===` might give false positive is not that much of a worry.

packages/common/src/converter/serialization-context.ts
THardy98: It's only used in `worker.ts` at the moment, if that's what you're asking.

It calls `withPayloadCodecContext` so necessarily I don't think it can be in workflow context (or is that what you're implying?)

packages/test/src/payload-converters/serialization-context-converter.ts
```diff
@@ -0,0 +1,107 @@
...
+  toPayload(value: unknown): Payload {
+    return defaultPayloadConverter.toPayload(appendTraceStep(value, `to:${contextTag(this.context)}`));
+  }
+
+  fromPayload<T>(payload: Payload): T {
+    const decoded = defaultPayloadConverter.fromPayload(payload);
+    return appendTraceStep(decoded, `from:${contextTag(this.context)}`) as T;
+  }
+
+  withContext(context: SerializationContext): PayloadConverter {
```
mjameswh, line 74: Kind of disappointing that users _must_ systematically write this function (which will basically always be the same code if they are context aware) and that they must make `toPayload` and `fromPayload` (or similar in other types of serializers) deal with the possibility of being called without a context, which we know should never happen if they support context and using a recent sdk.

packages/common/src/converter/serialization-context.ts
mjameswh: Yeah, that's what I means. I think we should explicitly state that assumption in the typedoc. That's an internal API only, and should be obvious to TS SDK maintainers, but can be easily missed by our colleagues that are less familiar with implications of the TS' main thread vs workerthreads design.

To clarify my concern: I _think_ calling this function from within the workflow sandbox would actually work without any immediate issue, but could result in inclusion of unnecessary code in the workflow bundle, thus increasing the memory footprint. That would be an easily missed bug, hence worth a warning.

packages/worker/src/worker.ts
```diff
@@ -1078,6 +1085,14 @@ export class Worker {
                       this.options.namespace,
                       this.options.taskQueue
                     );
+                    const context: ActivitySerializationContext = {
+                      namespace: info.workflowNamespace,
+                      activityId: info.activityId,
+                      ...(info.workflowExecution.workflowId ? { workflowId: info.workflowExecution.workflowId } : {}),
```
mjameswh: nit: Why the intermediate object+spread operator in this case (and `workflowType` below)? I don't think there's any reason to differentiate the "property set to undefined" vs "property not set at all" cases here, so you can take a shorter form.

packages/worker/src/worker.ts
```diff
@@ -1757,7 +1773,15 @@ export class Worker {
             let payload: Payload;
             try {
               try {
-                payload = await encodeToPayload(this.options.loadedDataConverter, details);
+                const context: ActivitySerializationContext = {
+                  namespace: info.workflowNamespace,
+                  activityId: info.activityId,
+                  ...(info.workflowExecution.workflowId ? { workflowId: info.workflowExecution.workflowId } : {}),
```
mjameswh: Same comment as before.

packages/worker/src/worker.ts
THardy98: The current behavior should be correct — if we fail, it's because either:
  1. IllegalStateError — we don't have meaningful context to apply
  2. extractActivityInfo threw — we don't have valid info to build context from

In both cases, context-free encoding is the only option since the information needed to construct the context doesn't exist yet

packages/common/src/converter/serialization-context.ts
THardy98: Forgot to respond to this - I agree, the usage makes this clear. I've added a type discriminant

THardy98: I was following the precedent set in Python where IIRC this wasn't the case (which in fact only has namespace and workflow_id). But I agree with your sentiment:

> Or put another way, how can that information be useful to serializers on an activity, but not needed on a workflow?

Added. I wonder if the concern was that workflow serialization context would only include fields that were certain to be populated.

packages/common/src/converter/serialization-context.ts
THardy98: Yeah that's a good point. Potentially worth filing a features issue for

packages/test/src/payload-converters/serialization-context-converter.ts
THardy98: > Kind of disappointing that users must systematically write this function (which will basically always be the same code if they are context aware)

Yeah, it is a bit annoying. That being said, users can also include their own app-specific context into there payload converters this way, so it's not necessarily the case that this will "always be the same code", though I do see your sentiment. In any case, I'm not sure I mind the 6-line boilerplate enough to deviate. But open to your thoughts

> ...that they must make toPayload and fromPayload (or similar in other types of serializers) deal with the possibility of being called without a context, which we know should never happen if they support context and using a recent sdk.

But it's exactly as you say, if they are on a recent sdk and want to support context, they don't need to handle being called without a context because it should always happen

PR #1951, opened 2026-02-26
feat(opentelemetry) add @temporalio/interceptors-opentelemetry-v2

### LINKED ISSUE #1658: [Feature Request] Add support for opentelemetry v2

chris-olszewski:
## What was changed
Add a package to provide interceptors that support OpenTelemetry v2.

First commit is a straight copy of the v1 with package bumps and the necessary changes the came along with those bumps.

Note: Histories created with the `@temporalio/interceptors-opentelemetry` will not be compatible with `@temporalio/interceptors-opentelemetry` and vice-versa. The crux of the issue is that OpenTelemetry changed their usage of `Math.random()`. The random number generator was shared by OpenTelemetry for generating span ids as well as the `uuid` function which is used for generating child workflow ids. This can cause NDE. To prevent future breakages from changes to `Math.random` usage, we now provide an `IdGenerator` to OpenTelemetry that uses its own random number generator. This doesn't offer a huge amount of value as we are still vulnerable to changes to the amount of calls to `Math.random` outside of id creation, but it does offer protection if the number of ids created changes.

Due to histories already being incompatible, I removed the yield point injections from the interceptors to have a cleaner starting point.

## Why?
This was done in a separate package to allow us to support both v1 and v2 without forcing users to upgrade.

## Checklist

1. Closes #1658 

2. How was this tested:
Copied over relevant tests and added an additional test to verify that replaying workflow history that has uuid call doesn't trigger NDE in the future. 

3. Any docs updates needed?


packages/interceptors-opentelemetry-v2/src/workflow/index.ts
```diff
@@ -0,0 +1,334 @@
...
+let contextManager: undefined | ContextManager = undefined;
+
+function getTracer(): otel.Tracer {
+  if (contextManager === undefined) {
+    contextManager = new ContextManager();
+  }
+  if (tracer === undefined) {
+    const provider = new tracing.BasicTracerProvider({
+      idGenerator: new DeterministicIdGenerator(),
+      spanProcessors: [new tracing.SimpleSpanProcessor(new SpanExporter())],
```
tconley1428: Allow customization.

packages/interceptors-opentelemetry-v2/src/plugin.ts
```diff
@@ -0,0 +1,77 @@
...
+export class OpenTelemetryPlugin extends SimplePlugin {
+  constructor(readonly otelOptions: OpenTelemetryPluginOptions) {
+    const workflowInterceptorsPath = require.resolve('./workflow-interceptors');
+    const interceptorOptions = otelOptions.tracer ? { tracer: otelOptions.tracer } : {};
+    super({
+      name: 'OpenTelemetryPlugin',
+      clientInterceptors: {
+        workflow: [new OpenTelemetryWorkflowClientInterceptor(interceptorOptions)],
+      },
+      workerInterceptors: {
```
mjameswh, line 42: We need to add nexus interceptors

contrib/interceptors-opentelemetry-v2/src/index.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @module
+ */
+
+export * from './plugin';
+export * from './workflow';
+export * from './worker';
+export {
+  OpenTelemetryWorkflowClientInterceptor,
+  /** deprecated: Use OpenTelemetryWorkflowClientInterceptor instead */
```
mjameswh, line 16: You can get rid of that legacy thing.

packages/interceptors-opentelemetry-v2/src/plugin.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A plugin that adds OpenTelemetry tracing.
+ *
+ * Configures Client, Activity, and Workflow interceptors for trace propagation and injects
+ * a span exporter sink for Workflow spans.
+ *
+ * @experimental Plugins is an experimental feature; APIs may change without notice.
+ */
+export class OpenTelemetryPlugin extends SimplePlugin {
```
mjameswh, line 33: Are we not injecting bundler options?

Btw, we now have a `preloadedModules` bundler option, which may help reduce the overhead associated with per-workflow loading of the OpenTelemetry library.

contrib/interceptors-opentelemetry-v2/src/workflow/workflow-imports-impl.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Real workflow imports for otel interceptors.
+ * This replaces the stub via webpack alias when bundled.
+ *
+ * @module
+ */
+export { inWorkflowContext, proxySinks, workflowInfo, AsyncLocalStorage, ContinueAsNew } from '@temporalio/workflow';
+export { alea, type RNG } from '@temporalio/workflow/lib/alea';
```
mjameswh, line 8: Use the new random streams api once it has landed.

contrib/interceptors-opentelemetry-v2/src/workflow/id-generator.ts
```diff
@@ -0,0 +1,54 @@
+import type { IdGenerator } from '@opentelemetry/sdk-trace-base';
+import { alea, type RNG } from './workflow-imports';
```
mjameswh, line 2: Use the new Random Streams API, once it has landed.

PR #1955, opened 2026-03-08
fix: respect pre-loaded LoadedDataConverter in Worker.create()

### LINKED ISSUE #1953: [Bug] `Worker.create()` ignores pre-loaded `LoadedDataConverter`, always re-loads via `require()`

We use a custom `PayloadConverter` that registers `instanceof`-based type transformers (e.g. `Decimal` via superjson). Our project runs under ESM (`--import=tsx`). We need `instanceof` checks in the payload converter to work against values created by activity code in the same process.

### Describe the bug

`Worker.create()` unconditionally calls `loadDataConverter()` on the `dataConverter` option, which uses `require()` to load the module at `payloadConverterPath`. This creates a separate CJS module cache from the ESM-imported application code. As a result, `instanceof` checks in custom payload converters silently fail -- classes loaded via `require()` are different objects from the same classes loaded via `import`.

The `Client` already handles this correctly by checking `isLoadedDataConverter()` before calling `loadDataConverter()`:

```ts
// @temporalio/client/src/base-client.ts
const dataConverter = options?.dataConverter ?? {};
this.loadedDataConverter = isLoadedDataConverter(dataConverter)
  ? dataConverter
  : loadDataConverter(dataConverter);
```

The Worker does not have this check:

```ts
// @temporalio/worker/src/worker-options.ts:1059
loadedDataConverter: loadDataConverter(opts.dataConverter),
```

The failure is completely silent -- `instanceof` returns `false`, the custom serializer is skipped, and values fall through to plain JSON serialization, losing their type information. There is no error or warning.

### Minimal Reproduction

See the three attached files for a self-contained reproduction using `@temporalio/testing`. Dependencies: `@temporalio/worker`, `@temporalio/testing`, `decimal.js`.

- **`bug-repro-payload-converter.ts`** -- a minimal custom `PayloadConverter` that detects `Decimal` values via `instanceof` and wraps them with format metadata for round-tripping.
- **`bug-repro-workflows.ts`** -- a minimal workflow that calls an activity returning a `Decimal`, then inspects the result type.
- **`bug-repro.test.ts`** -- two test cases:
  1. **"BUG: without workaround"** -- demonstrates that `Decimal` silently degrades to a plain string (`typeof result === 'string'`, `instanceof Decimal === false`, Decimal methods lost).
  2. **"WORKAROUND: replacing CJS converter"** -- demonstrates that mutating `worker.options.loadedDataConverter.payloadConverter` after `Worker.create()` fixes the issue.

Run with: `npx tsx --test bug-repro.test.ts`

**What happens:** The activity creates a `Decimal` using the ESM module (`decimal.mjs`). The Worker serializes it using the payload converter loaded via `require()`, where `instanceof Decimal` checks against the CJS class (`decimal.js`). These are different class objects:

```ts
import { Decimal as EsmDecimal } from 'decimal.js';   // decimal.mjs
const { Decimal: CjsDecimal } = require('decimal.js'); // decimal.js

EsmDecimal === CjsDecimal;                // false
new EsmDecimal('1') instanceof CjsDecimal; // false
```

So `instanceof` returns `false`, the custom serializer is skipped, and the `Decimal` is serialized as a plain JSON string. On the receiving side, calling `.isZero()` on the deserialized value throws `isZero is not a function`.

**Workaround:** Mutate the loaded converter after `Worker.create()`:

```ts
import { payloadConverter } from './payload-converter';

const worker = await Worker.create({ /* ... */ });
worker.options.loadedDataConverter.payloadConverter = payloadConverter;
```

### Environment/Versions

- OS and processor: M1 Mac (also reproducible on x86 Linux)
- SDK version: `@temporalio/worker` 1.15.0
- Using Temporal Cloud (but also reproducible with the local test server from `@temporalio/testing`)

#### Repro Test Files

- [bug-repro-payload-converter.ts](https://github.com/user-attachments/files/25780854/bug-repro-payload-converter.ts)
- [bug-repro-workflows.ts](https://github.com/user-attachments/files/25780853/bug-repro-workflows.ts)
- [bug-repro.test.ts](https://github.com/user-attachments/files/25780852/bug-repro.test.ts)

#### **Suggested fix** -- apply the same `isLoadedDataConverter` guard that the Client already uses:

```diff
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -1,4 +1,4 @@
-import { loadDataConverter } from '@temporalio/common/lib/internal-non-workflow';
+import { isLoadedDataConverter, loadDataConverter } from '@temporalio/common/lib/internal-non-workflow';

@@ -1056,7 +1056,9 @@
     defaultHeartbeatThrottleIntervalMs: msToNumber(opts.defaultHeartbeatThrottleInterval),
-    loadedDataConverter: loadDataConverter(opts.dataConverter),
+    loadedDataConverter: isLoadedDataConverter(opts.dataConverter)
+      ? opts.dataConverter
+      : loadDataConverter(opts.dataConverter),
     activities,
```

This lets users pass a pre-loaded `LoadedDataConverter` directly, bypassing `require()` and keeping the same module cache as the activity code:

```ts
import { payloadConverter } from './payload-converter';
import { defaultFailureConverter } from '@temporalio/common';

  // ...
  dataConverter: {
    payloadConverter,       // ESM-imported instance
    failureConverter: defaultFailureConverter,
    payloadCodecs: [],
  },
});
```

This is especially impactful for packages with separate CJS/ESM entry points (like `decimal.js`, `neverthrow`, etc.) where Node.js loads them into separate module caches, causing all `instanceof`-based type detection to silently break.

MaxwellCalkin:
## What was changed

`Worker.create()` now checks if the `dataConverter` option is already a pre-loaded `LoadedDataConverter` before calling `loadDataConverter()`. This applies the same `isLoadedDataConverter` guard that `BaseClient` already uses in `base-client.ts`.

Changes in `packages/worker/src/worker-options.ts`:
- Import `isLoadedDataConverter` alongside `loadDataConverter`
- Update `WorkerOptions.dataConverter` type to accept `DataConverter | LoadedDataConverter`
- Guard the `loadDataConverter()` call with `isLoadedDataConverter()` check

## Why?

`loadDataConverter()` uses `require()` to load converter modules. Under ESM (`--import=tsx`, etc.), `require()` creates a separate CJS module cache from `import`-loaded code. This means the same class (e.g. `Decimal` from `decimal.js`) exists as two different objects — one in the CJS cache used by the worker's converter, and one in the ESM cache used by activity code. All `instanceof`-based type detection silently breaks, with no error or warning.

The `Client` already handles this correctly:

```ts
// @temporalio/client/src/base-client.ts
this.loadedDataConverter = isLoadedDataConverter(dataConverter)
  ? dataConverter
  : loadDataConverter(dataConverter);
```

The Worker was missing this same guard.

## Checklist

1. Closes #1953
2. How was this tested: Verified the fix matches the pattern already used and tested in `base-client.ts`. The `isLoadedDataConverter` function checks for the `payloadConverter` property which is present on `LoadedDataConverter` but not on `DataConverter`.
3. Any docs updates needed? No

---

**Transparency note:** This PR was authored by Claude Opus 4.6, an AI model by Anthropic, operating under the direction of a human supervisor. See https://maxcalkin.com/ai for details.

### DIFF
--- a/packages/worker/src/worker-options.ts
+++ b/packages/worker/src/worker-options.ts
@@ -11,7 +11,7 @@ import {
   WorkerDeploymentVersion,
 } from '@temporalio/common';
 import { Duration, msOptionalToNumber, msToNumber } from '@temporalio/common/lib/time';
-import { loadDataConverter } from '@temporalio/common/lib/internal-non-workflow';
+import { isLoadedDataConverter, loadDataConverter } from '@temporalio/common/lib/internal-non-workflow';
 import { LoggerSinks } from '@temporalio/workflow';
 import { Context } from '@temporalio/activity';
 import { native } from '@temporalio/core-bridge';
@@ -166,12 +166,16 @@ export interface WorkerOptions {
   shutdownForceTime?: Duration;
 
   /**
-   * Provide a custom {@link DataConverter}.
+   * Provide a custom {@link DataConverter} or a pre-loaded {@link LoadedDataConverter}.
    *
    * When bundling workflows ahead of time, make sure to provide custom payload and failure
    * converter paths as options to `bundleWorkflowCode`.
+   *
+   * Passing a pre-loaded {@link LoadedDataConverter} avoids the `require()`-based loading
+   * path, which is useful when running under ESM where `require()` creates a separate module
+   * cache from `import`, causing `instanceof` checks to silently fail.
    */
-  dataConverter?: DataConverter;
+  dataConverter?: DataConverter | LoadedDataConverter;
 
   /**
    * Provide a custom {@link WorkerTuner}.
@@ -1056,7 +1060,9 @@ export function compileWorkerOptions(
     isolateExecutionTimeoutMs: msToNumber(opts.isolateExecutionTimeout),
     maxHeartbeatThrottleIntervalMs: msToNumber(opts.maxHeartbeatThrottleInterval),
     defaultHeartbeatThrottleIntervalMs: msToNumber(opts.defaultHeartbeatThrottleInterval),
-    loadedDataConverter: loadDataConverter(opts.dataConverter),
+    loadedDataConverter: isLoadedDataConverter(opts.dataConverter)
+      ? opts.dataConverter
+      : loadDataConverter(opts.dataConverter),
     activities,
     nexusServiceRegistry: nexusServiceRegistryFromOptions(opts),
     enableNonLocalActivities: opts.enableNonLocalActivities && activities.size > 0,


mjameswh:
Hey @MaxwellCalkin. Thanks for this submission.

Two concerns:

1. We can't review at your PR until you've agreed to the CLA. Please take at the message posted by [CLAassistant](https://github.com/temporalio/sdk-typescript/pull/1955#issuecomment-4019633481) above. Make sure that the GitHub account you use to complete the process corresponds to the same email address as the one recorded on the git commits.

2. Regarding [the underlying feature request](#1953), I have to clarify that the fact that `Worker` does not currently accept a `LoadedDataConverter` is _not_ an oversight. It is by design because _Payload Converters_ and _Failure Converters_ (two of the three components of Data Converters) must execute inside the Workflow sandbox context. It is simply not possible to pass live objects (including modules) from the main thread to the workflow sandbox, and therefore, Payload Converters and Failure Converters must be injected into the sandbox as modules to be loaded.

    We could maybe consider accepting `LoadedDataConverter` in `WorkerOptions` in the case of a Worker that only runs Activities and/or Nexus Services (i.e. no Workflow), or in the case of a Workflow Worker that is given an existing workflow bundle (i.e. not using the `workflowsPath` property). I suppose we could also consider some form of "dual data converter" object, that would include both live objects and file paths for _Payload Converters_ and _Failure Converters_. We'll however need to discuss and validate that design internally before we commit to that.

mjameswh:
We discussed this internally, and decided to reject this PR.

We want to reconsider the whole workflow bundling/sandboxing experience in the near future. Fixing the way that data converters are provided will be part of that larger effort.

PR #1959, opened 2026-03-10
Update sdk-core

VegetarianOrc:
## What was changed

Update the sdk-core submodule and adapt the TypeScript SDK to breaking changes in the core API: replace RetryClient with Connection, update error types (ClientInitError → ClientConnectError), wrap search attributes in { indexedFields: ... }, and update codec runner and tests.

## Why?

Upcoming changes changes to Nexus error serialization requires a newer version of core.

## Checklist

1. How was this tested:

Existing tests run.

### DIFF
--- a/packages/core-bridge/src/client.rs
+++ b/packages/core-bridge/src/client.rs
@@ -261,541 +261,160 @@ async fn client_invoke_workflow_service(
     use temporalio_client::grpc::WorkflowService;
 
     match call.rpc.as_str() {
-        "CountActivityExecutions" => {
-            rpc_call!(
-                connection,
-                call,
-                count_activity_executions,
-                workflow_service
-            )
-        }
-        "CountSchedules" => {
-            rpc_call!(connection, call, count_schedules, workflow_service)
-        }
-        "CountWorkflowExecutions" => {
-            rpc_call!(
-                connection,
-                call,
-                count_workflow_executions,
-                workflow_service
-            )
-        }
-        "CreateSchedule" => {
-            rpc_call!(connection, call, create_schedule, workflow_service)
-        }
-        "CreateWorkflowRule" => {
-            rpc_call!(connection, call, create_workflow_rule, workflow_service)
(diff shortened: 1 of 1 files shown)


packages/core-bridge/src/client.rs
```diff
@@ -240,11 +233,16 @@ impl TryIntoJs for tonic::Status {
 }
 
 macro_rules! rpc_call {
-    ($retry_client:ident, $call:ident, $call_name:ident) => {
+    ($connection:ident, $call:ident, $call_name:ident, $service_accessor:ident) => {
```
chris-olszewski, line 239: I know we have AI to easily bang this out now, but may as well default to `workflow_service` since that's by far the most common service we call.
```suggestion
    ($connection:ident, $call:ident, $call_name:ident) => {
        rpc_call!($connection, $call, $call_name, workflow_service)
    };
    ($connection:ident, $call:ident, $call_name:ident, $service_accessor:ident) => {
```

packages/workflow/src/workflow.ts
```diff
@@ -402,7 +402,7 @@ function startChildWorkflowExecutionNextHandler({
         cronSchedule: options.cronSchedule,
         searchAttributes:
           options.searchAttributes || options.typedSearchAttributes // eslint-disable-line @typescript-eslint/no-deprecated
-            ? encodeUnifiedSearchAttributes(options.searchAttributes, options.typedSearchAttributes) // eslint-disable-line @typescript-eslint/no-deprecated
+            ? { indexedFields: encodeUnifiedSearchAttributes(options.searchAttributes, options.typedSearchAttributes) } // eslint-disable-line @typescript-eslint/no-deprecated
```
chris-olszewski, line 405: Note to self: now all calls of `encodeUnifiedSearchAttributes` have the form of `{ indexedFields: encodeUnifiedSearchAttributes(...) }`. It might make sense to move this into the encoding function

PR #1963, opened 2026-03-11
Allow WorkerDeploymentOptions without defaultVersioningBehavior when useWorkerVersioning is false

### LINKED ISSUE #1962: [Bug] Unable to set `defaultVersioningBehavior` to `UNSPECIFIED` in `workerDeploymentOptions`

I want to attach a build id to workers just so we can see what version is running in the temporal UI but I dont want to use all of the worker/workflow versioning stuff yet.

### Describe the bug

`defaultVersioningBehavior` only allows auto upgrade and pinned, not unspecified, which is supported by the server.
https://temporalio.slack.com/archives/C01DKSMU94L/p1773250476620249

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]

mjameswh:
## Summary

- Make `WorkerDeploymentOptions` a discriminated union so that `defaultVersioningBehavior` is required when `useWorkerVersioning` is `true` but must be absent when `useWorkerVersioning` is `false`
- This lets users attach a build ID to workers for visibility in the Temporal UI without opting into the full worker versioning feature
- Update the native bridge (TypeScript and Rust) to propagate `None`/`null` for `defaultVersioningBehavior` when versioning is disabled

Fixes #1962

## Test plan

- [x] Added compile-time type tests verifying that `defaultVersioningBehavior` is required with `useWorkerVersioning: true` and rejected with `useWorkerVersioning: false`
- [x] Existing `test-worker-deployment-versioning` integration tests pass (6/6)
- [x] Full build passes (TypeScript + Rust)
- [x] Verify with a live Temporal server that a worker with `useWorkerVersioning: false` and a build ID correctly reports to the UI without causing workflow hangs

### DIFF
--- a/packages/common/src/internal-non-workflow/codec-helpers.ts
+++ b/packages/common/src/internal-non-workflow/codec-helpers.ts
@@ -357,6 +357,17 @@ export function noopEncodeMap<K extends string>(
   return map as Record<K, EncodedPayload> | null | undefined;
 }
 
+export function noopEncodeSearchAttrs<K extends string>(
+  attrs: temporal.api.common.v1.ISearchAttributes | null | undefined
+): temporal.api.common.v1.ISearchAttributes | null | undefined {
+  if (!attrs) {
+    return attrs;
+  }
+  return {
+    indexedFields: noopEncodeMap(attrs.indexedFields),
+  };
+}
+
 /**
  * Mark all values in the map as decoded.
  * Use this for headers, which we don't encode.
--- a/packages/core-bridge/Cargo.lock
+++ b/packages/core-bridge/Cargo.lock
@@ -373,6 +373,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs"
+version = "6.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3e8aa94d75141228480295a7d0e7feb620b1a5ad9f12bc40be62411e38cce4e"
+dependencies = [
+ "dirs-sys",
+]
+
+[[package]]
+name = "dirs-sys"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e01a3366d27ee9890022452ee61b2b63a67e6f13f58900b651ff5665f0bb1fab"
+dependencies = [
+ "libc",
+ "option-ext",
+ "redox_users",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "displaydoc"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -474,3 +477,52 @@ async function setCurrentDeploymentVersion(
...
+    useWorkerVersioning: false,
+  };
+
+  const invalidWithBehavior: WorkerDeploymentOptions = {
+    version: { buildId: '1.0', deploymentName: 'my-deployment' },
+    useWorkerVersioning: false,
+    // @ts-expect-error defaultVersioningBehavior must not be specified when useWorkerVersioning is false
+    defaultVersioningBehavior: 'AUTO_UPGRADE',
+  };
+
```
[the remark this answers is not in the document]
mjameswh: Ignoring, really not worth it.

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -474,3 +477,86 @@ async function setCurrentDeploymentVersion(
...
+  // @ts-expect-error defaultVersioningBehavior is required when useWorkerVersioning is true
+  const missingBehavior: WorkerDeploymentOptions = {
+    version: { buildId: '1.0', deploymentName: 'my-deployment' },
+    useWorkerVersioning: true,
+  };
+
+  t.pass();
+});
+
+test('WorkerDeploymentOptions with useWorkerVersioning false does not allow defaultVersioningBehavior', (t) => {
```
chris-olszewski, line 542: I've been using `skip` for tests that just want to exercise typechecking. Any preference on this vs having a no-op test?

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -474,3 +477,86 @@ async function setCurrentDeploymentVersion(
...
+    connection: nativeConnection,
+  });
+
+  const handle = await client.workflow.start('successString', {
+    taskQueue,
+    workflowId: 'versioning-off-build-id-' + randomUUID(),
+  });
+
+  await worker.runUntil(handle.result());
+  assert.equal(await handle.result(), 'success');
```
chris-olszewski: Reason for using `assert` over `t.equal` or was that just Claude being different? No need for `t.pass` if we use AVA's assertions.

mjameswh: Good catch, it's indeed just something claude did without reason, and I missed it.

packages/test/src/test-worker-deployment-versioning.ts
mjameswh: No strong opinion, really, but if I have to choose, I'd lean toward not marking those as `skip`, only because ava would count those in the number of "skipped" tests that it reports after execution.

PR #1968, opened 2026-03-16
Infer Workflow Result Type from WorkflowRunOperationHandler

VegetarianOrc:
## What was changed

`startWorkflow` now returns WorkflowHandle<WorkflowResultType<T>> instead of WorkflowHandle<T>, so the handler's output type is inferred as the workflow's return type (e.g. string) rather than the workflow function type. A new workflowResultType brand on WorkflowHandle makes WorkflowHandle<X> structurally distinct from WorkflowHandle<Y>, enabling TypeScript to catch type mismatches when explicit type params contradict the actual workflow.

## Why?

Previously the `WorkflowRunOperationHandler` was inferring the return type to be the workflow instead of the workflow's output type. This required you to explicitly type the `WorkflowRunOperationHandler` in order to fulfill Nexus service contracts.

## Checklist

1. How was this tested:

A test with the different combinations of typing has been added to assert compilation errors and correct inferrence.


packages/nexus/src/workflow-helpers.ts
```diff
@@ -123,6 +139,11 @@ export type WorkflowRunOperationStartHandler<I, O> = (
 /**
  * A Nexus Operation implementation that is backed by a Workflow run.
  *
+ * The type parameter `O` represents the operation's output type. When the handler uses
```
mjameswh: Let's skip that comment.

packages/test/src/test-nexus-handler.ts
```diff
@@ -790,6 +790,90 @@ test('WorkflowRunOperationHandler does not accept WorkflowHandle from WorkflowCl
...
 });
 
+export async function echoWorkflow(input: string): Promise<string> {
+  return input;
+}
+
+test('WorkflowRunOperationHandler infers correct output type from typed workflow function', async (t) => {
+  // When constructing WorkflowRunOperationHandler without explicit type parameters using a typed
+  // workflow function, the operation output type should be inferred as the workflow's return type
+  // (e.g. string), not the workflow function type (e.g. (input: string) => Promise<string>).
```
mjameswh: The last part of this comment will be unhelpful to future readers. It puts emphasis on what will then be a former bug, actually creating confusion.

Not totally convinced of the pertinence of the first two lines, but there's not doing any harm.

```suggestion
```

packages/nexus/src/workflow-helpers.ts
```diff
@@ -31,6 +36,17 @@ export interface WorkflowHandle<_T> {
...
+  /**
+   * Type brand that carries the workflow's result type, making `WorkflowHandle<X>` structurally
+   * distinct from `WorkflowHandle<Y>` so TypeScript can catch type mismatches.
+   *
+   * @internal
+   * @hidden
+   *
+   * @experimental Nexus support in Temporal SDK is experimental.
+   */
+  readonly [workflowResultType]: T;
```
bergundy, line 49: Seems like a better structure IMHO.

```suggestion
  readonly [workflowResultType]: WorkflowResult<T>;
```

VegetarianOrc: I explored this option, but it actually conflicts with explicit typing. Including that restricts T to `T extends Workflow` which has knock on effects that cause any explicitly provided type parameters to also extend Workflow. For example:

```ts  
// Explicit type parameters should also work correctly.
  const _explicitStringOp: nexus.OperationHandler<string, string> = new temporalnexus.WorkflowRunOperationHandler<
    string,
    string
  >(async (ctx, input) => {
    return await temporalnexus.startWorkflow(ctx, echoWorkflow, {
      args: [input],
      workflowId: 'test',
    });
  });
```

Instead would need to be 

```ts  
// Explicit type parameters should also work correctly.
  const _explicitStringOp: nexus.OperationHandler<string, string> = new temporalnexus.WorkflowRunOperationHandler<
    string,
    typeof echoWorkflow
  >(async (ctx, input) => {
    return await temporalnexus.startWorkflow(ctx, echoWorkflow, {
      args: [input],
      workflowId: 'test',
    });
  });
```

VegetarianOrc:
@Evanthx 
> Should there be unit tests for these? Or are they tested in some other way?

There are some existing tests for the functionality of the `WorkflowRunOperationHandler`. In this PR I included some compile time tests to ensure that patterns expected to fail do and patterns expected to succeed do not produce compile time errors.

PR #1972, opened 2026-03-17
Upgrade on CAN

THardy98:
## What was changed
Propagate `targetWorkerDeploymentVersionChanged` and CAN suggested reasons

## Why?
TS support for version trampolining

## Checklist

1. Part of https://github.com/temporalio/features/issues/738

2. How was this tested:
Integration tests

3. Any docs updates needed?
Probably


packages/test/src/test-integration-split-one.ts
```diff
@@ -763,6 +763,10 @@ test.serial('Workflow can read WorkflowInfo', configMacro, async (t, config) =>
     workflowId: handle.workflowId,
     historyLength: 3,
     continueAsNewSuggested: false,
+    targetWorkerDeploymentVersionChanged: false,
+    ...(res.suggestedContinueAsNewReasons != null && {
+      suggestedContinueAsNewReasons: res.suggestedContinueAsNewReasons,
+    }),
     // values ignored for the purpose of comparison
     historySize: res.historySize,
```
chris-olszewski, line 768: We can move `suggestedContinueAsNewReasons` to the section for values we don't compare. Also I'm not sure the reasoning for the conditional setting, `suggestedContinueAsNewReasons` should never be `null`.
```suggestion
    targetWorkerDeploymentVersionChanged: false,
    // values ignored for the purpose of comparison
    suggestedContinueAsNewReasons: res.suggestedContinueAsNewReasons,
    historySize: res.historySize,
```

packages/test/src/test-nexus-codec-converter-errors.ts
```diff
@@ -108,6 +108,5 @@ test('Nexus operation converter failure is not retried', async (t) => {
     const handlerError = nexusFailure.cause as nexus.HandlerError;
     t.is(handlerError.type, 'BAD_REQUEST');
     t.false(handlerError.retryable);
-    t.regex(handlerError.message, /Intentional payload converter failure for testing/);
```
chris-olszewski, line 111: I forget if this assertion is going to come back with @VegetarianOrc's changes or not. If it will be updated/refined, then it might be useful to leave a comment about when we can add it back.

packages/test/src/test-worker-deployment-versioning.ts
```diff
@@ -473,6 +473,135 @@ async function setCurrentDeploymentVersion(
...
+) {
+  return await asyncRetry(
+    async () => {
+      const resp = await client.workflowService.describeWorkerDeployment({
+        namespace: client.options.namespace,
+        deploymentName,
+      });
+      const routingConfig = resp.workerDeploymentInfo?.routingConfig;
+      const currentBuildId = routingConfig?.currentDeploymentVersion?.buildId ?? '';
+      const rampingBuildId = routingConfig?.rampingDeploymentVersion?.buildId ?? '';
```
chris-olszewski: Slightly against just defaulting to `''` here in case we ever need to identify between a server sent `''` and our default. (realize this is just a test helper)
```suggestion
      const currentBuildId = routingConfig?.currentDeploymentVersion?.buildId;
      const rampingBuildId = routingConfig?.rampingDeploymentVersion?.buildId;
```

PR #1973, opened 2026-03-18
💥 Nexus Error Serialization Changes

VegetarianOrc:
## What was changed
  - **Failure converter updates** (`packages/common/src/converter/failure-converter.ts`):
    - `failureToError`: Nexus `HandlerError` now preserves the original `ProtoFailure` via `originalFailure` and resolves error types through the `HandlerErrorType` enum instead of raw string casting.
    - `errorToFailure`: When a `HandlerError` carries an `originalFailure`, it is restored directly to a `ProtoFailure` without re-encoding.
    - Added `nexusFailureToTemporalFailure` and `temporalFailureToNexusFailure` private helpers.

  - **Worker Nexus conversions** (`packages/worker/src/nexus/conversions.ts`):
    - `handlerErrorToProto` and `operationErrorToProto` now return `ProtoFailure` directly via `encodeErrorToFailure`.
    - Fixed bug in gRPC status code switch statement where comma-separated `case` values silently matched only the last value.
    - `coerceToHandlerError` now provides a default message for non-retryable `ApplicationFailure`.

  - **Task completion responses** (`packages/worker/src/nexus/index.ts`, `packages/worker/src/worker.ts`):
    - Changed from `error`/`operationError` fields to `failure` fields in Nexus task completion protos.
    - `requestId` is now required; throws `IllegalStateError` if missing instead of falling back to `undefined`.
    - Threads `requestDeadline` from `IPollNexusTaskQueueResponse` into `nexus.OperationContext`.

  - **Test changes**:
    - Migrated Nexus handler HTTP-based tests to workflow-caller integration tests (`test-nexus-workflow-caller.ts`).
    - Added `helpers-nexus.ts` with `unwrapHandlerErrorCause` and `innermostHandlerError` for old/new server compatibility.
    - Added unit tests for `coerceToHandlerError` (all gRPC status codes) and `operationErrorToProto` in `test-nexus-handler.ts`.
    - Added round-trip unit test for `HandlerError` through the failure converter.
    - Updated error assertions across all Nexus test files to match the new serialization format.

## Why?

  Nexus error serialization now uses Temporal ProtoFailure end-to-end, letting the TypeScript SDK pass failures directly rather than converting to/from Nexus-specific proto types. Core handles backward compatibility for Nexus completions with older servers. This also fixes a JS bug where case (a, b, c): in switch statements silently evaluated only c via the comma operator, causing incorrect gRPC status code mapping.

## Checklist

How was this tested:

  - New and updated integration tests in `test-nexus-workflow-caller.ts` that exercise error serialization end-to-end through workflow executions
  - New unit tests for `coerceToHandlerError` covering all gRPC status codes, `ApplicationFailure`, and passthrough of `HandlerError`
  - New unit tests for `operationErrorToProto` covering failed, canceled, and no-cause states
  - New round-trip unit test for `HandlerError` through the failure converter via `originalFailure`
  - Updated `test-nexus-handler.ts`, `test-nexus-codec-converter-errors.ts`, and `test-workflow-nexus-cancellation.ts`

### DIFF
--- a/packages/test/src/test-nexus-workflow-caller.ts
+++ b/packages/test/src/test-nexus-workflow-caller.ts
@@ -80,49 +80,49 @@ const clientOperationTypeSafetyCheckerService = nexus.service('typeSafetyService
 // Caller workflows
 
 export async function syncOpCaller(endpoint: string): Promise<string> {
-  const client = workflow.createNexusClient({ endpoint, service: syncOpService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: syncOpService });
   return await client.executeOperation('testSyncOp', 'hello');
 }
 
 export async function cancelSyncOpCaller(endpoint: string): Promise<void> {
-  const client = workflow.createNexusClient({ endpoint, service: blockingOpService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: blockingOpService });
   return await client.executeOperation('blockingOp', undefined, {
     cancellationType: 'TRY_CANCEL',
   });
 }
 
 export async function errorOpCaller(endpoint: string, outcome: string): Promise<string> {
-  const client = workflow.createNexusClient({ endpoint, service: errorOpService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: errorOpService });
   return await client.executeOperation('op', outcome);
 }
 
 export async function asyncOpCaller(endpoint: string): Promise<string> {
-  const client = workflow.createNexusClient({ endpoint, service: asyncOpService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: asyncOpService });
   return await client.executeOperation('asyncOp', 'hello');
 }
 
 export async function loggerOpCaller(endpoint: string): Promise<string> {
-  const client = workflow.createNexusClient({ endpoint, service: loggerService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: loggerService });
   return await client.executeOperation('loggerOp', 'hello');
 }
 
 export async function getClientCaller(endpoint: string): Promise<boolean> {
-  const client = workflow.createNexusClient({ endpoint, service: getClientService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: getClientService });
   return await client.executeOperation('getClientOp', undefined);
 }
 
 export async function operationInfoCaller(endpoint: string): Promise<{ namespace: string; taskQueue: string }> {
-  const client = workflow.createNexusClient({ endpoint, service: operationInfoService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: operationInfoService });
   return await client.executeOperation('operationInfoOp', undefined);
 }
 
 export async function linkCallbackCaller(endpoint: string): Promise<void> {
-  const client = workflow.createNexusClient({ endpoint, service: linkCallbackService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: linkCallbackService });
   return await client.executeOperation('startTargetWorkflow', undefined);
 }
 
 export async function cancelAppFailureCaller(endpoint: string): Promise<void> {
-  const client = workflow.createNexusClient({ endpoint, service: cancelErrorService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: cancelErrorService });
   try {
     await client.executeOperation('cancelThrowsAppFailure', undefined, {
       cancellationType: 'WAIT_CANCELLATION_REQUESTED',
@@ -134,7 +134,7 @@ export async function cancelAppFailureCaller(endpoint: string): Promise<void> {
 }
 
 export async function cancelHandlerErrorCaller(endpoint: string): Promise<void> {
-  const client = workflow.createNexusClient({ endpoint, service: cancelErrorService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: cancelErrorService });
   try {
     await client.executeOperation('cancelThrowsHandlerError', undefined, {
       cancellationType: 'WAIT_CANCELLATION_REQUESTED',
@@ -146,12 +146,12 @@ export async function cancelHandlerErrorCaller(endpoint: string): Promise<void> 
 }
 
 export async function requestDeadlineStartCaller(endpoint: string): Promise<boolean> {
-  const client = workflow.createNexusClient({ endpoint, service: requestDeadlineService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: requestDeadlineService });
   return await client.executeOperation('checkDeadlineOnStart', undefined);
 }
 
 export async function requestDeadlineCancelCaller(endpoint: string): Promise<void> {
-  const client = workflow.createNexusClient({ endpoint, service: requestDeadlineService });
+  const client = workflow.createNexusServiceClient({ endpoint, service: requestDeadlineService });
   try {
     await client.executeOperation('checkDeadlineOnCancel', undefined, {
       cancellationType: 'WAIT_CANCELLATION_REQUESTED',
@@ -168,7 +168,7 @@ export async function multiOpCaller(
   action: string,
   cancellationType?: workflow.NexusOperationCancellationType
 ): Promise<string> {
-  const client = workflow.createNexusClient({
+  const client = workflow.createNexusServiceClient({
     endpoint,
     service: multiOpService,
   });
@@ -182,7 +182,7 @@ export async function multiOpCaller(
 }
 
 export async function callNonExistentService(endpoint: string): Promise<string> {
-  const client = workflow.createNexusClient({
+  const client = workflow.createNexusServiceClient({
     endpoint,
     service: nonExistentService,
   });
@@ -192,7 +192,7 @@ export async function callNonExistentService(endpoint: string): Promise<string> 
 export async function clientOperationTypeSafetyCheckerWorkflow(endpoint: string): Promise<void> {
   const Service = clientOperationTypeSafetyCheckerService;
   const operations = Service.operations;
-  const client = workflow.createNexusClient({
+  const client = workflow.createNexusServiceClient({
     endpoint,
     service: Service,
   });
@@ -976,7 +976,7 @@ test('inbound cancelOperation interceptor can modify input', async (t) => {
   }
 });
 
-test('NexusClient is type-safe in regard to Operation Definitions', async (t) => {
+test('NexusServiceClient is type-safe in regard to Operation Definitions', async (t) => {
   const { createWorker, executeWorkflow, registerNexusEndpoint } = helpers(t);
   const { endpointName, endpointIdentifier } = await registerNexusEndpoint();
   try {
--- a/packages/test/src/test-nexus-codec-converter-errors.ts
+++ b/packages/test/src/test-nexus-codec-converter-errors.ts
@@ -17,7 +17,7 @@ const testService = nexus.service('codec-converter-test', {
 });
 
 export async function nexusEchoCaller(endpoint: string): Promise<string> {
-  const client = workflow.createNexusClient({
+  const client = workflow.createNexusServiceClient({
     endpoint,
     service: testService,
   });
--- a/packages/test/src/test-nexus-operation-timeouts.ts
+++ b/packages/test/src/test-nexus-operation-timeouts.ts
@@ -22,7 +22,7 @@ const scheduleToStartService = nexus.service('nexus-schedule-to-start-timeout-te
 });
 
 export async function scheduleToStartTimeoutCallerWorkflow(endpoint: string): Promise<string> {
-  const client = workflow.createNexusClient({
+  const client = workflow.createNexusServiceClient({
     endpoint,
     service: scheduleToStartService,
   });
@@ -72,7 +72,7 @@ const startToCloseService = nexus.service('nexus-start-to-close-timeout-test-ser
 });
 
 export async function startToCloseTimeoutCallerWorkflow(endpoint: string): Promise<string> {
-  const client = workflow.createNexusClient({
+  const client = workflow.createNexusServiceClient({
     endpoint,
     service: startToCloseService,
   });
--- a/packages/test/src/test-workflow-nexus-cancellation.ts
+++ b/packages/test/src/test-workflow-nexus-cancellation.ts
@@ -44,7 +44,7 @@ export async function cancellationTestCallerWorkflow(
 ): Promise<void> {
   const { cancellationType, operationName: nexusOperationName } = scenario;
   try {
-    const client = workflow.createNexusClient({ endpoint, service });
+    const client = workflow.createNexusServiceClient({ endpoint, service });
     await client.executeOperation(nexusOperationName, scenario, { cancellationType });
     throw ApplicationFailure.nonRetryable('Unexpected Success');
   } catch (err) {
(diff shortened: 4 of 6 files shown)


packages/common/src/converter/failure-converter.ts
```diff
@@ -238,15 +239,18 @@ export class DefaultFailureConverter implements FailureConverter {
...
-        // TODO(nexus/error): Maybe set a default message here, once we've decided on error handling.
-        failure.message ?? undefined,
-        {
-          cause: this.optionalFailureToOptionalError(failure.cause, payloadConverter),
-          retryableOverride,
-        }
-      );
+      const rawErrorType = failure.nexusHandlerFailureInfo.type || '';
+      const resolvedType: nexus.HandlerErrorType =
+        rawErrorType in nexus.HandlerErrorType
```
bergundy: Same as the nexus-rpc PR, use `Object.hasOwn` instead.

packages/common/src/converter/failure-converter.ts
```diff
@@ -469,4 +472,37 @@ export class DefaultFailureConverter implements FailureConverter {
   optionalErrorToOptionalFailure(err: unknown, payloadConverter: PayloadConverter): ProtoFailure | undefined {
     return err ? this.errorToFailure(err, payloadConverter) : undefined;
   }
+
+  private nexusFailureToTemporalFailure(failure: nexus.Failure, retryable: boolean): ProtoFailure {
+    if (failure.metadata?.type === 'temporal.api.failure.v1.Failure') {
+      return failure.details as ProtoFailure;
```
Evanthx: Should you add a check to see if details is undefined? Not sure how that would surface back up or if it's needed, but thought it worth asking

packages/worker/src/nexus/conversions.ts
```diff
@@ -69,79 +73,30 @@ class PayloadSerializer implements nexus.Serializer {
...
-    // Create an error without capturing a stack trace.
-    const wrapped = Object.create(ApplicationFailure.prototype);
-    wrapped.message = err.message;
-    wrapped.stack = err.stack;
-    wrapped.nonRetryable = true;
-    cause = wrapped;
+): Promise<ProtoFailure> {
+  let newError: Error;
+  if (err.state === 'canceled') {
+    newError = new CancelledFailure(err.message, undefined, err.cause);
```
Evanthx, line 52: Not an issue you introduced but I was reading some code when looking at this and CancelledFailure is defined twice in packages/client/src - once in index.ts and a second time in workflow-client.ts. So that feels like an issue waiting to happen!

packages/worker/src/nexus/conversions.ts
```diff
@@ -151,24 +106,30 @@ export function coerceToHandlerError(err: unknown): nexus.HandlerError {
...
+    });
   }
 
   if (err instanceof ServiceError) {
     if (isGrpcServiceError(err.cause)) {
       switch (err.cause.code) {
         case status.INVALID_ARGUMENT:
           return new nexus.HandlerError('BAD_REQUEST', undefined, { cause: err });
-        case (status.ALREADY_EXISTS, status.FAILED_PRECONDITION, status.OUT_OF_RANGE):
+        case status.ALREADY_EXISTS:
```
Evanthx, line 90: I find logic like this (and the lines following) is usually better in the relevant class - ie, nexus.HandlerError in this case - if possible. What happens over time is that you make a change to HandlerError to add some cases, but there is no good way to track down every place using HandlerError, so not every place gets updated. But if you are in HandlerError making changes, you'll see the method right there and make the relevant changes.

Quinn-With-Two-Ns: This logic is Temporal specific, not relevant to `HandlerError` which is a generic Nexus error. That being said if we do one day inline the Nexus SDK into temporal then it would be a good idea to move it closer, but we haven't committed to that yet

packages/worker/src/nexus/conversions.ts
VegetarianOrc: I could be wrong here, but I only see one definition of CancelledFailure from `packages/common/src/failure.ts`. It's imported and re-exported elsewhere but I think that's the only definition.

Evanthx: Gah, you're right. Chalk this one up to my being not that familiar with Typescript - I was reading code and using the editor to go to the definition of some variables and also searching for some to see how they were used and ended up looking at both, and then confused myself!

packages/worker/src/nexus/conversions.ts
```diff
@@ -151,24 +106,30 @@ export function coerceToHandlerError(err: unknown): nexus.HandlerError {
...
+        case status.ABORTED:
+        case status.UNAVAILABLE:
           return new nexus.HandlerError('UNAVAILABLE', undefined, { cause: err });
-        case (status.CANCELLED,
-        status.DATA_LOSS,
-        status.INTERNAL,
-        status.UNKNOWN,
-        status.UNAUTHENTICATED,
-        status.PERMISSION_DENIED):
+        case status.CANCELLED:
```
Quinn-With-Two-Ns, line 97: Is there a reason this was changed? Just curious

VegetarianOrc: This was actually a bug that Claude found where the case statement was only applied to the last statement since the parens cause the expression to evaluate to just the last value.

```js
function switcher1(input) {
    switch (input) {
        case (1, 2, 3):
            console.log("123");
            break;
        default:
            console.log("456");
            break;
    }
}

function switcher2(input) {
    switch (input) {
        case 1:
        case 2:
        case 3:
            console.log("123");
            break;
        default:
            console.log("456");
            break;
    }
}

console.log("in parens")
switcher1(1);
switcher1(2);
switcher1(3);
console.log("\nas case statements")
switcher2(1);
switcher2(2);
switcher2(3);
```

```sh
>node tmp.js
in parens
456
456
123

as case statements
123
123
123
```

packages/common/src/converter/failure-converter.ts
bergundy: We already check `!failure.details` above I see, but perhaps a `failure.details != null` check is a bit more precise here.

packages/test/src/test-nexus-workflow-caller.ts
```diff
@@ -48,13 +253,451 @@ export async function handler(action: string): Promise<string> {
...
+        () => executeWorkflow(errorOpCaller, { args: [endpointName, 'NonRetryableApplicationFailure'] }),
+        { instanceOf: WorkflowFailedError }
+      );
+      t.true(
+        err instanceof WorkflowFailedError &&
+          err.cause instanceof NexusOperationFailure &&
+          err.cause.cause instanceof nexus.HandlerError &&
+          err.cause.cause.type === 'INTERNAL'
+      );
+      if (
```
bergundy: Is this necessary given that `t.true` will assert that the condition is true? https://github.com/avajs/ava/blob/02c63fe2a7f9aff60aece616c556d35118f50fb6/types/assertions.d.cts#L380

packages/worker/src/worker.ts
```diff
@@ -1305,7 +1306,8 @@ export class Worker {
   private async handleNexusRunTask(
     task: temporal.api.workflowservice.v1.IPollNexusTaskQueueResponse,
     base64TaskToken: string,
-    protobufEncodedTask: ArrayBuffer
+    protobufEncodedTask: ArrayBuffer,
+    requestDeadline: Date | undefined
```
chris-olszewski, line 1310: Not impactful, but it seems like we prefer optional args to `T | undefined` in this codebase.
```suggestion
    requestDeadline?: Date
```

packages/test/src/helpers-nexus.ts
```diff
@@ -0,0 +1,20 @@
+import * as nexus from 'nexus-rpc';
+
+/**
+ * Older servers wrap handler errors in an extra HandlerError; newer servers do not.
```
chris-olszewski, line 4: Should we normalize these errors to just never have the additional `HandlerError` layer? I don't expect users to often be inspecting `cause` chains incredibly close, but it would be nice to have a consistent error shape for users regardless of server version.

packages/common/src/converter/failure-converter.ts
```diff
@@ -469,4 +471,40 @@ export class DefaultFailureConverter implements FailureConverter {
   optionalErrorToOptionalFailure(err: unknown, payloadConverter: PayloadConverter): ProtoFailure | undefined {
     return err ? this.errorToFailure(err, payloadConverter) : undefined;
   }
+
+  private nexusFailureToTemporalFailure(failure: nexus.Failure, retryable: boolean): ProtoFailure {
+    if (failure.metadata?.type === 'temporal.api.failure.v1.Failure') {
+      if (failure.details == null) {
+        throw new TypeError("missing details for Nexus Failure of type 'temporal.api.failure.v1.Failure'");
+      }
+      return failure.details as ProtoFailure;
```
chris-olszewski: I don't think the cast is necessary
```suggestion
      return failure.details;
```

packages/common/src/converter/failure-converter.ts
```diff
@@ -469,4 +471,40 @@ export class DefaultFailureConverter implements FailureConverter {
...
+    }
+  }
+
+  private temporalFailureToNexusFailure(failure: ProtoFailure): nexus.Failure {
+    return {
+      message: failure.message ?? '',
+      metadata: { type: 'temporal.api.failure.v1.Failure' },
+      // Store the full ProtoFailure as the Nexus failure details so it can be round-tripped
+      // losslessly back to a ProtoFailure via nexusFailureToTemporalFailure.
+      details: { ...failure } as Record<string, unknown>,
```
chris-olszewski: Cast unnecessary
```suggestion
      details: { ...failure },
```

packages/common/src/converter/failure-converter.ts
```diff
@@ -386,28 +386,30 @@ export class DefaultFailureConverter implements FailureConverter {
...
+            case true:
+              retryBehavior = encodeNexusHandlerErrorRetryBehavior('RETRYABLE');
+              break;
+            case false:
+              retryBehavior = encodeNexusHandlerErrorRetryBehavior('NON_RETRYABLE');
+              break;
+          }
 
-        return {
-          // TODO(nexus/error): Maybe set a default message here, once we've decided on error handling.
```
chris-olszewski, line 400: Just making sure this TODO no longer applies and the decision was to keep the base error message.

packages/test/src/test-nexus-workflow-caller.ts
```diff
@@ -146,26 +789,25 @@ test('Nexus Operation from a Workflow', async (t) => {
...
-          err.cause.cause.cause.details[0] === 'a detail'
-      );
+      assert(err instanceof WorkflowFailedError);
+      assert(err.cause instanceof NexusOperationFailure);
+      assert(err.cause.cause instanceof nexus.HandlerError);
+      t.is(err.cause.cause.type, 'INTERNAL');
+      const appFailure = unwrapHandlerErrorCause(err.cause.cause);
+      assert(appFailure instanceof ApplicationFailure);
+      t.is(appFailure.message, 'test asked to fail');
+      t.deepEqual(appFailure.details, ['a detail']);
```
chris-olszewski, line 806:
```suggestion
      t.true(err instanceof WorkflowFailedError);
      t.true(err.cause instanceof NexusOperationFailure);
      t.true(err.cause.cause instanceof nexus.HandlerError);
      t.is(err.cause.cause.type, 'INTERNAL');
      const appFailure = unwrapHandlerErrorCause(err.cause.cause);
      t.true(appFailure instanceof ApplicationFailure);
      t.is(appFailure.message, 'test asked to fail');
      t.deepEqual(appFailure.details, ['a detail']);
```

packages/worker/src/worker.ts
mjameswh: > t seems like we prefer optional args to T | undefined in this codebase.

Not sure whether that statement is generally true of our internal APIs, but in this specific case, I would say `T | undefined` is definitely better than the optional arg syntax. A caller that wouldn't provide that argument is most certainly a caller that is doing something incorrect.

packages/common/package.json
```diff
@@ -15,7 +15,7 @@
     "@temporalio/proto": "workspace:*",
     "long": "^5.2.3",
     "ms": "3.0.0-canary.1",
-    "nexus-rpc": "^0.0.1",
+    "nexus-rpc": "^0.0.2",
```
mjameswh, line 18: I should have thought of asking this before, but didn't `nexus-rpc@0.0.2` introduced any breaking changes vs 0.0.1?

packages/worker/src/worker.ts
```diff
@@ -1347,7 +1349,7 @@ export class Worker {
         e instanceof nexus.HandlerError ? e : new nexus.HandlerError('INTERNAL', undefined, { cause: e });
       return {
         taskToken,
-        error: await handlerErrorToProto(this.options.loadedDataConverter, handlerError),
+        failure: await handlerErrorToProto(this.options.loadedDataConverter, handlerError),
```
mjameswh, line 1352: Unless I'm missing something, this implies newer TS SDK releases won't be able to do Nexus against older servers.

I'm personally ok with that given that Nexus support is still marked as experimental, but can you please confirm this is the direction that your team has agreed on?

packages/test/src/test-nexus-handler.ts
```diff
@@ -13,13 +12,12 @@ import {
...
-  convertWorkflowEventLinkToNexusLink,
-  convertNexusLinkToWorkflowEventLink,
-} from '@temporalio/nexus/lib/link-converter';
-import { isBun, cleanStackTrace, compareStackTrace, getRandomPort } from './helpers';
+import { ServiceError } from '@temporalio/client';
+import { coerceToHandlerError, operationErrorToProto } from '@temporalio/worker/lib/nexus/conversions';
+import { getRandomPort } from './helpers';
 
 export interface Context {
   httpPort: number;
```
mjameswh: You should be able to completely remove `httpPort` and `getRandomPort`.

mjameswh: Same for the `allowedAddresses` dynamic config.

packages/common/package.json
VegetarianOrc: Yes it did, but the changes were all removals. Between this PR and #1979 they are all accounted for.

packages/worker/src/worker.ts
VegetarianOrc: The server support negotiation happens in Core in order to simplify Lang integrations. Lang always returns the `failure` property and, if necessary, Core will convert it back to the `error` property.

packages/common/src/converter/failure-converter.ts
VegetarianOrc: That's correct

packages/test/src/test-nexus-workflow-caller.ts
VegetarianOrc: Spoke offline about this. Decided to keep the asserts for the type narrowing.

packages/test/src/test-nexus-handler.ts
VegetarianOrc: Removed these!

packages/worker/src/worker.ts
VegetarianOrc: Decided to leave it as is here because I think I'm with James. It's at least a moment to think of "should I have a value for this?" I also like that an agent won't be able to just omit it and we'd miss it in review

packages/common/package.json
mjameswh: Ok, but may those removal cause build time errors to Temporal SDK users that would have e.g. Temporal SDK 1.15.0 + Nexus 0.0.2?

VegetarianOrc: Yes that's accurate but my understanding is that a user would have to manually update to Nexus 0.0.2 in that case since Temporal 1.15.0 is pinned to `^0.0.1`. Does that align with your understanding?

PR #1979, opened 2026-03-23
:boom: Add Nexus Inbound Interceptors

VegetarianOrc:
## What was changed

 - Added executeStartOperation and executeCancelOperation methods to NexusInboundCallsInterceptor, replacing the previous generic execute interceptor.
  - Bypassed nexus-rpc's ServiceRegistry in favor of a local Map<string, ServiceHandler> with explicit duplicate-name
  validation. The worker now resolves service handlers directly, which simplifies the call stack and enables interceptor
  insertion between service lookup and handler invocation.
  - Simplified decodePayloadIntoLazyValue into a direct decodePayload function, removing the PayloadSerializer/LazyValue adapter layer that was needed for the ServiceRegistry path.
  - Updated stack trace expectations in tests to reflect the shorter call stack (no more ServiceRegistry frames).
  - Added tests for: duplicate nexus service registration rejection, calling a nonexistent service (NOT_FOUND), inbound start interceptor modifying input/result, and inbound cancel interceptor modifying the token.

## Why?

- Align Nexus interceptor functionality with that of other SDKs
- Removing the usage of `nexusrpc.ServiceRegistry` was required to allow interceptors to act on the deserialized input. This change also aligns with desired future state for Nexus in Temporal SDKs.

## Checklist

1. How was this tested:

  - Added integration test for inbound executeStartOperation interceptor can modify input and result
  - Added integration test for inbound executeCancelOperation interceptor can modify input
  - Added test for calling a nonexistent service returns NOT_FOUND
  - Added test for Worker.create rejects duplicate nexus service registrations
  - Updated existing stack trace assertions to match the new (shorter) call stack

### DIFF
--- a/packages/worker/src/interceptors.ts
+++ b/packages/worker/src/interceptors.ts
@@ -131,19 +131,19 @@ export type NexusInterceptors = {
  * @experimental Nexus support in Temporal SDK is experimental.
  */
 export type NexusInboundCallsInterceptor = {
-  executeStartOperation?: (
+  startOperation?: (
     input: ExecuteNexusOperationStartInput,
-    next: Next<NexusInboundCallsInterceptor, 'executeStartOperation'>
+    next: Next<NexusInboundCallsInterceptor, 'startOperation'>
   ) => Promise<ExecuteNexusOperationStartOutput>;
 
-  executeCancelOperation?: (
+  cancelOperation?: (
     input: ExecuteNexusOperationCancelInput,
-    next: Next<NexusInboundCallsInterceptor, 'executeCancelOperation'>
+    next: Next<NexusInboundCallsInterceptor, 'cancelOperation'>
   ) => Promise<void>;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


packages/worker/src/interceptors.ts
```diff
@@ -131,29 +131,44 @@ export type NexusInterceptors = {
  * @experimental Nexus support in Temporal SDK is experimental.
  */
 export type NexusInboundCallsInterceptor = {
-  execute?: (
-    input: NexusExecuteInput,
-    next: Next<NexusInboundCallsInterceptor, 'execute'>
-  ) => Promise<NexusExecuteOutput>;
+  executeStartOperation?: (
```
mjameswh: Why `execute` as a prefix (here and below)? In other TS interceptors, the hook names are simpler.

VegetarianOrc: The inbound interceptors for Workflow and Activity define the `execute` hook. I thought the prefix aligned with that and also reflects the naming from the Python SDK.

PR #1984, opened 2026-03-26
fix(otel): avoid TypeError if user code throws non-error

### LINKED ISSUE #1975: [Bug] maybeAddErrorToSpan fails when err is undefined
### Bug Report

`maybeAddErrorToSpan` in `@temporalio/interceptors-opentelemetry@1.15.0` fails with `TypeError: Cannot read properties of undefined (reading 'message')` when a workflow catches a non-Error value (e.g. `undefined`).

### Reproduction

Any workflow that catches an `undefined` error value (e.g. from a child workflow failure or a `throw undefined`) will trigger the failure:

```typescript
// In a workflow
try {
  await executeChild('someChildWorkflow', { ... });
} catch (err) {
  // If err is undefined, the OTel interceptor's wrapWithSpan
  // calls maybeAddErrorToSpan(undefined, span) which crashes
}
```

### Root Cause

In `packages/interceptors-opentelemetry/src/instrumentation.ts` line 93:

```typescript
span.setStatus({ code: statusCode, message: (err as Error).message ?? String(err) });
```

When `err` is `undefined`, `(undefined as Error).message` throws a `TypeError` **before** the `??` operator can evaluate. The `??` only guards against the `.message` property being nullish, not the `err` object itself.

### Impact

- The `TypeError` escapes `wrapWithSpan` and becomes the `workflowTaskError`
- Temporal retries the workflow task (deterministic replay), hits the same code path, creating an **infinite crash loop**
- The **original error is permanently masked** by the TypeError, making it impossible to diagnose the real failure

### Introduced By

PR #1632 (merged 2025-02-27) added error messages to span status using the unsafe `(err as Error).message ?? String(err)` pattern. 

### Suggested Fix

```typescript
// Before (crashes when err is undefined/null)
span.setStatus({ code: statusCode, message: (err as Error).message ?? String(err) });

// After (safe for all error types)
span.setStatus({ code: statusCode, message: err instanceof Error ? (err.message ?? String(err)) : String(err ?? 'unknown error') });
```

This handles:
- `undefined` / `null` -> `"unknown error"`
- Non-Error thrown values (strings, numbers) -> their string representation
- Error objects with `.message = undefined` -> falls back to `String(err)`
- Normal Error objects -> `.message` as expected

### Environment

- `@temporalio/interceptors-opentelemetry`: 1.15.0 (latest)
- `@temporalio/worker`: 1.15.0
- Node.js: 20.x

chris-olszewski:
## What was changed
Remove cast to `Error` and instead check if there's a `message` attached to the thrown object and use that, if not we stringify whatever was thrown.

## Why?
It is not suggested to throw non-Errors, but we should do our best effort to forward the message from whatever was thrown.

## Checklist

1. Closes #1975

2. How was this tested:
Added test to verify we forward message from a variety of thrown objects.

3. Any docs updates needed?
N/A

### DIFF
--- a/packages/interceptors-opentelemetry/src/instrumentation.ts
+++ b/packages/interceptors-opentelemetry/src/instrumentation.ts
@@ -90,7 +90,8 @@ function maybeAddErrorToSpan(err: any, span: otel.Span, acceptableErrors?: (err:
   const isBenignErr = err instanceof ApplicationFailure && err.category === ApplicationFailureCategory.BENIGN;
   if (acceptableErrors === undefined || !acceptableErrors(err)) {
     const statusCode = isBenignErr ? otel.SpanStatusCode.UNSET : otel.SpanStatusCode.ERROR;
-    span.setStatus({ code: statusCode, message: (err as Error).message ?? String(err) });
+    const message = err != null && typeof err.message === 'string' ? err.message : String(err);
+    span.setStatus({ code: statusCode, message });
     span.recordException(err);
   } else {
     span.setStatus({ code: otel.SpanStatusCode.OK });


packages/interceptors-opentelemetry/src/instrumentation.ts
```diff
@@ -90,7 +90,8 @@ function maybeAddErrorToSpan(err: any, span: otel.Span, acceptableErrors?: (err:
   const isBenignErr = err instanceof ApplicationFailure && err.category === ApplicationFailureCategory.BENIGN;
   if (acceptableErrors === undefined || !acceptableErrors(err)) {
     const statusCode = isBenignErr ? otel.SpanStatusCode.UNSET : otel.SpanStatusCode.ERROR;
-    span.setStatus({ code: statusCode, message: (err as Error).message ?? String(err) });
+    const message = err != null && typeof err.message === 'string' ? err.message : String(err);
```
mjameswh, line 93: We have other places in the code base where we were using the same pattern. We should fix those too.

PR #1992, opened 2026-04-02
Named random streams

### LINKED ISSUE #1895: [Feature Request] Add workflow API to create independent deterministic pseudo-number sequence
### Context
* Some Workflow Plugins need to call `Math.random()` or `workflow.uuid4()`.
* If the workflow code also use these methods, then the plugin and the workflow code both pull values out of the same PRNG sequence.
* That may result in NDEs if a workflow is replayed after adding/removing a Plugin.

Add the following to the Workflow API:

1. API to create a distinct deterministic PRNG instance
1. Possibility of generating UUIDs from that distinct PRNG instances (i.e. rather than only from the workflow’s global PRNG).
1. Possibility of activating a given PRNG instance as the “global” one within a given execution scope. This is required to ensure that third-party, non-Temporal libraries used by Temporal plugins don’t disrupt the workflow’s actual PRNG sequence.

Also needs to automatically wrap around any query handler / update validator / any readonly context

eamsden:
## What was changed
Added an API to get named streams of pseudorandomness, and to run worker code with a named stream patched in in place of workflow randomness.

## Why?
Plugins which use Math.random() or workflow.uuid4(), or call libraries which invoke Math.random(), steal randomness from the workflow's deterministic PRNG stream. This means that running with or without a plugin that uses randomness can cause the workflow to observe different random values and thus cause non-deterministic execution. We are adding an API that allows plugin authors to use randomness, or use libraries which use randomness, without changing the randomness observed by the workflow.

## Checklist

1. Closes #1895 

2. How was this tested:
[tbw]

3. Any docs updates needed?
API docs will need to be updated, and we should likely write a strong section in the [Plugins Guide](https://docs.temporal.io/develop/plugins-guide) on using separate random streams.


packages/workflow/src/random-streams.ts
```diff
@@ -0,0 +1,44 @@
...
+    return fillWithRandom(() => this.random(), bytes);
+  }
+}
+
+export function getRandomStream(name: string): WorkflowRandomStream {
+  const activator = assertInWorkflowContext('Workflow.getRandomStream(...) may only be used from workflow context.');
+  return new ActivatorRandomStream(activator, name);
+}
+
+export function withRandomStream<T>(_name: string, _fn: () => T): T {
```
mjameswh: I'd make that a method on `WorkflowRandomStream` rather than a top-level function that accepts the name. Or support both.

My intuition is that most users would prefer keeping the instance in a module variable rather systematically refer to the name.

packages/workflow/src/random-streams.ts
```diff
@@ -0,0 +1,44 @@
+import { IllegalStateError } from '@temporalio/common';
+
+import { assertInWorkflowContext } from './global-attributes';
+import type { Activator } from './internals';
+import { fillWithRandom, uuid4FromRandom } from './random-helpers';
+
+/**
+ * A deterministic PRNG stream scoped to the current Workflow execution.
+ */
+export interface WorkflowRandomStream {
```
mjameswh, line 19: Might as well expose a global variable of type `WorkflowRandomStream` that would correspond to the "default" random sequence of the workflow execution.

packages/workflow/src/alea.ts
```diff
@@ -67,6 +69,23 @@ export function alea(seed: number[]): RNG {
   return xg.next.bind(xg);
 }
 
+const RANDOM_STREAM_SEED_PREFIX = Array.from(encode('temporal-workflow-random-stream-v1'));
```
mjameswh, line 72: For the sake of legal clarity, I'd rather not add anything in the middle of this file. The `Alea` class above and `Mash` class below are taken from https://github.com/davidbau/seedrandom/blob/released/lib/alea.js, as indicated by the license banner at the top this file. I'd suggest you move the `RNG` type and the `alea`, `encodeU32` and `deriveAleaSeed` functions to the top of this file, with a separator before the "this is taken" banner. I think that should be sufficient. Or move that to `random-stream.ts`.

packages/workflow/src/internals.ts
```diff
@@ -478,11 +497,49 @@ export class Activator implements ActivationHandler {
...
+    );
+  }
+
+  public withCurrentRandom<T>(randomSource: ScopedWorkflowRandomSource, fn: () => T): T {
+    return this.withRandomSource(randomSource, fn);
+  }
+
+  public bindCurrentRandom<T extends (...args: any[]) => any>(fn: T): T {
+    const randomSource = this.currentRandomStorage?.getStore();
+    return ((...args: Parameters<T>) => this.withRandomSource(randomSource, () => fn(...args))) as unknown as T;
```
mjameswh, line 536: Why did you need the cast to `unknown` then `T`?

packages/workflow/src/random-streams.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+class ActivatorRandomStream implements WorkflowRandomStream {
+  constructor(
+    protected readonly activator: Activator,
+    protected readonly name: string
+  ) {}
+
+  random(): number {
+    // TODO: Cache the resolved RNG on this wrapper once named-stream reseeding can refresh cached streams safely.
```
mjameswh, line 65: Do you still intend to cache the namespaced RNG? That would be slightly desirable, but not a big thing, really.

packages/workflow/src/index.ts
```diff
@@ -136,6 +136,8 @@ export {
   WorkflowInfo,
 } from './interfaces';
 export { proxySinks, Sink, SinkCall, SinkFunction, Sinks } from './sinks';
+export type { WorkflowRandomStream } from './random-streams';
```
mjameswh, line 139: Please add an `@experimental` annotation on all new APIs.

packages/workflow/src/worker-interface.ts
```diff
@@ -3,13 +3,14 @@
  *
  * @module
  */
-import type { WorkflowFunctionWithOptions } from '@temporalio/common';
+import type { WorkflowFunctionWithOptions, WorkflowFunctionWithOptions } from '@temporalio/common';
 import { encodeVersioningBehavior, IllegalStateError } from '@temporalio/common';
 import { composeInterceptors } from '@temporalio/common/lib/interceptors';
```
mjameswh, line 8: Ooops... I quickly fixed some merge conflicts in GH's web interface, and only realized a second too late that I had introduced this incorrect import.

mjameswh (commented):
In a follow up PR, we should use a distinct, private random sequence for internal calls to `uuid4()` (e.g. when assigning a default workflow id to child workflows). That is, by far, the call that is the most likely to result in an NDE if the default random sequence is misused.

That will however require adding an SDK flag to preserve compatibility with previous workflow executions.

mjameswh (commented):
I'd like to see a test to confirm the behavior on workflow reset.

That coujld be something like this:

1. Start a workflow
2. The workflow calls named random() and uuid() (both from a named sequence), and startChildWorkflow(). Capture all three to an array (let's call that A).
3. On a new WFT, the workflow starts a timer.
4. The workflow calls random(), uuid() and startChildWorkflow(). Capture all three (let's call that B)
5 New timer.
6. The test harness extracts A and B.
7. Test harness resets the workflow to the WFT of the timer created at step 3. The workflow will replay step 2, so the randoms geneerated at that step are recreated. Let's call that C.
8. The reseted workflow calls random(), uuid() and startChildWorkflow(). Capture all three, to D.
9. The test harness extracts C and D out of the replayed workflow.
10. Confirm that:
  - A == C
  - A != B
  - C != D
  - B != D

mjameswh (changes requested):

packages/workflow/src/random-streams.ts
eamsden: Doesn't seem strongly necessary.

packages/test/src/test-interceptors.ts
```diff
@@ -259,6 +259,7 @@ if (RUN_INTEGRATION_TESTS) {
         ApplicationFailure: Expected anything other than 1
             at $CLASS.nonRetryable (common/src/failure.ts)
             at Object.continueAsNew (test/src/workflows/interceptor-example.ts)
+            at composedNext (common/src/interceptors.ts)
```
chris-olszewski, line 262: I think we should update `DROPPED_STACK_FRAMES_PATTERNS` to look for `composedNext` in addition to `next` so we don't pollute the stack trace

packages/workflow/src/random-streams.ts
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Get a named deterministic random stream for the current workflow execution.
+ *
+ * Named streams are derived from the workflow seed and a stable stream name,
+ * without consuming the workflow's default `Math.random()` stream. Repeated
+ * calls with the same `name` within a workflow execution refer to the same
+ * logical stream state, including across activations.
+ *
+ * This is the preferred entry point for workflow plugins and interceptors that
+ * need their own deterministic entropy. Use stable package- or module-style
```
chris-olszewski, line 123:
```suggestion
 * need their own deterministic entropy. Use stable package or module-style
```

chris-olszewski (approved):
Super minor fixup wrt the stack traces, but otherwise looks good to me!

mjameswh (approved):
LGTM

PR #1995, opened 2026-04-04
Lambda Worker contrib package

Sushisource:
- New @temporalio/lambda-worker package that wraps the full per-invocation lifecycle for running
  Temporal workers inside AWS Lambda (connect, create worker, poll, graceful shutdown)
  - Ships with Lambda-tuned concurrency defaults, automatic deadline management, structured JSON
  logging via Powertools, and optional OTEL helpers for ADOT integration
  - Worker Deployment Versioning is always enabled (defaults to PINNED)
  - Includes comprehensive unit tests (~625 lines) covering config/defaults, shutdown hooks, deadline
   handling, connection lifecycle, and connection options

### DIFF
--- a/packages/core-bridge/src/worker.rs
+++ b/packages/core-bridge/src/worker.rs
@@ -868,7 +868,7 @@ mod custom_slot_supplier {
                     }
                     Err(err) => {
                         warn!("Error reserving slot: {err:?}");
-                        tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+                        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
                     }
                 }
             }


02strich (approved):

packages/lambda-worker/src/json-logger.ts
```diff
@@ -0,0 +1,61 @@
...
+      case 'INFO':
+        this.ptLogger.info(message, meta as Record<string, unknown>);
+        break;
+      case 'WARN':
+        this.ptLogger.warn(message, meta as Record<string, unknown>);
+        break;
+      case 'ERROR':
+        this.ptLogger.error(message, meta as Record<string, unknown>);
+        break;
+    }
```
chris-olszewski, line 47: Not that I ever expect us to add a new log level
```suggestion
      default:
        level satisfies never;
        break;
    }
```

packages/lambda-worker/README.md
```diff
@@ -0,0 +1,115 @@
...
+## Quick start
+
+```typescript
+// handler.ts
+import { runWorker } from '@temporalio/lambda-worker';
+import * as activities from './activities';
+
+export const handler = runWorker({ deploymentName: 'my-service', buildId: 'v1.0' }, (config) => {
+  config.workerOptions.taskQueue = 'my-task-queue';
+  config.workerOptions.workflowBundle = require('./workflow-bundle.json');
```
chris-olszewski:
```suggestion
  config.workerOptions.workflowBundle = { code: require('./workflow-bundle.js') };
```

packages/lambda-worker/src/types.ts
```diff
@@ -0,0 +1,75 @@
...
+   * Worker options, pre-populated with Lambda-appropriate defaults.
+   *
+   * You must set at least `taskQueue` (or set the `TEMPORAL_TASK_QUEUE` env var). Typically you'll
+   * also set `activities` and `workflowBundle` (prefer pre-bundled workflows over `workflowsPath`
+   * to avoid bundling overhead on Lambda cold starts).
+   *
+   * The following fields are managed by settings elsewhere and will be overridden per-invocation:
+   * `connection`, `namespace`, `identity`, `workerDeploymentOptions`.
+   */
+  workerOptions: Partial<WorkerOptions>;
```
chris-olszewski: Not sure how beneficial this would be, but could convey that `connection` etc aren't respected via the type
```suggestion
   * You must set at least `taskQueue` (or set the `TEMPORAL_TASK_QUEUE` env var). Typically you'll
   * also set `activities` and `workflowBundle` (prefer pre-bundled workflows over `workflowsPath`
   * to avoid bundling overhead on Lambda cold starts).
   *
   * The following fields are managed by settings elsewhere and will be overridden per-invocation:
   * `connection`, `namespace`, `identity`, `workerDeploymentOptions`.
   */
  workerOptions: Partial<Omit<WorkerOptions, 'connection' | 'namespace' | 'identity' | 'workerDeploymentOptions'>>;
```

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,190 @@
...
+    },
+    shutdownHooks: [],
+  };
+
+  const envTaskQueue = process.env['TEMPORAL_TASK_QUEUE'];
+  if (envTaskQueue) {
+    config.workerOptions.taskQueue = envTaskQueue;
+  }
+
+  const connectConfig = deps.loadConnectConfig(config.envConfigOptions);
```
chris-olszewski: `config.envConfigOptions` will always be undefined at this point

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,190 @@
...
+  deps: WorkerDeps
+): LambdaHandler {
+  const config: LambdaWorkerConfig = {
+    workerOptions: {
+      ...LAMBDA_WORKER_DEFAULTS,
+      workerDeploymentOptions: {
+        version,
+        useWorkerVersioning: true,
+        defaultVersioningBehavior: 'PINNED',
+      },
```
chris-olszewski, line 91: Do we want to allow this to get modified by the user? If not, we should probably have a `userConfig` object that gets passed in and then later fully resolved into the config required to run the serverless worker.

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,190 @@
...
+    const connection = await deps.connect(connectionOptions);
+    try {
+      // Build worker options: user config + managed fields
+      const namespace = config.namespace ?? connectConfig.namespace ?? 'default';
+      const workerOptions: WorkerOptions = {
+        ...config.workerOptions,
+        connection,
+        namespace,
+        identity,
+      } as WorkerOptions;
```
chris-olszewski: Would feel better if this was a smaller cast than it currently is. I think the `connection` is the only type that actually needs massaging. Or beef up the connection mock definition, but that might be overkill.

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,190 @@
...
+
+      const worker = await deps.createWorker(workerOptions);
+
+      await worker.runUntil(
+        new Promise<void>((resolve) => {
+          setTimeout(resolve, workTimeMs);
+        })
```
chris-olszewski: Could use the stdlib promise timers: https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options
```suggestion
        setTimeout(workTimeMs)
```

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,190 @@
...
+ * @example
+ * ```ts
+ * import { runWorker } from '@temporalio/lambda-worker';
+ * import * as activities from './activities';
+ *
+ * export const handler = runWorker(
+ *   { buildId: 'v1.0.0', deploymentName: 'my-service' },
+ *   (config) => {
+ *     config.workerOptions.taskQueue = 'my-task-queue';
+ *     config.workerOptions.workflowBundle = require('./workflow-bundle.json');
```
chris-olszewski:
```suggestion
 *     config.workerOptions.workflowBundle = { code: require('./workflow-bundle.js') };
```

packages/lambda-worker/src/otel.ts
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Apply batteries-included OTel defaults: tracing via OTLP gRPC exporter.
+ * Registers a flush/shutdown hook on the config.
+ *
+ * Requires `@opentelemetry/*` peer dependencies to be installed.
+ *
+ * @example
+ * ```ts
+ * import { runWorker } from '@temporalio/lambda-worker';
+ * import { applyDefaults } from '@temporalio/lambda-worker/lib/otel';
```
chris-olszewski: We should re-export these to as an `otel` module to avoid having users reach into `lib`. I think if you make `otel` it's own [package entry point](https://nodejs.org/api/packages.html#package-entry-points) you could avoid the dynamic imports as well.

packages/lambda-worker/src/types.ts
Sushisource: Ah yeah good call

packages/lambda-worker/src/lambda-worker.ts
Sushisource: We do. They should be able to change the default behavior. They can't set useWorkerVersioning to false, though, and that's enforced at runtime. I can probably enforce that at compile time though so I'll see about doing that.

packages/lambda-worker/src/otel.ts
Sushisource: Let me know if what I did is what you're thinking of

Sushisource:
@claude review

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,184 @@
...
+  };
+
+  const envTaskQueue = process.env['TEMPORAL_TASK_QUEUE'];
+  if (envTaskQueue) {
+    config.workerOptions.taskQueue = envTaskQueue;
+  }
+
+  const connectConfig = deps.loadConnectConfig(config.envConfigOptions);
+
+  configure(config);
```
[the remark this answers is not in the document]
Sushisource: Sort of... real fix would be to ensure connectConfig is set into config so user sees it in their handler

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,178 @@
...
+        shutdownDeadlineBufferMs,
+        workTimeMs,
+      });
+    }
+
+    const identity = `${context.awsRequestId}@${context.invokedFunctionArn}`;
+
+    const connection = await deps.connect(config.connectionOptions ?? {});
+    try {
+      const workerOptions: WorkerOptions = {
```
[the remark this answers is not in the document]
Sushisource: Doesn't make sense - remaining time is remaining time, we can't just exclude the worker setup time

packages/lambda-worker/src/otel.ts
```diff
@@ -0,0 +1,82 @@
...
+  const traceExporter = new OTLPTraceExporter({
+    url: options?.endpoint,
+  });
+
+  const sdk = new NodeSDK({ resource, traceExporter });
+  sdk.start();
+
+  config.shutdownHooks.push(async () => {
+    await traceExporter.forceFlush();
+  });
```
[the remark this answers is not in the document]
Sushisource: forceFlush() doesn't exist on NodeSDK

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,181 @@
...
+  config.namespace = connectConfig.namespace ?? 'default';
+
+  configure(config);
+
+  const taskQueue = config.workerOptions.taskQueue;
+  if (!taskQueue) {
+    throw new Error(
+      'taskQueue is required: set config.workerOptions.taskQueue in the configure callback ' +
+        'or set the TEMPORAL_TASK_QUEUE environment variable'
+    );
```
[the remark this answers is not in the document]
Sushisource: Meh. No reasonable person would call it twice anyway, but maybe worth documenting.

packages/lambda-worker/src/otel.ts
```diff
@@ -0,0 +1,79 @@
...
+export function applyDefaults(config: LambdaWorkerConfig, options?: OtelOptions): void {
+  // Dynamic require to avoid failure when optional peer deps are not installed
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { NodeSDK } = require('@opentelemetry/sdk-node');
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { OTLPTraceExporter } = require('@opentelemetry/exporter-trace-otlp-grpc');
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { Resource } = require('@opentelemetry/resources');
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { ATTR_SERVICE_NAME } = require('@opentelemetry/semantic-conventions');
```
chris-olszewski: All of these could be lifted to normal `import`s at the top of the module now that you have `otel` as a separate entrypoint.
```
import { NodeSDK } from '@opentelemetry/sdk-node';
// etc
```
This would bring any `MODULE_NOT_FOUND` errors due to missing peer deps happen at module resolution time as opposed to when calling `applyDefaults`. Caveat is that `applyTracing` can no longer be used without the otel deps, but that might just indicate it belongs somewhere else.

packages/lambda-worker/src/lambda-worker.ts
```diff
@@ -0,0 +1,183 @@
...
+    const remainingMs = context.getRemainingTimeInMillis();
+    const workTimeMs = remainingMs - shutdownDeadlineBufferMs;
+
+    if (workTimeMs <= MINIMUM_WORK_TIME_MS) {
+      throw new Error(
+        `Insufficient time for Lambda worker: ${remainingMs}ms remaining, ` +
+          `${shutdownDeadlineBufferMs}ms shutdown buffer, only ${workTimeMs}ms work time ` +
+          `(minimum: ${MINIMUM_WORK_TIME_MS}ms)`
+      );
+    }
```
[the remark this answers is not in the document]
Sushisource: Not exactly important

chris-olszewski (approved):

PR #1996, opened 2026-04-06
Add `SerializationContext` API for core payload/failure conversion paths

THardy98:
## What was changed
Added `SerializationContext` API for core payload/failure conversion paths (i.e. workflows and activities).

Scope of this PR is limited to core workflow/activity boundaries (as an effort to split: https://github.com/temporalio/sdk-typescript/pull/1950 into easier to review chunks).

Specifically, the boundaries implemented in this PR are:
 | Boundary | Payload Context | Failure Context |
  | --- | --- | --- |
  | Workflow start args | Workflow | N/A |
  | Workflow result | Workflow | N/A |
  | Workflow failure | N/A | Workflow |
  | Query args/result | Workflow | N/A |
  | Update args/result | Workflow | N/A |
  | Signal receive payload | Workflow | N/A |
  | Remote activity args/result | Activity | Activity |
  | Remote activity failure observed by workflow | N/A | Activity |
  | Remote activity heartbeat details | Activity | N/A |
  | Local activity args/result | Activity (`isLocal=true`) | Activity (`isLocal=true`) |
  | Local activity failure observed by workflow | N/A | Activity (`isLocal=true`) |
  | Child workflow start args/result | Child workflow | Child workflow |
  | Child workflow failure observed by parent | N/A | Child workflow |
  | External workflow signal payload (success) | Target workflow | Target workflow |
  | External workflow signal failure | N/A | Target workflow |
  | External workflow cancel failure | N/A | Target workflow |
  | Continue-as-new args | Workflow | N/A |
  | Workflow memo upsert encode path | Workflow | N/A |
  | Timer summary encode path | Workflow | N/A |
  | Activity pre-start heartbeat-details decode failure | N/A | Activity |

**Note for reviewers**:
The diff looks big but about half of it is the test suite at `test-serialization-context.ts`. It might be easier to review this PR starting there, to see what the serialization context across these boundaries look like.

----

1. Part of  [#1661]

2. How was this tested:

- Comprehensive test suite at `test-serialization-context.ts` for the boundaries implemented
- Also a test in `test-worker-heartbeating.ts` to check serialization context is correct when `activity start` task fails (specifically when it fails during `extractActivityInfo` for a failed activity heartbeat, we want to check that the serialization context is propagated through the failure correctly).


mjameswh (commented):
At workflow-client.ts:1615 (new code), we pass the unbound data converter to `executionInfoFromRaw` so that we can later decode a workflow's memo, which is correct.

But then, I'd expect `executionInfoFromRaw` to bind workflow id when actually calling the data converter.

packages/common/src/converter/serialization-context.ts
```diff
@@ -0,0 +1,107 @@
+import type { LoadedDataConverter } from './data-converter';
+import type { FailureConverter } from './failure-converter';
+import type { PayloadConverter } from './payload-converter';
+
+/**
+ * Context for payloads owned by a workflow.
```
mjameswh, line 6:
```suggestion
 * Context provided to data converters for payloads owned by a workflow.
```

packages/workflow/src/workflow.ts
```diff
@@ -168,6 +205,9 @@ const validateLocalActivityOptions = validateActivityOptions;
 function scheduleActivityNextHandler({ options, args, headers, seq, activityType }: ActivityInput): Promise<unknown> {
   const activator = getActivator();
   validateActivityOptions(options);
+  const activityId = options.activityId ?? `${seq}`;
+  const context = activitySerializationContext(activator.info, activityId, false);
```
mjameswh, line 209: What's the use case for "activity id" in the activity serialization context?

THardy98: standalone activities

mjameswh: I see. But in the case of activities scheduled from a workflow, I'm not sure passing seq as activity id is desirable. I don't see how users could possibly do good use of that information, yet they will like be some that will misuse it (e.g. by assuming that "activity id" would be globally unique, which `seq` won't be.

THardy98: Discussed offline - there is already precedent (for a long long time) where the activity ID falls back to `seq` if it is not already set.

mjameswh (approved):

PR #1998, opened 2026-04-06
:boom: Nexus OpenTelemetry support

VegetarianOrc:
## What was changed

- Added OpenTelemetry interceptors for inbound Nexus operations.
- Added context propagation for Nexus calls.
- Added extractContextFromNexusHeaders and nexusHeadersWithContext helpers for propagating OTel context through Nexus headers.
- Registered the new Nexus interceptors in OpenTelemetryPlugin
- Added new span names: RunStartNexusOperation and RunCancelNexusOperation
- Fixed the outbound StartNexusOperation span name delimiter from :service:operation to :service/operation
- Renamed interceptor input/output types from ExecuteNexusOperation* to NexusStartOperation*/NexusCancelOperation* for consistency
- Added an OtelSdkContext helper to tear down the global TracerProvider between serial OTel tests

## Why?

Nexus operations previously had no inbound instrumentation. The workflow could create an outbound span when starting a Nexus operation, but trace context was not propagated to the worker handling the operation, breaking distributed traces across Nexus service boundaries. This adds the missing inbound/outbound interceptors and context propagation so that Nexus operations appear as connected spans in a single trace.

## Checklist

1. How was this tested:
    - Added a new integration test (Otel nexus inbound interceptor creates spans for startOperation and cancelOperation) that:
      - Starts an async Nexus operation from a workflow, then cancels it
      - Asserts that StartNexusOperation, RunStartNexusOperation, and RunCancelNexusOperation spans all exist
      - Verifies parent-child span relationships and shared trace IDs across the workflow-to-worker boundary
    - Existing OTel tests updated to use OtelSdkContext for proper global TracerProvider lifecycle management

2. Any docs updates needed?
- On release, the Nexus Otel docs will need updates to reflect the new functionality.

### DIFF
--- a/packages/test/src/test-otel.ts
+++ b/packages/test/src/test-otel.ts
@@ -331,12 +331,6 @@ if (RUN_INTEGRATION_TESTS) {
                   return nexus.HandlerStartOperationResult.async(ctx.requestId!);
                 },
                 async cancel(_ctx, _token): Promise<void> {},
-                async getInfo() {
-                  throw new Error('not implemented');
-                },
-                async getResult() {
-                  throw new Error('not implemented');
-                },
               },
             }),
           ],
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -3712,10 +3712,6 @@ packages:
     resolution: {integrity: sha512-IWjIExdVYlmwXuzHdY/Q3lXCv1gbqoAXPazQhy2w4Xgtgha3H0OOujEESVPQcFUFMWm+pAk2gKnb57g8S41JZg==}
     engines: {node: '>= 20.0.0'}
 
-  nexus-rpc@0.0.2:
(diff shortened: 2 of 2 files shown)


packages/test/src/test-otel.ts
```diff
@@ -704,10 +865,7 @@ if (RUN_INTEGRATION_TESTS) {
         },
       };
 
-      const otel = new opentelemetry.NodeSDK({
-        resource: staticResource,
-        traceExporter: wrappedExporter,
-      });
+      const otel = new OtelSdkContext({ resource: staticResource, traceExporter: wrappedExporter });
```
chris-olszewski, line 866: You should also extend this test to make sure that our nexus handlers are complete.

https://github.com/temporalio/sdk-typescript/blob/a564a3c311e641ab15dc686d969edc8b307236b2/packages/test/src/test-otel.ts#L898

mjameswh (approved):

Issue #1958, opened 2026-03-10
[Feature Request] Workflow Bundler: Allow user to define modules to be preloaded


mjameswh:
### Describe the solution you'd like

- Add a `preloadModules` property to `BundlerOptions`. Those modules should be preloaded during early initialization of the "reuse v8 context" workflow vm executor. This will significantly reduce the per-workflow memory footprint, as well as per-workflow initialization time.

PR #1999, opened 2026-04-06
Extend bundler to allow pre-loading modules across reusable VM

THardy98:
## What was changed
This change adds a `preloadModules?: string[]` option on `BundleOptions` and `WorkerOptions.bundleOptions`.

`preloadModules` allows you to load selected modules once into a reusable V8 context. This means preloaded module state can be shared across workflow executions running in that context instead of being reinitialized per workflow.

Under the hood, the bundler now emits a `__TEMPORAL__.preloadModules()` hook alongside the existing generated bundle hooks. In reusable VM mode, `ReusableVMWorkflowCreator` calls that hook immediately after evaluating the bundle (necessarily, the `__TEMPORAL__` bundle needs to exist first), but before taking the pristine snapshot / deep-freezing shared state (so preload modules can be added to the shared module cache).

We also reject configurations where the same module is both ignored and preloaded.

**Note**: naturally, this only applies to workers that enable `reuseV8Context` (we are only to share/reuse modules across a reused VM context).

## Why?
The issue explains the motivation. It reduces per-workflow memory footprint (assuming preloaded modules are used across multiple workflows) and per-workflow startup time (as we preload modules once at worker creation, instead of for each workflow).

1. Closes [#1958]

2. How was this tested:
Tests in `test-bundler.ts`

3. Any docs updates needed?
No

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -488,6 +488,10 @@ export class WithStartWorkflowOperation<T extends Workflow> {
       this[withStartWorkflowOperationResolve] = resolve;
       this[withStartWorkflowOperationReject] = reject;
     });
+    // Suppress unhandled rejection: if executeUpdateWithStart fails before any
+    // response, we reject this promise AND re-throw to the caller. Without this
+    // no-op catch the rejected promise has no handler and Node 15+ terminates.
+    this.workflowHandlePromise.catch(() => undefined);
   }
 
   public async workflowHandle(): Promise<WorkflowHandle<T>> {
--- a/packages/test/src/test-integration-update.ts
+++ b/packages/test/src/test-integration-update.ts
@@ -166,6 +166,36 @@ test('updateWithStart failure: invalid argument', async (t) => {
   }
 });
 
+// Regression test for https://github.com/temporalio/sdk-typescript/issues/1960
+// when executeUpdateWithStart fails at the transport level, workflowHandlePromise is
+// rejected. If the caller never awaits workflowHandle(), that rejection was unhandled,
+// causing Node 15+ to terminate the process.
+test('executeUpdateWithStart: no unhandledRejection when caller ignores workflowHandle', async (t) => {
+  const startOp = new WithStartWorkflowOperation(workflowWithUpdates, {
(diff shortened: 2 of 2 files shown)


packages/worker/src/worker-options.ts
```diff
@@ -581,6 +581,17 @@ export interface WorkerOptions {
...
     ignoreModules?: string[];
+
+    /**
+     * List of modules to preload once during reusable V8 context bootstrap.
+     *
+     * This option is only beneficial when {@link WorkerOptions.reuseV8Context} is enabled.
+     * Preloaded modules are shared across workflows that execute in the same reusable V8 context,
+     * and their module scope runs before a workflow activator exists.
+     *
+     * > NOTE: This is an advanced option that should be used with care.
```
mjameswh: We need to be explicit about why this "should be used with care". How is this option dangerous?

I'd suggest adding something along the line of (feel free to rephrase):

```
 * > NOTE: This is an advanced option that should be used with care.
           Preloading modules that internally stores some form of
           per-workflow state will very likely cause workflow context
           leak, which may result in non-deterministic behavior and/or
           cause other unexpected beheviours.
```

packages/worker/src/workflow/bundler.ts
```diff
@@ -386,6 +405,17 @@ export interface BundleOptions {
...
   ignoreModules?: string[];
 
+  /**
+   * List of modules to load once during reusable V8 context bootstrap.
+   *
+   * Preloaded modules are shared across workflows that execute in the same reusable V8 context.
+   * Module scope in these modules runs before a workflow activator exists, so this option is only
+   * appropriate for modules that are safe to initialize that early.
+   *
+   * > NOTE: This is an advanced option that should be used with care.
```
mjameswh: See my comment on the same property in `WorkerOptions`. It applies here too.

packages/worker/src/worker-options.ts
THardy98: updated

packages/worker/src/workflow/bundler.ts
THardy98: updated

PR #2015, opened 2026-04-17
chore: enforce type only imports

chris-olszewski:
## What was changed
Follow up for #2010, update our eslint settings to enforce identifiers only used for typing get imported as `import type`.

Done by enabling [consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports/)

## Why?
General good hygiene to use `import type` to identify what imports are not present at runtime.

## Checklist

1. Closes N/A

2. How was this tested:
Changes eslint config, then ran `pnpm run lint` to fixup the codebase.

3. Any docs updates needed?
N/A

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -22,7 +22,7 @@ export default tseslint.config(
     },
     rules: {
       eqeqeq: ['error', 'always', { null: 'ignore' }],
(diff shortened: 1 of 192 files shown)


mjameswh (approved):
Good idea, thanks.

PR #2020, opened 2026-04-21
feat(workflow): Add support for failing workflow execution on some error types

### LINKED ISSUE #1358: [Feature Request] Configurable workflow failure exception types

See https://github.com/temporalio/features/issues/322

mjameswh:
## What was changed

- Add support for failing workflow _execution_ (rather than workflow _task_) on some error types, notably Non-Determinism errors. Resolves #1358.

### DIFF
--- a/packages/test/src/test-workflow-fail-on-errors-policy.ts
+++ b/packages/test/src/test-workflow-fail-on-errors-policy.ts
@@ -1,0 +1,250 @@
+import asyncRetry from 'async-retry';
+import type { WorkflowHandle } from '@temporalio/client';
+import { WorkflowFailedError } from '@temporalio/client';
+import * as workflow from '@temporalio/workflow';
+import { ApplicationFailure } from '@temporalio/common';
+import { helpers, makeTestFunction } from './helpers-integration';
+
+const test = makeTestFunction({
+  workflowsPath: __filename,
+});
+
+////////////////////////////////////////////////////////////////////////////////
+// Test fixtures: custom error classes and workflows
+////////////////////////////////////////////////////////////////////////////////
+
+export class CustomWorkflowError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'CustomWorkflowError';
+  }
+}
+
+export class CustomWorkflowSubError extends CustomWorkflowError {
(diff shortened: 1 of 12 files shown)


lg-epitech:
Would love this feature, great work 🚀

packages/test/src/test-workflow-fail-on-errors-policy.ts
```diff
@@ -0,0 +1,250 @@
...
+  const worker = await createWorker({
+    maxCachedWorkflows: 0,
+    workflowFailureErrorTypes: { '*': ['NondeterminismError'] },
+  });
+  await worker.runUntil(async () => {
+    const err = (await t.throwsAsync(executeWorkflow(nondeterministicWorkflow), {
+      instanceOf: WorkflowFailedError,
+    })) as WorkflowFailedError;
+    t.true(err.cause instanceof ApplicationFailure);
+    t.regex(err.cause?.message ?? '', /[Nn]ondeterminism/);
```
chris-olszewski, line 248: Do we know what is triggering the capitalization change?

packages/worker/src/worker-options.ts
```diff
@@ -530,6 +530,37 @@ export interface WorkerOptions {
...
+   * including non-determinism errors.
+
+   * Note that {@link TemporalFailure} subclasses (with the exception of {@link ApplicationFailure})
+   * and cancellation errors that bubbles out of the Workflow always fail the Workflow Execution,
+   * regardless of either this and the {@link WorkflowDefinitionOptions.failureExceptionTypes} settings.
+   * 
+   * @experimental
+   */
+  workflowFailureErrorTypes?: Record<
+    '*' | string,
```
chris-olszewski: Should we do a similar trick for the key?
```suggestion
    '*' | (string & {}),
```

chris-olszewski (approved):
Makes sense to me. 2 very minor questions

packages/worker/src/worker-options.ts
mjameswh: Good catch. Fixed.

MatejFrnka:
Nice, thank you for this feature. Any idea when the next release is coming?

PR #2024, opened 2026-04-24
feat(openai-agents): add @temporalio/openai-agents package

xumaple:
## Summary

Adds `@temporalio/openai-agents`, a Temporal plugin that runs OpenAI Agents SDK agents as durable workflows. The agent loop (model reasoning, tool dispatch, handoffs) executes deterministically inside the workflow sandbox; each model invocation is dispatched to a Temporal activity where the real `ModelProvider` lives.


contrib/openai-agents/README.md
```diff
@@ -0,0 +1,480 @@
+# OpenAI Agents SDK Integration for Temporal (TypeScript)
+
+Run [OpenAI Agents SDK](https://github.com/openai/openai-agents-js) agents as Temporal workflows. The agent loop (model reasoning, tool dispatch, handoffs) runs durably in a workflow, model calls run as Temporal activities, and the whole thing survives worker crashes, retries failed model calls automatically, and scales horizontally.
+
+## Install
+
+```bash
+npm install @temporalio/openai-agents @openai/agents-core @openai/agents-openai openai
```
xumaple: Is pnpm the proper command to use here? What's the convention, to match our internal tooling or go with the common one?

brianstrauch: My vote would be to go with the more common one.

brianstrauch: (See the root README)

contrib/openai-agents/src/worker/activities.ts
```diff
@@ -0,0 +1,140 @@
...
+      const stopHeartbeat = startAdaptiveHeartbeat();
+
+      try {
+        const response = await model.getResponse(fromSerializedModelRequest(input.request));
+        return toSerializedModelResponse(response);
+      } catch (error) {
+        if (error instanceof APIError) {
+          const status = error.status;
+          const headers = error.headers;
+
```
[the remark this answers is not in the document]
xumaple: bad comment. We don't want to play blind wack-a-mole with error APIs, if openAI comes out with a new error type or interface we will support that when the time comes.

contrib/workflow-polyfills/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+if (inWorkflowContext()) {
+  // Headers polyfill
+  if (typeof globalThis.Headers === 'undefined') {
+    (globalThis as any).Headers = Headers;
+  }
+
+  // ReadableStream polyfill
+  // eslint-disable-next-line @typescript-eslint/no-require-imports,import/no-unassigned-import
+  require('web-streams-polyfill/polyfill');
```
brianstrauch: I think some of these are quite big and a common directory within contrib/ might be a better place for this file. Currently, anyone who imports `@temporalio/workflow` gets this whether they want it or not.

contrib/openai-agents/src/workflow/convert-agent.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (typeof t === 'function') {
+      throw ApplicationFailure.create({
+        message:
+          `Agent '${agent.name}': Provided tool is a raw function, not a tool object. ` +
+          'Did you mean to use tool() or activityAsTool()?',
+        type: 'AgentsWorkflowError',
+        nonRetryable: true,
+      });
+    }
+  }
```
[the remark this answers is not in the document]
xumaple: FunctionTools are supported. PR description is updated now

packages/openai-agents/src/worker/plugin.ts
```diff
@@ -0,0 +1,147 @@
...
+  // Inject the workflow trace-interceptor module so pre-built bundles pick it
+  // up without callers wiring workflowInterceptorModules themselves.
+  override configureBundler(options: BundleOptions): BundleOptions {
+    return super.configureBundler({
+      ...options,
+      workflowInterceptorModules: [
+        ...(options.workflowInterceptorModules ?? []),
+        require.resolve('../workflow/trace-interceptor'),
+      ],
+    });
```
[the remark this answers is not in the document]
xumaple: Verified the bundler de-dupes via [...new Set(...)] (packages/worker/src/workflow/bundler.ts:168),
    so the duplication wasn't functionally observable. But the configureBundler override was redundant —
    SimplePlugin.configureBundler already appends workerInterceptors.workflowModules. Removed the dead override.

packages/openai-agents/src/workflow/prng.ts
```diff
@@ -0,0 +1,34 @@
+// Agent-SDK-shaped span ID derived from a Temporal-supplied unique input ID
+// (queryId/updateId), orthogonal to the workflow PRNG stream.
+export function spanIdFromInputId(inputId: string): string {
+  const hex = inputId.toLowerCase().replace(/[^0-9a-f]/g, '');
+  return `span_${hex.padEnd(24, '0').slice(0, 24)}`;
```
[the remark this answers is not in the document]
xumaple: Good comment. fixed.

contrib/openai-agents/src/workflow.ts
```diff
@@ -0,0 +1,12 @@
+// Workflow-safe exports — importable from workflow code running in the V8 sandbox.
+export { TemporalOpenAIRunner } from './workflow/runner';
+export type { TemporalRunOptions } from './workflow/runner';
```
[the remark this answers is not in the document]
xumaple: PR description was stale, updated

contrib/openai-agents/package.json
```diff
@@ -0,0 +1,104 @@
...
+      "import": "./lib/workflow/load-polyfills.js",
+      "require": "./lib/workflow/load-polyfills.js",
+      "default": "./lib/workflow/load-polyfills.js"
+    },
+    "./workflow-interceptor": {
+      "types": "./lib/workflow/trace-interceptor.d.ts",
+      "import": "./lib/workflow/trace-interceptor.js",
+      "require": "./lib/workflow/trace-interceptor.js",
+      "default": "./lib/workflow/trace-interceptor.js"
+    }
```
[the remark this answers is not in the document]
xumaple: PR description was stale, updated

packages/openai-agents/src/worker/activities.ts
```diff
@@ -0,0 +1,140 @@
...
+export function createModelActivity(modelProvider: ModelProvider): {
+  invokeModelActivity: (input: InvokeModelActivityInput) => Promise<SerializedModelResponse>;
+} {
+  return {
+    async invokeModelActivity(input: InvokeModelActivityInput): Promise<SerializedModelResponse> {
+      const model = await Promise.resolve(modelProvider.getModel(input.modelName));
+
+      const stopHeartbeat = startAdaptiveHeartbeat();
+
+      try {
```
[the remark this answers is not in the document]
xumaple: fixed

contrib/workflow-polyfills/src/index.ts
xumaple: Good suggestion. done.

pnpm-workspace.yaml
```diff
@@ -19,6 +19,8 @@ packages:
   - packages/workflow
   - contrib/ai-sdk
   - contrib/interceptors-opentelemetry
+  - contrib/openai-agents
+  - contrib/workflow-polyfills
```
brianstrauch: It looks like ai-sdk and openai-agents both use everything inside workflow-polyfills, but will that always be the case? I wonder if there might be cases where we're overimporting.

By the way, I'm discussing (with @chris-olszewski in https://github.com/temporalio/sdk-typescript/pull/2039#discussion_r3249788983) the benefits of polyfills vs native classes. Maybe that change could fix any issues with the size of the polyfills.

packages/workflow/src/workflow.ts
```diff
@@ -1143,6 +1144,33 @@ export function uuid4(): string {
...
+ * core-provided randomness seed.
+ *
+ * Calling this does not consume `Math.random()` state, and the returned RNG's
+ * output stream is unaffected by subsequent `Math.random()` calls. The RNG
+ * auto-reseeds when core sends an `updateRandomSeed` activation, so callers
+ * may cache the returned reference across activations. When `updateRandomSeed`
+ * fires, the RNG resets to the start of the new seed's stream — any prior
+ * position is discarded.
+ *
+ * Mirrors Python's `workflow.new_random()`.
```
brianstrauch: I think SDK compatibility is assumed, although I suppose it's good to know that this exists in sdk-python for the purposes of reviewing this PR
```suggestion
```

packages/test/tsconfig.json
```diff
@@ -52,6 +52,12 @@
     },
     {
       "path": "../nyc-test-coverage"
+    },
+    {
+      "path": "../../contrib/ai-sdk"
```
brianstrauch: Interesting that this wasn't being included before. Can we just include `../../contrib` or is that bad practice?

contrib/openai-agents/src/worker/otel-sink-bridge.ts
```diff
@@ -0,0 +1,112 @@
...
+    kind: (s.kind as otel.SpanKind) ?? otel.SpanKind.INTERNAL,
+    spanContext: () => spanContext,
+    parentSpanId: s.parentSpanId,
+    startTime: s.startTime,
+    endTime: s.endTime,
+    status: s.status as otel.SpanStatus,
+    attributes: s.attributes as otel.Attributes,
+    links: s.links.map(reconstructLink),
+    events: s.events.map(reconstructEvent),
+    duration: [s.endTime[0] - s.startTime[0], s.endTime[1] - s.startTime[1]],
```
brianstrauch: The format here is [seconds, nanoseconds] so if you did [1, 0] - [0, 1] you would get [1, -1] instead of [0, 999999999]

brianstrauch: Claude's fix: use hrTimeDuration from @opentelemetry/core

brianstrauch:
Other thoughts:
* I assume you're going to wait for #2039
* Did you consider adding nexusOperationAsTool and workflowMethodAsTool for Python compatibility?

packages/test/tsconfig.json
xumaple: These are project references, not include globs — they need an explicit tsconfig.json target, which we don't have in contrib/. But also, I think it's missing maybe because ai-sdk was only very recently moved out of packages/ into the new contrib/ dir, so I guess it was just missed. In the future, AI should be very capable of following the pattern for future contrib stuff so I'm not too worried about this. Let me know if you think otherwise.

contrib/ai-sdk/src/load-polyfills.ts
```diff
@@ -1,18 +1,3 @@
...
-  // Attach the polyfill as a Global function
-  if (!('structuredClone' in globalThis)) {
-    // eslint-disable-next-line @typescript-eslint/no-require-imports
-    const structuredClone = require('@ungap/structured-clone');
-    globalThis.structuredClone = structuredClone.default;
-  }
-}
+// Web-API polyfills (Headers, ReadableStream, structuredClone, crypto.randomUUID)
+// eslint-disable-next-line import/no-unassigned-import
+import '@temporalio/workflow-polyfills';
```
chris-olszewski: This polyfill contains more than what is necessary for the AI SDK integration. IMO we should polyfill the bare minimum

contrib/workflow-polyfills/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  // structuredClone polyfill
+  if (!('structuredClone' in globalThis)) {
```
chris-olszewski: I'm not sure sharing a few lines of polyfill are worth a new public facing package. I would rather dupe these so each package gets exactly what it needs.

contrib/openai-agents/package.json
```diff
@@ -0,0 +1,105 @@
...
+    "url": "git+https://github.com/temporalio/sdk-typescript.git",
+    "directory": "contrib/openai-agents"
+  },
+  "homepage": "https://github.com/temporalio/sdk-typescript/tree/main/contrib/openai-agents",
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "src",
+    "lib"
```
chris-olszewski: Strip out the test files from the shipped package
```suggestion
    "src",
    "lib",
    "!src/__tests__",
    "!lib/__tests__"

```

contrib/openai-agents/package.json
```diff
@@ -0,0 +1,105 @@
...
+      "require": "./lib/index.js",
+      "default": "./lib/index.js"
+    },
+    "./workflow": {
+      "types": "./lib/workflow.d.ts",
+      "import": "./lib/workflow.js",
+      "require": "./lib/workflow.js",
+      "default": "./lib/workflow.js"
+    },
+    "./load-polyfills": {
```
chris-olszewski: Do we need to expose the polyfills and require users to manually import them? Could we copy the pattern from the AI SDK and import the polyfills in entrypoints of the package?

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,456 @@
+# OpenAI Agents SDK Integration for Temporal
+
+Run [OpenAI Agents SDK](https://github.com/openai/openai-agents-js) agents as Temporal Workflows. You write agents the same way: same `Agent` class, same `tool()`s, same handoffs, guardrails, and tracing APIs. The plugin runs the agent loop in a Workflow and dispatches each model call as an Activity, so Workflows survive Worker crashes and model calls retry automatically. Workflows replay safely, meaning no duplicate model calls.
+
+## Install
+
+```bash
+npm install @temporalio/openai-agents @openai/agents-core @openai/agents-openai openai
```
chris-olszewski, line 11: Just to cover the big 3
```suggestion
# Or `pnpm add`/`yarn add`
npm install --save @temporalio/openai-agents @openai/agents-core @openai/agents-openai openai
```

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,456 @@
...
+
+Three files: a Workflow, a Worker, and a Client.
+
+### Workflow (`workflows.ts`)
+
+```ts
+// MUST be the first import in any Workflow file using this package.
+// It installs EventTarget/Event/CustomEvent and a working AsyncLocalStorage
+// that the agent SDK needs before any agent-SDK code is evaluated.
+import '@temporalio/openai-agents/load-polyfills';
```
chris-olszewski: Does `@openai/agents-core` do stuff on import that requires our polyfills?

contrib/openai-agents/package.json
```diff
@@ -0,0 +1,105 @@
...
+    "workflow",
+    "ai",
+    "openai",
+    "agents"
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT",
+  "dependencies": {
+    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/sdk-trace-base": "^1.25.1",
```
chris-olszewski: Unsure if worth the squeeze, but this is forcing all users to take OTEL dependencies even if they don't enable them. For Lambda we defined a separate entrypoint that allows us to declare these as optional peer deps: https://github.com/temporalio/sdk-typescript/blob/d2d307ca87e6cf16c23621f52a1697b764a59f90/packages/lambda-worker/package.json#L12

chris-olszewski (changes requested):
Partial review, sorry I didn't get too far past the package level. Will pick up review this afternoon.

contrib/workflow-polyfills/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@temporalio/workflow-polyfills",
```
mjameswh: Please don't introduce a new package for this just yet. For now, duplicate the polyfill usage in each package.

It is very likely that third parties will duplicate the code of our `ai-sdk` and `openai-agents` packages as a base to develop other AI integrations. And so the `@temporalio/workflow-polyfills` package would implicitly become a public contract, from which we'll never be able to walk back without risking breaking those third-party packages.

There's clearly a need for us to provide some of those polyfills out-of-the-box, but we'll have to think this holistically.

packages/workflow/src/workflow.ts
```diff
@@ -1143,6 +1144,33 @@ export function uuid4(): string {
...
+ */
+export function newRandom(): RNG {
```
mjameswh: Don't. We already have WIP to add a proper API for this: #1992.

contrib/openai-agents/src/__tests__/test-openai-agents.ts
```diff
@@ -0,0 +1,433 @@
+// Integration tests that need a real Worker. Pure-function tests live in
```
mjameswh, line 1: Those tests belong into `contrib/openai-agents/src/__tests__`.

packages/test/tsconfig.json
mjameswh: > Can we just include ../../contrib or is that bad practice?

Not possible.

> But also, I think it's missing maybe because ai-sdk was only very recently moved out of packages/ into the new contrib/ dir, so I guess it was just missed.

A contrib's tests should now be defined inside that contrib's `__test__` directory. As a result, there should be no need for this reference.

mjameswh (dismissed):

pnpm-workspace.yaml
mjameswh: Don't add a `workflow-polyfills` package just now. See my other comment for details.

packages/workflow/src/workflow.ts
mjameswh: Don't duplicate Python's `new_random()` API. See my other comment for details.

mjameswh:
Note that I only skimmed over the actual `@temporalio/openai-agents` package, as that's under the AI team's responsibility. I raised several concerns regarding things outside of that package.

mjameswh:
You'll need to add the `contrib/openai-agents` path to the `.github/CODEOWNERS` file.

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,698 @@
...
+
+  console.log(result);
+}
+
+main();
+```
+
+## The runner
+
+`TemporalOpenAIRunner` is a drop-in replacement for Agent SDK's `Runner` class. `run(agent, input, options?)` takes the same options (`maxTurns`, `context`, `session`, guardrails via `runConfig`, etc.) with three exceptions:
```
chris-olszewski:
```suggestion
`TemporalOpenAIRunner` is a drop-in replacement for Agent SDK's `Runner` class. `run(agent, input, options?)` takes the same options with three exceptions:
```

contrib/openai-agents/src/workflow/trace-interceptor.ts
```diff
@@ -0,0 +1,322 @@
...
+}
+
+/** Outbound interceptor: injects agent trace/span context into outbound headers under `__openai_span` and (for child Workflows / continueAsNew) the plugin config under `__openai_agents_config`. */
+export class OpenAIAgentsTraceOutboundInterceptor implements WorkflowOutboundCallsInterceptor {
+  async scheduleActivity(
+    input: ActivityInput,
+    next: Next<WorkflowOutboundCallsInterceptor, 'scheduleActivity'>
+  ): Promise<unknown> {
+    // Fire-and-forget infra Activities (e.g. stateful MCP session) need an
+    // INSTANT span — a lifetime span would swallow every subsequent Activity
```
tconley1428: It sounds like that should just be a span which isn't put on the current context, not an instant one.

contrib/openai-agents/src/worker/agent-sink-bridge.ts
```diff
@@ -0,0 +1,110 @@
...
+    parent
+  );
+  if (!event.forwardToOtel) markSuppressOtelBridge(span);
+  span.start();
+  if (event.span.error) span.setError(event.span.error);
+  span.end();
+}
+
+async function dispatchToHostProcessors(event: AgentTracingSinkEvent): Promise<void> {
+  if (event.kind === 'trace_start') {
```
tconley1428: Use a switch?

contrib/openai-agents/src/worker/agent-sink-bridge.ts
```diff
@@ -0,0 +1,110 @@
...
+    return;
+  }
+  if (event.kind === 'span_complete') {
+    dispatchSpanComplete(event);
+    return;
+  }
+}
+
+export function makeAgentTracingSink(): InjectedSink<AgentTracingSink> {
+  return {
```
tconley1428, line 114: There seems to be a lot of shenanigans to try to pipe through on the event whether it should be emitted to otel. That can just be part of the sink construction. Or even separate sinks.

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,698 @@
...
+
+- `runConfig.model` takes a model name string, not a `Model` instance. The `ModelProvider` you register on the Worker (via the plugin's `modelProvider` option) resolves the name to a `Model` inside the model Activity, so the `Model` instance itself never has to cross the Workflow/Activity boundary.
+- No `signal`. Cancel the Workflow with Temporal's `CancellationScope` instead. Passing `signal` is a type error.
+- No streaming. `runStreamed()` is not supported yet.
+
+### Run state
+
+`RunState` is the Agent SDK's snapshot of an in-flight run: input items, generated items, the current Agent, pending interruptions, etc. `TemporalOpenAIRunner.run` accepts a `RunState` directly as the second argument, mirroring upstream `Runner.run` — useful for human-in-the-loop approval flows where the run must pause across a `continueAsNew` boundary while a Signal or Update collects approval.
```
chris-olszewski:
```suggestion
`RunState` is the Agent SDK's snapshot of an in-flight run. `TemporalOpenAIRunner.run` accepts a `RunState` directly as the second argument, mirroring upstream `Runner.run` — useful for human-in-the-loop approval flows where the run must pause across a `continueAsNew` boundary while a Signal or Update collects approval.
```

contrib/openai-agents/src/worker/agent-sink-bridge.ts
xumaple: yes, I've completely reworked the sync stuff, haven't pushed the rework yet.

xumaple: I think we do need some sort of flag for whether to pipe an event to otel though, that gets plumbed all the way. Because we could have one worker managing multiple workflows that have different settings on whether otel gets mirrored or not. 

But it should be a simple bool flag on the span now in the most recent push, no more multiple paths for everything.

contrib/openai-agents/src/__tests__/test-openai-agents.ts
xumaple: moved the tests.

contrib/workflow-polyfills/package.json
xumaple: done.

contrib/openai-agents/src/worker/agent-sink-bridge.ts
xumaple: done

contrib/openai-agents/src/workflow/trace-interceptor.ts
xumaple: changed this to a span not on current context. thanks for the suggestion

packages/workflow/src/workflow.ts
xumaple: waiting on this PR to be completed.

contrib/openai-agents/package.json
xumaple: I moved the sdk-trace-base to an opt-in, but the opentelemtry/api stuff is pretty tightly coupled to the entire tracing module, I left it alone. Is that okay for you?

contrib/openai-agents/README.md
xumaple: not sure I fully understand this comment, but I did refactor this so the plugin prebundles the imports and user no longer needs to do this weird "first import" thing.

contrib/openai-agents/README.md
xumaple: done

contrib/openai-agents/package.json
xumaple: done

contrib/workflow-polyfills/src/index.ts
xumaple: duped

contrib/ai-sdk/src/load-polyfills.ts
xumaple: undid all this

contrib/openai-agents/package.json
xumaple: I think I did this properly, but not sure... let me know if something is missing.

contrib/openai-agents/src/worker/otel-sink-bridge.ts
xumaple: done

packages/test/tsconfig.json
xumaple: reverted this so AI-SDK shouldn't be affected in this change at all anymore.

contrib/openai-agents/src/workflow/trace-interceptor.ts
xumaple: Hm actually this is not possible based on the current API that openai exposes for us. because we don't have a way of pushing out a span until it's completed. Will wait on them to give us a new API before I call this completed

contrib/openai-agents/package.json
```diff
@@ -0,0 +1,132 @@
...
+    "@ungap/structured-clone": "^1.3.1",
+    "headers-polyfill": "^4.0.3",
+    "web-streams-polyfill": "^4.2.0"
+  },
+  "peerDependencies": {
+    "@openai/agents-core": "~0.11.5",
+    "@openai/agents-openai": "~0.11.5",
+    "@opentelemetry/sdk-trace-base": "^1.25.1",
+    "openai": "^6.35.0",
+    "zod": "^4.0.0"
```
chris-olszewski: Unsure if this was from before, but I'm not finding any usage of `zod` in this package so I don't think it should be a peer dependency.

contrib/openai-agents/README.md
xumaple: readme was completely rewritten. marking this resolved

contrib/openai-agents/README.md
xumaple: readme was rewritten. marking this resolved

contrib/openai-agents/package.json
xumaple: removed it

contrib/openai-agents/src/workflow/trace-interceptor.ts
xumaple: fixed after openai gave us a new API

packages/workflow/src/workflow.ts
xumaple: done

packages/workflow/package.json
```diff
@@ -24,7 +24,10 @@
   "dependencies": {
     "@temporalio/common": "workspace:*",
     "@temporalio/proto": "workspace:*",
-    "nexus-rpc": "^0.0.2"
+    "@ungap/structured-clone": "^1.3.0",
```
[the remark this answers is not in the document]
xumaple: done

contrib/openai-agents/src/client/trace-interceptor.ts
```diff
@@ -0,0 +1,135 @@
...
+  addTemporalSpans?: boolean;
+  /** Emit OTel spans on the Workflow side for agent-SDK trace/span events. @default false */
+  useOtelInstrumentation?: boolean;
+  /** Model Activity options propagated to the Workflow via the config header. */
+  modelParams?: SerializableModelActivityOptions;
+}
+
+// Concrete `next` types: avoid the circular `Next<this, ...>` resolution that
+// triggers when a class method's own signature feeds back into `Next`.
+type NextStartWithDetails = (input: WorkflowStartInput) => Promise<WorkflowStartOutput>;
```
tconley1428: Not clear to me why this would be needed for the first time.

contrib/openai-agents/src/common/mcp-types.ts
```diff
@@ -0,0 +1,89 @@
...
+  description?: string;
+  arguments?: Array<{
+    name: string;
+    description?: string;
+    required?: boolean;
+  }>;
+}
+
+/** Shape registered with `StatelessMCPServerProvider` to serve stateless MCP requests. */
+export interface StatelessMCPServerFactory {
```
tconley1428: I'm not sure why these types exist. This clearly isn't a factory, and no such interface equivalent is in python

contrib/openai-agents/src/common/mcp-types.ts
```diff
@@ -0,0 +1,89 @@
...
+  cleanup(): Promise<void>;
+  listTools(): Promise<MCPToolDefinition[]>;
+  callTool(toolName: string, args: Record<string, unknown> | null): Promise<MCPCallToolResult[]>;
+  listPrompts?(): Promise<MCPPromptDefinition[]>;
+  getPrompt?(name: string, args: Record<string, unknown> | null): Promise<unknown>;
+}
+
+/** Workflow-side MCP server handle. */
+export interface TemporalMCPServer extends MCPServer {
+  listPrompts(factoryArgument?: unknown): Promise<MCPPromptDefinition[]>;
```
tconley1428: I don't see how or why we would be _adding_ prompt support

contrib/openai-agents/src/common/model-activity-options.ts
```diff
@@ -0,0 +1,50 @@
...
+export interface ModelSummaryProvider {
+  /** Generate a human-readable summary for the model Activity shown in the Temporal UI. */
+  provide(
+    agent: Agent<any, any> | undefined,
+    instructions: string | undefined,
+    input: string | AgentInputItem[]
+  ): string;
+}
+
+export interface ModelActivityOptions {
```
tconley1428: This should probably be based on the existing activity options. In TS we can replace the summary.

contrib/openai-agents/src/common/model-activity-options.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+export type SerializableModelActivityOptions = Omit<ModelActivityOptions, 'summaryOverride'> & {
+  /** Static summary string used as a tracing summary override. */
+  summaryOverride?: string;
+};
+
+export const DEFAULT_MODEL_ACTIVITY_OPTIONS: SerializableModelActivityOptions = {
+  startToCloseTimeout: '60s',
+  useLocalActivity: false,
+  cancellationType: ActivityCancellationType.TRY_CANCEL,
```
tconley1428: Why are you setting this at all?

contrib/openai-agents/src/common/trace-context.ts
```diff
@@ -0,0 +1,91 @@
...
+    spanId: header.otelSpanId,
+    traceFlags: otel.TraceFlags.SAMPLED,
+    isRemote: true,
+  };
+  const ctx = otel.trace.setSpanContext(otel.context.active(), otelSpanContext);
+  return otel.context.with(ctx, restored);
+}
+
+/**
+ * Fresh ALS scope preserving current trace, clearing current span. Used by
```
tconley1428: ALS?

contrib/openai-agents/src/common/tracing-bridge.ts
```diff
@@ -0,0 +1,142 @@
...
+  const hex = agentSpanId.startsWith('span_') ? agentSpanId.slice(5) : agentSpanId.replace(/-/g, '');
+  return hex.padStart(16, '0').slice(0, 16).toLowerCase();
+}
+
+/** OTel span ID for a trace's root span, derived deterministically from the agent-SDK trace ID so host and Workflow sides agree without explicit propagation. */
+export function agentTraceIdToOtelRootSpanId(agentTraceId: string): string {
+  return agentTraceIdToOtelTraceId(agentTraceId).slice(0, 16);
+}
+
+export function spanNameFromData(data: SpanData): string {
```
tconley1428, line 41: This seems like a strange function to need.

contrib/openai-agents/src/worker/seeded-ids.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export const seedsStorage = new AsyncLocalStorage<SeededIdStore>();
+
+/** Scopes `traceSeed`/`spanSeed` for one `tracer.startSpan(...)` call. */
+export function withSeededIds<T>(traceSeed: string | undefined, spanSeed: string | undefined, fn: () => T): T {
+  return seedsStorage.run({ trace: traceSeed, span: spanSeed }, fn);
+}
+
+export function randomHex(len: number): string {
```
tconley1428: This seems like an odd function to have here.

contrib/openai-agents/src/common/tracing-bridge.ts
```diff
@@ -0,0 +1,142 @@
...
+  return hex.padStart(32, '0').slice(0, 32).toLowerCase();
+}
+
+export function agentSpanIdToOtelSpanId(agentSpanId: string): string {
+  const hex = agentSpanId.startsWith('span_') ? agentSpanId.slice(5) : agentSpanId.replace(/-/g, '');
+  return hex.padStart(16, '0').slice(0, 16).toLowerCase();
+}
+
+/** OTel span ID for a trace's root span, derived deterministically from the agent-SDK trace ID so host and Workflow sides agree without explicit propagation. */
+export function agentTraceIdToOtelRootSpanId(agentTraceId: string): string {
```
tconley1428, line 37: This is a different methodology than Python. Can you elaborate on why?

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,610 @@
...
+
+> ⚠️ **This package is currently at an experimental release stage.** ⚠️
+
+Run [OpenAI Agents SDK](https://github.com/openai/openai-agents-js) agents as Temporal Workflows. Agent orchestration runs in the Workflow; model calls run as Activities, so they retry durably and are not repeated during Workflow replay.
+
+## Install
+
+```bash
+# Or `pnpm add`/`yarn add`
+npm install --save @temporalio/openai-agents @openai/agents-core @openai/agents-openai openai
```
brianstrauch: > As of [npm 5.0.0](http://blog.npmjs.org/post/161081169345/v500), installed modules are added as a dependency by default, so the --save option is no longer needed.

Hasn't been needed since 2017

```suggestion
npm install @temporalio/openai-agents @openai/agents-core @openai/agents-openai openai
```

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,610 @@
...
+
+**Calling `setDefaultOpenAITracingExporter()` after constructing the plugin resets and breaks the plugin's tracing configurations.**
+
+### OpenTelemetry
+
+Install `@opentelemetry/sdk-trace-base` before importing the OTel subpath:
+
+```bash
+# Or `pnpm add`/`yarn add`
+npm install --save @opentelemetry/sdk-trace-base
```
brianstrauch:
```suggestion
npm install @opentelemetry/sdk-trace-base
```

xumaple: chris asked for the --save so I will default to that

xumaple: oh, just saw the other comment... will check in with him on convention for this repo

brianstrauch: Maybe we still support npm 5? If so, I'll add `--save` to my PRs too.

contrib/openai-agents/src/client/trace-interceptor.ts
xumaple: removed all this

contrib/openai-agents/src/common/model-activity-options.ts
xumaple: done

contrib/openai-agents/src/common/trace-context.ts
xumaple: asynclocalstorage. wrote out the acronym now

contrib/openai-agents/src/worker/seeded-ids.ts
xumaple: moved

contrib/openai-agents/src/common/model-activity-options.ts
xumaple: removed

contrib/openai-agents/src/common/mcp-types.ts
xumaple: removed

contrib/openai-agents/src/common/mcp-types.ts
```diff
@@ -0,0 +1,29 @@
+import type { MCPServer } from '@openai/agents-core';
+
+export const MCP_LIST_TOOLS_SUFFIX = '-list-tools';
+export const MCP_CALL_TOOL_SUFFIX = '-call-tool';
+export const MCP_SERVER_SESSION_SUFFIX = '-server-session';
+export const MCP_STATEFUL_SUFFIX = '-stateful';
```
DABH, line 6: Why no `MCP_STATELESS_SUFFIX` for parity?  The current duplicate-name check (worker/plugin.ts:73) only catches collisions between MCP providers themselves, not against the broader activity registry. Adding MCP_STATELESS_SUFFIX = '-stateless' would give party and close the collision I think.

contrib/openai-agents/src/worker/seeded-ids.ts
DABH: Why have this function here vs. in `otel.ts` next to `TemporalIdGenerator` (the only consumer)?  Consider moving?

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,578 @@
...
+
+Enable the upstream hosted exporter before constructing the plugin:
+
+```ts
+import { setDefaultOpenAITracingExporter } from '@openai/agents-openai';
+
+setDefaultOpenAITracingExporter();
+```
+
+**Calling `setDefaultOpenAITracingExporter()` after constructing the plugin resets and breaks the plugin's tracing configurations.**
```
DABH: This seems like a potential footgun here - could we add some checks so that users cannot do this (or they get an error when they try to?)?

contrib/openai-agents/src/common/tracing-bridge.ts
xumaple: in TS because we need to emit spans through sinks, and the sinks pull span/trace events asynchronously, there's a chance that spans appear in the sink before their root trace. So we define that worker and workflow each derives the OTel trace-id and root-span-id deterministically from the agent's trace-id to solve this.

contrib/openai-agents/src/workflow/activity-backed-model.ts
```diff
@@ -0,0 +1,126 @@
...
+      const input: InvokeModelActivityInput = {
+        modelName: this.modelName,
+        request: wire,
+      };
+
+      const modelSummary = this.modelParams.summary;
+      if (modelSummary && typeof modelSummary !== 'string') {
+        const provider = modelSummary as ModelSummaryProvider;
+        const systemInstructions = request.systemInstructions;
+        const summary = provider.provide(this.agent, systemInstructions, request.input);
```
DABH, line 111: So provider.provide() runs inside getResponse() which executes in workflow code iiuc.

What if the user's provide() calls Date.now()/Math.random()/etc. - non-determinism issue?  Can this be warned about in the README maybe, or does it make sense to run the summary computation on the activity side?

contrib/openai-agents/src/workflow/session.ts
```diff
@@ -0,0 +1,50 @@
...
+  initialItems?: AgentInputItem[];
+}
+
+/**
+ * Workflow-sandbox-safe `Session` for `TemporalOpenAIRunner.run`. Upstream
+ * `MemorySession` depends on Node-only utilities and is rejected at runtime.
+ */
+export class WorkflowSafeMemorySession implements Session {
+  private readonly sessionId: string;
+  private items: AgentInputItem[];
```
DABH, line 16: The session keeps this field on the workflow heap. I haven't tested this, but I think that continueAsNew JSON-serializes its args, dropping class identity, meaning this field won't survive CAN. (Python sessions don't have this issue because they're Pydantic models.) If you verify this behavior, I suggest the README should document explicitly or ideally provide a serialize()/fromSerialized() pair, similar to the `RunState` resume pattern.

contrib/openai-agents/README.md
```diff
@@ -0,0 +1,578 @@
...
+    plugins: [plugin],
+  });
+
+  await worker.run();
+}
+
+main();
+```
+
+Register the same plugin type on the Client so model parameters and tracing options propagate to new Workflows:
```
DABH, line 64: Typically from what I've seen the pattern is registering plugins either on the client or the worker.  Why do we have to register on both here?  Why does the SDK not auto propagate Client->Worker?

contrib/openai-agents/src/worker/seeded-ids.ts
```diff
@@ -0,0 +1,21 @@
...
+  const bytes = new Uint8Array(len / 2);
+  for (let i = 0; i < bytes.length; i++) bytes[i] = Math.floor(Math.random() * 256);
```
DABH: Switch to crypto.randomBytes(len/2) (Node ≥ 14) for cryptographic quality — the current implementation has 4 bits less entropy than expected (Math.random returns 53-bit, but it's being truncated through Math.floor(... * 256)).

contrib/openai-agents/src/worker/plugin.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export type MCPServerProvider = StatelessMCPServerProvider | StatefulMCPServerProvider;
+
+/**
+ * Options controlling trace interceptor behavior on both the Activity side
+ * (directly) and the Workflow side (via header propagation).
+ */
+export interface OpenAIAgentsPluginInterceptorOptions {
+  /** Wrap calls in `temporal:*` custom spans. @default false */
+  addTemporalSpans?: boolean;
```
DABH, line 32: So, if users don't specify anything, I think this comes out `false`, whereas in Python it's `true` by default.  Can we resolve that inconsistency?

If you want cross-SDK parity, the smallest change is one line: flip client/trace-interceptor.ts:39, worker/trace-interceptor.ts:41, and workflow/span-helpers.ts:9 from === true to !== false (and update workflow/trace-interceptor.ts:76 to ?? true), then update the @default false JSDoc on worker/plugin.ts:31 to @default true.

DABH (changes requested):

contrib/openai-agents/src/common/tracing-bridge.ts
xumaple: I think it's needed to derive the name, what's the issue?

contrib/openai-agents/README.md
xumaple: done

contrib/openai-agents/README.md
xumaple: nevermind, chris has said remove is fine

contrib/openai-agents/src/workflow/activity-backed-model.ts
xumaple: Date.now() and Math.random() are both polyfilled in TS to be deterministic. but point taken, will look into documenting no determinism allowed here. I would assume that users would assume so from it being explicitly run from within workflow though..

contrib/openai-agents/README.md
xumaple: this is talking about the starter file, the one that calls client.execute_workflow - that client also needs the plugin so that it can determine the exact set of args/settings to pass to the workflow.

contrib/openai-agents/src/worker/plugin.ts
xumaple: it should already be false in Python by default. That's our consistent standard throughout all telemetry plugins.

contrib/openai-agents/src/worker/seeded-ids.ts
xumaple: this utility is for generating span ids, I think we don't need high entropy for cryptographic security here, as long as no collisions (which we have) it should be fine?

contrib/openai-agents/src/workflow/session.ts
xumaple: I believe the serialize/deserialize methods on this class are `getItems` and `initialItems` which follow openai's MemorySession pattern (which is the object that ours attempts to emulate with replay safety). So what you're describing here already exists and is the expected path for the continue-as-new use case.

contrib/openai-agents/README.md
xumaple: It requires quite a bit of monkey patching to be able to do this, which is why I opted for documentation instead. 

I've pinged openai to see if they can refactor to make our lives easier. Will keep this open for the time being.

contrib/openai-agents/src/common/mcp-types.ts
xumaple: added

contrib/openai-agents/src/worker/seeded-ids.ts
xumaple: moved

contrib/openai-agents/src/workflow/activity-backed-model.ts
xumaple: added a jsdoc line on ModelSummaryProvider interface's provide() method.

contrib/openai-agents/src/worker/plugin.ts
xumaple: apparently python is actually true by default (checking with Tim on why now) but it doesn't matter, our default should be false for these. will leave this one be.

contrib/openai-agents/src/worker/seeded-ids.ts
DABH: Seems fair to me

DABH (approved):
All my comments have been sufficiently addressed, thanks!

contrib/openai-agents/src/common/mcp-types.ts
xumaple: good call out. I've removed them

contrib/openai-agents/README.md
xumaple: Actually, digging in some more, realized that setDefaultOpenAITracingExporter is simply not the right API to recommend here. Updated the readme to recommend the proper API to achieve what we want here.

chris-olszewski (approved):
All changes to repo config/test helpers look good to me.

PR #2027, opened 2026-04-24
Add standalone Nexus operations client

VegetarianOrc:
## What was changed

Add a standalone Nexus operations client to the TypeScript SDK.

  - New NexusClient sub-client on Client, accessible via client.nexus.
  - NexusClient.createServiceClient({ endpoint, service }) returns a typed NexusServiceClient<T> built from a nexus-rpc
  ServiceDefinition
  - NexusClient.getHandle(operationId, { runId? }) to rehydrate a handle for an existing operation.
  - NexusOperationHandle<O> supports result(), describe(), cancel(), and terminate().
  - Listing / counting APIs: NexusClient.list(options) (async iterable) and NexusClient.count(options).
  - New types in packages/client/src/nexus-types.ts
  - Interceptor support via ClientInterceptors.nexus / NexusClientInterceptor with inputs for start, get-result, describe, cancel, terminate, list, and count.
  - New error classes: NexusOperationFailureError, NexusOperationAlreadyStartedError, NexusOperationNotFoundError.
  - Re-exports wired in packages/client/src/index.ts

  ## Checklist

  1. How was this tested:
      - Added packages/test/src/test-nexus-standalone.ts with integration tests exercising startOperation / executeOperation, result, describe, cancel, terminate, list, count, id-reuse and id-conflict policies, search attributes, and interceptor.
      - Added packages/test/src/test-nexus-standalone-types.ts with compile-time assertions that createServiceClient produces correctly typed start/execute signatures and handles generic inference from nexus.OperationDefinition.
  2. Any docs updates needed?
    - Yes SANO will need documentation for all languages.


packages/client/src/interceptors.ts
```diff
@@ -243,13 +252,102 @@ export type CreateScheduleOutput = {
...
+export interface GetNexusOperationResultInput {
+  readonly operationId: string;
+  readonly runId?: string;
+}
+
+/** Input for {@link NexusClientInterceptor.describe}. */
+export interface DescribeNexusOperationInput {
+  readonly operationId: string;
+  readonly runId?: string;
+  readonly longPollToken?: Uint8Array;
```
Quinn-With-Two-Ns: Do we need to expose this?

VegetarianOrc: I thought the UI needed this from the TS SDK, but I was wrong about that. Will remove it.

Quinn-With-Two-Ns (approved):

packages/client/src/helpers.ts
```diff
@@ -170,6 +170,27 @@ export function getGrpcStatusDetails(err: GrpcServiceError): google.rpc.Status['
...
+/**
+ * Try to extract the runId field from a gRPC ALREADY_EXISTS error's details containing
+ * NexusOperationExecutionAlreadyStartedFailure. Returns undefined if the error details
+ * do not include this failure type or cannot be decoded.
+ */
+export function extractNexusOperationAlreadyStartedRunId(err: GrpcServiceError): string | undefined {
+  try {
+    for (const entry of getGrpcStatusDetails(err) ?? []) {
+      if (!entry.type_url || !entry.value) continue;
+      const type = entry.type_url.replace(/^type.googleapis.com\//, '') as ErrorDetailsName;
```
chris-olszewski:
```suggestion
      const type = entry.type_url.replace(/^type.googleapis.com\//, '');
```

packages/test/src/test-nexus-standalone.ts
```diff
@@ -0,0 +1,492 @@
...
+    // Visibility has update delay, repeat query until the operation count is as expected.
+    let groupedCount: NexusOperationExecutionCount = { count: 0, groups: [] };
+    await waitUntil(async () => {
+      groupedCount = await client.nexus.count(`Endpoint="${endpointName}" GROUP BY ExecutionStatus`);
+      return groupedCount.count === 3;
+    }, 10000);
+
+    let count = 0;
+    for (const group of groupedCount.groups) {
+      // Groups should add up to the total count
```
chris-olszewski, line 262: We never actually make the this assertion

packages/test/src/test-nexus-standalone.ts
```diff
@@ -0,0 +1,492 @@
...
+      for (const value of group.groupValues) {
+        t.is(value.type, SearchAttributeType.KEYWORD);
+      }
+    }
+
+    const listed = new Set<string>();
+    for await (const op of client.nexus.list({ query: `Endpoint="${endpointName}"` })) {
+      listed.add(op.operationId);
+    }
+    t.true(opIds.size === listed.size && [...opIds].every((id) => listed.has(id)));
```
chris-olszewski: Minor, but just using an array for ids and sorting before comparison would allow us to use `deepEqual` which provides a far better assertion failure.

packages/client/src/nexus-client.ts
```diff
@@ -0,0 +1,768 @@
...
+      nexusHeader: input.headers ?? {},
+      userMetadata,
+    };
+    let res: temporal.api.workflowservice.v1.IStartNexusOperationExecutionResponse;
+    try {
+      res = await this.connection.workflowService.startNexusOperationExecution(req);
+    } catch (err: unknown) {
+      if (isGrpcServiceError(err) && err.code === grpcStatus.ALREADY_EXISTS) {
+        throw new NexusOperationAlreadyStartedError(input.id, extractNexusOperationAlreadyStartedRunId(err));
+      }
```
chris-olszewski: Would it be worth/useful to add this logic into `rethrowGrpcError` so all callers get this by default?

packages/client/src/nexus-client.ts
```diff
@@ -0,0 +1,768 @@
...
+
+/**
+ * A reference to an operation in a Nexus service.
+ *
+ * Operations may be referenced either by their property key in the service
+ * definition, or by passing the operation definition object itself.
+ */
+type OperationReference<T extends nexus.ServiceDefinition> =
+  | nexus.OperationKey<T['operations']>
+  | T['operations'][nexus.OperationKey<T['operations']>];
```
chris-olszewski: I am sorry I didn't catch this earlier, but we should do what the other nexus client does. This lets some silly things through where you could use an arg type against a different operation.
```
async function _assertion(op: 'mySyncOp' | typeof myService.operations.myOtherOp) {
    const nexusClient = client.nexus.createServiceClient({
      endpoint: 'my-endpoint',
      service: myService,
    });

    // This currently compiles, but it would call mySyncOp with a string if
    // op resolves to 'mySyncOp'.
    const _output: MyOutput | number = await nexusClient.executeOperation(op, 'string-only-input', {
      id: 'op-1',
      scheduleToCloseTimeout: '10s',
    });
```

chris-olszewski (changes requested):
Only real blocker from my perspective it to match the existing Nexus client typing strategy

packages/client/src/nexus-client.ts
VegetarianOrc: Great feedback, tyvm.

I've added this assertion to the type tests and modified the interface for `NexusServiceClient` to use overloads rather than the union types. I also added a few more type assertions to try out both the operation reference path and the key path.

The union types (or the inline equivalent) are still needed to properly type the implementation of the `NexusServiceClient` interface as it has to implement both overloads. Since they're not used anywhere else now I've colocated them there. If you think it'd be better I can remove them and define them inline on the implementing functions. I was on the fence, but I decided that I do think they add clarity to rather than having similar symbols tied into the function definition.

packages/client/src/workflow-client.ts
```diff
@@ -1332,6 +1332,23 @@ export class WorkflowClient extends BaseClient {
       );
     }
 
+    // Server currently only supports workflow_event and batch_job
+    // link types. This filter should be removed or adapted as
+    // server-side support comes online.
+    // See https://github.com/temporalio/temporal/issues/10345
```
chris-olszewski, line 1337: Thinking more about this, we need to make sure server exposes the new behavior behind a capability. Otherwise old server users will not be able to upgrade to a newer SDK.

packages/test/src/test-nexus-standalone-types.ts
```diff
@@ -0,0 +1,247 @@
...
+test('Union of operation output types produces type error', async (t) => {
+  async function _assertion(op: 'mySyncOp' | typeof myService.operations.myOtherOp) {
+    const nexusClient = client.nexus.createServiceClient({
+      endpoint: 'my-endpoint',
+      service: myService,
+    });
+
+    // This currently compiles, but it would call mySyncOp with a string if
+    // op resolves to 'mySyncOp'.
+
```
chris-olszewski:
```suggestion
```

packages/client/src/nexus-types.ts
```diff
@@ -0,0 +1,459 @@
...
+   * If provided, targets this specific run of the operation. If absent, targets the latest run.
+   */
+  runId?: string;
+}
+
+/**
+ * Options for {@link NexusOperationHandle.describe}
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface DescribeNexusOperationOptions {}
```
chris-olszewski, line 459: Just confirming this is here for future extensibility

packages/client/src/workflow-client.ts
```diff
@@ -1331,6 +1331,23 @@ export class WorkflowClient extends BaseClient {
...
+    if (internalOptions?.links != null) {
+      internalOptions.links = internalOptions.links.filter(
```
chris-olszewski: Minor, but prefer constructing a local copy of `links` and using it instead of mutating the inputs to the function.

packages/client/src/workflow-client.ts
```diff
@@ -1331,6 +1331,23 @@ export class WorkflowClient extends BaseClient {
...
+        (link) => link.workflowEvent != null || link.batchJob != null
+      );
+    }
+    if (internalOptions?.completionCallbacks != null) {
+      internalOptions.completionCallbacks?.forEach((cb) => {
```
chris-olszewski: Minor, but prefer constructing a local copy of `completionCallbacks` and using it instead of mutating the inputs to the function.

chris-olszewski (approved):
Overall looks good to me. Would be nice to have a quick follow up for the link filtering once we have an idea of the new server behavior.

packages/client/src/nexus-types.ts
VegetarianOrc: Confirmed

PR #2029, opened 2026-04-28
💥 Standalone Activities support

### LINKED ISSUE #1851: [Feature Request] Support standalone activities

Allow executing activities without a workflow. The clients must support the new (upcoming) server APIs for starting, managing and getting results of activities. The workers must support running activities outside of workflow context.

### Additional context

Server API PR (WIP): https://github.com/temporalio/api/pull/640

maciejdudko:
## 💥 Breaking change

This PR alters Activity Info interface in backward-incompatible way by making `workflowExecution`, `workflowNamespace` and `workflowType` properties optional. As long as Standalone Activities are not being used, it is safe to use `!` to remove type errors, however the recommended migration path is to check the `inWorkflow` property and handle both cases - this way the activity can be run in standalone context.

## What was changed
Implemented support for Standalone Activities.

Telemetry support will be done separately, see #2031

## Why?

Feature request: https://github.com/temporalio/features/issues/706

## Checklist

1. Closes #1851 

2. How was this tested:

- New test file `test-standalone-activities.ts`
- New tests added to `test-async-completion.ts`

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -267,10 +267,14 @@ export interface Info {
    * ID of the current run of this activity. Can be used to differentiate between different activity executions that
    * share the same ID. Activities started by a Workflow don't have activity run ID - instead, they can be identified by
    * workflow ID and workflow run ID; see {@link workflowExecution}
+   *
+   * @experimental Standalone Activities are experimental. APIs may be subject to change.
    */
   readonly activityRunId?: string;
   /**
    * Whether this activity was started by a workflow
+   *
+   * @experimental Standalone Activities are experimental. APIs may be subject to change.
    */
   readonly inWorkflow: boolean;
 }
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -153,11 +153,17 @@ export const asyncLocalStorage: AsyncLocalStorage<Context> = (globalThis as any)
  * Holds information about the current Activity Execution. Retrieved inside an Activity with `Context.current().info`.
  */
 export interface Info {
+  /**
+   * Task token associated with this activity execution. Can be used for asynchronous completion
+   */
   readonly taskToken: Uint8Array;
   /**
-   * Base64 encoded `taskToken`
+   * Base64 encoded {@link taskToken}
    */
   readonly base64TaskToken: string;
+  /**
+   * ID of this activity
+   */
   readonly activityId: string;
   /**
    * Exposed Activity function name
@@ -165,6 +171,8 @@ export interface Info {
   readonly activityType: string;
   /**
    * The namespace this Activity is running in
+   *
+   * @deprecated Use {@link namespace} instead
    */
   readonly activityNamespace: string;
   /**
@@ -176,20 +184,22 @@ export interface Info {
    */
   readonly isLocal: boolean;
   /**
-   * Information about the Workflow that scheduled the Activity
+   * Information about the Workflow that scheduled the Activity. Not set if the activity was not started by a Workflow
    */
-  readonly workflowExecution: {
+  readonly workflowExecution?: {
     readonly workflowId: string;
     readonly runId: string;
   };
   /**
-   * The namespace of the Workflow that scheduled this Activity
+   * The namespace of the Workflow that scheduled this Activity. Not set if the activity was not started by a Workflow
+   *
+   * @deprecated Use {@link namespace} instead
    */
-  readonly workflowNamespace: string;
+  readonly workflowNamespace?: string;
   /**
-   * The module name of the Workflow that scheduled this Activity
+   * The module name of the Workflow that scheduled this Activity. Not set if the activity was not started by a Workflow
    */
-  readonly workflowType: string;
+  readonly workflowType?: string;
   /**
    * Timestamp for when this Activity was first scheduled.
    * For retries, this will have the timestamp of the first attempt.
@@ -249,6 +259,20 @@ export interface Info {
    * version), but it may still be defined server-side.
    */
   readonly retryPolicy?: RetryPolicy;
+  /**
+   * The namespace this Activity is running in
+   */
+  readonly namespace: string;
+  /**
+   * ID of the current run of this activity. Can be used to differentiate between different activity executions that
+   * share the same ID. Activities started by a Workflow don't have activity run ID - instead, they can be identified by
+   * workflow ID and workflow run ID; see {@link workflowExecution}
+   */
+  readonly activityRunId?: string;
+  /**
+   * Whether this activity was started by a workflow
+   */
+  readonly inWorkflow: boolean;
 }
 
 /**
--- a/packages/client/src/activity-client.ts
+++ b/packages/client/src/activity-client.ts
@@ -292,7 +292,7 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
       idReusePolicy: encodeActivityIdReusePolicy(input.options.idReusePolicy),
       idConflictPolicy: encodeActivityIdConflictPolicy(input.options.idConflictPolicy),
       searchAttributes,
-      header: input.headers,
+      header: { fields: input.headers },
       userMetadata: await encodeUserMetadata(this.dataConverter, input.options.summary, undefined),
       priority: input.options.priority ? compilePriority(input.options.priority) : undefined,
     };
--- a/packages/client/src/activity-client.ts
+++ b/packages/client/src/activity-client.ts
@@ -1,0 +1,599 @@
+import { status } from '@grpc/grpc-js';
+import type {
+  LoadedDataConverter,
+  Next,
+  Priority,
+  RetryPolicy,
+  SearchAttributePair,
+  TypedSearchAttributes,
+} from '@temporalio/common';
+import { compilePriority, compileRetryPolicy, decodePriority, decompileRetryPolicy } from '@temporalio/common';
+import type { Duration } from '@temporalio/common/lib/time';
+import { msOptionalToTs, optionalTsToDate, optionalTsToMs } from '@temporalio/common/lib/time';
+import { SymbolBasedInstanceOfError } from '@temporalio/common/lib/type-helpers';
+import { composeInterceptors } from '@temporalio/common/lib/interceptors';
+import {
+  decodeTypedSearchAttributes,
+  encodeUnifiedSearchAttributes,
+  searchAttributePayloadConverter,
+} from '@temporalio/common/lib/converter/payload-search-attributes';
+import { convertDeploymentVersion } from '@temporalio/worker/lib/utils';
+import {
+  decodeArrayFromPayloads,
+  decodeFromPayloadsAtIndex,
+  decodeOptionalFailureToOptionalError,
+  encodeToPayloads,
+  encodeUserMetadata,
+} from '@temporalio/common/lib/internal-non-workflow';
+import { uuid4 } from '@temporalio/workflow';
+import type { temporal } from '@temporalio/proto';
+import type {
+  ActivityCancelInput,
+  ActivityClientInterceptor,
+  ActivityCountInput,
+  ActivityDescribeInput,
+  ActivityGetResultInput,
+  ActivityListInput,
+  ActivityStartInput,
+  ActivityTerminateInput,
+} from './interceptors';
+import type { AsyncCompletionClientOptions } from './async-completion-client';
+import { AsyncCompletionClient } from './async-completion-client';
+import type {
+  ActivityDescription,
+  ActivityExecutionInfo,
+  ActivityIdConflictPolicy,
+  ActivityIdReusePolicy,
+  CountActivityExecutions,
+} from './types';
+import {
+  decodeActivityExecutionStatus,
+  decodePendingActivityState,
+  encodeActivityIdConflictPolicy,
+  encodeActivityIdReusePolicy,
+} from './types';
+import { rethrowKnownErrorTypes } from './helpers';
+import { isGrpcServiceError, ServiceError } from './errors';
+
+/**
+ * Thrown when an Activity with the given ID is not known to Temporal Server.
+ */
+@SymbolBasedInstanceOfError('ActivityNotFoundError')
+export class ActivityNotFoundError extends Error {}
+
+/**
+ * Thrown by the client while waiting on Activity execution result if execution completes with failure.
+ * The failure is stored in the `cause` property.
+ */
+@SymbolBasedInstanceOfError('ActivityExecutionFailedError')
+export class ActivityExecutionFailedError extends Error {
+  constructor(
+    message: string,
+    public readonly cause: Error | undefined,
+    public readonly activityId: string,
+    public readonly runId?: string
+  ) {
+    super(message);
+  }
+}
+
+/**
+ * Options used to configure {@link ActivityClient}
+ */
+export interface ActivityClientOptions extends AsyncCompletionClientOptions {
+  interceptors?: ActivityClientInterceptor[];
+}
+
+/**
+ * Client for starting and managing Activities, and for asynchronous completion and heartbeating of Activities.
+ * Includes all functionality of {@link AsyncCompletionClient}.
+ *
+ * Typically this client should not be instantiated directly, instead create the high level {@link Client} and use
+ * {@link Client.activity} to interact with Activities.
+ */
+export class ActivityClient extends AsyncCompletionClient {
+  private readonly interceptedHandlers: {
+    [K in keyof Required<ActivityClientInterceptor>]: Next<ActivityClientInterceptor, K>;
+  };
+
+  constructor(options?: ActivityClientOptions) {
+    super(options);
+
+    const interceptors = options?.interceptors ?? [];
+    this.interceptedHandlers = {
+      start: composeInterceptors(interceptors, 'start', this.startHandler.bind(this)),
+      getResult: composeInterceptors(interceptors, 'getResult', this.getResultHandler.bind(this)),
+      describe: composeInterceptors(interceptors, 'describe', this.describeHandler.bind(this)),
+      cancel: composeInterceptors(interceptors, 'cancel', this.cancelHandler.bind(this)),
+      terminate: composeInterceptors(interceptors, 'terminate', this.terminateHandler.bind(this)),
+      list: composeInterceptors(interceptors, 'list', this.listHandler.bind(this)),
+      count: composeInterceptors(interceptors, 'count', this.countHandler.bind(this)),
+    };
+  }
+
+  /**
+   * Starts new Standalone Activity execution.
+   *
+   * @param activity Name of the activity to start.
+   * @param options Options controlling the start and execution of the activity.
+   * @returns Handle to the started activity. The handle's `runId` property will be set to the started run.
+   */
+  async start<O = any>(activity: string, options: ActivityOptions): Promise<ActivityHandle<O>> {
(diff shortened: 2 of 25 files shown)


packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+      scheduleToCloseTimeout: msOptionalToTs(input.options.scheduleToCloseTimeout),
+      scheduleToStartTimeout: msOptionalToTs(input.options.scheduleToStartTimeout),
+      startToCloseTimeout: msOptionalToTs(input.options.startToCloseTimeout),
+      heartbeatTimeout: msOptionalToTs(input.options.heartbeatTimeout),
+      retryPolicy: input.options.retry ? compileRetryPolicy(input.options.retry) : undefined,
+      input: { payloads: await encodeToPayloads(this.dataConverter, ...(input.options.args || [])) },
+      idReusePolicy: encodeActivityIdReusePolicy(input.options.idReusePolicy),
+      idConflictPolicy: encodeActivityIdConflictPolicy(input.options.idConflictPolicy),
+      searchAttributes: { indexedFields: searchAttributes },
+      header: input.headers,
```
Sushisource: Additional conversion needs to happen here I think. `input.Headers` doesn't equal the type in the proto. Test should exist to catch this.

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+    const handle = await this.start(activity, options);
+    return handle.result();
+  }
+
+  /**
+   * Creates an Activity handle from ID and optionally from run ID. If `runId` is not set, the handle will refer to the
+   * newest Activity run with the given Activity ID.
+   *
+   * Note 1: this function always succeeds. If the provided ID is invalid, the error will only be thrown when calling
+   * the handle's methods.
```
Sushisource:
```suggestion
   * Note 1: this function always succeeds. If the provided ID is invalid, an error will only be thrown when calling
   * the handle's methods.
```

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+import type { Duration } from '@temporalio/common/lib/time';
+import { msOptionalToTs, optionalTsToDate, optionalTsToMs } from '@temporalio/common/lib/time';
+import { SymbolBasedInstanceOfError } from '@temporalio/common/lib/type-helpers';
+import { composeInterceptors } from '@temporalio/common/lib/interceptors';
+import {
+  decodeTypedSearchAttributes,
+  encodeUnifiedSearchAttributes,
+  searchAttributePayloadConverter,
+} from '@temporalio/common/lib/converter/payload-search-attributes';
+import { convertDeploymentVersion } from '@temporalio/worker/lib/utils';
```
Sushisource: This makes an import circle I think. This function should be shoved down into common

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+   */
+  id: string;
+  /**
+   * Task queue to run this activity on.
+   */
+  taskQueue: string;
+  /**
+   * Input arguments to pass to the activity.
+   */
+  args?: any[];
```
Sushisource: Why aren't we generic over input to make it type safe? Not necessarily blocking just want some context on the decision

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+  /**
+   * If set, specifies maximum time between successful heartbeats.
+   */
+  heartbeatTimeout?: Duration;
+  retry?: RetryPolicy;
```
Sushisource, line 522: Missing docstring

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+  /**
+   * Is set, specifies total time the activity is allowed to run, including retries.
+   *
+   * Note: it is required to set at least one of {@link startToCloseTimeout} and {@link scheduleToCloseTimeout}.
```
Sushisource, line 526: Since this is brand new I think we should take the opportunity to enforce this requirement at compile time.

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+  encodeActivityIdReusePolicy,
+} from './types';
+import { rethrowKnownErrorTypes } from './helpers';
+import { isGrpcServiceError, ServiceError } from './errors';
+
+/**
+ * Thrown when an Activity with the given ID is not known to Temporal Server.
+ */
+@SymbolBasedInstanceOfError('ActivityNotFoundError')
+export class ActivityNotFoundError extends Error {}
```
Sushisource: It looks like this dupes an existing error in `async-completion-client.ts`. We only need one.

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+
+      return {
+        count: resp.count?.toNumber() ?? 0,
+        groups: resp.groups?.map((g) => ({
+          count: g.count?.toNumber() ?? 0,
+          groupValues: g.groupValues?.map((v) => searchAttributePayloadConverter.fromPayload(v)),
+        })),
+      };
+    } catch (err) {
+      this.rethrowGrpcError(err, 'Failed to request activity cancellation');
```
Sushisource: Copy-paste error

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+      scheduleToStartTimeout: msOptionalToTs(input.options.scheduleToStartTimeout),
+      startToCloseTimeout: msOptionalToTs(input.options.startToCloseTimeout),
+      heartbeatTimeout: msOptionalToTs(input.options.heartbeatTimeout),
+      retryPolicy: input.options.retry ? compileRetryPolicy(input.options.retry) : undefined,
+      input: { payloads: await encodeToPayloads(this.dataConverter, ...(input.options.args || [])) },
+      idReusePolicy: encodeActivityIdReusePolicy(input.options.idReusePolicy),
+      idConflictPolicy: encodeActivityIdConflictPolicy(input.options.idConflictPolicy),
+      searchAttributes: { indexedFields: searchAttributes },
+      header: input.headers,
+      userMetadata: await encodeUserMetadata(this.dataConverter, input.options.summary, undefined),
```
Sushisource, line 310: Looks like there's no way to set details?

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+    }
+    validateActivityOptions(input.options);
+
+    try {
+      const resp = await this.workflowService.startActivityExecution(
+        await this.buildStartActivityExecutionRequest(input)
+      );
+      return this.createHandle(input.options.id, resp.runId);
+    } catch (err) {
+      this.rethrowGrpcError(err, 'Failed to start activity');
```
Sushisource, line 282: I think we need an explicit AlreadyStarted error like the other SDKs provide

packages/client/src/types.ts
```diff
@@ -213,3 +220,159 @@ export const [encodeQueryRejectCondition, decodeQueryRejectCondition] = makeProt
...
+  status: ActivityExecutionStatus;
+  typedSearchAttributes: TypedSearchAttributes;
+  taskQueue: string;
+  executionDurationMs?: number;
+}
+
+/**
+ * Return type of {@link ActivityClient.describe}
+ */
+export interface ActivityDescription extends ActivityExecutionInfo {
```
Sushisource: Other SDKs are calling this `ActivityExecutionDescription` which would also be consistent with the other type names I think

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+        }
+      } catch (err) {
+        this.rethrowGrpcError(err, 'Failed to get activity result');
+      }
+
+      if (failedErr) {
+        throw failedErr;
+      }
+    }
+  }
```
dandavison, line 351: It looks like this long-polling loop will fail to wait for longer than the server's long-poll timeout due to not handling DEADLINE_EXCEEDED. Can you compare to the implementations in Go and Python and add an integration test that demonstrates it can wait for longer than the server timeout, and for longer than the per-attempt client-side deadline? And also that it responds correctly when the activity is cancelled on the server. 

**Python**:

```python
        while True:
            try:
                res = await self._client.workflow_service.poll_activity_execution(
                    req,
                    retry=True,
                    metadata=rpc_metadata,
                    timeout=rpc_timeout,
                )
                if res.HasField("outcome"):
                    self._known_outcome = res.outcome
                    return
            except RPCError as err:
                if err.status == RPCStatusCode.DEADLINE_EXCEEDED:
                    # Deadline exceeded is expected with long polling; retry
                    continue
                elif err.status == RPCStatusCode.CANCELLED:
                    raise asyncio.CancelledError() from err
                else:
                    raise
            except asyncio.CancelledError:
                raise
```

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+    ...buildActivityExecutionInfoCommonPart(info),
+    rawInfo: info,
+    runState: decodePendingActivityState(info.runState),
+    scheduleToCloseTimeoutMs: optionalTsToMs(info.scheduleToCloseTimeout),
+    scheduleToStartTimeoutMs: optionalTsToMs(info.scheduleToStartTimeout),
+    startToCloseTimeoutMs: optionalTsToMs(info.startToCloseTimeout),
+    heartbeatTimeoutMs: optionalTsToMs(info.heartbeatTimeout),
+    retryPolicy: decompileRetryPolicy(info.retryPolicy)!,
+    lastHeartbeatTime: optionalTsToDate(info.lastHeartbeatTime),
+    lastStartedTime: optionalTsToDate(info.scheduleTime),
```
dandavison: copy-paste error. (Suggest adding a test that fails for it)

```suggestion
    lastStartedTime: optionalTsToDate(info.lastStartedTime),
```

packages/client/src/activity-client.ts
dandavison: Can you also check that activity cancellation is handled correctly in this loop?

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+      taskQueue: { name: input.options.taskQueue },
+      scheduleToCloseTimeout: msOptionalToTs(input.options.scheduleToCloseTimeout),
+      scheduleToStartTimeout: msOptionalToTs(input.options.scheduleToStartTimeout),
+      startToCloseTimeout: msOptionalToTs(input.options.startToCloseTimeout),
+      heartbeatTimeout: msOptionalToTs(input.options.heartbeatTimeout),
+      retryPolicy: input.options.retry ? compileRetryPolicy(input.options.retry) : undefined,
+      input: { payloads: await encodeToPayloads(this.dataConverter, ...(input.options.args || [])) },
+      idReusePolicy: encodeActivityIdReusePolicy(input.options.idReusePolicy),
+      idConflictPolicy: encodeActivityIdConflictPolicy(input.options.idConflictPolicy),
+      searchAttributes: { indexedFields: searchAttributes },
```
Sushisource: Should be conditionally set iff `searchAttributes` is defined

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,599 @@
...
+  }
+
+  /**
+   * Starts new Standalone Activity execution.
+   *
+   * @param activity Name of the activity to start.
+   * @param options Options controlling the start and execution of the activity.
+   * @returns Handle to the started activity. The handle's `runId` property will be set to the started run.
+   */
+  async start<O = any>(activity: string, options: ActivityOptions): Promise<ActivityHandle<O>> {
```
Sushisource: The generic type isn't actually bound to the activity name in any meaningful way. We need a more comprehensive solution here that also types inputs

packages/worker/src/worker.ts
```diff
@@ -2225,6 +2222,9 @@ async function extractActivityInfo(
     ),
     priority: decodePriority(start.priority),
     retryPolicy: decompileRetryPolicy(start.retryPolicy),
+    namespace: start.workflowNamespace,
+    activityRunId: !inWorkflow ? start.activityId : undefined,
```
Sushisource: This looks wrong, `activityId` instead of `activityRunId` - should be a test to catch that

Sushisource (commented):
Overall making sense but we need to find a way to provide better type safety

packages/client/src/activity-client.ts
maciejdudko: This matches current workflow activities implementation. I realize we are being a little inconsistent between SDKs here.

packages/client/src/activity-client.ts
maciejdudko: It's non-trivial to achieve, especially if we want docstrings to get preserved. We can consider it for a future improvement.

packages/client/src/activity-client.ts
```diff
@@ -91,7 +82,7 @@ export interface ActivityClientOptions extends AsyncCompletionClientOptions {
  * Typically this client should not be instantiated directly, instead create the high level {@link Client} and use
  * {@link Client.activity} to interact with Activities.
  */
-export class ActivityClient extends AsyncCompletionClient {
+export class ActivityClient extends AsyncCompletionClient implements TypedActivityClient<any> {
```
Sushisource, line 87: I think generally in SDKs we call the typed one just "XxxClient" and the untyped one "UntypedXxxClient" - we want to push people to prefer the typed one.

Might also make sense to have a constructor which accepts the type and gives a typed client. That said probably most people are using the `workflowclient.activity` way of getting it, and I think `workflowclient.activity.typedClient<T>()` is totally reasonable.

packages/client/src/activity-client.ts
```diff
@@ -111,14 +102,25 @@ export class ActivityClient extends AsyncCompletionClient {
...
   }
 
+  /**
+   * Returns this client as a {@link TypedActivityClient}. It enables strong type checking of Activity name, arguments
+   * and result based on the provided Activity interface. Note that no new client object is created - this method only
+   * affects type annotations.
+   * @template T Activity interface to use for type checking. The returned client can only start activities present in
+   * this interface.
+   */
+  typedClient<T>(): TypedActivityClient<T> {
```
Sushisource: Could potentially just called this `typed`, but I'm fine either way.

Sushisource (commented):
LMK if this is ready for final review or not and I can make a last pass

packages/client/src/activity-client.ts
maciejdudko: The server behavior for polling timeout is to return empty response. This is handled correctly. The code throws exception on client-side deadline, but it could be argued it's the expected behavior - we can revisit is some other time, though it does match behavior of workflow result polling. Client-side cancellation is handled correctly.

Added tests to verify long long polling and client-side cancellation.

packages/client/src/activity-client.ts
maciejdudko: I like `typed`. Done.

packages/client/src/activity-client.ts
maciejdudko: Yes, `client.activity.typed<...>()` (after rename) will be the typical use. Especially considering that users would likely want to use a single client instance with many different activity interfaces.

maciejdudko:
@Sushisource PR is ready for final review.

packages/client/src/activity-client.ts
Sushisource: Ah, I guess you can only set them from inside the activity? We should support it in both places, but can be a follow up PR. Let's make sure we track it.

Sushisource (approved):
Looking good on my end but @mjameswh or @chris-olszewski should approve too

packages/activity/src/index.ts
```diff
@@ -249,6 +259,24 @@ export interface Info {
...
+   *
+   * @experimental Standalone Activities are experimental. APIs may be subject to change.
+   */
+  readonly activityRunId?: string;
+  /**
+   * Whether this activity was started by a workflow
+   *
+   * @experimental Standalone Activities are experimental. APIs may be subject to change.
+   */
+  readonly inWorkflow: boolean;
```
chris-olszewski, line 279: Non-blocking just since this is currently experimental, but making `Info` enforce that `inWorkflow: true` implies that workflowExecution etc are present would be nice

e.g.

```
  type Info = BaseInfo & (
    | { readonly inWorkflow: true;
        readonly workflowExecution: { workflowId: string; runId: string };
        readonly workflowNamespace: string;
        readonly workflowType: string;
        readonly activityRunId?: undefined }
    | { readonly inWorkflow: false;
        readonly activityRunId: string;
        readonly workflowExecution?: undefined;
        readonly workflowNamespace?: undefined;
        readonly workflowType?: undefined }
  );
```

packages/client/src/helpers.ts
```diff
@@ -108,8 +108,8 @@ export function decodeCountWorkflowExecutionsResponse(
   };
 }
 
-type ErrorDetailsName = `temporal.api.errordetails.v1.${keyof typeof temporal.api.errordetails.v1}`;
-type FailureName = `temporal.api.failure.v1.${keyof typeof temporal.api.failure.v1}`;
+export type ErrorDetailsName = `temporal.api.errordetails.v1.${keyof typeof temporal.api.errordetails.v1}`;
+export type FailureName = `temporal.api.failure.v1.${keyof typeof temporal.api.failure.v1}`;
```
chris-olszewski, line 112: I don't believe this needs to be made public
```suggestion
type FailureName = `temporal.api.failure.v1.${keyof typeof temporal.api.failure.v1}`;
```

packages/client/src/activity-client.ts
```diff
@@ -0,0 +1,725 @@
...
+      const resp = await this.workflowService.startActivityExecution(
+        await this.buildStartActivityExecutionRequest(input)
+      );
+      return this.createHandle(input.options.id, resp.runId);
+    } catch (err) {
+      if (isGrpcServiceError(err) && err.code === grpcStatus.ALREADY_EXISTS) {
+        for (const entry of getGrpcStatusDetails(err) ?? []) {
+          if (!entry.type_url || !entry.value) continue;
+          if (
+            (trimGrpcTypeUrl(entry.type_url) as ErrorDetailsName) ===
```
chris-olszewski, line 269: Just to be more honest with the cast
```suggestion
            (trimGrpcTypeUrl(entry.type_url) as ErrorDetailsName | '') ===
```

packages/test/src/activities/helpers.ts
```diff
@@ -4,8 +4,11 @@ import { Context } from '@temporalio/activity';
 
 function getSchedulingWorkflowHandle(): WorkflowHandle {
   const { info, client } = Context.current();
+  if (!info.inWorkflow) {
+    throw new Error('Not in workflow');
```
chris-olszewski, line 8: Not major, but we usually throw `IllegalStateError` when attempting to use workflow only code in non-workflow contexts

chris-olszewski (approved):
Lgtm just minor suggestions that are not blocking

packages/client/src/helpers.ts
maciejdudko: It's not public, just privately exported within package. It's so it can be used in `activity-client.ts`.

packages/activity/src/index.ts
maciejdudko: This won't really work in a way that's helpful to the users, because `ActivityContext.current().info` still has to be the most general type and TS compiler doesn't support type assertions on `this`.

PR #2030, opened 2026-04-29
feat(worker): allow async configureWorker and configureReplayWorker plugin hooks

wardpeet:
## What was changed

The two `WorkerPlugin` hooks that participate in worker construction —
`configureWorker` and `configureReplayWorker` — now accept asynchronous
implementations:

```ts
configureWorker?(options: WorkerOptions): WorkerOptions | Promise<WorkerOptions>;
configureReplayWorker?(options: ReplayWorkerOptions): ReplayWorkerOptions | Promise<ReplayWorkerOptions>;
```

The two call sites in `Worker` (`Worker.create` and `Worker.constructReplayWorker`)
now `await` the hook result.

## Why?

Synchronous-only hooks force plugin authors to perform any asynchronous
setup eagerly outside the hook and reassemble the result on the
synchronous path. Common cases that hit this constraint:

- Loading configuration from a remote source (feature flags, secrets
  manager, etc.) before deciding which interceptors / activities to
  register.
- Dynamically resolving a workflow bundle path that depends on an
  `async` factory.
- Initializing instrumentation libraries that expose async APIs
  (e.g. async exporters, providers built with `await`).

Allowing the hook itself to return a `Promise` keeps that work inside
the plugin where it belongs and removes the awkward workaround.

This is a **non-breaking change**: existing synchronous implementations
continue to work as-is, because `await` is transparent on non-Promise
values. None of the in-repo plugins (`SimplePlugin`,
`OpenTelemetryWorkerPlugin`, `ExamplePlugin` in tests) needed to change.

## Checklist

1. Closes N/A — this is a small DX improvement on top of the plugin API
   introduced in #1794.

2. How was this tested:
   Added a new `AsyncExamplePlugin` in `packages/test/src/test-plugins.ts`
   that mirrors the existing `ExamplePlugin` but implements
   `configureWorker` as an `async` function which yields to the event
   loop (`await new Promise(setImmediate)`) before returning the
   modified options. Added two new tests that mirror the existing sync
   plugin tests:
   - `Basic plugin with async hooks` — pre-bundles workflow code with
     the async plugin, then constructs a `Worker` with the same plugin,
     and runs a workflow end-to-end. This exercises the new `await` in
     `Worker.create` — without it, `worker.options.taskQueue` would be
     a `Promise` rather than a string and the assertion would fail
     before the workflow ever ran.
   - `Bundler plugins are passed from worker with async hooks` —
     same scenario but lets the worker drive bundling internally.

   Existing sync plugin tests (`Basic plugin`, `Bundler plugins are
   passed from worker`, the OTEL plugin tests, `SimplePlugin` tests)
   continue to pass unchanged, demonstrating backwards compatibility.

3. Any docs updates needed?
   No — the JSDoc on `WorkerPlugin.configureWorker` /
   `configureReplayWorker` already describes them as "hooks called when
   creating a worker" without committing to a sync-only contract, and
   the change is additive at the type level.


chris-olszewski (approved):
Thanks for this contribution!

wardpeet:
@chris-olszewski, is the current error something I should look at? Seems like the other platforms went through.

chris-olszewski:
Nope, that appears to be a flake. Sorry for delay in getting this merged. We will get it in after Replay is finished

chris-olszewski:
@wardpeet Could you please sign the CLA?

PR #2039, opened 2026-05-06
contrib: add @temporalio/workflow-streams

brianstrauch:
## Summary                                                                                                
                                                                                                            
  Introduces `@temporalio/workflow-streams` — a durable, offset-addressed                            
  event channel for Temporal workflows, built on Signals (publish) and polling
  Updates (subscribe). Mirrors [the equivalent contrib package in sdk-python](https://github.com/temporalio/sdk-python/tree/main/temporalio/contrib/workflow_streams) so                              
  the same wire format and conventions work cross-SDK.                                                      
                                                                                                            
  A workflow holds an append-only log of `(topic, data)` entries; external                                  
  clients (activities, starters, other workflows) publish and subscribe via
  the workflow handle. Cost scales with durable batches, not tokens; ~100ms                                 
  per round-trip. Suited for chat tokens, agent reasoning streams, progress                                 
  events. Not for ultra-low-latency voice.                                                                  
                                                                                                            
  ## What's in the package                                                                                  
                  
  - **`WorkflowStream`** — workflow-side. Registers publish/poll/offset                                     
    handlers, append-only log with optional `truncate()`,
    `continueAsNew()` helper, dedup by `(publisher_id, sequence)`.                                          
  - **`WorkflowStreamClient`** — external-side. Buffered publishing with                                    
    `batchInterval` / `maxBatchSize` / `forceFlush`, async `flush()` barrier,                               
    retry with `maxRetryDuration` and explicit data-loss surfacing,                                         
    `subscribe()` that follows continue-as-new and exits cleanly on                                         
    workflow completion, `await using` disposal.                                                            
  - **`TopicHandle` / `WorkflowTopicHandle`** — bind a topic to a value type                                
    once; subsequent `publish` / `subscribe` don't repeat the name.                                         
  - **Cross-SDK interop** — handler names (`__temporal_workflow_stream_*`)                                  
    and base64-of-protobuf-Payload wire format match sdk-python; tests in                                   
    `test-contrib-workflow-stream-interop.ts` cover the round-trip.                                         
                                                                                                            
  ## Side changes                                                                                           
                                                                                                            
  - `bundler.ts` strips the `node:` URI scheme so `node:stream/web` and                                     
    similar imports route through the existing alias map.
  - Workflow sandbox injects Web Streams globals (`ReadableStream`,                                         
    `WritableStream`, `TransformStream`, queuing strategies) — needed by                                    
    `@temporalio/ai-sdk`'s `ReadableStream` use; lets `ai-sdk` drop the                                     
    `web-streams-polyfill` workaround.                                                                      
  - `throwIfReservedName` allowlists the three `__temporal_workflow_streams_*`
    handler names so contrib packages can register under that namespace; the                                
    dispatch-time prefix check is narrowed to only fire when a default
    handler would otherwise catch the call.                                                                 
                                                                                                            
  ## Test plan                                                                                              
                                                                                                            
  - [ ] `pnpm -C packages/test exec ava lib/test-contrib-workflow-streams.js` — 21 functional tests          
  - [ ] `pnpm -C packages/test exec ava lib/test-contrib-workflow-streams-interop.js` — 10 cross-SDK wire
  tests                                                                                                     
  - [ ] `pnpm -C packages/test exec ava lib/test-ai-sdk.js` — 8 ai-sdk tests covering the bundler/sandbox
  changes                                                                                                   
  - [ ] `pnpm -C packages/test exec ava lib/test-integration-workflows.js --match "*reserved prefix*"` —
  confirm reserved-prefix guard still rejects user-defined `__temporal_*` handlers


packages/ai-sdk/package.json
xumaple, line 35: The PR description says this can be removed, but I think you forgot to actually remove it?

packages/contrib-workflow-streams/README.md
```diff
@@ -0,0 +1,265 @@
...
+activity context. Bind a topic handle on the client and publish through it,
+the same way as on the workflow side:
+
+```typescript
+import { Context } from '@temporalio/activity';
+import { WorkflowStreamClient } from '@temporalio/contrib-workflow-streams';
+
+export async function streamEvents(): Promise<void> {
+  await using client = WorkflowStreamClient.fromActivity({ batchInterval: '2 seconds' });
+  client.start();
```
xumaple: I wasn't a part of the review process for workflow streams, so maybe there's some context I'm missing here. The explicit requirement for user to call WorkflowStreamClient.start() seems unnecessary, if the user forgets to call it then it silently just accumulates in the buffer. Is there a reason we don't just defer until the first publish, and start the client then if it's not already started? That way user can choose to call `start` if they want to define when it happens; otherwise it just happens lazily?

packages/ai-sdk/package.json
brianstrauch: Nice find! Removed

packages/contrib-workflow-streams/README.md
brianstrauch: Awesome suggestion. When we use `await using` in our examples it auto-stops the client when it goes out of scope. If we auto-start as well our examples are cleaner and closer in style to sdk-python which uses `async with` (same idea)

contrib/ai-sdk/src/load-polyfills.ts
```diff
@@ -7,9 +7,6 @@ if (inWorkflowContext()) {
     globalThis.Headers = Headers;
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-require-imports,import/no-unassigned-import
-  require('web-streams-polyfill/polyfill');
```
chris-olszewski, line 11: Note this is change for all ai-sdk users since the polyfill unconditionally replaces native classes: https://github.com/MattiasBuelens/web-streams-polyfill

contrib/ai-sdk/src/load-polyfills.ts
```diff
@@ -7,9 +7,6 @@ if (inWorkflowContext()) {
...
-  // Attach the polyfill as a Global function
```
chris-olszewski, line 12: This comment still applies

packages/core-bridge/Cargo.lock
```diff
@@ -414,7 +408,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
```
chris-olszewski, line 417: I'm not sure I understand why these lockfile changes are in here?

packages/common/src/reserved.ts
```diff
@@ -2,6 +2,17 @@ export const TEMPORAL_RESERVED_PREFIX = '__temporal_';
 export const STACK_TRACE_QUERY_NAME = '__stack_trace';
 export const ENHANCED_STACK_TRACE_QUERY_NAME = '__enhanced_stack_trace';
 
+/**
+ * Wire identifiers used by first-party SDK contrib packages. These
+ * bypass the {@link TEMPORAL_RESERVED_PREFIX} check at registration time.
+ */
+const INTERNAL_HANDLER_NAME_ALLOWLIST: ReadonlySet<string> = new Set([
```
chris-olszewski: I'm in agreement with copilot here, I think we want to minimize how these names could possibly be used to their exact use cases by attaching the expected entity type to the names.

contrib/ai-sdk/src/activities.ts
```diff
@@ -75,6 +94,116 @@ export function createActivities(provider: ProviderV3, mcpClientFactories?: McpC
...
+          default:
+            // tool-call, tool-result, file, source — collect as content
+            if (
+              'type' in part &&
+              (part.type === 'tool-call' ||
+                part.type === 'tool-result' ||
+                part.type === 'file' ||
+                part.type === 'source')
+            ) {
+              content.push(part as LanguageModelV3Content);
```
chris-olszewski: I don't think the cast is necessary
```suggestion
              content.push(part);
```

contrib/ai-sdk/src/activities.ts
```diff
@@ -75,6 +94,116 @@ export function createActivities(provider: ProviderV3, mcpClientFactories?: McpC
...
+      }
+
+      return {
+        content,
+        finishReason,
+        usage,
+        warnings,
+        request: streamResult.request,
+        response: responseMetadata ? { ...responseMetadata, ...streamResult.response } : streamResult.response,
+      } as InvokeModelResult;
```
chris-olszewski: I don't think the cast is necessary
```suggestion
      };
```

contrib/workflow-streams/src/types.ts
```diff
@@ -0,0 +1,283 @@
...
+}
+
+/**
+ * Wire representation of a WorkflowStreamItem (base64 of serialized Payload).
+ *
+ * The `offset` field is populated by the poll handler from the item's
+ * position in the global log. It is unused in the `getState()` snapshot
+ * (offsets there are re-derivable from `base_offset + index`).
+ */
+export interface _WorkflowStreamWireItem {
```
chris-olszewski: `_` prefix is odd for a public facing interface

contrib/workflow-streams/src/types.ts
```diff
@@ -0,0 +1,283 @@
...
+//     bytes data = 2;
+//   }
+//
+// Map entries are encoded as embedded messages:
+//   message MapEntry { string key = 1; bytes value = 2; }
+//
+// Wire format is hand-rolled here (rather than reusing `@temporalio/proto`'s
+// generated class) to avoid pulling the protobufjs runtime into the workflow
+// sandbox. The schema is a fixed public API — the manual encoder cannot
+// silently go out of sync with server-side expectations.
```
chris-olszewski: This is slightly concerning, but I understand it is necessary. It would be good to have some [features](https://github.com/temporalio/features/tree/main) tests to verify that the SDK/server implementations stay in sync.

contrib/ai-sdk/src/activities.ts
```diff
@@ -75,6 +94,116 @@ export function createActivities(provider: ProviderV3, mcpClientFactories?: McpC
...
+      let finishReason: LanguageModelV3FinishReason = { unified: 'other', raw: undefined };
+      let usage: LanguageModelV3Usage = {
+        inputTokens: { total: undefined, noCache: undefined, cacheRead: undefined, cacheWrite: undefined },
+        outputTokens: { total: undefined, text: undefined, reasoning: undefined },
+      };
+      const warnings: SharedV3Warning[] = [];
+      let responseMetadata: Record<string, unknown> | undefined;
+
+      let currentText = '';
+      let currentReasoning = '';
```
chris-olszewski: AI is pointing out that there is an `id` on the parts we receive from the stream and the parts might be interspersed. I think we might need to have a map between these ids and the current state.

https://ai-sdk.dev/docs/ai-sdk-ui/streaming-data#data-part-reconciliation

contrib/workflow-streams/src/types.ts
```diff
@@ -0,0 +1,283 @@
...
+    const d = i + 3 < clean.length ? B64.indexOf(clean.charAt(i + 3)) : 0;
+    if (a < 0 || b < 0 || c < 0 || d < 0) {
+      throw new TypeError(`Invalid base64 input: non-alphabet character at offset ${i}`);
+    }
+    out[j++] = (a << 2) | (b >> 4);
+    if (j < len) out[j++] = ((b << 4) | (c >> 2)) & 0xff;
+    if (j < len) out[j++] = ((c << 6) | d) & 0xff;
+  }
+  return out;
+}
```
chris-olszewski: Was this vendored from somewhere we should attribute?

Also, I find it odd to have implementations in a `types.ts` file

contrib/workflow-streams/src/__tests__/test-workflow-streams-interop.ts
```diff
@@ -0,0 +1,212 @@
...
+      0x08, // inner key: field 1, len 8
+      ...encoder.encode('encoding'),
+      0x12,
+      0x0a, // inner value: field 2, len 10
+      ...encoder.encode('json/plain'),
+      0x12,
+      0x07, // data: field 2, len 7
+      ...encoder.encode('"hello"'),
+    ])
+  );
```
chris-olszewski, line 52: Should we just use the actual payload encoding logic instead of handrolling an example?

chris-olszewski (commented):
Partial review. Apologies, lots to go through in this PR.

contrib/ai-sdk/src/load-polyfills.ts
brianstrauch: That's intended. Hopefully it's ok? From what I understand, native classes are the better approach here and supported as of Node 18 (sdk-typescript requires Node 20+).

packages/core-bridge/Cargo.lock
brianstrauch: I probably messed something up while following the setup instructions... fixed.

contrib/workflow-streams/src/types.ts
brianstrauch: Deferring question to @jssmith and moving implementation to codec.ts

brianstrauch: It's generated by Claude

contrib/workflow-streams/src/__tests__/test-workflow-streams-interop.ts
brianstrauch: If we used `encodePayloadProto`, the assertion would reduce to `encodePayloadWire(p) === encodeBase64(encodePayloadProto(p))`, which by definition is true, and the test would pass if both functions regressed at the same time

contrib/workflow-streams/src/types.ts
brianstrauch: Will include in a follow-up PR!

contrib/ai-sdk/src/load-polyfills.ts
mjameswh: > supported as of Node 18 (sdk-typescript requires Node 20+).

Yes... but no. That's a bit tricky.

The thing is that most APIs that are added to Node are only exposed to Node's code executing in the context of the main thread, and sometime in the context of a worker thread. But new APIs rarely get exposed to [VMs](https://nodejs.org/api/vm.html), which is what the workflow sandbox is built on.

That's why we need polyfills, even for many APIs that have been supported by Node for a very long time.

brianstrauch: TIL! Reverting...

contrib/workflow-streams/README.md
```diff
@@ -0,0 +1,261 @@
...
+## Quick Start
+
+### Workflow side
+
+Construct `new WorkflowStream()` at the start of your workflow function, then
+get a typed handle for each topic via `stream.topic<T>(name)` and call
+`publish` on the handle:
+
+```typescript
+import { WorkflowStream } from '@temporalio/workflow-streams';
```
chris-olszewski: We need to have multiple entrypoints for `@temporalio/workflow-streams` so workflow code can avoid pulling in the activity code. Right now attempting to use `@temporalio/workflow-streams` in workflow code will fail:
```
 workflow-streams › workflow streams can be bundled

  Promise rejected with:

  Error {
    message: `Your Workflow code (or a library used by your Workflow code) is importing the following disallowed modules:␊
      - 'crypto'␊
      - '@temporalio/activity'␊
      - '@temporalio/client'␊
    These modules can't be used in workflow context as they might break determinism.HINT: Consider the following options:␊
     • Make sure that activity code is not imported from workflow code. Use \`import type\` to import activity function signatures.␊
     • Move code that has non-deterministic behaviour to activities.␊
     • If you know for sure that a disallowed module will not be used at runtime, add its name to 'WorkerOptions.bundlerOptions.ignoreModules' in order to dismiss this warning.␊
    See also: https://typescript.temporal.io/api/namespaces/worker#workflowbundleoption and https://docs.temporal.io/typescript/determinism.`,
  }

```

This didn't show up earlier as our test helper for creating test workers allowlists these modules so we're able to write workflow code inline with tests 🙃 

Repro:
```
test('workflow streams can be bundled', async (t) => {
  await t.notThrowsAsync(
    bundleWorkflowCode({
      workflowsPath: require.resolve('./workflows/workflow-streams'),
    })
  );
});
```

chris-olszewski (changes requested):
I think we'll need to switch over to having a split `workflow` and `client` entrypoints for this package to avoid adding the client code to workflow bundles. 

https://nodejs.org/api/packages.html#package-entry-points

contrib/workflow-streams/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+The `WorkflowStream` constructor registers the `__temporal_workflow_stream_publish` signal,
+`__temporal_workflow_stream_poll` update, and `__temporal_workflow_stream_offset` query handlers on your workflow.
+Any value the default payload converter can serialize (JSON, `Uint8Array`, or
+a pre-built `Payload`) can be passed to `publish`. The type parameter `T` is
+purely a compile-time annotation — TypeScript has no runtime type
+representation, so per-topic type uniformity isn't enforced at runtime
+(unlike sdk-python). Repeated calls to `stream.topic('foo')` return the same
+handle instance, so a stale `T` annotation can't introduce a duplicate
+publisher.
```
chris-olszewski: Few notes on this section:
 - "Any value the default payload converter can serialize (JSON, `Uint8Array`, or a pre-built `Payload`)" is a little confusing as it mixes encoding targets of `T` with `Payload` that can be provided directly
 - TypeScript devs should be pretty knowledgeable that types are not present at runtime, I don't think we need to spill too much ink on that.
 - I don't understand the "stale `T` annotations can't introduce a duplicate publisher", are we attempting to convey that topics are identified completely by name?
```suggestion
Any value the default payload converter can serialize or
a pre-built `Payload` can be passed to `publish`. The type parameter `T` is only a compile-time annotation. Repeated calls to `stream.topic('foo')` return the same
handle instance, so a stale `T` annotation can't introduce a duplicate
publisher.
```

contrib/workflow-streams/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Activity side (publishing)
+
+Use `WorkflowStreamClient.fromActivity()` with `await using` for batched publishing
```
chris-olszewski: `await using` syntax support was only added in [Node 24](https://nodejs.org/en/blog/release/v24.0.0), but can be used in [Typescript 5.2+](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html) as the symbols existed in Node 20+.

(This is a note to myself and any other reviewers as I was concerned that this would cause issues with Node 20/22 support)

contrib/workflow-streams/src/client.ts
```diff
@@ -0,0 +1,553 @@
...
+    return instance;
+  }
+
+  /**
+   * Create a WorkflowStreamClient targeting the current activity's parent workflow.
+   *
+   * Must be called from within an activity. The Temporal client and
+   * parent workflow id are taken from the activity context.
+   */
+  static fromActivity(options?: WorkflowStreamClientOptions): WorkflowStreamClient {
```
chris-olszewski: Notice that python's equivalent is `from_within_activity`. I think that reads a little bit better as `fromActivity` sounds like close to a conversion method name.
```suggestion
  static fromWithinActivity(options?: WorkflowStreamClientOptions): WorkflowStreamClient {
```

contrib/workflow-streams/README.md
```diff
@@ -0,0 +1,261 @@
...
+  console.log(item.topic, item.offset, item.data);
+  if (isDone(item.data)) break;
+}
+```
+
+For raw `Payload` access — multi-topic subscriptions whose payload types
+differ, or fine-grained control over decoding — call
+`WorkflowStreamClient.subscribe(topics?, fromOffset?)` directly. The yielded
+items have `data: Payload` carrying encoding metadata; decode with
+`defaultPayloadConverter.fromPayload<T>(item.data)` per-topic.
```
chris-olszewski:
```suggestion
For raw `Payload` access call
`WorkflowStreamClient.subscribe(topics?, fromOffset?)` directly. The yielded
items have `data: Payload` carrying encoding metadata; decode with
`defaultPayloadConverter.fromPayload<T>(item.data)` per-topic.
```

contrib/workflow-streams/README.md
```diff
@@ -0,0 +1,261 @@
...
+list (or omit the argument) to receive messages from all topics. Publishing to
+a topic implicitly creates it.
+
+## Continue-as-new
+
+Carry both your application state and workflow stream state across continue-as-new
+boundaries:
+
+```typescript
+import { continueAsNew, workflowInfo } from '@temporalio/workflow';
```
chris-olszewski:
```suggestion
import { workflowInfo } from '@temporalio/workflow';
```

contrib/workflow-streams/README.md
```diff
@@ -0,0 +1,261 @@
...
+    taskQueue: 'other-tq',
+  });
+  await continueWithOptions({
+    itemsProcessed,
+    streamState: stream.getState(),
+  });
+}
+```
+
+## API Reference
```
chris-olszewski: Not major, but unsure if this is helpful compared to our gen'd docsite. This is more likely to get out of sync.

contrib/workflow-streams/README.md
brianstrauch: Done!

contrib/workflow-streams/src/topic-handle.ts
```diff
@@ -0,0 +1,107 @@
...
+   *   bound type.
+   * @param options.forceFlush If true, wake the flusher to send
+   *   immediately (fire-and-forget — does not block the caller).
+   */
+  publish(value: T | Payload, options?: { forceFlush?: boolean }): void {
+    // Cast through `unknown` so the internal publisher accepts the value;
+    // the per-handle T is compile-time only.
+    (
+      this.client as unknown as { _publishToTopic(name: string, value: unknown, forceFlush: boolean): void }
+    )._publishToTopic(this.name, value, options?.forceFlush ?? false);
```
chris-olszewski: Cast is not necessary
```suggestion
    this.client._publishToTopic(this.name, value, options?.forceFlush ?? false);
```

contrib/workflow-streams/src/__tests__/workflows/workflow-streams.ts
```diff
@@ -0,0 +1,178 @@
...
+export async function workflowSidePublishWorkflow(count: number): Promise<void> {
+  const stream = new WorkflowStream();
+  const events = stream.topic('events');
+  let closed = false;
+  setHandler(closeSignal, () => {
+    closed = true;
+  });
+  const encoder = new TextEncoder();
+  for (let i = 0; i < count; i++) {
+    events.publish(encoder.encode(`item-${i}`));
```
chris-olszewski: Is there a reason all of the workflows in this file publish `Uint8Array` explicitly instead of just passing the string at the callsite? We should have coverage of both.

contrib/workflow-streams/src/topic-handle.ts
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @param value Value to publish. Goes through the workflow's default
+   *   payload converter. A pre-built {@link Payload} bypasses
+   *   conversion, regardless of the handle's bound type.
+   */
+  publish(value: T | Payload): void {
+    (this.stream as unknown as { _publishToTopic(name: string, value: unknown): void })._publishToTopic(
+      this.name,
+      value
+    );
```
chris-olszewski:
```suggestion
    this.stream._publishToTopic(this.name, value);
```

contrib/workflow-streams/src/client.ts
```diff
@@ -0,0 +1,556 @@
...
+    let handle = this.topicHandles.get(name);
+    if (handle === undefined) {
+      handle = new TopicHandle<T>(this, name);
+      this.topicHandles.set(name, handle as TopicHandle<unknown>);
+    }
+    return handle as TopicHandle<T>;
+  }
+
+  /** @internal Used by {@link TopicHandle.publish}. */
+  _publishToTopic(topic: string, value: unknown, forceFlush: boolean): void {
```
chris-olszewski: I think we can avoid exposing this method publicly if we instead pass a closure that calls this to `TopicHandle` variants.
```suggestion
  /** @internal Used by {@link TopicHandle.publish}. */
  private publishToTopic(topic: string, value: unknown, forceFlush: boolean): void {
```

contrib/workflow-streams/src/client.ts
```diff
@@ -0,0 +1,556 @@
...
+   *
+   * Repeated calls with the same name return the same handle instance.
+   * The type parameter ``T`` is purely a compile-time annotation — see
+   * the module note in {@link TopicHandle} for the difference from
+   * sdk-python's runtime type-uniformity check.
+   */
+  topic<T = unknown>(name: string): TopicHandle<T> {
+    let handle = this.topicHandles.get(name);
+    if (handle === undefined) {
+      handle = new TopicHandle<T>(this, name);
```
chris-olszewski: In relation to making `publishToTopic` private
```suggestion
      handle = new TopicHandle<T>(this, name, (topic, value, forceFlush) => this.publishToTopic(topic, value, forceFlush));
```

contrib/workflow-streams/src/workflow.ts
```diff
@@ -0,0 +1,370 @@
...
+    let handle = this.topicHandles.get(name);
+    if (handle === undefined) {
+      handle = new WorkflowTopicHandle<T>(this, name);
+      this.topicHandles.set(name, handle as WorkflowTopicHandle<unknown>);
+    }
+    return handle as WorkflowTopicHandle<T>;
+  }
+
+  /** @internal Used by {@link WorkflowTopicHandle.publish}. */
+  _publishToTopic(topic: string, value: unknown): void {
```
chris-olszewski: See comment in `client.ts` for how we can keep this method private.
```suggestion
  /** @internal Used by {@link WorkflowTopicHandle.publish}. */
  private publishToTopic(topic: string, value: unknown): void {
```

contrib/workflow-streams/README.md
brianstrauch: Agree, rewriting the last sentence to: "The type parameter `T` is a compile-time annotation and doesn't affect handle identity."

chris-olszewski (approved):

PR #2042, opened 2026-05-07
feat(extstore): add base types and configuration for extstore

cconstable:
Not ready for full review yet. Seeking targeted feedback about general shape of the implementation.


packages/core-bridge/src/client.rs
```diff
@@ -264,13 +264,19 @@ async fn client_invoke_workflow_service(
         "CountActivityExecutions" => rpc_call!(connection, call, count_activity_executions),
         "CountSchedules" => rpc_call!(connection, call, count_schedules),
         "CountWorkflowExecutions" => rpc_call!(connection, call, count_workflow_executions),
-        "CountNexusOperationExecutions" => rpc_call!(connection, call, count_nexus_operation_executions),
+        "CountNexusOperationExecutions" => {
+            rpc_call!(connection, call, count_nexus_operation_executions)
+        }
```
cconstable: All this got formatted when I ran `pnpm format`. Will separate this out into a different commit.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,284 @@
...
+  if (payloads.length === 0 || externalStorage === undefined) return payloads;
+
+  const { drivers, driverSelector, driversByName, payloadSizeThreshold } = externalStorage;
+  const fallbackDriver: StorageDriver | undefined =
+    drivers.length === 1 && driverSelector === undefined ? drivers[0] : undefined;
+
+  const batchController = new AbortController();
+  const batchSignal = abortSignal
+    ? AbortSignal.any([batchController.signal, abortSignal])
+    : batchController.signal;
```
cconstable, line 74: Here is the cancellation logic: 

If the user supplies an `abortSignal` for cancelling extstore, all drivers should respect that. Additionally, if a single driver fails, we should cancel the other drivers. I think trying to partially recover here is tricky and we should just bubble up the failure and let the activity decide what to do. I'm using `AbortSignal.any` to combine the caller-supplied abort signal ("cancel everything") with our internally create abort signal ("cancel because something else failed") and this new composite signal is what is passed to each driver. `AbortSignal.any` required a bump of `@types/node` which ended up breaking something in `iterators-utils.ts` but was something that actually needed to be fixed to begin with.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,301 @@
...
+// ============================================================================
+
+/** @internal */
+export const EXTSTORE_REFERENCE_MESSAGE_TYPE = 'temporal.api.sdk.v1.ExternalStorageReference';
+
+/** @internal */
+export const EXTSTORE_REFERENCE_ENCODING = encodingTypes.METADATA_ENCODING_PROTOBUF_JSON;
+
+/** Pre-1.0 reference encoding. Read-only — never written by current implementations. @internal */
+export const LEGACY_EXTSTORE_REFERENCE_ENCODING = 'json/external-storage-reference';
```
cconstable: I saw this was referenced in some older go and python implementations and included it for backwards compatibility. Since that was pre-GA I thinking it might not be necessary to include here but wanted to bring it up.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,302 @@
...
+/**
+ * Context handed to {@link StorageDriver.store} (and to the selector).
+ *
+ * @experimental
+ */
+export interface StorageDriverStoreContext {
+  /** Aborts the in-flight operation; siblings are cancelled on first error. */
+  abortSignal?: AbortSignal;
+  /** Identity of the workflow / activity that produced the payloads. */
+  target?: SerializationContext;
```
jmaeagle99: The target field should be a new type that is not a serialization context. See the analogous field in Python: https://github.com/temporalio/sdk-python/blob/main/temporalio/converter/_extstore.py#L151

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,302 @@
...
+ * Mounted on {@link DataConverter.externalStorage}.
+ *
+ * @experimental
+ */
+export class ExternalStorage {
+  readonly drivers: StorageDriver[];
+  readonly driverSelector?: StorageDriverSelector;
+  readonly payloadSizeThreshold: number;
+  /** Built at construction; used by the retrieve path and by selector validation. */
+  readonly driversByName: ReadonlyMap<string, StorageDriver>;
```
jmaeagle99: This should be private/# without public accessor, but somehow an internal accessor.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,302 @@
...
+// ============================================================================
+// Configuration
+// ============================================================================
+
+/**
+ * Default {@link ExternalStorage.payloadSizeThreshold}: 256 KiB.
+ *
+ * @experimental
+ */
+export const DEFAULT_PAYLOAD_SIZE_THRESHOLD = 256 * 1024;
```
jmaeagle99, line 136: I don't think this should be exported.

packages/common/src/converter/extstore.ts
jmaeagle99: Let's drop this for now. I think that the prerelease format will have incredible low usage. We can add back if we get feedback.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,302 @@
...
+    this.driversByName = driversByName;
+  }
+}
+
+// ============================================================================
+// Wire format
+// ============================================================================
+
+/** @internal @experimental */
+export const EXTSTORE_REFERENCE_MESSAGE_TYPE = 'temporal.api.sdk.v1.ExternalStorageReference';
```
jmaeagle99: I think that all of these constants shouldn't necessarily be exported.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,290 @@
...
+
+  return result;
+}
+
+// ============================================================================
+// Internal helpers
+// ============================================================================
+
+function payloadProtoSize(payload: Payload): number {
+  return PayloadProto.encode(payload).finish().length;
```
jmaeagle99: Couple of AI suggestions here:
- the `.finish()` call does an array allocation and copy. And then were just checking the length. Expensive since every single payload will go through this. The writer that is produced from the `encode(...)` call might have a `len` field that might be able to use.
- Less exact, but maybe a good indication of pass through: check `payload.data` length first (and add some headroom for metadata) to see if we really need to serialize the entire payload here. We didn't do this optimization in other languages, and it's more heuristic based, so just something to think about.

packages/common/src/internal-non-workflow/codec-helpers.ts
jmaeagle99: I haven't gone through all of these to check if they are correct, but I do have some architectural feedback. We'll want to enable concurrent external store and retrieval operations for workflow activations and completions since these will have such a large impact on the workflow task duration. As far as I can tell, workflow activation and completion do not use the utility functions that have been updated, which I think is coincidental but also may lead to inconsistency of usage given how these are similarly named.

jmaeagle99: We might need to more cleanly separate the functions that only encode/decode with codecs vs ones that encode/decode and store/retrieve. The former would be exclusively used by things that require concurrent external storage visiting (like workflow activation and completion) where as the latter would be used by things that are okay with sequential processing (everything else).

jmaeagle99: Oh, and if not clear, we need to make sure workflow activation and completion messages (and their descendants) have their payloads be considered for external storage, regardless of serial vs concurrent visiting. This is were the bulk of the user benefit will be seen.

packages/common/src/internal-non-workflow/external-storage-runner.ts
jmaeagle99: > I think trying to partially recover here is tricky and we should just bubble up the failure and let the activity decide what to do.

Agreed, we need payloads to be externalized deterministically so customers see the expected behavior. Partial recovery means some payloads are not externally store (when they should have) and would likely cause confusion when they see their workflows still bloating even when external storage was configured.

Let's wait on feedback for if we need a less strict and likely opt-in behavior.

packages/common/src/internal-non-workflow/external-storage-runner.ts
cconstable: It's a good call on the array allocation, I didn't catch that and `len` gives us the same result.

> Less exact, but maybe a good indication of pass through: check payload.data length first (and add some headroom for metadata) to see if we really need to serialize the entire payload here.

I don't think any perf optimizations would be worth losing accuracy here.

packages/common/src/internal-non-workflow/codec-helpers.ts
cconstable: Agree with all points. I didn't finish pushing updates to this file (beyond this initial integration) because of what you mentioned about cleanly separating them. I was beginning to get a little confused myself. Planning on splitting out the codec + extstore function to a different file and will likely push an update for this Tuesday.

PR #2047, opened 2026-05-12
Support protobuf-es (@bufbuild/protobuf v2) payloads

### LINKED ISSUE #2045: Add @bufbuild/protobuf (protobuf-es) payload converters

The SDK ships payload converters for [`protobufjs`](https://github.com/protobufjs/protobuf.js) only (`@temporalio/common/lib/protobufs`). Users who generate their proto types with [`@bufbuild/protobuf`](https://github.com/bufbuild/protobuf-es) v2 can't pass those types to a Temporal client or workflow without writing their own payload converter. The two libraries produce incompatible message shapes (`Schema` + `create(Schema, init)` vs `Message.create(init)`), so the existing converters can't be reused.

A parallel converter trio under `@temporalio/common/lib/protobufs-es`, mirroring the existing `…/protobufs` exports:

- `ProtobufEsBinaryPayloadConverter`
- `ProtobufEsJsonPayloadConverter`
- `DefaultPayloadConverterWithProtobufsEs`

**Today** — every protobuf-es user writes their own (~150 LOC) converter:

```ts
// payload-converter.ts
import { create, toBinary, fromBinary, toJson, fromJson, createRegistry } from '@bufbuild/protobuf';
// ...custom Binary + JSON converters, composite ordering,
//    messageType metadata, Registry forwarding for google.protobuf.Any...
```

**Proposed** — first-class support, matching the protobufjs ergonomics:

```ts
// payload-converter.ts
import { createRegistry } from '@bufbuild/protobuf';
import { DefaultPayloadConverterWithProtobufsEs } from '@temporalio/common/lib/protobufs-es';
import { MyMessageSchema } from './gen/messages_pb';

export const payloadConverter = new DefaultPayloadConverterWithProtobufsEs({
  registry: createRegistry(MyMessageSchema),
});
```

Sketch of the desired properties:

- `@bufbuild/protobuf` as an **optional peer dependency** so users on the protobufjs path pay nothing.
- Wire-compatible with the existing protobufjs converters:
  - A payload produced by either converter (binary or JSON) round-trips correctly through the other.
  - Same `encoding` markers (`binary/protobuf`, `json/protobuf`) and fully-qualified `messageType` metadata.
  - Composite ordering matches `DefaultPayloadConverterWithProtobufs` (Undefined → Binary → ProtobufJson → ProtobufBinary → Json), so on-the-wire converter selection is identical.
- Constructor accepts either a pre-built `Registry` or a schema array (calls `createRegistry` internally).

### Additional context

The existing [`docs/protobuf-libraries.md`](https://github.com/temporalio/sdk-typescript/blob/main/docs/protobuf-libraries.md) is the SDK's original (Dec '21) decision log for protobuf library choice. It defines three criteria — (A) TS types, (B) runtime-checkable message instances, (C) spec-compliant proto3 JSON — and notes that protobufjs scores A+B but not C, which is why the SDK bolts on `proto3-json-serializer` and `patchProtobufRoot`. `@bufbuild/protobuf` v2 satisfies A+B+C natively, and its generated output addresses every item in that doc's "Future work" section (spec-compliant `to/fromJSON`, fully-qualified `typename`, "this is a generated file" header).

### Concrete use case driving this

I'm building [`protoc-gen-ts-temporal`](https://github.com/nu-sync/protoc-gen-ts-temporal) and [`protoc-gen-rust-temporal`](https://github.com/nu-sync/protoc-gen-rust-temporal) (TypeScript and Rust analogues of [`protoc-gen-go-temporal`](https://cludden.github.io/protoc-gen-go-temporal/)) so that a single `.proto` file generates type-safe Temporal clients, workers, and activity stubs across languages.

I'm trying to build a Node service that accepts a webhook and starts a workflow whose worker is implemented in Rust, with **no intermediate Rust HTTP server** that exists only to translate the webhook into a workflow start. For that to work, the TS client and the Rust worker have to agree on the wire format of the workflow args. `protoc-gen-ts-temporal` is built on `@bufbuild/protoc-gen-es`, so its generated message types are `@bufbuild/protobuf` v2.

wcygan:
Resolves https://github.com/temporalio/sdk-typescript/issues/2045.

I would like to use this in my project [protoc-gen-ts-temporal](https://github.com/nu-sync/protoc-gen-ts-temporal) which is similar to [protoc-gen-go-temporal](https://cludden.github.io/protoc-gen-go-temporal/) from "[Go Code Generation with Temporal & Protobufs](https://temporal.io/code-exchange/go-code-generation-with-temporal-and-protobufs)"

This would be useful to me because I could run an [HTTP server in TypeScript](https://docs.deno.com/runtime/fundamentals/http_server/) and use a typed client (this PR + [protoc-gen-ts-temporal](https://github.com/nu-sync/protoc-gen-ts-temporal)) to begin a [workflow that executes on a Rust worker](https://docs.temporal.io/develop/rust/workers/worker-process)

## Summary

Adds these in `@temporalio/common/lib/protobufs-es`:

1. `ProtobufEsBinaryPayloadConverter`
2. `ProtobufEsJsonPayloadConverter`
3. `DefaultPayloadConverterWithProtobufsEs`

Wire-compatible with the protobufjs path — binary bytes are byte-identical; JSON is canonical proto3.


mjameswh:
Are there any reason why this would need to be part of the SDK itself?

wcygan:
> Are there any reason why this would need to be part of the SDK itself?

Hey @mjameswh I started by mirroring the existing layout. It might not be strictly necessary, but it might help with discoverability since the SDK already ships the [protobufjs converter](https://github.com/temporalio/sdk-typescript/blob/main/packages/common/src/converter/protobuf-payload-converters.ts). I also reused some internal APIs like [PayloadConverterWithEncoding](https://github.com/temporalio/sdk-typescript/blob/e704c48fe73d6698a7c517d0d2c0e57aa613030f/packages/common/src/converter/payload-converter.ts#L142) in the new code.

I opened this PR because I see myself reusing this stack with Rust / TypeScript in a few different projects:
1. [Temporal TypeScript SDK](https://github.com/temporalio/sdk-typescript) (durable execution)
2. [ConnectRPC](https://connectrpc.com/) (transport)
3. [Buf CLI](https://buf.build/product/cli) (codegen via https://github.com/bufbuild)

I feel like supporting https://github.com/bufbuild/protobuf-es natively in the SDK would be a really nice user experience so that clients don't need too many workarounds to enable it; I am aiming for something close to a "it just works" experience

---

**Sidenote**: I am also admittedly not the most familiar with the node ecosystem's packaging conventions (I primarily write Java and Rust, only using TypeScript where absolutely necessary), so I'm open to whatever direction you prefer, and I'm happy to support this work

wcygan:
I also have a [prototype](https://github.com/nu-sync/protoc-gen-ts-temporal/blob/validate-sdk-pr-2047/validation-notes.md) of this working locally with [protoc-gen-ts-temporal](https://github.com/nu-sync/protoc-gen-ts-temporal)

mjameswh:
> the SDK already ships the protobufjs converter

Yeah, to be honest, I’d say the existing `ProtobufPayloadConverter` lives in the SDK mostly for legacy reasons. Given the opportunity to make that decision again today, I don’t think we would include it directly in the core SDK package.

The main concern is not with `protobuf-es` specifically — the proposal itself makes sense, and I can see why it would be useful. The issue is more about where optional integrations like this should live.

Every optional feature we ship as part of the SDK carries long-term maintenance, compatibility, testing, release, and support expectations. Over time, that does not scale very well: it reduces maintainer bandwidth for core SDK work, makes the package surface area harder to reason about, and blurs the line between “core Temporal SDK behavior” and “integration code for a particular ecosystem/toolchain.” It also makes us more cautious about future changes, because users reasonably expect anything shipped in the SDK itself to follow the SDK’s reliability and compatibility standards.

We’re trying to move toward a repo/ecosystem model that is more scalable for us and more open to community contributions. In particular, we’ve been extracting optional features either into the contrib area or into external community-maintained repositories, depending on the case. We’re also encouraging integrations like this to use the newly introduced Plugin API, which should significantly improve the DX for users of optional features without requiring those features to live directly in the core SDK package.

For external community-maintained integrations, we’ve started the [Temporal Code Exchange](https://temporal.io/code-exchange)￼ initiative to give those contributions better visibility, so users can still discover them without everything needing to live directly in the SDK repo.

So, what about this specific pull request? I think that, at this time, an external community-maintained package may be the better fit than adding this to our `contrib` directory.

The main reason is that this converter seems closely tied to the protoc codegen plugin you’re building. In practice, the interesting evolution is likely to happen at that integration layer: what the generated types look like, how schemas are registered, how cross-language payload compatibility is exposed, what conventions the generator establishes, etc.

By contrast, the Temporal payload converter API itself is relatively stable. If we put only the converter piece in the SDK repo, but the generator and surrounding workflow live elsewhere, ownership gets split in a way that would be inefficient for both teams. Changes to the end-to-end experience would need coordination across two repositories, and in some cases you’d need SDK maintainer involvement for changes that are really driven by the generator’s design.

So my intuition, at least for now, is that the best ownership model is for the `protobuf-es` Temporal integration to live near the codegen/tooling that drives it. That way, our team can evolve the whole user experience together, while still depending on the SDK’s stable payload-converter/plugin APIs. Once the integration has matured and the boundaries are clearer, we can always revisit whether some part of it belongs closer to the SDK itself.

If you’re open to that direction, I’d be happy to help figure out what the right contrib/plugin shape should look like.

PR #2048, opened 2026-05-13
Add standalone activity startDelay support

eamsden:
Wire ActivityOptions.startDelay into StartActivityExecutionRequest and validate that it is non-negative.

Port the standalone activity delayed-start integration coverage from Python, including the server version/config needed to assert the scheduled-to-started delay.

## What was changed
- Delayed-start is surfaced as `ActivityOptions.startDelay`
- An integration test checks that the option has the desired effect

## Why?
Standalone activities now have a delayed-start feature. These changes expose this feature from the TS SDK

## Checklist

1. Closes 

2. How was this tested:
A new integration test was added and validated that new option has the desired effect of delaying the start of a standalone activity.

3. Any docs updates needed?
No.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,7 +214,6 @@ jobs:
         env:
           RUN_INTEGRATION_TESTS: true
           REUSE_V8_CONTEXT: ${{ matrix.reuse-v8-context }}
-          TEMPORAL_SERVICE_ADDRESS: localhost:7233
 
           # For Temporal Cloud + mTLS tests
           TEMPORAL_CLOUD_MTLS_TEST_TARGET_HOST: ${{ vars.TEMPORAL_CLIENT_NAMESPACE }}.tmprl.cloud:7233
@@ -239,7 +238,6 @@ jobs:
         env:
           RUN_INTEGRATION_TESTS: true
           REUSE_V8_CONTEXT: ${{ matrix.reuse-v8-context }}
-          TEMPORAL_SERVICE_ADDRESS: localhost:7233
 
       # FIXME: Move samples tests to a custom activity
       # Sample 1: hello-world to local server
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,6 +188,8 @@ jobs:
 
       - name: Install Temporal CLI
         uses: temporalio/setup-temporal@1059a504f87e7fa2f385e3fa40d1aa7e62f1c6ca # v0
+        with:
+          version: ${{ env.TESTS_CLI_VERSION }}
 
       - name: Run Temporal CLI
         working-directory: ${{ runner.temp }}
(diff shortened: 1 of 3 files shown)


packages/test/src/test-standalone-activities.ts
```diff
@@ -175,6 +177,49 @@ if (RUN_INTEGRATION_TESTS) {
...
+          startDelay: startDelayMs,
+        });
+
+        t.is(await handle.result(), 'hello');
+
+        const description = await handle.describe();
+        t.is(description.status, ActivityExecutionStatus.COMPLETED);
+        t.truthy(description.scheduleTime);
+        t.truthy(description.lastStartedTime);
+        t.true(description.lastStartedTime!.getTime() - description.scheduleTime!.getTime() >= startDelayMs - 500);
```
THardy98: why `- 500` ?

THardy98 (approved):

packages/test/src/test-standalone-activities.ts
maciejdudko: Accounting for possible time discrepancies in server internal bookkeeping (protection against CI flakiness).

packages/test/src/test-standalone-activities.ts
```diff
@@ -175,6 +177,49 @@ if (RUN_INTEGRATION_TESTS) {
     t.is(err?.cause?.message, 'failure');
   });
 
+  test('Start activity with start delay', async (t) => {
+    const env = await createLocalTestEnvironment({
+      server: {
+        executable: {
+          type: 'cached-download',
+          version: startDelayTestCliVersion,
+        },
```
maciejdudko: This has very high chance of getting stale. Instead, use the default env that uses the existing server and update [TESTS_CLI_VERSION](https://github.com/temporalio/sdk-typescript/blob/a9a7eaf203f9934ac42e2370bc30c9faa3673e46/.github/workflows/ci.yml#L25) and [--dynamic-config-value](https://github.com/temporalio/sdk-typescript/blob/a9a7eaf203f9934ac42e2370bc30c9faa3673e46/.github/workflows/ci.yml#L206) in `.github/workflows/ci.yml`.

maciejdudko (changes requested):
Looks good! The test needs different server setup (see comment) but otherwise LGTM!

.github/workflows/ci.yml
```diff
@@ -214,7 +214,6 @@ jobs:
         env:
           RUN_INTEGRATION_TESTS: true
           REUSE_V8_CONTEXT: ${{ matrix.reuse-v8-context }}
-          TEMPORAL_SERVICE_ADDRESS: localhost:7233
```
mjameswh: Why? This may slow down test execution.

If that's because some tests requires features not enabled on the server ambient server, adjust the `setup-temporal` action above instead.

maciejdudko (approved):

.github/workflows/ci.yml
eamsden: This is a reversion of a change made within this PR that was done in an attempt to debug CI and actually broke it further.

mjameswh (approved):

PR #2058, opened 2026-05-18
Serialization context plumbing for async client

THardy98:
## What was changed
Add serialization context support for async completion client

## Why?
Expand serialization context coverage.

1. Part of https://github.com/temporalio/sdk-typescript/issues/1661

2. How was this tested:
Integration tests

3. Any docs updates needed?
No

### DIFF
--- a/packages/test/src/test-async-completion.ts
+++ b/packages/test/src/test-async-completion.ts
@@ -12,11 +12,14 @@ import {
   Connection,
 } from '@temporalio/client';
 import type { Info } from '@temporalio/activity';
-import { rootCause } from '@temporalio/common';
+import type { ActivitySerializationContext, Payload } from '@temporalio/common';
+import { ApplicationFailure, defaultPayloadConverter, fromPayloadsAtIndex, rootCause } from '@temporalio/common';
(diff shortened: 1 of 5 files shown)


packages/test/src/test-async-completion.ts
```diff
@@ -61,13 +66,177 @@ async function makeNotFoundTaskToken(conn: Connection, namespace: string): Promi
...
+  await client.activity.heartbeat({ workflowId: 'wf1', activityId: 'act1' }, makeContextTrace('details'));
+
+  const payloads = recorded.recordActivityTaskHeartbeatById?.details?.payloads;
+  t.deepEqual(fromPayloadsAtIndex<ContextTrace<string>>(defaultPayloadConverter, 0, payloads), {
+    label: 'details',
+    trace: [enc('details', ctx)],
+  });
+});
+
+test('AsyncCompletionClient heartbeat with task token uses explicit activity context', async (t) => {
```
chris-olszewski, line 225: Not that it is super interesting, but a case asserting behavior for task token w/o explicit serialization context would be a good add for completeness.

chris-olszewski (approved):

PR #2060, opened 2026-05-18
Serialization context support for codecs

THardy98:
## What was changed
Adds serialization context support for codecs

## Why?
parity between converters and codecs for serialization context support

1. Part of https://github.com/temporalio/sdk-typescript/issues/1661

2. How was this tested:
Integration test suites

3. Any docs updates needed?
No


mjameswh:
That's not new, but `workflow-codec-runner.ts‎` is worrying me.

That's something we'll have to look into at some point. But not today.

mjameswh (approved):

PR #2077, opened 2026-05-21
chore: Multiple dependencies cleanup

mjameswh:
## What was changed

- Remove dependency on the 'uuid' package
- Remove dependency on the 'abort-controller' package
- Bumped NPM package 'webpack' to v5.106.2
- Bumped NPM package 'source-map-loader' to v5.0.0
- Bumped GHA action 'actions/cache' from v5.0.4 to v5.0.5
- Bumped GHA action 'actions/upload-artifact' from v7.0.0 to v7.0.1
- Other various dependencies cleanups


packages/meta/src/index.ts
```diff
@@ -12,3 +12,5 @@ export * as testing from '@temporalio/testing';
 export * as opentelemetry from '@temporalio/interceptors-opentelemetry';
 export * as envconfig from '@temporalio/envconfig';
 export * as plugin from '@temporalio/plugin';
+export * as cloud from '@temporalio/cloud';
+export * as aisdk from '@temporalio/ai-sdk';
```
chris-olszewski, line 16: Are we updating the meta package still? I thought it was frozen in time

packages/meta/package.json
```diff
@@ -6,6 +6,7 @@
     "@temporalio/activity": "workspace:*",
     "@temporalio/ai-sdk": "workspace:*",
     "@temporalio/client": "workspace:*",
+    "@temporalio/cloud": "workspace:*",
```
chris-olszewski, line 9: Same question regarding if we should update this package

chris-olszewski (approved):
LGTM, just wondering about if it makes sense to update the `meta` package if we never mean to publish it.

packages/meta/src/index.ts
mjameswh: The meta package is no longer getting published. However, it is used as part of the API doc generation process; that basically controls what gets included in the generated site. Kind of hacky, to be honest, but that works with very little downside, so no justification to put efforts on changing this at the moment.

PR #2085, opened 2026-05-27
Add Temporal Operation Handler

VegetarianOrc:
## What was changed
Added `TemporalOperationHandler` for Nexus operations backed by Temporal workflows, including sync/async result helpers, a Nexus-aware client wrapper, operation-token parsing, default workflow cancellation, and public exports from `@temporalio/nexus`.

Added unit and integration coverage for token validation, type inference, sync/async operation execution, cancellation behavior, invalid tokens, duplicate async starts, and retrying after a failed workflow start.

  ## Why?
This gives Nexus handlers a first-class way to return synchronous results or start a Temporal workflow as the asynchronous backing operation, while preserving correct token validation and cancellation behavior.

  ## Checklist

  1. How was this tested:
  - New tests added
  - Existing tests

### DIFF
--- a/packages/nexus/src/token.ts
+++ b/packages/nexus/src/token.ts
@@ -5,35 +5,45 @@
  * @internal
  * @hidden
  */
-export interface WorkflowRunOperationToken {
+export interface OperationToken {
   /**
-   * Version of the token, by default we assume we're on version 1, this field is not emitted as part of the output,
+   * Version of the token, by default we assume we're on version 0, this field is not emitted as part of the output,
    * it's only used to reject newer token versions on load.
    */
   v?: number;
 
   /**
-   * Type of the Operation. Must be OPERATION_TOKEN_TYPE_WORKFLOW_RUN.
+   * Type of the Operation.
    */
   t: OperationTokenType;
 
   /**
-   * Namespace of the workflow.
+   * Namespace of the operation.
    */
   ns: string;
 
   /**
    * ID of the workflow.
    */
-  wid: string;
+  wid?: string;
 }
-type OperationTokenType = (typeof OperationTokenType)[keyof typeof OperationTokenType];
 
 /**
  * @internal
  * @hidden
  */
-const OperationTokenType = {
+export interface WorkflowRunOperationToken extends OperationToken {
+  t: typeof OperationTokenType.WORKFLOW_RUN;
+  wid: string;
+}
+
+export type OperationTokenType = (typeof OperationTokenType)[keyof typeof OperationTokenType];
+
+/**
+ * @internal
+ * @hidden
+ */
+export const OperationTokenType = {
   WORKFLOW_RUN: 1,
 } as const;
 
@@ -50,11 +60,11 @@ export function generateWorkflowRunOperationToken(namespace: string, workflowId:
 }
 
 /**
- * Load and validate a workflow run Operation token.
+ * Load and validate the common fields of an Operation token.
  */
-export function loadWorkflowRunOperationToken(data: string): WorkflowRunOperationToken {
+export function loadOperationToken(data: string): OperationToken {
   if (!data) {
-    throw new TypeError('invalid workflow run token: token is empty');
+    throw new TypeError('invalid operation token: token is empty');
   }
 
   let decoded: string;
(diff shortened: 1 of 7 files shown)


packages/nexus/src/workflow-helpers.ts
```diff
@@ -159,3 +160,188 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+  }
+}
+
+/**
+ * A handler function for the {@link TemporalOperationHandler} constructor.
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export type TemporalOperationStartHandler<I, O> = (
+  ctx: nexus.StartOperationContext,
```
Quinn-With-Two-Ns: I think we don't want to take `nexus.StartOperationContext` here right? I believe we adjusted this in Python to be a temporal exported type?

packages/nexus/src/workflow-helpers.ts
```diff
@@ -159,3 +160,188 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+  client: TemporalNexusClient,
+  input: I
+) => Promise<TemporalOperationResult<O>>;
+
+/**
+ * Options for customizing a {@link TemporalOperationHandler}.
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export interface TemporalOperationHandlerOptions {
```
Quinn-With-Two-Ns, line 307:
```suggestion
  cancelWorkflowRun?: (ctx: nexus.CancelOperationContext, client: Client, workflowId: string) => Promise<void>;
```

One thought I had was if we want to expose more options when cancelling a workflow run, can we do that with this current API?

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: I've added interfaces that implement the nexusrpc interfaces now to avoid leaking these types.

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: Moved to an options object rather than a plain string for the workflowId argument to allow more options to get added.

packages/nexus/src/token.ts
```diff
@@ -64,26 +82,57 @@ export function loadWorkflowRunOperationToken(data: string): WorkflowRunOperatio
...
   }
 
-  if (token.t !== OperationTokenType.WORKFLOW_RUN) {
-    throw new TypeError(`invalid workflow token type: ${token.t}, expected: ${OperationTokenType.WORKFLOW_RUN}`);
+  if (typeof token !== 'object' || token == null) {
+    throw new TypeError(`invalid operation token: expected object, got ${typeof token}`);
   }
   if (token.v !== undefined && token.v !== 0) {
-    throw new TypeError('invalid workflow run token: "v" field should not be present');
+    throw new TypeError('invalid operation token: "v" field should not be present');
```
chris-olszewski, line 96: Just clarifying: `0` is valid because that's the current version we expect and would reject any newer versions?

If so, a more naunced message than "v should not be present" might be beneficial.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -154,3 +166,188 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+   * @experimental Nexus support in Temporal SDK is experimental.
+   */
+  startWorkflow<T extends Workflow>(
+    workflowTypeOrFunc: string | T,
+    workflowOptions: WorkflowStartOptions<T>
+  ): Promise<TemporalOperationResult<WorkflowResultType<T>>>;
+}
+
+class TemporalNexusClientImpl implements TemporalNexusClient {
+  private asyncOperationStarted = false;
```
chris-olszewski, line 231: Running feels a little more accurate than started considering we clear it on completion.
```suggestion
  private asyncOperationRunning = false;
```

packages/nexus/src/workflow-helpers.ts
```diff
@@ -154,3 +166,188 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+   * Starts a workflow run as the asynchronous backing operation for the current Nexus Operation.
+   *
+   * @experimental Nexus support in Temporal SDK is experimental.
+   */
+  public async startWorkflow<T extends Workflow>(
+    workflowTypeOrFunc: string | T,
+    workflowOptions: WorkflowStartOptions<T>
+  ): Promise<TemporalOperationResult<WorkflowResultType<T>>> {
+    return await this.withAsyncOperationStartReservation(async () => {
+      const handle = await startWorkflow(this.startOperationContext, workflowTypeOrFunc, workflowOptions);
```
chris-olszewski, line 254: Do we need to be concerned with a `startWorkflow` request failing only on the response side (e.g. client timeout) which would clear `asyncOperationStarted`?

chris-olszewski (approved):
Nothing blocking, basically just questions from my end.

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: The normal retry path is covered by using the Nexus context's request ID as an idempotency key. So to trip on this a user would have to catch the response side failure and ignore it.

packages/nexus/src/token.ts
VegetarianOrc: That is correct. Definitely open to alternative error messages. Perhaps, just something along the lines of "unsupported version"?

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: Each handler invocation gets a new `TemporalNexusClient` that will have this flag cleared, but it isn't cleared until it's garbage collected (or the error scenario). I think I'll keep this as started since the context it's running in is the "start operation" handler and the intent is to return an async token. Plus that's how it's named in other SDKs 😅.

THardy98 (approved):

Quinn-With-Two-Ns (approved):

THardy98 (approved):
double approve (on the fix)

PR #2088, opened 2026-05-27
feat(extstore): Add foundational types, config, and initial logic for external storage.

cconstable:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/common/src/converter/extstore.ts
+++ b/packages/common/src/converter/extstore.ts
@@ -1,0 +1,328 @@
+import { temporal } from '@temporalio/proto';
+import Long from 'long';
+import { decode, encode } from '../encoding';
+import { ValueError } from '../errors';
+import type { Payload } from '../interfaces';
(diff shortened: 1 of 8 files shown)


packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+// ============================================================================
+// Configuration
+// ============================================================================
+
+/**
+ * Default {@link ExternalStorage.payloadSizeThreshold}: 256 KiB.
+ *
+ * @experimental
+ */
+export const DEFAULT_PAYLOAD_SIZE_THRESHOLD = 256 * 1024;
```
cconstable, line 136: It was mentioned that maybe this shouldn't be exposed but I think it's a helpful constant to have and I would rather folks use a constant rather than hardcoding numbers. Additionally the tests use this.

PR #2090, opened 2026-05-27
fix(lint): add .npmrc that fixes issue with linter memory.

cconstable:
## What was changed
Fixes:
```
$ pnpm run lint
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,10 @@ packages/*/package-lock.json
 contrib/*/package-lock.json
 /sdk-node.iml
 *~
+
+# tooling
 .claude/settings.local.json
+mise.toml
 
 # One test creates persisted SQLite DBs; they should normally be deleted automatically,
 # but may be left behind in some error scenarios.
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+node-options=--max-old-space-size=8192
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -12,7 +12,10 @@ packages/*/package-lock.json
 contrib/*/package-lock.json
 /sdk-node.iml
 *~
+
+# tooling
 .claude/settings.local.json
+mise.toml
```
cconstable, line 18: surprised we didn't have this here given how many people at temporal use mise

.npmrc
```diff
@@ -0,0 +1 @@
+node-options=--max-old-space-size=8192
```
brianstrauch: Can we remove this line from the CI configuration and verify lints/tests still work on this branch?

brianstrauch: I count at least three usages

cconstable: Can you explain what you mean? There were no issues in CI before this fix AFAIK and if the checks pass we should be good yes? Locally, I verified that I was receiving memory errors before the fix.

brianstrauch: Yep, here's one instance that we shouldn't need any more: https://github.com/temporalio/sdk-typescript/blob/85cf60674b0fdee800badd4e3f59f6f642fdbf1b/.github/workflows/conventions.yml#L56

mjameswh: Yes, we added `NODE_OPTIONS=` in GHA workflows two weeks ago.

cconstable: Oh I see, yep I can remove those and we will see if it passes CI. I would imagine yes.

brianstrauch (approved):

chris-olszewski (approved):

PR #2091, opened 2026-05-27
contrib: add @temporalio/strands-agents plugin

brianstrauch:
## Summary

Adds `@temporalio/strands-agents`, a Temporal plugin that runs [Strands Agents](https://strandsagents.com/) inside Temporal Workflows. Model invocations, MCP tool calls, and `activityAsTool`/`activityAsHook` calls all dispatch through Temporal Activities for durable execution and Temporal-managed retries.

## API

```ts
// Workflow side
import { TemporalAgent, TemporalMCPClient, workflow as strandsWorkflow } from '@temporalio/strands-agents';

const agent = new TemporalAgent({
  model: 'bedrock',
  tools: [
    strandsWorkflow.activityAsTool('getWeather', { description: '…', inputSchema: { … } }),
    new TemporalMCPClient({ server: 'filesystem' }),
  ],
});
const result = await agent.invoke(prompt);

// Worker side
new StrandsPlugin({
  models: { bedrock: () => new BedrockModel({ … }) },
  mcpClients: { filesystem: () => new McpClient({ … }) },
});
```

Mirrors `temporalio[strands-agents]` from sdk-python (`TemporalAgent`, `TemporalMCPClient`, `activity_as_tool`, `activity_as_hook`, `auto_heartbeater`).

## Workflow-bundle plumbing

`@strands-agents/sdk@1.3.0`'s index transitively pulls `fs`, `node:*`, and MCP transport modules into any workflow that imports `Agent`/`McpClient`. `StrandsPlugin.configureBundler` handles this by:

- Ignoring `fs` (statically imported from `vended-plugins/skills` and `vended-tools/file-editor`, both unreachable from workflow code).
- Replacing `node:fs/promises`/`os`/`path`/`process`/`stream` and `@modelcontextprotocol/sdk/client/{stdio,sse}.js` with an empty stub via `NormalModuleReplacementPlugin` (dynamic imports inside `mcp-config.js`'s server-only code paths).
- Disabling async chunks / code-splitting so webpack's JSONP chunk loader (which uses `self`/`document`) never ships in the bundle.

`load-polyfills.ts` installs `Headers`, `web-streams-polyfill`, `@ungap/structured-clone`, and a deterministic `crypto.randomUUID` backed by `workflow.uuid4()` — `@ungap/structured-clone` calls `crypto.randomUUID()` internally and Temporal's sink path uses `structuredClone` for log payloads, so without the polyfill the first `logger.warn` from inside an agent crashes the workflow.

`TemporalMCPClient.listTools()` returns lightweight `TemporalMCPTool` wrappers (lazy-required to break the cycle) because `McpTool` isn't re-exported from `@strands-agents/sdk`'s public index. **If [strands-agents/sdk-typescript#1108](https://github.com/strands-agents/sdk-typescript/pull/1108) merges**, `temporal-mcp-tool.ts` can be deleted entirely and `TemporalMCPClient.listTools()` can return real `McpTool` instances bound to itself — picking up the built-in content mapping for images, embedded resources, and URL-elicitation errors that our wrapper currently elides.

## Tests

9 ava integration tests in `contrib/strands/src/__tests__/test-strands.ts`:
- Model dispatch via `invokeModel` activity
- `activityAsTool` end-to-end
- `TemporalMCPClient` listTools + callTool through per-server activities
- In-workflow `tool()` (Strands `FunctionTool`) running inside the sandbox
- `activityAsHook` on `AfterToolCallEvent`
- `structuredOutputSchema` via the `strands_structured_output` tool
- Activity-tool interrupt round-trip through the failure converter + workflow-signal resume
- `streamingTopic` events consumed via `WorkflowStreamClient.subscribe`
- Captured-history replay (validates the deterministic `crypto.randomUUID` polyfill stays replay-safe)

## Test plan
- [x] `pnpm build` clean
- [x] `pnpm test` in `contrib/strands` — 9/9 passing
- [x] Reviewer-side smoke: install in a downstream project and run a real agent against Bedrock


pnpm-workspace.yaml
```diff
@@ -31,6 +32,7 @@ minimumReleaseAge: 20160
 minimumReleaseAgeExclude:
   - '@temporalio/*'
   - nexus-rpc
+  - '@strands-agents/sdk'
```
mjameswh: Why a cooldown-period exclusion?

contrib/strands/src/__tests__/test-strands.ts
```diff
@@ -0,0 +1,492 @@
...
+    });
+    t.is(result, 'hello from stub');
+  });
+});
+
+test('TemporalAgent without `model:` resolves to the single registered factory', async (t) => {
+  // Fix for the case where the caller registered exactly one model factory
+  // but omitted `model:` on the agent. Previously this raised inside
+  // ModelActivity.getModel because `defaultName` was only set on the
+  // implicit-Bedrock branch; now the plugin uses the single key as default.
```
JasonSteving99: nit: this comment seems to be referencing back to some old state during development. Probably worth rewording

contrib/strands/src/__tests__/test-strands.ts
```diff
@@ -0,0 +1,492 @@
...
+        models: {
+          test: () =>
+            new StubModel([
+              toolCallTurn('getWeather', 'call_1', { location: 'Tokyo' }),
+              textTurn('the weather is sunny'),
+            ]),
+        },
+      }),
+    ],
+    activities: { getWeather },
```
JasonSteving99, line 250: I don't think this test asserts that the `getWeather` activity was actually called. Mind adding an assert for that?

contrib/strands/src/__tests__/test-strands.ts
```diff
@@ -0,0 +1,492 @@
...
+            ]),
+        },
+        mcpClients: { testServer: factory },
+      }),
+    ],
+  });
+
+  await worker.runUntil(async () => {
+    const result = await executeWorkflow(mcpAgent, { args: ['list files'] });
+    t.is(result, 'found files');
```
JasonSteving99, line 304: same as above test, I think this test could use an assert that the mcp tool was _actually_ called here

I think skimming the below tests this may be generally true for more of the following tests as well

contrib/strands/src/model-activity.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+ * Input for the `invokeModel` activity. Mirrors `Model.stream(messages, options)`
+ * with the chosen factory name carried alongside.
+ */
+export interface InvokeModelInput {
+  modelName?: string;
+  messages: unknown;
+  options?: unknown;
```
JasonSteving99: do these have to be `unknown`?

contrib/strands/src/temporal-activity-tool.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Options accepted by {@link activityAsTool}.
+ *
+ * `name` defaults to the activity function's `.name`. `description` defaults
+ * to an empty string and SHOULD be supplied so the model has signal about
+ * when to call the tool. `inputSchema` is the JSON Schema that describes
+ * the activity's single argument; pass a literal JSON Schema object, or use
+ * a Zod schema with {@link https://github.com/StefanTerdell/zod-to-json-schema | zod-to-json-schema}
+ * to derive one.
```
JasonSteving99: is it possible to do this for the user automatically?

contrib/strands/src/temporal-activity-tool.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export interface ActivityAsToolOptions {
+  name?: string;
+  description?: string;
+  inputSchema?: unknown;
```
JasonSteving99: It looks like the type strands is expecting is `JSONSchema` defined [here](https://github.com/strands-agents/sdk-typescript/blob/0f99011408c45dcc6ca403794f60c05a1ac61eb8/strands-ts/src/tools/types.ts#L29). Can we use that instead of `unknown`?

contrib/strands/src/plugin.ts
```diff
@@ -0,0 +1,167 @@
...
+      invokeModel: (input: never) => modelActivity.invokeModel(input),
+      invokeModelStreaming: (input: never) => modelActivity.invokeModelStreaming(input),
+    };
+
+    const mcpClients = options.mcpClients ?? {};
+    for (const [server, factory] of Object.entries(mcpClients)) {
+      const list = buildListToolsActivity(server);
+      const call = buildCallToolActivity(server, factory);
+      activities[`${server}-listTools`] = list as (...args: never[]) => unknown;
+      activities[`${server}-callTool`] = call as (...args: never[]) => unknown;
```
JasonSteving99: could call the utils exported by `contrib/strands/src/temporal-mcp-client.ts`

contrib/strands/src/temporal-mcp-client.ts
```diff
@@ -0,0 +1,188 @@
...
+  factory: () => McpClient
+): (input: CallToolInput) => Promise<JSONValue> {
+  const name = callToolActivityName(server);
+  const fn = async (input: CallToolInput): Promise<JSONValue> => {
+    ActivityContext.current().log.debug(`Calling MCP tool ${input.toolName} on server ${server}`, {
+      activityId: activityInfo().activityId,
+    });
+    const client = factory();
+    try {
+      await client.connect();
```
JasonSteving99: I didn't read *that* closely to fully parse this out, but seeing explicit `client.connect()` and `clien.disconnect()` calls wrapping mcp tool calls makes me wonder if this is less efficient than strands' default behavior here? I assume there'd be a longer lived client connection across potentially multiple successive mcp calls? That's just an assumption, so if I assumed wrong stop reading here..

If that's the case, could you consider caching the already-open connection so that we don't need to reconnect repeatedly in the case where sticky workflow task routing hits the same worker instance for multiple successive mcp tool calls? 

Here's [a similar example](https://github.com/temporalio/sdk-python/blob/bea727c7ec02aed2408ef00659a3aeec1d9f5f06/temporalio/contrib/openai_agents/sandbox/_sandbox_client_provider.py#L88-L92) in openai agents plugin.

Note that if you did this, your `McpClient::connect` / `McpClient::disconnect` override impls wouldn't be NOPs anymore as you'd be able to follow the full strands lifecycle (connect would add to the cache, disconnect would delete from cache)

contrib/strands/src/temporal-mcp-client.ts
```diff
@@ -0,0 +1,188 @@
...
+import type { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
+import * as workflow from '@temporalio/workflow';
+import type { ActivityOptions } from '@temporalio/workflow';
+import { activityInfo, Context as ActivityContext } from '@temporalio/activity';
+
+/** Cached tool descriptor; one per MCP tool, populated at worker startup. */
+export interface McpToolInfo {
+  name: string;
+  description: string;
+  inputSchema: unknown;
```
JasonSteving99: similar to other comment, I think this `inputSchema: unknown` could be `JSONSchema`

contrib/strands/src/temporal-mcp-tool.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  constructor(server: string, info: McpToolInfo, activityOptions: ActivityOptions) {
+    super();
+    this.server = server;
+    this.name = info.name;
+    this.description = info.description || `Tool which performs ${info.name}`;
+    this.toolSpec = {
+      name: info.name,
+      description: this.description,
+      inputSchema: info.inputSchema as never,
```
JasonSteving99: probably shouldn't need this `as never` if inputSchema is properly typed as `JSONSchema`

pnpm-workspace.yaml
brianstrauch: Removed!

contrib/strands/src/model-activity.ts
brianstrauch: Nope, and glad you mentioned this, there were a ton of other instances where we were lazy about typing.

contrib/strands/src/temporal-activity-tool.ts
brianstrauch: Nice suggestion, also applied to MCP

contrib/strands/src/temporal-mcp-client.ts
brianstrauch: Great suggestion, will make a note to apply this to other plugins where needed too. Added but the overrides still need to be NOPs, since there would be no way to guarantee that a connect, tool call, and disconnect all map to the same worker.

contrib/strands/src/__tests__/test-strands.ts
```diff
@@ -250,12 +235,23 @@ test('invokeModel reconstructs Message instances from wire data', async (t) => {
 test('activityAsTool dispatches a registered activity from the agent loop', async (t) => {
   const { createWorker, executeWorkflow } = helpers(t);
 
+  // The final string is the stub's scripted second turn, so it can't prove the
+  // activity ran. Instead capture the second stream() call
```
JasonSteving99: I still think this isn't necessarily validating what we need it to. With this not actually validating that the activity literally gets called, we wouldn't notice if some bug were introduced where the model returns a tool call, but this plugin somehow fails to invoke the activity. Right now this test is moreso testing the `StubModel` impl than the plugin itself.

Can you wrap the underlying activity (`getWeather`) def in something that logs the args that it was called with so we can assert that here? Should be fairly straightforward to even make an edit directly in `./activities/strands` to collect that. 

same comment on other similar tests

contrib/strands/src/__tests__/test-strands.ts
```diff
@@ -302,17 +307,89 @@ test('TemporalMCPClient lists and calls tools through per-server activities', as
...
+  const counters = { factoryCalls: 0, connects: 0, disconnects: 0, callTools: 0 };
+  class CountingMcpClient extends StubMcpClient {
+    override async connect(): Promise<void> {
+      counters.connects++;
+    }
+    override async disconnect(): Promise<void> {
+      counters.disconnects++;
+    }
+    override async callTool(tool: { name: string }, args: JSONValue): Promise<JSONValue> {
+      counters.callTools++;
```
JasonSteving99, line 323: Let's also make sure we track the args to assert it's expected.

contrib/strands/src/temporal-mcp-client.ts
```diff
@@ -133,10 +133,75 @@ export function buildListToolsActivity(server: string): (input: { server: string
...
+
 /**
- * Builds the per-server `{server}-callTool` activity. Opens an MCP session
- * through the user-supplied factory, invokes the named tool, and returns
- * the raw result for the workflow to forward as a Strands tool result.
+ * How long an idle MCP connection is kept open before it's disconnected. The
+ * timer resets on every {@link callToolActivityName | callTool} that reuses the
+ * connection. Exported for tests.
+ */
+export const MCP_CONNECTION_IDLE_MS = 5 * 60 * 1000;
```
JasonSteving99, line 189: this is a good idea! How about we wire this idle timeout through as a plugin config option?

contrib/strands/src/temporal-mcp-client.ts
```diff
@@ -133,10 +133,75 @@ export function buildListToolsActivity(server: string): (input: { server: string
...
+  }
+}
+
+function getConnection(server: string, factory: () => McpClient): Promise<CachedConnection> {
+  let entry = CONNECTIONS.get(server);
+  if (entry === undefined) {
+    entry = (async () => {
+      const client = factory();
+      await client.connect();
+      const tools = await client.listTools();
```
JasonSteving99: I'm not sure how i feel about the preemptive `client.listTools()` call here. If all the sdk asked for was a connection, maybe it's not going to ever call `client.listTools()` and this cache warming is wasted effort. Could you just extract some generic ttl caching utility and have two separate caches one for tools, and one for clients? 

But on another note, is it even right to cache tools at all? MCP servers may redeploy or something and start publishing different tools, no?

contrib/strands/src/temporal-mcp-client.ts
```diff
@@ -147,17 +212,17 @@ export function buildCallToolActivity(
     ActivityContext.current().log.debug(`Calling MCP tool ${input.toolName} on server ${server}`, {
       activityId: activityInfo().activityId,
     });
-    const client = factory();
+    const { client, tools } = await getConnection(server, factory);
+    const tool = tools.find((t) => t.name === input.toolName);
+    if (tool === undefined) {
+      throw new Error(`MCP tool '${input.toolName}' not found on server '${server}'`);
+    }
```
JasonSteving99: I actually don't think you need this tool call activity to be listing the tools before calling the tool. As best as I can tell, [the underlying `callTool` impl in strands](https://github.com/strands-agents/sdk-typescript/blob/0f99011408c45dcc6ca403794f60c05a1ac61eb8/strands-ts/src/mcp.ts#L412) makes MCP tool calls using literally just the tool name and the args which you already have here in the `CallToolInput`. 

Is there another specific reason why this running on Temporal requires doing a tool lookup before calling each time? I could just be missing something (also sorry for not catching this on first review pass)

contrib/strands/src/temporal-mcp-client.ts
brianstrauch: Fixed the first comment, so now we're in line with sdk-python on this.

I really like the tool caching approach because it (1) gets us a little closer performance-wise to native Strands and (2) matches the philosophy of durable execution a bit better. If we allowed MCP redeployments mid-workflow they could introduce breaking changes that break replay.

contrib/strands/src/temporal-mcp-client.ts
brianstrauch: Fixed

JasonSteving99 (approved):

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -159,6 +159,19 @@ export function injectGlobals(context: vm.Context): void {
...
+  // emitted by TypeScript and shipped by some dependencies — work inside the
+  // workflow sandbox. This must run before the global scope is frozen.
+  // `Symbol.for` yields a stable registry symbol, keeping replay deterministic.
+  // The context's built-in `Symbol` isn't reliably reachable from this outer
+  // scope, so run the assignment inside the context.
+  vm.runInContext(
+    `if (typeof Symbol.dispose !== 'symbol') Symbol.dispose = Symbol.for('nodejs.dispose');
+     if (typeof Symbol.asyncDispose !== 'symbol') Symbol.asyncDispose = Symbol.for('nodejs.asyncDispose');`,
+    context
+  );
```
chris-olszewski: I think the comment might be misleading unless these symbols are included in the subset of Node globals provided to the VM.

We should do a similar pattern as above so these aren't enumerable.
```
       Object.defineProperty(Symbol, 'asyncDispose', {
         value: Symbol.for('nodejs.asyncDispose'),
         writable: false,
         enumerable: false,
         configurable: false,
       });
```

chris-olszewski (approved):
Talked with James, `Symbol` polyfill seems to be fine. Once changed to define Symbols via `Object.defineProperty` fine to merge from SDK perspective.

chris-olszewski:
Another you could take is to polyfill the Symbols as part of this plugin's `preloadModules` setting. This would avoid needing to change the isolate environment on every workflow.

PR #2092, opened 2026-05-28
External Storage: foundational types and config

cconstable:
## What

- Foundational types, configuration, and wire-format codec for **External Storage** in `@temporalio/common`. 
- Plumbing only, no public API is exposed yet.
- `converter/extstore.ts` (new): the `StorageDriver` contract, the `ExternalStorage` config (driver registry, optional selector, size threshold).
- `converter/internal-non-workflow/extstore-helpers.ts` (new): encode/decode  `temporal.api.sdk.v1.ExternalStorageReference`. Pulls in proto types.
- `proto/scripts/compile-proto.ts`: include `external_storage.proto` in the `@temporalio/proto` bundle.
- Tests for config validation and the reference-payload round-trip.

### Visibility
Nothing is reachable from a public entry point yet

## Why

First phase of External Storage for the TypeScript SDK, part of the Large Payloads work. Integration into the payload pipeline comes later.


packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+// ============================================================================
+// Configuration
+// ============================================================================
+
+/**
+ * Default {@link ExternalStorage.payloadSizeThreshold}: 256 KiB.
+ *
+ * @experimental
+ */
+export const DEFAULT_PAYLOAD_SIZE_THRESHOLD = 256 * 1024;
```
cconstable: @jmaeagle99 I think this could be helpful for users of the SDK and it's used in the tests as well. Are there any big drawbacks to leaving this exported?

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+  /** The activity type name, if available. */
+  readonly type?: string;
+}
+
+/**
+ * Identity of the workflow or activity that produced the payloads being stored.
+ *
+ * @experimental
+ */
+export type StorageDriverTarget = StorageDriverWorkflowInfo | StorageDriverActivityInfo;
```
jmaeagle99: nit: We called this `StorageDriverTargetInfo` in Go and I don't think Python has a named sum type for this. I would make it consistent with either one of those.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+  readonly name: string;
+  readonly type: string;
+  store(context: StorageDriverStoreContext, payloads: Payload[]): Promise<StorageDriverClaim[]>;
+  retrieve(context: StorageDriverRetrieveContext, claims: StorageDriverClaim[]): Promise<Payload[]>;
+}
+
+/**
+ * User-supplied function that picks the destination driver for a given
+ * payload, or returns `null` to keep the payload inline. Required when
+ * more than one driver is registered.
```
jmaeagle99: >  Required when more than one driver is registered.

nit: This is a behavior of `ExternalStorage` and not of the selector itself. I would make the comment there rather than here, if it isn't already there.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+
+/**
+ * Build a {@link StorageDriverTarget} from a codec-layer {@link SerializationContext}.
+ * Bridges the codec pipeline (which threads `SerializationContext`) into the storage
+ * driver API (which expects the narrower, extstore-specific target shape).
+ *
+ * @internal
+ * @experimental
+ */
+export function storageTargetFromContext(ctx: SerializationContext | undefined): StorageDriverTarget | undefined {
```
jmaeagle99: `SerializationContext` is only available/popoulated under symmetrical access to the underlying information (where the information is available on both the caller and the callee side). I would not tie `StorageDriverTarget` specifically to `SerializationContext` since it doesn't have that symmetry requirement; it is best effort of providing as much information as possible and is only use on the caller/storing side.

packages/common/src/converter/extstore.ts
jmaeagle99: It being used in our tests shouldn't necessarily mean that we make it a publicly exposed and documented value. It is important for our testing because we want to make sure that system exhibits different behaviors above/below the threshold value, but I'm not convinced that customers would be testing that.

What customer scenario does exposing this value actually enable?

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    if (driversByName.size > 1 && driverSelector === undefined) {
+      throw new ValueError(
+        'ExternalStorage.driverSelector is required when more than one driver is registered'
+      );
+    }
+
+    this.drivers = [...drivers];
+    this.driverSelector = driverSelector;
```
jmaeagle99: Consider making an artificial driver selector if driver count is 1 and no selector was specified. Then observers of the `ExternalStorage` type don't have to reason about whether a selector is used or not and how they interpret the driver list when there is no selector.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+    this.payloadSizeThreshold = payloadSizeThreshold;
+    this.driversByName = driversByName;
+  }
+
+  /** Look up a registered driver by name. Returns `undefined` if no driver with that name is registered. */
+  getDriver(name: string): StorageDriver | undefined {
```
jmaeagle99: Why does this function return `StorageDriver | undefined` whereas a driver selector returns `StorageDriver | null`? Not necessarily suggesting a change, more for my understanding.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+const EXTSTORE_REFERENCE_MESSAGE_TYPE_BYTES = encode(EXTSTORE_REFERENCE_MESSAGE_TYPE);
+
+/** 
+ * True if the payload is a External Storage reference:
+ * temporal.api.sdk.v1.ExternalStorageReference
+ * 
+ * @internal @experimental 
+ * */
+export function isReferencePayload(payload: Payload): boolean {
+  if (payload.externalPayloads && payload.externalPayloads.length > 0) {
```
jmaeagle99: I don't think we should include this in the check of if something is an external storage reference. Someone can create a payload whose content is `ExternalStorageReference` and not set this additional data on the payload and it is still a valid storage reference.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,328 @@
...
+
+/**
+ * Decode a reference payload (v1 proto encoding or legacy JSON encoding). Throws
+ * {@link ValueError} if the payload is not a reference or is malformed; callers
+ * should gate on {@link isReferencePayload} first.
+ *
+ * @internal
+ * @experimental
+ */
+export function decodeReferencePayload(payload: Payload): DecodedReferencePayload {
```
jmaeagle99: I think these encode and decode functions should perform the same conversions as `ProtobufJsonPayloadConverter`, whether that's copying the implementation or constructing one of these for use within external storage. I'm not certain that the current implementation of these methods is correct and interoperable with the other SDKs.

packages/common/src/index.ts
```diff
@@ -11,6 +11,16 @@ export * from './activity-options';
 export { ActivityCancellationDetailsOptions, ActivityCancellationDetails } from './activity-cancellation-details';
 export { SuggestContinueAsNewReason } from './continue-as-new';
 export * from './converter/data-converter';
+export { DEFAULT_PAYLOAD_SIZE_THRESHOLD, ExternalStorage, StorageDriverClaim } from './converter/extstore';
```
jmaeagle99: Any reason why these are not included in the export on the next line?

packages/test/src/test-extstore-helpers.ts
```diff
@@ -0,0 +1,109 @@
...
+  const converter = makeConverter(externalStorage);
+  const context: WorkflowSerializationContext = {
+    type: 'workflow',
+    namespace: 'ns',
+    workflowId: 'wf-1',
+  };
+
+  const payloads = await encodeToPayloadsWithContext(converter, context, ['hello', { v: 42 }]);
+  t.is(payloads!.length, 2);
+  t.is(driver.storeCalls.length, 1);
```
jmaeagle99: Should this be 2? I think that each value should be externally storage since the threshold is 0.

packages/test/src/test-extstore-helpers.ts
```diff
@@ -0,0 +1,109 @@
...
+
+test('decodeArrayFromPayloads raises TMPRL1105 when no externalStorage and reference payload received', async (t) => {
+  const driver = makeFakeDriver();
+  const writerStorage = new ExternalStorage({ drivers: [driver], payloadSizeThreshold: 0 });
+  const writerConv = makeConverter(writerStorage);
+  const storedPayloads = await encodeToPayloads(writerConv, 'big');
+
+  const readerConv = makeConverter(undefined);
+  await t.throwsAsync(() => decodeArrayFromPayloads(readerConv, storedPayloads!), {
+    instanceOf: Error,
```
jmaeagle99: Either the test should describe more generically what it's testing (e.g. remove TMPRL1105 from test title) or it should check for the specific error code or error type.

packages/test/src/test-extstore-helpers.ts
```diff
@@ -0,0 +1,109 @@
...
+    kind: 'workflow',
+    namespace: 'ns',
+    id: 'wf-1',
+  });
+
+  const decoded = await decodeArrayFromPayloads(converter, payloads!);
+  t.deepEqual(decoded, ['hello', { v: 42 }]);
+});
+
+test('encodeToPayloads is a no-op when externalStorage is undefined', async (t) => {
```
jmaeagle99: Can you add that isReferencePayload returns false?

packages/test/src/test-extstore-helpers.ts
```diff
@@ -0,0 +1,109 @@
...
+  const failure = await encodeErrorToFailure(
+    converter,
+    ApplicationFailure.create({ message: 'boom', details: [bigPayload] }),
+    context
+  );
+
+  const details = failure.applicationFailureInfo?.details?.payloads ?? [];
+  t.is(details.length, 1);
+  t.truthy(details[0]!.externalPayloads?.length);
+
```
jmaeagle99: Can you add that isReferencePayload returns true?

packages/test/src/test-extstore-helpers.ts
```diff
@@ -0,0 +1,109 @@
...
+// Ensure we don't accidentally regress the default loaded converter when
+// externalStorage isn't passed.
+test('loadDataConverter passes through externalStorage when set', (t) => {
+  const driver = makeFakeDriver();
+  const externalStorage = new ExternalStorage({ drivers: [driver] });
+  const loaded = loadDataConverter({ externalStorage });
+  t.is(loaded.externalStorage, externalStorage);
+});
+
+// Silence the unused-imports warning for converter types we keep around
```
jmaeagle99: Why are these needed?

packages/test/src/extstore-fake-driver.ts
```diff
@@ -0,0 +1,57 @@
...
+  const memory = new Map<string, Payload>();
+  let nextId = 0;
+  const driver: FakeDriver = {
+    name,
+    type: name,
+    storeCalls: [],
+    retrieveCalls: [],
+    async store(context, payloads) {
+      driver.storeCalls.push({ context, payloads });
+      if (opts.onStore) return opts.onStore(payloads);
```
jmaeagle99: I feel that we should do one of the following:
- Don't let onStore and onRetrieve return anything and just use it for error simulation
- Split this driver into two: one that performs in-memory storage and one that allows for generic callbacks.  Only one of them would need to support error simulation.

packages/common/src/converter/extstore.ts
cconstable: I'll go with `StorageDriverTargetInfo`. I do think the prefix might be odd in some languages.

packages/common/src/converter/extstore.ts
cconstable: Agreed it shouldn't _just_ be exposed for tests but I could imagine folks deciding to use Temporal sanctioned defaults for things like this and then writing some extensions/tooling that reference it. It is more API surface area though. 

My questions were "is this value important for users to know to understand how this feature works?" and "can I imagine someone hard-coding this number somewhere?".

I don't feel too strongly either way.

packages/common/src/converter/extstore.ts
cconstable: This should be returning `null` imo. Thanks!

packages/common/src/converter/extstore.ts
cconstable: That is true. This is leftover from refactoring.

packages/test/src/test-extstore-helpers.ts
cconstable: artifact of lint warnings I was having. removing.

packages/common/src/converter/extstore.ts
cconstable: Thanks for this catch. I went ahead with constructing one here.

packages/common/src/converter/data-converter.ts
```diff
@@ -60,12 +71,14 @@ export interface LoadedDataConverter {
   payloadConverter: PayloadConverter;
   failureConverter: FailureConverter;
   payloadCodecs: PayloadCodec[];
+  /** @experimental */
+  externalStorage?: ExternalStorage;
```
cconstable: Pushing an update so this either isn't publicly exposed or will throw immediately.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,282 @@
...
+import { patchProtobufRoot } from '@temporalio/proto/lib/patch-protobuf-root';
+import { decode } from '../encoding';
+import { ValueError } from '../errors';
+import type { Payload } from '../interfaces';
+import { ProtobufJsonPayloadConverter } from './protobuf-payload-converters';
+import { encodingTypes, METADATA_ENCODING_KEY, METADATA_MESSAGE_TYPE_KEY } from './types';
+
+const ExternalStorageReferenceProto = proto.temporal.api.sdk.v1.ExternalStorageReference;
+const PayloadProto = proto.temporal.api.common.v1.Payload;
+const claimConverter = new ProtobufJsonPayloadConverter(patchProtobufRoot(proto));
```
jmaeagle99: nit: storageReferenceConverter

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,282 @@
...
+
+/**
+ * Reference returned from {@link StorageDriver.store}. `claimData` is an
+ * opaque key/value map the driver uses to retrieve the payload later.
+ *
+ * @experimental
```
jmaeagle99, line 9: Are these types in this package meant to be referenced directly by customers or will they be referenced by them via other re-exports? If the latter, can these be marked internal or would that make the re-exports internal as well? Maybe a question for TS owners.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,282 @@
...
+export function isReferencePayload(payload: Payload): boolean {
+  const encodingValue = readMetadataString(payload, METADATA_ENCODING_KEY);
+  if (encodingValue === EXTSTORE_REFERENCE_ENCODING) {
+    return readMetadataString(payload, METADATA_MESSAGE_TYPE_KEY) === EXTSTORE_REFERENCE_MESSAGE_TYPE;
+  }
+  return false;
+}
+
+/** Parsed contents of a reference payload. `sizeBytes` is `0` for legacy payloads. @internal @experimental */
+export interface DecodedReferencePayload {
```
jmaeagle99: internal experimental?

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+  return {
+    driverName: ref.driverName,
+    claimData: { ...ref.claimData },
+    sizeBytes: readSizeBytes(payload),
+  };
+}
+
+/** Encode a reference payload to wire format. @internal @experimental */
+export function encodeReferencePayload({
```
jmaeagle99: internal experimental?

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,282 @@
...
+  getDriver(name: string): StorageDriver | null {
+    return this.driversByName.get(name) ?? null;
+  }
+}
+
+// ============================================================================
+// Wire format
+// ============================================================================
+
+/** @internal @experimental */
```
jmaeagle99: If these const are not exported, then I don't think they need internal or experimental.

jmaeagle99 (approved):
Left a few more mostly nitpick comments. Two things:
- Update the PR description to reflect the current state of the PR
- Hold off until a TS owner reviews and approves

packages/common/src/converter/extstore.ts
cconstable: I believe `@internal` will simple not produce docs when something tagged as `@internal` is included in a wildcard reexport e.g. `export * from`. 

My rules for annotating were not mutually exclusively:
1. if (WIP || new) && exported -> mark it as `@experimental`
2. if it shouldn't show up in public docs  -> mark it as `@internal`

cconstable: I should add that I felt it was OK to mark something as `@internal` now and then promote it to public later once it's actually usable. Maybe @mjameswh can weigh in on that.

mjameswh: > I felt it was OK to mark something as @internal now and then promote it to public later once it's actually usable.

I agree.

mjameswh: > I believe `@internal` will simple not produce docs when something tagged as `@internal`

Right, I just checked to confirm: the TypeScript language server itself does not honor the `@internal`, `@hidden`, and `@ignore` doc annotations. All three only affect generated API docs.

We can apparently configure TSC to strip symbols flagged with `@internal` when generating public `.d.ts` files. We'll have to experiment with that option to determine whether that is appropriate for us. Another possibility would be to post-process our compiled packages with API Extractor, which would allow for much more powerful transformations.

mjameswh: As a general rule, though, I generally encourage being explicit on public reexports. Avoid wildcard reexport.

It's a tiny bit more efforts for us, but provides much stronger visibility into what our public API surface really is.

Doing that does not eliminate 100% of use cases for `@internal` symbols, but that helps quite a lot.

packages/common/src/converter/extstore.ts
```diff
@@ -0,0 +1,295 @@
...
+import { patchProtobufRoot } from '@temporalio/proto/lib/patch-protobuf-root';
+import { decode } from '../encoding';
+import { ValueError } from '../errors';
+import type { Payload } from '../interfaces';
+import { ProtobufJsonPayloadConverter } from './protobuf-payload-converters';
+import { encodingTypes, METADATA_ENCODING_KEY, METADATA_MESSAGE_TYPE_KEY } from './types';
+
+const ExternalStorageReferenceProto = proto.temporal.api.sdk.v1.ExternalStorageReference;
+const PayloadProto = proto.temporal.api.common.v1.Payload;
+const storageReferenceConverter = new ProtobufJsonPayloadConverter(patchProtobufRoot(proto));
```
mjameswh: > `new ProtobufJsonPayloadConverter(patchProtobufRoot(proto));`

I don't think you should need to call `patchProtobufRoot()` here. The export of `@temporalio/proto` should already have been processed.

cconstable: Will remove, thanks!

packages/common/src/internal-non-workflow/extstore-helpers.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Throws {@link ValueError} if the payload is not a reference or is malformed.
+ * Callers should gate on {@link isReferencePayload} first.
+ *
+ * @internal
+ * @experimental
+ */
+export function decodeReferencePayload(payload: Payload): DecodedReferencePayload {
+  const ref = storageReferenceConverter.fromPayload<proto.temporal.api.sdk.v1.IExternalStorageReference>(payload);
+  if (!ref.driverName) {
+    // TODO: use the new stable error codes here
```
cconstable, line 55: I'll remove this comment.

PR #2094, opened 2026-05-28
External Storage: add external_storage_runner to handle core extstore logic

cconstable:
## What was changed
- Adds a `external_storage_runner` that handles core external storage logic.
- Adds some tentative errors and error codes. These just drafts and will change.

## Checklist
- New tests.

### DIFF
--- a/packages/common/src/errors.ts
+++ b/packages/common/src/errors.ts
@@ -73,3 +73,25 @@ export class NamespaceNotFoundError extends Error {
  */
 @SymbolBasedInstanceOfError('CompleteAsyncError')
 export class CompleteAsyncError extends Error {}
+
+/// -----------------------------------------------------------------------
+/// External Storage Errors
+/// -----------------------------------------------------------------------
+
+/**
+ * Thrown when a driver's `store()` or `retrieve()` raises. Wraps the
+ * underlying error in `cause` (network, auth, rate-limit, S3 5xx, etc).
+ *
+ * @experimental
+ */
+@SymbolBasedInstanceOfError('ExternalStorageDriverOperationFailedError')
+export class ExternalStorageDriverOperationFailedError extends Error {
+  constructor(
+    message: string,
+    public readonly driverName: string,
+    public readonly operation: 'store' | 'retrieve',
+    public readonly cause: unknown
+  ) {
+    super(message);
+  }
+}
--- a/packages/common/src/errors.ts
+++ b/packages/common/src/errors.ts
@@ -73,25 +73,3 @@ export class NamespaceNotFoundError extends Error {
  */
 @SymbolBasedInstanceOfError('CompleteAsyncError')
 export class CompleteAsyncError extends Error {}
-
-/// -----------------------------------------------------------------------
-/// External Storage Errors
-/// -----------------------------------------------------------------------
-
-/**
- * Thrown when a driver's `store()` or `retrieve()` raises. Wraps the
- * underlying error in `cause` (network, auth, rate-limit, S3 5xx, etc).
- *
- * @experimental
- */
-@SymbolBasedInstanceOfError('ExternalStorageDriverOperationFailedError')
-export class ExternalStorageDriverOperationFailedError extends Error {
-  constructor(
-    message: string,
-    public readonly driverName: string,
-    public readonly operation: 'store' | 'retrieve',
-    public readonly cause: unknown
-  ) {
-    super(message);
-  }
-}
--- a/packages/common/src/internal-non-workflow/external-storage-runner.ts
+++ b/packages/common/src/internal-non-workflow/external-storage-runner.ts
@@ -1,0 +1,222 @@
+/**
+ * Engine that drives External Storage `store` and `retrieve` operations.
+ *
+ * @module
+ */
+import { temporal } from '@temporalio/proto';
+
+import { StorageDriverClaim } from '../converter/extstore';
+import type {
+  ExternalStorage,
+  StorageDriver,
+  StorageDriverRetrieveContext,
+  StorageDriverStoreContext,
+  StorageDriverTargetInfo,
+} from '../converter/extstore';
+import { ExternalStorageDriverOperationFailedError, ValueError } from '../errors';
+import type { Payload } from '../interfaces';
+import { decodeReferencePayload, encodeReferencePayload, isReferencePayload } from './extstore-helpers';
+
+const PayloadProto = temporal.api.common.v1.Payload;
+
+/** @internal @experimental */
+export interface ExternalStoreOptions {
+  /** Identity of the workflow or activity that produced the payloads. */
+  target?: StorageDriverTargetInfo;
+  /** Aborts the in-flight store operation. */
+  abortSignal?: AbortSignal;
+}
+
+/** @internal @experimental */
+export interface ExternalRetrieveOptions {
+  /** Aborts the in-flight retrieve operation. */
+  abortSignal?: AbortSignal;
+}
+
+/**
+ * Drives External Storage operations against a configured {@link ExternalStorage}.
+ *
+ * @internal
+ * @experimental
+ */
+export class ExternalStorageRunner {
+  constructor(private readonly externalStorage: ExternalStorage) {}
+
+  /**
+   * Replace each payload above the configured size threshold with a reference payload.
+   * Payloads below the threshold (or that the selector keeps inline) pass through
+   * unchanged. Order is preserved.
+   */
+  async store(payloads: Payload[], options: ExternalStoreOptions = {}): Promise<Payload[]> {
+    if (payloads.length === 0) return payloads;
+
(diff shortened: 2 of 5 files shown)


packages/test/src/test-extstore-runner.ts
```diff
@@ -0,0 +1,235 @@
...
+  const strangerDriver = makeFakeDriver({ name: 'a' }); // same name, different identity
+  const externalStorage = new ExternalStorage({
+    drivers: [registeredDriver],
+    payloadSizeThreshold: 0,
+    driverSelector: () => strangerDriver,
+  });
+
+  await t.throwsAsync(() => runExternalStore({ externalStorage, payloads: [makePayload(1)] }), {
+    instanceOf: ExternalStorageSelectorInvalidDriverError,
+    message: /TMPRL1109/,
```
cconstable: Forgot to update these tests. Updating now.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,290 @@
...
+  abortSignal?: AbortSignal;
+}
+
+/**
+ * Replace each payload above the configured size threshold with a reference payload.
+ *
+ * @internal
+ * @experimental
+ */
+export async function runExternalStore({
```
jmaeagle99: Could we create a class that has store and retrieve methods on it? I'm thinking ahead to where we want to add more capabilities, such as caching, and not wanting to pass more things via parameters. So that thing that are shared across all invocations (ExternalStorage object, hypothetical cache object/configuration) are provided at construction and the remaining variable pieces (payloads, target, abort signal) are parameters to store and retrieve functions.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,290 @@
...
+export interface ExternalStoreOptions {
+  /** When `undefined` the runner is a no-op (payloads pass through unchanged). */
+  externalStorage?: ExternalStorage;
+
+  /** Identity forwarded to the selector and each driver via `StorageDriverStoreContext.target`. */
+  target?: StorageDriverTarget;
+
+  payloads: Payload[];
+
+  /** Composed via `AbortSignal.any` with an internal controller so a failing driver call aborts siblings. */
```
jmaeagle99: Let's not document implementation details, just expected behavior and usage.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,290 @@
...
+ * @internal
+ * @experimental
+ */
+export async function runExternalRetrieve({
+  externalStorage,
+  payloads,
+  abortSignal,
+}: ExternalRetrieveOptions): Promise<Payload[]> {
+  if (payloads.length === 0) return payloads;
+  if (!payloads.some(isReferencePayload)) return payloads;
```
jmaeagle99: I think we can remove this check to avoid multiple iteration of the payloads arg. And then inline the external storage check within the for loop.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,290 @@
...
+  const batchSignal = abortSignal
+    ? AbortSignal.any([batchController.signal, abortSignal])
+    : batchController.signal;
+
+  const retrieveCtx: StorageDriverRetrieveContext = { abortSignal: batchSignal };
+
+  interface RetrieveItem {
+    index: number;
+    claim: StorageDriverClaim;
+    /** `0` for legacy references that pre-date the size_bytes field. */
```
jmaeagle99: I think this comment is inaccurate. All externally stored payloads had set this additional data on the payload protobuf message.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,290 @@
...
+        `Driver '${group.driver.name}' returned ${retrieved.length} payloads for ${group.items.length} claims`,
+        group.driver.name,
+        'retrieve',
+        group.items.length,
+        retrieved.length
+      );
+    }
+    for (const [j, retrievedPayload] of retrieved.entries()) {
+      const item = group.items[j]!;
+      if (item.expectedSize > 0) {
```
jmaeagle99: I don't think this check is necessary and in fact might prevent retrieval of otherwise valid externally stored payloads.

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -0,0 +1,290 @@
...
+        const actual = payloadProtoSize(retrievedPayload);
+        if (actual !== item.expectedSize) {
```
jmaeagle99: This is performing extra validation that the other SDKs are not doing. Let's remove for now and add back if we feel this is necessary, but to all SDKs.

packages/common/src/errors.ts
```diff
@@ -73,3 +73,124 @@ export class NamespaceNotFoundError extends Error {
  */
 @SymbolBasedInstanceOfError('CompleteAsyncError')
 export class CompleteAsyncError extends Error {}
+
+/// -----------------------------------------------------------------------
+/// External Storage Errors (draft)
```
jmaeagle99: I'm not convinced we need specialized error types for validations that do not cross into workflow/activity user code. These mostly are carrying specific error messages and error codes and fields that are not consumed by the rest of the system. I'm not opposed to the information they carry or the error codes necessarily, just the new error types themselves if they aren't specially treated.

For errors that drivers or the driver selector may raise that we want to specifically know about, those make sense to have dedicated error types.

packages/common/src/errors.ts
```diff
@@ -73,3 +73,124 @@ export class NamespaceNotFoundError extends Error {
...
+ * but the worker/client has no `ExternalStorage` configured to resolve it.
+ *
+ * Code: `TMPRL-SDK-00001`
+ *
+ * @experimental
+ */
+@SymbolBasedInstanceOfError('ExternalStorageNotConfiguredError')
+export class ExternalStorageNotConfiguredError extends Error {
+  constructor(message: string) {
+    super(`[TMPRL-SDK-00001] ${message}`);
```
jmaeagle99: I think this is https://github.com/temporalio/rules/blob/main/rules/TMPRL1105.md

cconstable: According to the rules page this should just emit a log... it doesn't look like an error. Regardless, I'll add the log for it when we actually integrate that piece. I preemptively added it here thinking it would be used in this PR.

packages/common/src/errors.ts
```diff
@@ -73,3 +73,25 @@ export class NamespaceNotFoundError extends Error {
...
+/// -----------------------------------------------------------------------
+
+/**
+ * Thrown when a driver's `store()` or `retrieve()` raises. Wraps the
+ * underlying error in `cause` (network, auth, rate-limit, S3 5xx, etc).
+ *
+ * @experimental
+ */
+@SymbolBasedInstanceOfError('ExternalStorageDriverOperationFailedError')
+export class ExternalStorageDriverOperationFailedError extends Error {
```
jmaeagle99: I would remove this for now and then apply it across the board to all extstore implementations if we feel the need for it.

packages/common/src/errors.ts
```diff
@@ -73,3 +73,25 @@ export class NamespaceNotFoundError extends Error {
...
+
+/// -----------------------------------------------------------------------
+/// External Storage Errors
+/// -----------------------------------------------------------------------
+
+/**
+ * Thrown when a driver's `store()` or `retrieve()` raises. Wraps the
+ * underlying error in `cause` (network, auth, rate-limit, S3 5xx, etc).
+ *
+ * @experimental
```
jmaeagle99: `@internal` if we're keeping this. It isn't user-code catchable.

packages/test/src/test-extstore-runner.ts
```diff
@@ -0,0 +1,203 @@
...
+  );
+  const storedPayloads = await writerRunner.store([makePayload(1)]);
+
+  const readerDriver = makeFakeDriver({ name: 'different-name' });
+  const readerRunner = new ExternalStorageRunner(new ExternalStorage({ drivers: [readerDriver] }));
+
+  await t.throwsAsync(() => readerRunner.retrieve(storedPayloads), {
+    instanceOf: ValueError,
+  });
+});
```
jmaeagle99, line 200: Should have more retrieve tests around checking arity, mixed reference and pass through payloads in a single array, sibling driver abort causes all to abort, multiple references to the same driver are batched together, etc.

jmaeagle99 (approved):

packages/test/src/test-extstore-runner.ts
cconstable: Added.

PR #2099, opened 2026-06-03
feat(langsmith): add @temporalio/langsmith package

xumaple:


contrib/langsmith/src/run-tree.ts
```diff
@@ -0,0 +1,326 @@
...
+  // uuid4. Propagated to every child so the whole subtree draws from one source.
+  protected readonly random?: () => number;
+
+  constructor(config: RunTreeConfig, random?: () => number) {
+    super(ReplaySafeRunTree.fill(config, random));
+    this.random = random;
+  }
+
+  private static fill(config: RunTreeConfig, random?: () => number): RunTreeConfig {
+    const id = config.id ?? (random ? uuid4FromRandom(random) : uuid4());
```
DABH: So a small concern about the fallback to `uuid4()` here - on the recorded interceptor paths, if generating run IDs with `uuid4()`, this pulls from the same per-workflow random generator as the user's `Math.random()` / `uuid4()`. So just installing the plugin (or toggling `addTemporalRuns`) changes the random numbers a user's workflow sees, which I think could cause a `NondeterminismError` on replay.  `prngFromInputId` in `prng.ts` mints IDs from a separate, self-seeded PRNG so they never touch the main RNG.  Consider minting all plugin run IDs from a stream that doesn't advance the workflow's main RNG - either `prngFromInputId` or perhaps `WorkflowRandomStream` / `getRandomStream` from `@temporalio/workflow`?

DABH:
Should we try to organize the file hierarchy similar to other plugins for consistency's sake?  For instance, based off how we organize things for the OpenAI Agents SDK, we'd want to move things around in this plugin to be:
```
src/
  index.ts                  ← unchanged (Node entry; re-export from ./worker/plugin)

  client/
    client-interceptor.ts   ← move client-interceptor.ts

  worker/
    plugin.ts               ← move plugin.ts (incl. the webpack-alias helpers)
    activity-interceptor.ts ← move activity-interceptor.ts (activity + Nexus inbound)
    sinks.ts                ← SPLIT: createLangSmithSinks + toCreate/UpdateParams

  workflow/
    interceptors.ts         ← move workflow-interceptors.ts (the bundler-loaded module)
    run-tree.ts             ← SPLIT: ReplaySafeRunTree, _RootReplaySafeRunTreeFactory,
                              isReplaying, serializeRun, NOOP_CLIENT
    prng.ts                 ← move prng.ts unchanged (mirrors canonical workflow/prng.ts)

  common/
    sink-contract.ts        ← SPLIT: LANGSMITH_SINK_NAME + SerializedRun + LangSmithSinks
                              + EmitterConfig   ← the load-bearing move
    run-names.ts            ← SPLIT: all *RunName builders + RUN_TYPE (pure strings)
    run-helpers.ts          ← SPLIT: describeError, asOutputs, runHeaders, emitMarkerRun,
                              runTreeFromContext, buildRunTree (sandbox-safe, used both sides)
    propagation.ts          ← move propagation.ts unchanged
```
Thoughts?

contrib/langsmith/src/run-tree.ts
xumaple: this is a very good callout to get things off of the main RNG. but prng.ts is getting ripped out withe completion of PR #2141 . on it.

xumaple: done

xumaple:
Responding to David's comment about organizational strategy, I think with 8 non-test files here it's not necessary to split things out in the same way that Openai (39 non-test files) did. It would actually cause some files to have to be split and it's even more cumbersome. There are plenty of contrib packages in this repo that are smaller and don't split, so we're not breaking convention here staying flat.

DABH (approved):

chris-olszewski (approved):
Approving `packages/common` changes as well as general repo structure.

PR #2106, opened 2026-06-10
Add SAA support to TemporalOperationHandler

VegetarianOrc:
### What was changed

Added support for Activity-backed Nexus operations in the TemporalOperationHandler.

- TemporalNexusClient.startActivity(): starts a standalone activity as an async Nexus operation, propagating callbacks, request ID, and inbound/outbound links automatically
- NexusTypedActivityClient: strongly-typed variant (via .typedActivity<T>()) for type-safe activity name, args, and result
- New ActivityOperationToken (token type 2) with aid/rid fields to identify activity-backed operations
- cancelActivity option on TemporalOperationHandlerOptions with a default implementation that cancels the underlying activity
- Extracted shared startWithNexusOptions helper to deduplicate link conversion, callback attachment, and back-link propagation between workflow and activity start paths

### Why?

Temporal Nexus operations should be able to start and cancel activities as well as workflows, using a consistent operation token and backing-operation model across both paths.

### Checklist

1. How was this tested:
  - Expanded test-nexus-temporal-operation.ts with integration tests covering: activity start, idempotent re-start (conflict handling), cancellation, and bad-options error surfacing as BAD_REQUEST rather than internal retries
  - Token encode/decode round-trip tests in test-nexus-token-helpers.ts


Quinn-With-Two-Ns (approved):

packages/nexus/src/__tests__/test-nexus-token-helpers.ts
```diff
@@ -34,16 +49,34 @@ test('decode Operation token errors', (t) => {
...
+  const missingRIDToken = base64URLEncodeNoPadding('{"t":2,"ns":"ns","aid":"a"}');
+  t.throws(() => assertActivityOperationToken(loadOperationToken(missingRIDToken)), {
+    message: /invalid activity token: missing activity run ID \(rid\)/,
+  });
+});
+
+test('loadWorkflowRunOperationToken rejects activity token', (t) => {
+  const activityToken = generateActivityOperationToken('ns', 'a', 'r');
+  t.throws(() => loadWorkflowRunOperationToken(activityToken), {
+    message: /invalid workflow token type: 2, expected: 1/,
```
Evanthx, line 148: Stupid question ... is this right? Invalid token type 2, expected 1? If so then I might ask for a quick slack message explaining what this means! :-)

packages/nexus/src/token.ts
```diff
@@ -129,6 +171,21 @@ export function assertWorkflowRunOperationToken(token: OperationToken): asserts
...
 
+/**
+ * Assert that an OperationToken identifies an activity.
+ */
+export function assertActivityOperationToken(token: OperationToken): asserts token is ActivityOperationToken {
+  if (token.t !== OperationTokenType.ACTIVITY) {
+    throw new TypeError(`invalid activity token type: ${token.t}, expected: ${OperationTokenType.ACTIVITY}`);
+  }
+  if (!token.aid || typeof token.aid !== 'string') {
+    throw new TypeError('invalid activity token: missing activity ID (aid)');
```
Evanthx: Nit: For this and the next error string - might want to change the message to "invalid or missing token". The current message is a little contradictory.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -24,6 +35,59 @@ import {
...
+  token: string,
+  handler: (opts: InternalNexusStartOptions[typeof InternalNexusStartOptionsSymbol]) => Promise<T>
+) {
+  const links = Array<temporal.api.common.v1.ILink>();
+  if (ctx.inboundLinks?.length > 0) {
+    for (const l of ctx.inboundLinks) {
+      try {
+        links.push(convertNexusLinkToTemporalLink(l));
+      } catch (error) {
+        log.warn('failed to convert Nexus link to Workflow event link', { error });
```
Evanthx: This just swallows the error and continues - is that as desired?

packages/nexus/src/workflow-helpers.ts
```diff
@@ -24,6 +35,59 @@ import {
...
+      },
+    ];
+  }
+
+  const result = await handler(internalOptions);
+  if (internalOptions.backLink != null) {
+    try {
+      ctx.outboundLinks.push(convertTemporalLinkToNexusLink(internalOptions.backLink));
+    } catch (error) {
+      log.warn('failed to convert temporal link to Nexus link', { error });
```
Evanthx: Same, this just swallows the error and continues. Is that as desired?

packages/nexus/src/workflow-helpers.ts
```diff
@@ -174,6 +197,75 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+    };
+
+    try {
+      return await client.activity.start<R>(activity, startOptions);
+    } catch (err) {
+      // ActivityClient.validateActivityOptions throws TypeError if there are bad options specified.
+      // e.g. missing startToCloseTimeout or scheduleToCloseTimeout
+      // If uncaught, causes Nexus operations to retry until timeout or circuit breaker trips and
+      // reason is not easily discoverable.
+      if (err instanceof TypeError) {
```
Evanthx: Nit and just out of curiosity - should this catch all errors and wrap like this instead of just the one type?

packages/nexus/src/token.ts
```diff
@@ -64,6 +86,26 @@ export function generateWorkflowRunOperationToken(namespace: string, workflowId:
...
+    t: OperationTokenType.ACTIVITY,
+    ns: namespace,
+    aid: activityId,
+    rid: runId,
+  };
+  return encodeOperationToken(token);
+}
+
+/**
+ * Encode an OPerationToken as a string.
```
Evanthx: Nit: has OPerationToken, lowercase the P

packages/nexus/src/workflow-helpers.ts
```diff
@@ -174,6 +197,75 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+  activityOptions: ActivityOptions
+): Promise<ActivityHandle<R>> {
+  const { client, taskQueue } = getHandlerContext();
+
+  // Activity tokens included in nexus callback headers cannot have
+  // run ID as it's unknown until the activity has been started so we
+  // manually construct the token.
+  const token = encodeOperationToken({
+    t: OperationTokenType.ACTIVITY,
+    ns: client.options.namespace,
```
Evanthx: This gets the namespace from client.options.namespace. Line 386 gets it from the context though:
`const { namespace } = getHandlerContext();`
Should be fine but having it in two places doesn't seem great - this goes over http, right? So having duplicated tokens doesn't seem great at all. 

But this also doesn't introduce the duplication so ... it's just a nit.

Evanthx (changes requested):

packages/nexus/src/__tests__/test-nexus-token-helpers.ts
VegetarianOrc: Maybe not clear, but it is correct! Workflow tokens are expected to have the enum value `1`. The test generates an activity token (enum value `2`) and asserts that `loadWorkflowRunOperationToken` rejects it.

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: Yup, this preserves existing behavior.

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: Yup, this preserves existing behavior.

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: No, in general errors should bubble up and the Nexus worker will wrap them in a HandlerError. That wrapping assumes the error is retryable if it's an unknown type. In this specific case, that results in the error being retried a bunch and does not log or return the final error until several retries occur. This just makes it more obvious that you've accidentally forgotten to include one or the other instead of it seeming like it's hanging while the handler is retried.

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: Those wind up being the same value, but you're right to call it out. I've updated the various usages to use the handler context version.

The token does not go over HTTP from the client to the server. It is added to the callback headers which is included in the callback proto and sent in the gRPC request. The duplication is purposeful to help with the race condition where the thing backing the nexus operation (an activity in this case) completes before the StartOperation handler completes and returns the real token. In that case the history of the operation is adjusted to include the NexusOperationStarted event and uses the token from the callback headers. If the StartOperation handler completes first as it does most of the time, the token from the headers is unused and the token returned from the handler is used.

Evanthx (approved):
LGTM

packages/nexus/src/workflow-helpers.ts
```diff
@@ -374,6 +384,86 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+  }
+
+  const { taskQueue: userSpecifiedTaskQueue, ...rest } = activityOptions;
+  const startOptions: InternalActivityStartOptions = {
+    ...rest,
+    taskQueue: userSpecifiedTaskQueue || taskQueue,
+    [InternalActivityStartOptionsSymbol]: internalOptions,
+  };
+
+  const handle = await client.activity.start<R>(activity, startOptions);
```
JoshuaFrenchwood, line 461: validateOptions in ActivityOptions.startHandler will check if timeout options are set. If neither are then this will throw a TypeError. Will the Worker report these TypeErrors are retryable?

VegetarianOrc: Yes that's correct. The worker will translate any errors that bubble up appropriately. In this case the TypeError does become a retryable error. That is consistent behavior with workflows and other SDKs at the moment, but we can consider changing it.

JoshuaFrenchwood (approved):

packages/nexus/src/workflow-helpers.ts
```diff
@@ -374,6 +384,87 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
 }
 
+/**
+ * Options for starting an untyped activity using {@link TemporalNexusClient.startActivity}, this type is identical to the
+ * client's `ActivityOptions` with the exception that `taskQueue` is optional and defaults
+ * to the current worker's task queue.
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export type ActivityOptions = Replace<ClientActivityOptions, { taskQueue?: string }>;
```
THardy98, line 394: Can we name `NexusActivityOptions` or `NexusClientActivityOptions` given that this is intended only for Nexus?

Just `ActivityOptions` and `ClientActivityOptions` seem ambiguous if you don't read the docstrings

packages/nexus/src/workflow-helpers.ts
```diff
@@ -374,6 +384,87 @@ export class WorkflowRunOperationHandler<I, O> implements nexus.OperationHandler
...
+
+/**
+ * Options for starting a typed activity using {@link NexusTypedActivityClient.startActivity}, this type is identical to the
+ * client's `ActivityOptionsFor` with the exception that `taskQueue` is optional and defaults
+ * to the current worker's task queue.
+ *
+ * @experimental Nexus support in Temporal SDK is experimental.
+ */
+export type ActivityOptionsFor<T, N extends ActivityName<T>> = Replace<
```
THardy98, line 403: ditto

packages/nexus/src/workflow-helpers.ts
VegetarianOrc: I'm totally fine with the rename, but want to call out the that the workflow variant has this pre-existing pattern e.g. `WorkflowStartOptions`.

Happy to rename these though if you think it would make it clearer even with that inconsistency. LMK!

THardy98: i'm fine with not renaming then. Maybe something we reconsider in the future

THardy98 (approved):

chris-olszewski (approved):

PR #2107, opened 2026-06-10
fix(activity): correctly populate activity info task queue and namespace

chris-olszewski:
## What was changed
- Use `activityNamespace` over workflow namespace for populating namespace fields on `activityInfo`
- Switch to use an object instead of loose params to prevent incorrect calling of `extractActivityInfo` in the future

## Why?
In #1996 we accidentally flipped task queue and activity namespace in the call to `extractActivityInfo`

## Checklist

1. Closes N/A

2. How was this tested:
Added test to verify task queue and activity namespace work as expected.

3. Any docs updates needed?

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -1059,13 +1059,13 @@ export class Worker {
                         `Got start event for an already running activity: ${base64TaskToken}`
                       );
                     }
-                    info = await extractActivityInfo(
+                    info = await extractActivityInfo({
                       task,
-                      loadedDataConverter,
-                      this.options.namespace,
-                      this.options.taskQueue,
-                      context
-                    );
+                      dataConverter: loadedDataConverter,
+                      taskQueue: this.options.taskQueue,
+                      activityNamespace: this.options.namespace,
+                      context,
+                    });
 
                     const { activityType } = info;
                     // Use the corresponding activity if it exists, otherwise, fallback to default activity function (if exists)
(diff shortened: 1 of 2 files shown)


packages/worker/src/worker.ts
```diff
@@ -2247,7 +2253,7 @@ async function extractActivityInfo(
     ),
     priority: decodePriority(start.priority),
     retryPolicy: decompileRetryPolicy(start.retryPolicy),
-    namespace: start.workflowNamespace,
+    namespace: activityNamespace,
```
chris-olszewski: Unsure if we should fall back to `start.workflowNamespace` if `activityNamespace` is empty?

THardy98 (approved):
Thanks for the fix

packages/worker/src/worker.ts
THardy98: yeah that seems prudent, `activityNamespace` is only populated for SAA if i remember correctly

chris-olszewski: It appears that we want to default to `start.workflowNamespace` and only use the client namespace if the workflow namespace isn't present: https://github.com/temporalio/sdk-python/blob/d8675f5896258ce6e2b13059c4fa53562c614ab6/temporalio/worker/_activity.py#L643

Stereobit:
Would be great if we can get a release for this, I woke up to a bunch of issues this morning which turned out to be caused by it 🥲

THardy98:
> Would be great if we can get a release for this, I woke up to a bunch of issues this morning which turned out to be caused by it 🥲

@Stereobit  Patch release is planned today, sorry for the inconvenience!

THardy98:
@Stereobit [1.18.1 released](https://github.com/temporalio/sdk-typescript/releases/tag/v1.18.1) with the fix. Published on NPM. Thanks for your patience!

PR #2109, opened 2026-06-10
fix(openai-agents): correct misleading legacy-query comment in resolveQueryKey

xumaple:
Corrects a misleading comment in `resolveQueryKey` in the `@temporalio/openai-agents` trace interceptor. No logic change — comment only.

The fallback seed for legacy queries was annotated with a caveat claiming that two Queries answered in one Workflow Task would collide on the spanId. That scenario cannot occur on this branch:

- `queryId === 'legacy_query'` is produced only on Temporal Core's legacy single-Query Workflow-Task path. In `sdk-core`, the deprecated singular `PollWorkflowTaskQueueResponse.query` field is mapped to `legacy_query: Option<WorkflowQuery>` (a single optional query), and only that query is tagged with `LEGACY_QUERY_ID = "legacy_query"`.
- Core treats normal (batched) and legacy queries as mutually exclusive — it raises a fatal error if a poll response carries both, and `is_legacy_query()` matches only an activation with exactly one job that is a `legacy_query`.
- Modern batched queries each carry their own unique `queryId` and never reach this branch.

So legacy queries arrive one-per-Task, in separate Tasks answered at distinct wall-clock moments; seeding from `queryName` + the Task's wall-clock gives each its own seed. The fallback is sufficient, not best-effort. The comment is reworded to say so and the false collision caveat is removed.


DABH (approved):
LGTM

chris-olszewski (approved):

PR #2111, opened 2026-06-11
Add `disableTraceContextInMetricTags` to otel plugin options

THardy98:
## What was changed
Adds `disableTraceContextInMetricTags` to `OpenTelemetryPluginOptions` to allow for disabling otel tags (i.e. span/trace ids) from being added to custom metrics, thereby controlling the cardinality of their metrics.

2. How was this tested:
Added unit tests.

3. Any docs updates needed?
Probably not


contrib/interceptors-opentelemetry/src/plugin.ts
```diff
@@ -35,8 +41,15 @@ export interface OpenTelemetryPluginOptions extends InterceptorOptions {
  */
 export class OpenTelemetryPlugin extends SimplePlugin {
   constructor(readonly otelOptions: OpenTelemetryPluginOptions) {
-    const workflowInterceptorsPath = require.resolve('./workflow-interceptors');
+    const workflowInterceptorsPath = require.resolve(
+      otelOptions.disableTraceContextInMetricTags
+        ? './workflow-interceptors-no-metric-trace-context'
```
chris-olszewski: I don't love this, but not sure how we could get around it.

contrib/interceptors-opentelemetry/src/__tests__/test-otel.ts
```diff
@@ -947,6 +948,82 @@ if (RUN_INTEGRATION_TESTS) {
...
+    });
+
+  t.deepEqual(activityTags, { existing: 'activity' });
+  t.deepEqual(nexusTags, { existing: 'nexus' });
+  t.false('trace_id' in activityTags);
+  t.false('span_id' in activityTags);
+  t.false('trace_flags' in activityTags);
+  t.false('trace_id' in nexusTags);
+  t.false('span_id' in nexusTags);
+  t.false('trace_flags' in nexusTags);
```
chris-olszewski: These assertions are already covered by the deepEqual assertions

contrib/interceptors-opentelemetry/src/__tests__/test-otel.ts
```diff
@@ -947,6 +948,82 @@ if (RUN_INTEGRATION_TESTS) {
...
+
+  t.deepEqual(logAttributes, { ...traceContext, existing: 'attribute' });
+
+  t.deepEqual(workflowInterceptorModuleNames, ['workflow-interceptors-no-metric-trace-context']);
```
chris-olszewski: Just because we are loading a fully different module, I'd like test coverage that we actually load and use that module via the plugin.

chris-olszewski (changes requested):
We should have an integration test for this new plugin functionality. Current tests are a little too different from how the plugin would be used by an end user.

contrib/interceptors-opentelemetry/src/worker/index.ts
mjameswh: I'm not sure there's ever a use case for attaching these three attributes to metric tags. That's an overlook from my part when I added custom metrics support. I wouldn't add an option for that.

THardy98:
opening a fresh PR, direction has changed

Issue #2108, opened 2026-06-10
[Bug] Otel interceptor adds trace_id/span_id to custom metrics


andrewspinks:
### What are you really trying to do?

Report custom metrics when tracing is enabled

### Describe the bug

When using the @temporalio/interceptors-opentelemetry plugin and have tracing enabled, trace_id/span_id are added to custom metrics which leads to unbounded cardinality

### Minimal Reproduction

Add a custom metric to the interceptor sample:

https://github.com/andrewspinks/samples-typescript/commit/d7dfb284d6102366a3332502416f38c650825537

Example metrics exported:

<img width="1628" height="309" alt="Image" src="https://github.com/user-attachments/assets/a5bdbcf1-054b-40f2-963d-ecf49550c70c" />

### Environment/Versions

sdk-typescript: 1.18.0

THardy98:
Fix released in [1.18.1](https://github.com/temporalio/sdk-typescript/releases/tag/v1.18.1) and published to NPM

PR #2113, opened 2026-06-11
Remove OTel tags from metrics

THardy98:
## What was changed
DWISOTT - remove otel tags from metrics (i.e. spanId, traceId)

## Why?
Blows up metrics cardinality with no benefit.

2. How was this tested:
Added integration test

3. Any docs updates needed?
Probably not

### DIFF
--- a/contrib/interceptors-opentelemetry/src/__tests__/test-otel-metric-tags.ts
+++ b/contrib/interceptors-opentelemetry/src/__tests__/test-otel-metric-tags.ts
@@ -1,0 +1,116 @@
+import { randomUUID } from 'crypto';
+import * as opentelemetry from '@opentelemetry/sdk-node';
+import { InMemorySpanExporter, SimpleSpanProcessor } from '@opentelemetry/sdk-trace-base';
+import { SEMRESATTRS_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
+import test from 'ava';
+import { MetricsBuffer, Runtime, Worker } from '@temporalio/worker';
+import {
+  RUN_INTEGRATION_TESTS,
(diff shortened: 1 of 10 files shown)


packages/proto/package.json
```diff
@@ -10,7 +10,7 @@
     "lib/"
   ],
   "scripts": {
-    "build": "pnpm run \"/^build:.*/\"",
+    "build": "pnpm run build:protos && pnpm run build:ts",
```
THardy98, line 13: small change to enforce that protos are built before the actual package build, can run into race conditions where the build fails because protos haven't finished generating

packages/test/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "1.18.0",
   "description": "Temporal.io SDK Tests",
   "scripts": {
-    "build": "pnpm run \"/^build:.*/\"",
+    "build": "pnpm run build:protos && pnpm run build:ts",
```
THardy98, line 7: ditto (small change to enforce that protos are built before the actual package build, can run into race conditions where the build fails because protos haven't finished generating)

chris-olszewski (approved):

mjameswh (approved):

contrib/interceptors-opentelemetry/src/__tests__/test-otel-metric-tags.ts
THardy98, line 1: separated into it's own test file because it modifies `Runtime` which is shared amongst other tests in the process (modifying `metrics` was causing other tests to fail)

`ava` runs separate test files in their own process, so this file has isolation

THardy98:
Resolves: https://github.com/temporalio/sdk-typescript/issues/2108

PR #2114, opened 2026-06-11
:boom: Adding links to Nexus signals

Evanthx:
This adds links to Nexus signals - the receiving workflow has a link to the workflow sending the signal. and the sending workflow has a link to the receiving workflow.


packages/client/src/internal.ts
```diff
@@ -35,6 +36,13 @@ export interface InternalWorkflowStartOptions extends WorkflowOptions {
      */
     backLink?: temporal.api.common.v1.ILink;
 
+    /**
+     * Backlink copied by the client from the SignalWithStartWorkflowExecutionResponse.
+     * Only populated by servers that support CHASM signal backlinks (1.31 and up); left unset
+     * otherwise.
+     */
+    signalBackLink?: temporal.api.common.v1.ILink;
```
VegetarianOrc: Is there a scenario where backLink and signalBackLink are both populated? If not we should remove this and use the one field.

packages/client/src/workflow-client.ts
```diff
@@ -502,7 +506,7 @@ export class WithStartWorkflowOperation<T extends Workflow> {
  * Typically this client should not be instantiated directly, instead create the high level {@link Client} and use
  * {@link Client.workflow} to interact with Workflows.
  */
-export class WorkflowClient extends BaseClient {
+export class WorkflowClient extends BaseClient implements InternalWorkflowClientWithNexusLinks {
```
VegetarianOrc: We shouldn't implement this interface here as it leaks the method onto `client.workflow`. I don't have a great suggestion for exposing `signal` to do the internal options dance though. One option might be to hide the method behind an internal symbol. Let's sync with @chris-olszewski and come up with what feels like the best path here.

packages/client/src/interceptors.ts
```diff
@@ -89,6 +90,14 @@ export interface WorkflowSignalInput {
...
   readonly workflowExecution: WorkflowExecution;
   readonly headers: Headers;
+  /**
+   * SDK-internal options used by the Temporal Nexus helpers to forward inbound Nexus links onto
+   * the signal request and to capture the backlink returned on the response.
+   *
+   * @internal
+   * @hidden
+   */
+  readonly [InternalWorkflowSignalOptionsSymbol]?: InternalWorkflowSignalOptions[typeof InternalWorkflowSignalOptionsSymbol];
```
VegetarianOrc: We shouldn't leak these internal details here. instead we should define an internal interface in `internal.ts` that extends `WorkflowSignalInput` and adds this symbol.

packages/client/src/workflow-client.ts
```diff
@@ -1211,6 +1215,9 @@ export class WorkflowClient extends BaseClient {
   protected async _signalWorkflowHandler(input: WorkflowSignalInput): Promise<void> {
     const dataConverter = this.dataConverter;
     const context = this.workflowSerializationContext(input.workflowExecution.workflowId!);
+    const internalOptions = input[InternalWorkflowSignalOptionsSymbol];
```
VegetarianOrc:
```suggestion
    const internalOptions = (input as InternalWorkflowSignalInput)[InternalWorkflowSignalOptionsSymbol];
```

Going along with my above comment about not leaking this symbol, this should be cast to an internal type here and the internal options retrieved from that. In my suggested change I made up a name, but feel free to use something more appropriate.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -130,17 +128,114 @@ export async function startWorkflow<T extends Workflow>(
...
+  return links;
+}
+
+/**
+ * Pushes a backlink returned by an outbound Workflow RPC onto the operation's outbound links so the
+ * Nexus task handler attaches it to the StartOperationResponse, linking the caller Workflow's
+ * NexusOperation history event back to the callee Workflow's event. No-op when the server did not
+ * return a backlink (older servers, or CHASM signal backlinks disabled).
+ */
+function pushBacklink(ctx: nexus.StartOperationContext, backLink: temporal.api.common.v1.ILink | undefined): void {
```
VegetarianOrc: Nit: the `backLink` arg should probably not accept `undefined`. Call sites look like they have a side effect, but in reality may not do anything.

packages/test/src/test-nexus-signal-linking.ts
```diff
@@ -0,0 +1,322 @@
...
+  workflow.setHandler(pingSignal, (msg: string) => {
+    received.push(msg);
+  });
+  await workflow.condition(() => received.length >= expectedSignals);
+  return received.join(',');
+}
+
+export const pingSignal = workflow.defineSignal<[string]>('ping');
+
+export async function caller(endpoint: string, input: string): Promise<string> {
```
VegetarianOrc: Let's remove the concept of the mode switch here and have separate caller workflows and operation handlers as needed for each test. Ideally we can colocate these right above where they're used to make it easier to see the full picture of what runs as part of the test.

packages/test/src/test-nexus-signal-linking.ts
```diff
@@ -0,0 +1,322 @@
...
+  const client = workflow.createNexusServiceClient({ endpoint, service: signalingService });
+  const [mode, rest] = splitFirst(input, ':');
```
VegetarianOrc: My above comment stands about separating this out, but just want to call out that we should avoid encoding params as strings and instead use objects.

packages/test/src/test-nexus-signal-linking.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+  });
+});
+
+////////////////////////////////////////////////////////////////////////////////////////////////////
+// Test utilities
+
+async function fetchHistory(client: any, workflowId: string): Promise<temporal.api.history.v1.IHistory> {
+  return await client.workflow.getHandle(workflowId).fetchHistory();
+}
```
VegetarianOrc: Let's remove this helper and just get a handle where necessary to get the history. There's a few places where we already get or have a handle that then call this helper.

packages/test/src/test-nexus-signal-linking.ts
```diff
@@ -0,0 +1,322 @@
...
+      workflowId: callerWorkflowId,
+      args: [endpointName, `${MODE_MULTI_SIGNAL_WITH_START}:${calleeIds.join(',')}`],
+    });
+    t.is(await callerHandle.result(), `ok:multi:${calleeIds.join(',')}`);
+
+    for (const calleeId of calleeIds) {
+      t.is(await client.workflow.getHandle(calleeId).result(), 'multi-signal');
+    }
+
+    const callerHistory = await fetchHistory(client, callerWorkflowId);
```
VegetarianOrc:
```suggestion
    const callerHistory = await callerHandler.fetchHistory();
```

packages/test/src/test-nexus-signal-linking.ts
```diff
@@ -0,0 +1,322 @@
...
+      workflowId: callerWorkflowId,
+      args: [endpointName, `${MODE_ASYNC_SIGNAL_WITH_START}:${calleeWorkflowId}`],
+    });
+    t.is(await callerHandle.result(), 'async-started');
+
+    const calleeResult = await client.workflow.getHandle(calleeWorkflowId).result();
+    t.is(calleeResult, 'async-signal');
+
+    const callerHistory = await fetchHistory(client, callerHandle.workflowId);
+    const calleeHistory = await fetchHistory(client, calleeWorkflowId);
```
VegetarianOrc:
```suggestion
    const calleeHandle = client.workflow.getHandle(calleeWorkflowId);
    const calleeResult = await calleeHandle.result();
    t.is(calleeResult, 'async-signal');

    const callerHistory = await callerHandle.fetchHistory();
    const calleeHistory = await calleeHandle.fetchHistory();
```

VegetarianOrc (changes requested):
Unsure if the CI failures are directly related to these changes, but please take a look and ensure they're passing.

packages/client/src/internal.ts
Evanthx: Good catch, consolidated them

packages/client/src/workflow-client.ts
Evanthx: Changed as per our conversation!

packages/client/src/workflow-client.ts
Evanthx: Hopefully fixed as per our conversation!

packages/client/src/interceptors.ts
Evanthx: Fixed

packages/nexus/src/workflow-helpers.ts
Evanthx: Now checked

packages/test/src/test-nexus-signal-linking.ts
Evanthx: Done

packages/test/src/test-nexus-signal-linking.ts
Evanthx: Done

packages/test/src/test-nexus-signal-linking.ts
Evanthx: Done

chris-olszewski (approved):

VegetarianOrc (approved):

PR #2120, opened 2026-06-15
feat(google-adk-agents): add @temporalio/google-adk-agents package

xumaple:
Adds `@temporalio/google-adk-agents`, a contrib plugin that runs Google Agent Development Kit (`@google/adk`) agents as durable Temporal Workflows. The agent graph runs inside the Workflow and replays deterministically; the non-deterministic boundaries — every model call and every MCP tool call — are routed out to Activities, so the whole run gets Temporal's retries, timeouts, heartbeating, and crash-safe replay. For users it's a one-line change: wrap the agent's model in `TemporalLlm` and register the plugin.


.github/CODEOWNERS
```diff
@@ -4,6 +4,7 @@
 # someone opens a pull request.
 * @temporalio/sdk
 
-/contrib/ai-sdk/           @temporalio/sdk @temporalio/ai-sdk
-/contrib/openai-agents/    @temporalio/sdk @temporalio/ai-sdk
-/contrib/workflow-streams/ @temporalio/sdk @temporalio/ai-sdk
+/contrib/ai-sdk/             @temporalio/sdk @temporalio/ai-sdk
+/contrib/google-adk-agents/ @temporalio/sdk @temporalio/ai-sdk
+/contrib/openai-agents/     @temporalio/sdk @temporalio/ai-sdk
+/contrib/workflow-streams/  @temporalio/sdk @temporalio/ai-sdk
```
brianstrauch:
```suggestion
/contrib/ai-sdk/            @temporalio/sdk @temporalio/ai-sdk
/contrib/google-adk-agents/ @temporalio/sdk @temporalio/ai-sdk
/contrib/openai-agents/     @temporalio/sdk @temporalio/ai-sdk
/contrib/workflow-streams/  @temporalio/sdk @temporalio/ai-sdk
```

README.md
```diff
@@ -115,6 +115,7 @@ This monorepo contains the following packages:
 | [`packages/nexus/`](packages/nexus/)                                         | [`@temporalio/nexus`](https://www.npmjs.com/package/@temporalio/nexus)                                           |
 | [`contrib/interceptors-opentelemetry/`](contrib/interceptors-opentelemetry/) | [`@temporalio/interceptors-opentelemetry`](https://www.npmjs.com/package/@temporalio/interceptors-opentelemetry) |
 | [`contrib/ai-sdk/`](contrib/ai-sdk/)                                         | [`@temporalio/ai-sdk`](https://www.npmjs.com/package/@temporalio/ai-sdk)                                         |
+| [`contrib/google-adk-agents/`](contrib/google-adk-agents/)                   | [`@temporalio/google-adk-agents`](https://www.npmjs.com/package/@temporalio/google-adk-agents)                   |
```
brianstrauch: We have 11 contrib directories but only 3 are listed here. Not sure if this will scale nicely.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+Run [Google Agent Development Kit](https://github.com/google/adk-js) (`@google/adk`)
+agents as durable [Temporal](https://temporal.io) Workflows.
+
+Your ADK agent graph — `LlmAgent`, `SequentialAgent`/`ParallelAgent`/`LoopAgent`,
+`FunctionTool`s, `MCPToolset`s, the `Runner` loop — runs **inside the Workflow**
+and replays deterministically. Only the non-deterministic I/O boundaries are
+routed out to Activities:
+
+- every **model call** (`generateContentAsync`) becomes a retriable, observable
```
brianstrauch: Never seen "retryable" spelled this way. Looked it up and apparently it doesn't matter.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+  Activity, and
+- every **MCP tool call** (list-tools / call-tool) becomes an Activity.
```
brianstrauch: Mention that tools default to in-workflow, and the user can opt-in to running them as activities?

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+`assert`/`url`/`util` resolve to the sandbox overrides; all others alias to
+`false`). The model/MCP code paths that would actually _use_ those builtins
+never run in the Workflow — they run worker-side in Activities — so stripping the
+scheme is safe. The sandbox's determinism guard is preserved for **your** code:
+if your own Workflow imports a `node:` builtin, the build still fails with the
+usual "disallowed modules" error (only `node_modules`-originated imports are
+exempted).
+
+You don't configure any of this — registering the plugin is enough. Any
+`webpackConfigHook` you already pass is preserved and runs first.
```
brianstrauch: This is a pretty heavy comment. To what degree does the user actually need to know this?

.github/CODEOWNERS
xumaple: Applied — aligned the owner column. Rebasing onto main also folded in the new `strands` entry, so the block now lists all five ai-sdk-owned contrib packages.

README.md
xumaple: Good call — filled it out. The table now lists all published contrib packages. The four private `external-storage-*` packages are omitted since they are not published to npm (main's table already excluded them).

contrib/google-adk-agents/README.md
xumaple: Switched to `retryable` for consistency.

contrib/google-adk-agents/README.md
xumaple: Added to the intro: regular `FunctionTool`s run in the Workflow, and `activityAsTool` is how you expose an existing Temporal Activity to the agent as a tool.

contrib/google-adk-agents/README.md
xumaple: Agreed — dropped the "Workflow bundling" subsection. It was implementation detail users do not need; registering the plugin is enough, and the mechanics live in the code.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+## Hello world
+
+Take an agent you already have and change **one line** — wrap its model in
+`TemporalModel` — then register the plugin.
+
+### `workflows.ts`
+
+```typescript
+import { InMemoryRunner, LlmAgent, isFinalResponse, stringifyContent } from '@google/adk';
+import { TemporalModel } from '@temporalio/google-adk-agents';
```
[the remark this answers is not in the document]
xumaple: Fixed — the workflow-side snippet now imports `TemporalModel` from the `/workflow` subpath.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+    model: new TemporalModel('gemini-2.5-flash'),
+    tools: [new TemporalMcpToolSet({ name: 'filesystem' })],
+  });
+  ```
+
+- **Existing Activities as tools.** Already have a Temporal Activity? Expose it
+  to the agent with `activityAsTool` instead of re-declaring it:
+
+  ```typescript
+  import { activityAsTool } from '@temporalio/google-adk-agents';
```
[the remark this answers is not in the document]
xumaple: Fixed — `activityAsTool` is now imported from the `/workflow` subpath.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+## What this plugin gives you
+
+- **Durable model calls.** Swap `model: 'gemini-2.5-flash'` for
+  `model: new TemporalModel('gemini-2.5-flash')` and each inference runs as a
+  Temporal Activity with a per-model `RetryPolicy`, `startToCloseTimeout`, and
+  auto-heartbeat for slow / thinking-mode calls. Upstream `429`/`5xx` and
+  `retry-after` headers are honored; non-retryable `4xx` fail fast.
+- **Durable MCP tools.** `new TemporalMcpToolSet({ name })` routes tool
```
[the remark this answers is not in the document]
xumaple: Fixed — corrected to `TemporalMCPToolset`.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+        serverParams: { command: 'npx', args: ['-y', '@modelcontextprotocol/server-filesystem', '/data'] },
+      }),
+    },
+  });
+
+  // workflow / agent
+  const agent = new LlmAgent({
+    name: 'fs',
+    model: new TemporalModel('gemini-2.5-flash'),
+    tools: [new TemporalMcpToolSet({ name: 'filesystem' })],
```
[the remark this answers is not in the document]
xumaple: Fixed — corrected to `TemporalMCPToolset`.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+  `Math.random()` / `Date.now()`, which the Temporal Workflow sandbox makes
+  deterministic — so the agent loop replays without a custom determinism hook.
+
+### Testing your workflows
+
+Import test doubles from the `./testing` entry point to unit-test agents
+without a live model or MCP server:
+
+```typescript
+import { fakeModelProvider, mockMcpToolset } from '@temporalio/google-adk-agents/testing';
```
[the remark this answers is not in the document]
xumaple: Fixed — corrected to `mockMCPToolset`.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+const plugin = new GoogleAdkPlugin({
+  modelProvider: fakeModelProvider(),
+  mcpToolsets: {
+    weather: mockMcpToolset([
```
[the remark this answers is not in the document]
xumaple: Fixed — corrected to `mockMCPToolset`.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,225 @@
...
+and replay — transparently augments `bundlerOptions.webpackConfigHook`
+to strip the `node:` scheme **before** webpack's scheme detection, after which
+the sandbox's normal builtin policy applies (the three polyfilled builtins
+`assert`/`url`/`util` resolve to the sandbox overrides; all others alias to
+`false`). The model/MCP code paths that would actually _use_ those builtins
+never run in the Workflow — they run worker-side in Activities — so stripping the
+scheme is safe. The sandbox's determinism guard is preserved for **your** code:
+if your own Workflow imports a `node:` builtin, the build still fails with the
+usual "disallowed modules" error (only `node_modules`-originated imports are
+exempted).
```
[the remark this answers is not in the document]
xumaple: Addressed — the "Workflow bundling" section was removed entirely, so the inaccurate determinism-guard description is gone.

contrib/google-adk-agents/src/activities.ts
```diff
@@ -0,0 +1,371 @@
...
+}
+
+function mcpActivitiesForName(
+  name: string,
+  factory: MCPToolsetFactory
+): Record<string, (args: never) => Promise<unknown>> {
+  const resolveToolset = (): BaseToolset => {
+    const produced = factory();
+    return isBaseToolset(produced) ? produced : new MCPToolset(produced);
+  };
```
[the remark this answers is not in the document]
xumaple: Kept as-is — precise per-key typing does not actually remove the casts here. `noUncheckedIndexedAccess` makes every dynamic-key access `T | undefined`, so the call sites still need a `!`/cast regardless of value-type precision; and TS will not infer template-literal index signatures from the computed-property object literals that build these activities, so the producer would need a double-cast. Net, the precise version relocates and multiplies casts rather than removing them, so the `(args: never)` signature is the cleaner idiom.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,205 @@
...
+    description: 'Look up an order by id.',
+    parameters: { type: Type.OBJECT, properties: { orderId: { type: Type.STRING } } },
+  });
+  ```
+
+- **Streaming (SSE).** Streaming requires `streamingTopic` on `TemporalModel`:
+  set it to publish incremental `LlmResponse` chunks via
+  `@temporalio/workflow-streams` while the Workflow still receives the full
+  transcript. Requesting streaming without a `streamingTopic` throws a
+  non-retryable `GoogleAdkStreamingTopicRequired` error.
```
brianstrauch: Could this have its own section with a code snippet?

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,205 @@
...
+- **Human-in-the-loop.** Because the agent loop runs in the Workflow body, a
+  `LongRunningFunctionTool` can `await` a Temporal Signal or Update carrying a
+  human's result — no special shim required.
+- **Deterministic replay.** ADK's event IDs and timestamps funnel through
+  `Math.random()` / `Date.now()`, which the Temporal Workflow sandbox makes
+  deterministic — so the agent loop replays without a custom determinism hook.
```
brianstrauch: What would a custom determinism hook entail? Since we don't have one, maybe we can just leave that part out so we don't confuse anyone.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,205 @@
...
+## Composing with other plugins
+
+Temporal applies `plugins: [...]` in order. Place observability and governance
+plugins **before** this one so model/tool Activities are wrapped by them:
+
+```typescript
+new Client({
+  plugins: [
+    new OpenTelemetryPlugin(), // 1. observability (outermost)
+    new GovernancePlugin(), // 2. governance
```
brianstrauch: Do we have an example plugin for this? If not, rename OpenTelemetryPlugin() to ObservabilityPlugin() so users aren't looking for this.

pnpm-lock.yaml
```diff
@@ -2857,6 +3812,11 @@ packages:
     resolution: {integrity: sha512-w+5mJ3GuFL+NjVtJlvydShqE1eN3h3PbI7/5LAsYJP/2qtuMXjfL2LpHSRqo4b4eSF5K/DH1JXKUAHSB2UW50g==}
     engines: {node: '>= 0.4'}
 
+  esbuild@0.21.5:
```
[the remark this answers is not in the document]
xumaple: /fp we do not invoke esbuild's Deno CLI module. This package's tests run via ava, not esbuild; the flagged esbuild@0.21.5 was pulled in by vitest, which this package no longer uses, and is no longer in the lockfile. The only remaining esbuild (0.25.12) is a pre-existing main dependency and is not reachable via the Deno-CLI / NPM_CONFIG_REGISTRY path.

pnpm-lock.yaml
```diff
@@ -5068,10 +6763,75 @@ packages:
...
+      sass-embedded:
+        optional: true
+      stylus:
+        optional: true
+      sugarss:
+        optional: true
+      terser:
+        optional: true
+
+  vitest@1.6.1:
```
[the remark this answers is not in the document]
xumaple: /fp we do not run the Vitest UI server (nor on Windows). This package migrated its tests to ava and no longer depends on vitest at all — vitest is no longer present in the lockfile.

pnpm-lock.yaml
```diff
@@ -5068,10 +6763,75 @@ packages:
     resolution: {integrity: sha512-ZZKSmDAEFOijERBLkmYfJ+vmk3w+7hOLYDNkRCuRuMJGEmqYNCNLyBBFwWKVMhfwaEF3WOd0Zlw86U/WC/+nYw==}
     engines: {'0': node >=0.6.0}
 
+  vite-node@1.6.1:
+    resolution: {integrity: sha512-YAXkfvGtuTzwWbDSACdJSg4A4DZiAqckWe90Zapc/sEX3XvHcw1NdurM/6od8J207tSDqNbSsgdCacBgvJKFuA==}
+    engines: {node: ^18.0.0 || >=20.0.0}
+    hasBin: true
+
+  vite@5.4.21:
```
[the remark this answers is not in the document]
xumaple: /fp we do not run or expose a Vite dev server. vite was only a transitive dependency of vitest, which this package no longer uses — vite is no longer present in the lockfile.

brianstrauch (approved):

contrib/google-adk-agents/README.md
xumaple: done

contrib/google-adk-agents/README.md
xumaple: removed

contrib/google-adk-agents/README.md
xumaple: removed this

contrib/google-adk-agents/package.json
```diff
@@ -0,0 +1,103 @@
...
+    "@temporalio/common": "workspace:*",
+    "@temporalio/plugin": "workspace:*",
+    "@temporalio/workflow": "workspace:*",
+    "@temporalio/workflow-streams": "workspace:*",
+    "@ungap/structured-clone": "^1.2.0",
+    "headers-polyfill": "^4.0.0",
+    "web-streams-polyfill": "^4.0.0"
+  },
+  "peerDependencies": {
+    "@google/adk": "^1.2.0",
```
DABH: Suggest bumping to ^1.4.0, MCPToolset.close() was broken before this and we use that function in here

contrib/google-adk-agents/package.json
```diff
@@ -0,0 +1,103 @@
+{
+  "name": "@temporalio/google-adk-agents",
+  "version": "0.1.0",
```
DABH: Is this right?  Seems every other contrib package is at 1.21.1?

contrib/google-adk-agents/src/activities.ts
```diff
@@ -0,0 +1,308 @@
...
+  /** Reconstructs a `BaseLlm` from a model name; defaults to the ADK registry. */
+  modelProvider?: (model: string) => BaseLlm;
+}
+
+/** @internal */
+export function createModelActivities(options: ModelActivitiesOptions = {}): ModelActivities {
+  const resolveModel = (model: string): BaseLlm => options.modelProvider?.(model) ?? LLMRegistry.newLlm(model);
+
+  return {
+    async invokeModel(args: InvokeModelArgs): Promise<LlmResponse[]> {
```
DABH: So allegedly these names get registered globally and SimplePlugin merges activities by object spread — silent last-wins.  So for instance if a user uses both StrandsPlugin and GoogleAdkPlugin in their code then  one plugin's implementation will silently service the other's calls - but the arg shapes differ, so this fails at runtime in confusing ways, or worse.  Recommend either prefixing (googleAdk-invokeModel, matching the <name>-listTools style) or, at minimum, documenting the incompatibility. This is an ecosystem-level problem (Python's unprefixed invoke_model has it too), but TS is where two same-named plugins can actually be co-registered today.

Please verify this is really the case before doing anything though - but seems worth fixing if you can repro/verify.

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+```typescript
+import { InMemoryRunner, LlmAgent, isFinalResponse, stringifyContent } from '@google/adk';
+import { TemporalModel } from '@temporalio/google-adk-agents/workflow';
+
+export async function askAgent(prompt: string): Promise<string> {
+  const agent = new LlmAgent({
+    name: 'assistant',
+    // The only change from a vanilla ADK agent:
+    model: new TemporalModel('gemini-2.5-flash'),
```
DABH, line 41: If a user just puts 'gemini-2.5-flash' and forgets to wrap in TemporalModel(), ADK resolves it via LLMRegistry inside the workflow and attempts a live HTTP call from the sandbox. With the shims in place this fails with something like "fetch is not defined" or a gutted google-auth-library - loud, but cryptic, and pointing nowhere near the actual mistake.  Admittedly, I don't think we handle this very well in Go or Python either, but can we make some more helpful checks so that the user can immediately be pointed to where the error is?  Or even just add a README entry that's like 'if you see this bizarre error, it's probably because of this'?

contrib/google-adk-agents/package.json
xumaple: done

contrib/google-adk-agents/package.json
xumaple: fixed

contrib/google-adk-agents/src/activities.ts
xumaple: this is indeed an issue, good catch. renamed to adk-*.

contrib/google-adk-agents/README.md
xumaple: it's pretty hard to have a dynamic check for this, so just added a note in the README.

DABH (approved):

contrib/google-adk-agents/src/activities.ts
```diff
@@ -0,0 +1,309 @@
...
+        throw toApplicationFailure(err);
+      } finally {
+        try {
+          await toolset?.close();
+        } catch {
+          /* a close failure must not mask the primary result/error */
+        }
+        stopHeartbeat();
+      }
+    },
```
DABH, line 173: From Claude:
> Every MCP callTool activity pays a redundant listTools round trip and two fresh sessions. (src/activities.ts:144–173) Verified against ADK 1.4.0 source: the activity does getTools() (session #1: connect, tools/list, close) then tool.runAsync() (session #2: connect, tools/call, close); native ADK pays one session per call with no re-list. For stdio transports that's two subprocess spawns per tool call. @temporalio/ai-sdk just shipped the fix pattern (MCP connection reuse with mcpConnectionIdleTimeout, default 5 min) — consider adopting it or documenting the per-call cost. Related sharp edge: the activity calls toolset.close() on whatever the factory returned (:166–168), so a factory returning a shared long-lived BaseToolset gets closed after every call — document that factories must tolerate close-per-call, or skip closing toolsets the plugin didn't construct.

contrib/google-adk-agents/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "rootDir": "./src",
+    "lib": ["ES2022", "ESNext.Disposable", "DOM"],
+    "target": "ES2022"
+  },
+  "references": [
+    { "path": "../../packages/plugin" },
+    { "path": "../../packages/worker" },
+    { "path": "../../packages/workflow" },
+    { "path": "../../packages/activity" },
+    { "path": "../../packages/client" },
```
DABH, line 14: The unused @temporalio/client dep was removed so this line can be cut? (nit)

contrib/google-adk-agents/README.md
```diff
@@ -0,0 +1,285 @@
...
+from history rather than re-executed.
+
+## Operational notes
+
+- Register `GoogleAdkPlugin` on the Worker. Passing it directly to `Client` does
+  not register the model/MCP Activities. If composing plugins, place
+  observability and governance plugins before this one.
+- Model calls use Temporal retries. The plugin disables nested GenAI SDK retries
+  for model requests and honors `retry-after` where available.
+- Heartbeats are sent only when the Activity options include `heartbeatTimeout`.
```
DABH, line 264: This is contradicted by the streaming activity's unconditional per-chunk heartbeat (src/activities.ts:81).

contrib/google-adk-agents/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: MIT
+ *
+ * Worker-side API for the Google ADK Temporal plugin. The workflow-safe
+ * surface (`TemporalModel`, `TemporalMCPToolset`, `activityAsTool`) lives in
+ * the `@temporalio/google-adk-agents/workflow` entry point; test helpers in
+ * `@temporalio/google-adk-agents/testing`.
+ */
+
+/**
+ * @experimental The Google ADK plugin is an experimental feature; APIs may change without notice.
```
DABH, line 13: Nit: the @experimental block comment floats unattached to any declaration, so doc tooling won't pick it up.

contrib/google-adk-agents/src/model.ts
```diff
@@ -0,0 +1,222 @@
...
+    const wire = toWireRequest(llmRequest);
+    const streamingTopic = this.options.streamingTopic;
+
+    let responses: LlmResponse[];
+    if (stream) {
+      if (!streamingTopic) {
+        throw ApplicationFailure.nonRetryable(
+          `TemporalModel('${this.model}'): streaming was requested but no 'streamingTopic' is ` +
+            'configured. Set TemporalModelOptions.streamingTopic to publish incremental chunks.',
+          'GoogleAdkStreamingTopicRequired'
```
DABH, line 143: Nit: Error-type strings (GoogleAdkStreamingTopicRequired, GoogleAdkMCPToolNotFound, …) must be hand-typed by users matching failure.type.  Consider exporting constants from `/workflow` plus a small README error table.

contrib/google-adk-agents/src/mcp.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export interface TemporalMCPToolsetOptions {
+  /**
+   * Name selecting the worker-registered factory in
+   * `GoogleAdkPluginOptions.mcpToolsets`. Also names the pair of Activities
+   * (`<name>-listTools`, `<name>-callTool`).
+   */
+  name: string;
+  /** Restrict the advertised tools to these (post-prefix) names. */
+  toolFilter?: string[];
```
DABH, line 62: Nit: toolFilter appears to only support string[], not ADK's ToolPredicate (a pure predicate would be workflow-safe); worth a doc note if intentionally unsupported.

contrib/google-adk-agents/package.json
```diff
@@ -0,0 +1,108 @@
...
+    "@ungap/structured-clone": "^1.3.1",
+    "headers-polyfill": "^4.0.3",
+    "web-streams-polyfill": "^4.2.0"
+  },
+  "peerDependencies": {
+    "@google/adk": "^1.4.0",
+    "@google/genai": "^2.8.0"
+  },
+  "devDependencies": {
+    "@google/adk": "^1.4.0",
```
DABH, line 79: So, 1.4.x and 1.5.x are both supported. But `^1.4.0` auto-opts users into future minors, and a routine ADK minor can break bundle load in a hard-to-diagnose way (as happened when testing 1.5.0). Suggest bounding the dep to verified minors (>= 1.4, < 1.6)?

chris-olszewski (approved):
Reminder to add this to the `docs` package index so it shows up on the docs site

PR #2127, opened 2026-06-16
Add system Nexus signalWithStartWorkflow API

tconley1428:
## Summary
- add system Nexus TypeScript generation to build:protos using nex-gen
- expose generated signalWithStartWorkflow workflow API and checked-in generated review output
- serialize system Nexus envelopes with default JSON through the isolate, then protobuf-binary in the worker while applying codecs only to nested payloads
- add unit and integration coverage for codec/payload converter behavior

## Validation
- NEX_GEN_BIN=../nexus-api-gen/target/debug/nex-gen pnpm run build:protos
- pnpm --filter @temporalio/test run build:ts
- pnpm -C packages/test exec ava ./lib/test-workflow-codec-runner.js
- pnpm -C packages/test exec ava ./lib/test-system-nexus.js


scripts/gen-system-nexus.ts
```diff
@@ -0,0 +1,463 @@
...
+    let source = readFileSync(path, 'utf8');
+    source = source.replace(/((?:from|export .* from) ['"]\.{1,2}[^'"]*)\.ts(['"])/g, '$1$2');
+    source = source.replace(/from ['"]@temporalio\/workflow['"]/g, () =>
+      path.includes('/operations/') ? "from '../workflow-exports'" : "from './workflow-exports'"
+    );
+    writeFileSync(path, source);
+  }
+
+  writeFileSync(
+    resolve(workflowOutputDir, 'workflow-exports.ts'),
```
tconley1428: This is not okay.

packages/worker/src/system-nexus-payload-visitor.ts
```diff
@@ -0,0 +1,111 @@
+// Generated by scripts/gen-system-nexus.ts. DO NOT EDIT!
+
+import type { Payload } from '@temporalio/common';
+
+type PayloadVisitor = (payload: Payload) => Promise<Payload>;
+type PayloadsVisitor = (payloads: Payload[]) => Promise<Payload[]>;
+
+export const SYSTEM_NEXUS_OPERATIONS = [
```
tconley1428: Should be using the existing registry.

packages/worker/src/system-nexus-payload-visitor.ts
```diff
@@ -0,0 +1,111 @@
...
+  },
+] as const;
+
+export async function visitSystemNexusPayloads(
+  messageTypeName: string,
+  message: any,
+  visitPayload: PayloadVisitor,
+  visitPayloads: PayloadsVisitor
+): Promise<void> {
+  if (messageTypeName === 'temporal.api.workflowservice.v1.SignalWithStartWorkflowExecutionRequest') {
```
tconley1428: Use a switch

packages/worker/src/system-nexus-payload-visitor.ts
```diff
@@ -0,0 +1,111 @@
...
+      visitPayload,
+      visitPayloads
+    );
+    return;
+  }
+  throw new Error(`Unrecognized system Nexus proto message type: ${messageTypeName}`);
+}
+
+async function visit_temporal_api_common_v1_Payloads(
+  value: any,
```
tconley1428: Try to make this code all better.

packages/worker/src/system-nexus-payload-converter.ts
```diff
@@ -0,0 +1,132 @@
...
+  const metadata = payload.metadata ?? {};
+  if (decodeUtf8(metadata.encoding) !== 'binary/protobuf') {
+    throw new Error('System Nexus envelope must use binary/protobuf encoding');
+  }
+  if (decodeUtf8(metadata.messageType) !== messageTypeName) {
+    throw new Error(`System Nexus envelope type mismatch: expected ${messageTypeName}`);
+  }
+}
+
+function lookupMessageClass(messageTypeName: string): any {
```
tconley1428: See if there's a better way to do this with protojs

tconley1428: protoRoot.lookupType?

packages/worker/src/system-nexus-payload-converter.ts
```diff
@@ -0,0 +1,132 @@
...
+  payload: Payload | null | undefined,
+  context: SerializationContext
+): Promise<Payload | undefined> {
+  if (service == null || operation == null || payload == null) return undefined;
+  const op = operationDefinition(service, operation);
+  if (op == null) return undefined;
+
+  const messageType = lookupMessageClass(op.inputType);
+  const message = messageType.create(defaultPayloadConverter.fromPayload(payload) ?? {});
+  normalizePayloadBytes(message);
```
tconley1428, line 40: Check on normalization.

packages/worker/src/system-nexus-payload-converter.ts
```diff
@@ -0,0 +1,132 @@
...
+  await visitSystemNexusPayloads(
+    op.inputType,
+    message,
+    (single) => encode(codecs, [single], context).then(([encoded]) => encoded!),
+    (payloads) => encode(codecs, payloads, context)
+  );
+  return protoBinaryPayload(op.inputType, message);
+}
+
+export async function tryConvertSystemNexusOutputProtoBinaryToJsonPayload(
```
tconley1428, line 54: Is this actually needed?

tconley1428: If a response has a payload.

packages/worker/src/system-nexus-payload-converter.ts
```diff
@@ -0,0 +1,132 @@
...
+  payload: Payload | null | undefined,
+  context: SerializationContext
+): Promise<Payload | undefined> {
+  if (service == null || operation == null || payload == null) return undefined;
+  const op = operationDefinition(service, operation);
+  if (op == null) return undefined;
+
+  assertProtoBinaryPayload(payload, op.outputType);
+  const messageType = lookupMessageClass(op.outputType);
+  const message = messageType.decode(payload.data!);
```
tconley1428: Use the data converter

.github/workflows/ci.yml
```diff
@@ -206,7 +206,8 @@ jobs:
             --dynamic-config-value activity.startDelayEnabled=true \
```
tconley1428, line 206: Confirm manipulating generated files fails CI.

scripts/gen-system-nexus.ts
```diff
@@ -0,0 +1,509 @@
...
+}
+
+function commandExists(command: string): boolean {
+  return spawnSync(command, ['--help'], { stdio: 'ignore' }).status === 0;
+}
+
+function nexGenCommand(): string {
+  if (process.env.NEX_GEN_BIN) return process.env.NEX_GEN_BIN;
+  if (!commandExists('nex-gen')) {
+    run('cargo', ['install', '--locked', 'nex-gen', '--force']);
```
chris-olszewski: I'd prefer we don't have gen scripts mess with host systems. I think erroring in this case with how to install is just fine.

scripts/gen-system-nexus.ts
```diff
@@ -0,0 +1,509 @@
...
+function postProcessGeneratedWorkflowFiles(): void {
+  for (const path of [
+    resolve(workflowOutputDir, 'index.ts'),
+    resolve(workflowOutputDir, 'models.ts'),
+    resolve(workflowOutputDir, 'support.ts'),
+    resolve(workflowOutputDir, 'service.ts'),
+    resolve(workflowOutputDir, 'operations/signal-with-start-workflow.ts'),
+  ]) {
+    let source = readFileSync(path, 'utf8');
+    source = source.replace(/((?:from|export .* from) ['"]\.{1,2}[^'"]*)\.ts(['"])/g, '$1$2');
```
chris-olszewski: Is there a reason why `nexus-gen` creates these files with imports that specify `.ts` as the extension? I don't think that is correct unless we run TS files directly instead of building them to JS first.

packages/workflow/package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "@temporalio/common": "workspace:*",
     "@temporalio/proto": "workspace:*",
+    "long": "^5.2.3",
```
chris-olszewski: From what I can see we only use this for a type reference. If so we can move it to `devDependency`

scripts/gen-system-nexus.ts
```diff
@@ -0,0 +1,509 @@
...
+      '// Generated by scripts/gen-system-nexus.ts. DO NOT EDIT!',
+      '// Local proxy for @temporalio/workflow references emitted by nex-gen.',
+      '',
+      ...exports,
+      '',
+    ].join('\n')
+  );
+}
+
+function postProcessGeneratedWorkflowFiles(): void {
```
chris-olszewski, line 141: Would appreciate comment describing what/why post processing is necessary

packages/workflow/system-nexus/typescript/temporal_model_converters.ts
```diff
@@ -0,0 +1,357 @@
+import * as common from "@temporalio/common";
+import type { google, temporal } from "@temporalio/proto";
+import * as workflow from "@temporalio/workflow";
+import type Long from "long";
+
+function int64ToNumber(
+  value: Long | number | string | object | null | undefined,
```
chris-olszewski: Is there a reason why this helper accepts so many types, but it looks like `Long | null | undefined` are the only ones ever passed?

scripts/gen-system-nexus.ts
```diff
@@ -0,0 +1,509 @@
...
+    return {
+      service: operation.service,
+      operation: operation.operation,
+      inputType: operation.inputType,
+      outputType: operation.outputType,
+    };
+  });
+}
+
+function generateWorkerVisitor(operations: OperationInfo[], messages: Map<string, MessageInfo>): void {
```
chris-olszewski, line 462: Unsure how I feel about this. Could we piggyback on `protobufjs` reflection to just directly implement this instead of generating it?

tconley1428: We definitely could, but it isn't the path taken in every other SDK, where we generate the code for efficiency at runtime.

scripts/gen-system-nexus.ts
tconley1428: Because I don't know typescript!

.github/workflows/conventions.yml
```diff
@@ -47,9 +47,20 @@ jobs:
           env-vars: ''
           save-if: false
 
+      - name: Install nex-gen
+        run: cargo install --force --locked nex-gen --version 0.1.5
+
+      - name: Generate code
+        run: |
+          pnpm run build:protos
+          pnpm run build:system-nexus
```
chris-olszewski: With suggestions to include running system nexus as part of workflow build we don't need this explicit build step in CI
```suggestion
```

packages/workflow/package.json
chris-olszewski, line 24: We should invoke `gen-system-nexus` here similar to how the proto package invokes `build:protos` as part of its build step.

package.json
```diff
@@ -23,15 +23,17 @@
...
+    "build:system-nexus": "tsx ./scripts/gen-system-nexus.ts",
+    "build:generated": "pnpm run build:protos && pnpm run build:system-nexus",
     "test": "pnpm --recursive --parallel --workspace-concurrency=1 --stream run test",
     "test:watch": "pnpm --recursive --stream run test:watch",
     "ci-stress": "node ./packages/test/lib/load/run-all-stress-ci-scenarios.js",
     "ci-nightly": "node ./packages/test/lib/load/run-all-nightly-scenarios.js",
     "lint": "eslint --no-error-on-unmatched-pattern --fix && prettier --write . && pnpm --recursive --stream run lint",
     "lint:check": "eslint --no-error-on-unmatched-pattern && prettier --end-of-line auto --check .  && pnpm --recursive --stream run lint:check",
-    "lint:prune": "ts-prune --error -p tsconfig.prune.json --ignore \"used in module\" --skip \".d.ts\"",
+    "lint:prune": "ts-prune --error -p tsconfig.prune.json --ignore \"used in module|packages/workflow/src/nexus/system/generated\" --skip \".d.ts\"",
```
chris-olszewski, line 37: Note to self: verify dead code in workflow package isn't a significant increase

package.json
```diff
@@ -55,6 +57,7 @@
     "fs-extra": "^11.2.0",
     "json5": "^2.2.3",
     "prettier": "^3.1.1",
+    "protobufjs": "7.5.5",
```
chris-olszewski: If this is required for `gen-system-nexus` it should be added to the scripts package and not the top level.

chris-olszewski (commented):
Dumping some minor package hygiene suggestions

package.json
tconley1428: It's mostly some proto converters that aren't necessarily used, it shouldn't be substantial.

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -30,17 +35,15 @@ export class WorkflowCodecRunner {
...
   constructor(
     private readonly codecs: PayloadCodec[],
     private readonly workflowContext: WorkflowSerializationContext
   ) {}
 
-  private consumeContext<TContext extends SerializationContext>(
-    map: Map<number, TContext>,
-    seq: number | null | undefined
-  ): TContext | undefined {
+  private consumeContext<TContext>(map: Map<number, TContext>, seq: number | null | undefined): TContext | undefined {
```
chris-olszewski: Should we have nexus operations as a serialization context instead of dropping this constraint?

packages/workflow/system-nexus/typescript/temporal_model_converters.ts
```diff
@@ -0,0 +1,333 @@
...
+  duration: common.Duration,
+): google.protobuf.IDuration {
+  return common.msToTs(duration);
+}
+
+function typedSearchAttributePayload(
+  value: unknown,
+  type: common.SearchAttributeType,
+): common.Payload {
+  const payload = configuredPayloadConverter().toPayload(value);
```
chris-olszewski: We shouldn't be using user configured payload converters for search attributes since they need to be `json/plain` encoded regardless of what user has configured.

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -579,11 +604,19 @@ export class WorkflowCodecRunner {
...
-                              command.scheduleNexusOperation.input,
-                              this.workflowContext
-                            ),
+                            input:
+                              (await tryConvertSystemNexusInputJsonToProtoBinaryPayload(
+                                this.codecs,
+                                command.scheduleNexusOperation.service,
+                                command.scheduleNexusOperation.operation,
+                                command.scheduleNexusOperation.input,
+                                this.workflowContext
```
chris-olszewski, line 620: Is it correct that the serialization context should be the caller workflow and not the target workflow?

chris-olszewski (approved):
2 questions and I do think we want to avoid using user configured payload converter for the search attributes

packages/worker/src/workflow-codec-runner.ts
tconley1428: It is not. I filed an issue for follow up: https://github.com/temporalio/sdk-typescript/issues/2210

PR #2128, opened 2026-06-17
fix(worker): reject process global in workflow bundle

### LINKED ISSUE #1112: [Bug] Workflow Bundler does not deny usage of the `process` global variable
### Describe the bug

The Workflow Bundler does not forbid uses of the `process` global variable inside Workflow code, resulting in `ReferenceError: process is not defined` at Workflow execution.

### Minimal Reproduction

1. Add something like `process.getuid()` to any existing workflow (note that no `import` statement is added as `process` is a global variable exposed by Node).

2. Start the worker, and confirm the workflow code compiles without error:

```
[INFO] asset workflow-bundle-09308af9a4bcd2033177.js 726 KiB [emitted] [immutable] (name: main)
[INFO] runtime modules 937 bytes 4 modules
[INFO] modules by path ./node_modules/@temporalio/ 168 KiB
[INFO]   modules by path ./node_modules/@temporalio/common/lib/ 69.1 KiB 18 modules
[INFO]   modules by path ./node_modules/@temporalio/workflow/ 97.1 KiB
[INFO]     ./node_modules/@temporalio/workflow/lib/worker-interface.js 11 KiB [built] [code generated]
[INFO]     + 13 modules
[INFO]   ./node_modules/@temporalio/worker/lib/workflow-log-interceptor.js 2.27 KiB [built] [code generated]
[INFO] modules by path ./src/ 3.73 KiB
[INFO]   ./src/workflows-autogenerated-entrypoint.cjs 575 bytes [built] [code generated]
[INFO]   ./src/workflows.ts 2.9 KiB [built] [code generated]
[INFO]   ./src/shared.ts 270 bytes [built] [code generated]
[INFO] __temporal_custom_payload_converter (ignored) 15 bytes [built] [code generated]
[INFO] __temporal_custom_failure_converter (ignored) 15 bytes [built] [code generated]
[INFO] ./node_modules/long/umd/index.js 43.1 KiB [built] [code generated]
[INFO] ./node_modules/ms/index.js 2.95 KiB [built] [code generated]
[INFO] webpack 5.77.0 compiled successfully in 223 ms
[INFO] Workflow bundle created { size: '0.71MB' }
[INFO] Worker state changed { state: 'RUNNING' }
```

3. Start a Workflow excution, then observe an error message similar to this one in Worker's log:
```
[ERROR] Failed to activate workflow {
  runId: '6898d7d0-6622-4cca-928f-2aa2577e6221',
  error: /Users/.../samples-typescript/mutex/workflow-bundle-09308af9a4bcd2033177.js:4594
  const C = process.env.SOMETHING;
            ^
  
  ReferenceError: process is not defined
      at process (/Users/.../samples-typescript/mutex/src/workflows.ts:26:10)
      at __webpack_require__ (webpack/bootstrap:19:0)
      at importWorkflows (/Users/.../samples-typescript/mutex/src/workflows-autogenerated-entrypoint.cjs:9:9)
      at Object.initRuntime (/Users/.../samples-typescript/mutex/node_modules/@temporalio/workflow/src/worker-interface.ts:138:14)
      at evalmachine.<anonymous>:1:18
      at Script.runInContext (node:vm:134:12)
      at Object.runInContext (node:vm:284:6)
      at Proxy.<anonymous> (/Users/.../samples-typescript/mutex/node_modules/@temporalio/worker/lib/workflow/vm.js:36:46)
      at VMWorkflowCreator.createWorkflow (/Users/.../samples-typescript/mutex/node_modules/@temporalio/worker/lib/workflow/vm.js:43:24)
      at async handleRequest (/Users/.../samples-typescript/mutex/node_modules/@temporalio/worker/lib/workflow/workflow-worker-thread.js:42:13),
  workflowExists: false
}
```

nkgotcode:
## Summary

Reject workflow bundles that reference the Node `process` global directly.

Webpack previously allowed free `process` references through the workflow bundle, which only failed later inside the workflow sandbox. This adds a workflow bundler parser check for free `process` expressions while preserving locally shadowed bindings.

Fixes #1112.

## Testing

```bash
corepack pnpm install --frozen-lockfile
```

```bash
corepack pnpm -F @temporalio/worker run build
```

```bash
corepack pnpm -F @temporalio/test run build
```

Additional focused bundler checks were run with a minimal `@temporalio/core-bridge` cache stub:

- a workflow referencing free `process` is rejected as a disallowed module
- a workflow with a local `process` binding still bundles

Not run:

```bash
corepack pnpm -F @temporalio/worker... run build
```

This currently fails while building `@temporalio/core-bridge` because `protoc` is not installed in this environment.


packages/worker/src/workflow/bundler.ts
```diff
@@ -238,6 +238,22 @@ exports.importInterceptors = function importInterceptors() {
         },
       },
       plugins: [
+        {
+          apply: (compiler) => {
+            compiler.hooks.normalModuleFactory.tap('WorkflowCodeBundler', (normalModuleFactory) => {
+              for (const moduleType of ['javascript/auto', 'javascript/dynamic', 'javascript/esm']) {
+                normalModuleFactory.hooks.parser.for(moduleType).tap('WorkflowCodeBundler', (parser) => {
+                  const javascriptParser = parser as any;
```
chris-olszewski: Will need to add `import type { javascript } from 'webpack'`
```suggestion
                normalModuleFactory.hooks.parser.for(moduleType).tap('WorkflowCodeBundler', (javascriptParser: javascript.JavascriptParser) => {
```

chris-olszewski (changes requested):
Thanks for taking this on. Only a small suggested change to avoid the `as any` cast.

nkgotcode:
Addressed the review feedback in 448b528c by replacing the `as any` parser cast with `javascript.JavascriptParser` from webpack while preserving the hook callback signature required by the local webpack typings.\n\nValidation:\n- `corepack pnpm install --frozen-lockfile`\n- `corepack pnpm -F @temporalio/worker run build` still fails before completion because generated `@temporalio/proto` exports are missing in this fresh checkout; the previous bundler callback type error is gone after the update.\n- `corepack pnpm exec tsc --noEmit --pretty false packages/worker/src/workflow/bundler.ts --skipLibCheck --esModuleInterop --moduleResolution node --target ES2022 --module commonjs` still fails only on missing generated `@temporalio/proto` exports from imported worker utilities.

chris-olszewski (changes requested):
Please read my suggested change from the previous review. You are still casting unnecessarily.

nkgotcode:
Updated the review fix in 5bba954b to follow the requested shape directly.\n\nWhat changed:\n- Removed the parser cast entirely.\n- The parser callback parameter is now typed directly as `javascript.JavascriptParser`.\n- Marked the JavaScript module type list `as const` so TypeScript selects webpack's JavaScript parser hook overload instead of the generic parser fallback.\n\nValidation:\n- `corepack pnpm exec prettier --check packages/worker/src/workflow/bundler.ts` passed.\n- `corepack pnpm -F @temporalio/worker run build` still fails in this local checkout because generated `@temporalio/proto` exports are unavailable, but the previous `src/workflow/bundler.ts` callback type error is gone after the `as const` overload fix.

chris-olszewski (changes requested):
We will need this new hook to respect our allowlist so that our internal testing where we colocate workflow code with tests can pass.

nkgotcode:
Addressed the allowlist feedback from @chris-olszewski.

Changes made:
- The `process` parser hook now respects `ignoreModules` through the same module matching path used for disallowed imports.
- Added `process` to `baseBundlerIgnoreModules` so colocated workflow/test code using the shared test helper allowlist can bundle.
- Added a regression test that bundles a workflow referencing the `process` global with `baseBundlerIgnoreModules`.

Validation:
- `corepack pnpm exec prettier --check packages/worker/src/workflow/bundler.ts packages/test/src/test-bundler.ts packages/test-helpers/src/bundler.ts`
- `corepack pnpm run build:protos`
- `corepack pnpm -F @temporalio/test run build:protos`
- `corepack pnpm -F @temporalio/worker -F @temporalio/test-helpers -F @temporalio/test run build:ts`

I also tried the focused AVA bundler test locally, but this checkout is missing the `packages/core-bridge/releases/aarch64-apple-darwin/index.node` native prebuild, so AVA exits before test discovery.

PR #2129, opened 2026-06-17
extstore s3 driver

cconstable:
## What was changed?

  - Adds the built-in S3 driver for External Storage
  - `@temporalio/external-storage-s3`: S3StorageDriver + S3StorageDriverClient interface, no
  AWS dependency.
  - `@temporalio/external-storage-s3-aws-sdk`: AwsSdkS3StorageDriverClient, backed by
  @aws-sdk/client-s3 (peer dep).
  - `@experimental` not yet wired into the converter pipeline.

## Why?

  - The two-package split keeps the AWS SDK out of core, so users can supply their own S3
  client.

## Checklist

- New tests added

### DIFF
--- a/contrib/external-storage-s3-aws-sdk/package.json
+++ b/contrib/external-storage-s3-aws-sdk/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@temporalio/external-storage-s3-aws-sdk",
-  "version": "1.17.2",
+  "version": "1.18.1",
+  "private": true,
   "description": "AWS SDK (@aws-sdk/client-s3) client for the Temporal S3 external storage driver",
   "main": "lib/index.js",
   "types": "./lib/index.d.ts",
--- a/contrib/external-storage-s3-aws-sdk/package.json
+++ b/contrib/external-storage-s3-aws-sdk/package.json
@@ -1,0 +1,55 @@
+{
+  "name": "@temporalio/external-storage-s3-aws-sdk",
+  "version": "1.17.2",
+  "description": "AWS SDK (@aws-sdk/client-s3) client for the Temporal S3 external storage driver",
+  "main": "lib/index.js",
+  "types": "./lib/index.d.ts",
+  "keywords": [
+    "temporal",
+    "workflow",
+    "external storage",
+    "payload",
+    "s3",
+    "aws"
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT",
+  "scripts": {
+    "build": "tsc --build",
+    "test": "ava ./lib/__tests__/test-*.js"
+  },
+  "ava": {
+    "timeout": "60s"
+  },
+  "dependencies": {
+    "@temporalio/external-storage-s3": "workspace:*"
+  },
+  "peerDependencies": {
+    "@aws-sdk/client-s3": "^3.0.0"
+  },
+  "devDependencies": {
+    "@aws-sdk/client-s3": "^3.0.0",
+    "ava": "^5.3.1"
+  },
+  "engines": {
+    "node": ">= 20.3.0"
+  },
+  "bugs": {
+    "url": "https://github.com/temporalio/sdk-typescript/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/temporalio/sdk-typescript.git",
+    "directory": "contrib/external-storage-s3-aws-sdk"
+  },
+  "homepage": "https://github.com/temporalio/sdk-typescript/tree/main/contrib/external-storage-s3-aws-sdk",
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "src",
+    "lib",
+    "!src/__tests__",
+    "!lib/__tests__"
+  ]
+}
--- a/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
+++ b/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
@@ -20,7 +20,9 @@ export class AwsSdkS3StorageDriverClient implements S3StorageDriverClient {
 
   async objectExists(bucket: string, key: string, options?: S3RequestOptions): Promise<boolean> {
     try {
-      await this.client.send(new HeadObjectCommand({ Bucket: bucket, Key: key }), { abortSignal: options?.abortSignal });
+      await this.client.send(new HeadObjectCommand({ Bucket: bucket, Key: key }), {
+        abortSignal: options?.abortSignal,
+      });
       return true;
     } catch (err) {
       if (isNotFound(err)) {
--- a/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
+++ b/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
@@ -12,8 +12,6 @@ export class AwsSdkS3StorageDriverClient implements S3StorageDriverClient {
   constructor(private readonly client: AwsS3Client) {}
 
   describe(): Record<string, string> {
-    // v3 normalizes `region` to an async provider, so it is usually unavailable
-    // synchronously here; surface it only when it was supplied as a plain string.
     const region = this.client.config?.region;
     return typeof region === 'string' && region ? { clientRegion: region } : {};
   }
--- a/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
+++ b/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
@@ -1,5 +1,5 @@
 import { PutObjectCommand, GetObjectCommand, HeadObjectCommand } from '@aws-sdk/client-s3';
-import type { S3Client as AwsS3Client } from '@aws-sdk/client-s3';
+import type { S3Client } from '@aws-sdk/client-s3';
 import type { S3StorageDriverClient, S3RequestOptions } from '@temporalio/external-storage-s3';
 
 /**
@@ -9,7 +9,7 @@ import type { S3StorageDriverClient, S3RequestOptions } from '@temporalio/extern
  * @experimental
  */
 export class AwsSdkS3StorageDriverClient implements S3StorageDriverClient {
-  constructor(private readonly client: AwsS3Client) {}
+  constructor(private readonly client: S3Client) {}
 
   describe(): Record<string, string> {
     const region = this.client.config?.region;
--- a/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
+++ b/contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
@@ -1,0 +1,53 @@
+import { PutObjectCommand, GetObjectCommand, HeadObjectCommand } from '@aws-sdk/client-s3';
+import type { S3Client as AwsS3Client } from '@aws-sdk/client-s3';
+import type { S3StorageDriverClient, S3RequestOptions } from '@temporalio/external-storage-s3';
+
+/**
+ * An {@link S3StorageDriverClient} backed by an `@aws-sdk/client-s3` `S3Client`,
+ * for use with `S3StorageDriver`.
+ *
+ * @experimental
+ */
+export class AwsSdkS3StorageDriverClient implements S3StorageDriverClient {
+  constructor(private readonly client: AwsS3Client) {}
+
+  describe(): Record<string, string> {
+    // v3 normalizes `region` to an async provider, so it is usually unavailable
+    // synchronously here; surface it only when it was supplied as a plain string.
+    const region = this.client.config?.region;
+    return typeof region === 'string' && region ? { clientRegion: region } : {};
+  }
+
+  async objectExists(bucket: string, key: string, options?: S3RequestOptions): Promise<boolean> {
+    try {
+      await this.client.send(new HeadObjectCommand({ Bucket: bucket, Key: key }), { abortSignal: options?.abortSignal });
+      return true;
+    } catch (err) {
+      if (isNotFound(err)) {
+        return false;
+      }
+      throw err;
+    }
+  }
+
+  async putObject(bucket: string, key: string, data: Uint8Array, options?: S3RequestOptions): Promise<void> {
+    await this.client.send(new PutObjectCommand({ Bucket: bucket, Key: key, Body: data }), {
+      abortSignal: options?.abortSignal,
+    });
+  }
+
+  async getObject(bucket: string, key: string, options?: S3RequestOptions): Promise<Uint8Array> {
+    const response = await this.client.send(new GetObjectCommand({ Bucket: bucket, Key: key }), {
+      abortSignal: options?.abortSignal,
+    });
+    if (!response.Body) {
+      throw new Error(`S3 GetObject returned an empty body [bucket=${bucket}, key=${key}]`);
+    }
+    return response.Body.transformToByteArray();
+  }
+}
+
+function isNotFound(err: unknown): boolean {
+  const e = err as { name?: string; $metadata?: { httpStatusCode?: number } };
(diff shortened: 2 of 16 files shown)


contrib/external-storage-s3-aws-sdk/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@temporalio/external-storage-s3-aws-sdk",
+  "version": "1.18.1",
+  "private": true,
```
cconstable, line 4: prevents npm publish from exposing these until extstore is released. we'll want to flip this later.

contrib/external-storage-s3/src/driver.ts
```diff
@@ -0,0 +1,221 @@
...
+  const signal = external ? AbortSignal.any([external, controller.signal]) : controller.signal;
+  const tasks = makeTasks(signal);
+  try {
+    return await Promise.all(tasks);
+  } catch (err) {
+    controller.abort();
+    await Promise.allSettled(tasks);
+    throw err;
+  }
+}
```
cconstable, line 114: n.b. 
- S3 driver's `runAllWithAbortOnError` is concurrent and cancels in-flight sibling requests on the first failure (AWS SDK allows for cancelling mid-request). 
- GCS driver's `allSettledOrThrow` is concurrent but doesn't attempt mid-flight cancellation. `@google-cloud/storage` can't cancel an in-flight request. Abort in GCS is only check up-front.

contrib/external-storage-s3/src/__tests__/test-driver.ts
```diff
@@ -0,0 +1,207 @@
...
+};
+
+test('store then retrieve round-trips the payload bytes', async (t) => {
+  const driver = new S3StorageDriver({ client: new FakeS3Client(), bucket: 'b' });
+  const original = makePayload('"hello"');
+
+  const [claim] = await driver.store(workflowContext, [original]);
+  const [retrieved] = await driver.retrieve({}, [claim!]);
+
+  t.deepEqual(payloadBytes(retrieved!), payloadBytes(original));
```
jmaeagle99: Is encoding to bytes and then comparing the bytes correct? It might be _here_ because we are using the same exact version of the libraries to do this encoding. But in general I think deep object equality is needed to avoid reordering problems.

jmaeagle99: A lot of the tests are using the non-null assertion operator all over the place. Could we use something like `assert(retrieved)` and drop the operator? Might make it easier to understand a test failure if these things are returning `null` or `undefined`.

contrib/external-storage-s3/src/driver.ts
```diff
@@ -0,0 +1,221 @@
...
+ * Runs the per-payload operations concurrently. On the first failure, aborts the
+ * shared signal so in-flight sibling requests are cancelled, then waits for them
+ * to settle before propagating the original error.
+ */
+async function runAllWithAbortOnError<T>(
+  external: AbortSignal | undefined,
+  makeTasks: (signal: AbortSignal) => Promise<T>[]
+): Promise<T[]> {
+  const controller = new AbortController();
+  const signal = external ? AbortSignal.any([external, controller.signal]) : controller.signal;
```
jmaeagle99, line 105: Claude tells me that the use of `AbortSignal.any()` causes the node engine version to be a min of 20.3.0, which is higher than the other packages in this repo. Any alternatives that might be used instead to allow compat with 20.0.0?

contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
```diff
@@ -0,0 +1,53 @@
...
+ * for use with `S3StorageDriver`.
+ *
+ * @experimental
+ */
+export class AwsSdkS3StorageDriverClient implements S3StorageDriverClient {
+  constructor(private readonly client: AwsS3Client) {}
+
+  describe(): Record<string, string> {
+    const region = this.client.config?.region;
+    return typeof region === 'string' && region ? { clientRegion: region } : {};
```
jmaeagle99, line 16: `region` might also be a `Provider<string | undefined>` as documented at https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/ClientInputEndpointParameters/

Do we want to invoke the provider, check if the promise is completed (assuming the promise is cached), and read the value? If this is even possible.

contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
```diff
@@ -0,0 +1,53 @@
+import { PutObjectCommand, GetObjectCommand, HeadObjectCommand } from '@aws-sdk/client-s3';
+import type { S3Client as AwsS3Client } from '@aws-sdk/client-s3';
```
jmaeagle99: Is the type alias necessary?

contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
```diff
@@ -0,0 +1,53 @@
...
+import type { S3StorageDriverClient, S3RequestOptions } from '@temporalio/external-storage-s3';
+
+/**
+ * An {@link S3StorageDriverClient} backed by an `@aws-sdk/client-s3` `S3Client`,
+ * for use with `S3StorageDriver`.
+ *
+ * @experimental
+ */
+export class AwsSdkS3StorageDriverClient implements S3StorageDriverClient {
+  constructor(private readonly client: AwsS3Client) {}
```
jmaeagle99: I think the way we are using this client, it does not support multipart upload. We might either need to [depend on another package](https://github.com/aws/aws-sdk-js-v3/blob/main/lib/lib-storage/README.md) or write it ourselves using this client. There is varying recommendations of at what threshold multipart should be used, but it looks like AWS [says](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) to definitely use it at 100 MB or larger, which is above our max payload size. So maybe something to defer for now, what for feedback and/or measure with experimentation.

jmaeagle99 (commented):
LGTM with a few minor comments. Please seek an TS maintainer for review as well.

contrib/external-storage-s3/src/driver.ts
cconstable: Was just coming back to leave this comment 😂 I bumped the node engine version but forgot to push that PR because it needed some more discussion. tl;dr is that 20 is past EOL, had 60+ CVEs and I think it's in everyone's best interest to at least bump the minor version here.

contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
cconstable: Nope, this is an artifact from when I didn't have 2 packages and wanted to explicitly alias the official AWS client to avoid confusion. Removing.

contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
cconstable: Good callout. I've documented this. Agree on deferring for now.

contrib/external-storage-s3-aws-sdk/src/aws-sdk-client.ts
cconstable: My understanding is that region will be set to the string when the promise resolves which is essentially what you are suggesting. We can't check promise status.

contrib/external-storage-s3/src/__tests__/test-driver.ts
cconstable: I don't think I'm following. The payload bytes are what we are trying to test here and given claims are content-addressed I think this avoids any issues we might have with comparing anything else.

contrib/external-storage-s3/README.md
```diff
@@ -0,0 +1,159 @@
+# Amazon S3 External Storage Driver for the Temporal TypeScript SDK
+
+> ⚠️ **This package is experimental and may be subject to change.** ⚠️
+
+`@temporalio/external-storage-s3` stores and retrieves Temporal payloads in Amazon S3 via the [External Storage](../../README.md) feature.
```
THardy98: Why does `External Storage` link to the README? did you mean docs: https://docs.temporal.io/external-storage

contrib/external-storage-s3/README.md
THardy98, line 1: nice README!

contrib/external-storage-s3/src/driver.ts
```diff
@@ -0,0 +1,221 @@
...
+
+const PayloadProto = proto.temporal.api.common.v1.Payload;
+
+const DRIVER_TYPE = 'aws.s3driver';
+const DEFAULT_MAX_PAYLOAD_SIZE = 50 * 1024 * 1024;
+
+/** Picks the destination bucket for a given payload. Enables dynamic per-payload routing. */
+export type BucketSelector = (context: StorageDriverStoreContext, payload: Payload) => string;
+
+/** @experimental */
```
THardy98: Might be worth adding a docstring here

contrib/external-storage-s3/src/driver.ts
```diff
@@ -0,0 +1,221 @@
...
+ */
+export class S3StorageDriver implements StorageDriver {
+  readonly name: string;
+  readonly type = DRIVER_TYPE;
+  private readonly client: S3StorageDriverClient;
+  private readonly bucket: BucketSelector;
+  private readonly maxPayloadSize: number;
+
+  constructor(options: S3StorageDriverOptions) {
+    const { client, bucket, driverName, maxPayloadSize = DEFAULT_MAX_PAYLOAD_SIZE } = options;
```
THardy98: pedantic, can `driverName = DRIVER_TYPE` no? (same pattern you're using for `maxPayloadSize`)

contrib/external-storage-s3/src/__tests__/test-driver.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  // space -> %20, / -> %2F, !*'() kept, + -> %2B, = -> %3D, ~ -> %7E
+  t.true(
+    claim.claimData.key.startsWith(
+      "v0/ns/payments%20prod/wt/Capture%2FCharge!*'()/wi/order%2B123%3Dabc/ri/r%7E1/d/sha256/"
+    )
+  );
+});
+
+test('a target with no identity falls back to a bare digest key', async (t) => {
```
THardy98, line 111: maybe worth adding a case where there is a target, but with missing segments
asserting we encode as literal `null` string

contrib/external-storage-s3/src/driver.ts
```diff
@@ -0,0 +1,221 @@
...
+  maxPayloadSize?: number;
+}
+
+/**
+ * Percent-encodes a key segment, escaping everything outside S3's safe-character
+ * set (alphanumerics and `! - _ . * ' ( )`). An empty or absent value becomes the
+ * literal `null`.
+ */
+function encodeKeySegment(value: string | undefined): string {
+  if (!value) return 'null';
```
THardy98: did we have this as a sentinel const value in other SDKs?

just wondering if it's clearer if we have this as a `const` alongside `DRIVER_TYPE` and `DEFAULT_MAX_PAYLOAD_SIZE`

contrib/external-storage-s3/src/driver.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    const hashValue = createHash('sha256').update(payloadBytes).digest('hex');
+    const key = `v0${contextSegments}/d/sha256/${hashValue}`;
+
+    try {
+      if (!(await this.client.objectExists(bucket, key, { abortSignal }))) {
+        await this.client.putObject(bucket, key, payloadBytes, { abortSignal });
+      }
+    } catch (err) {
```
THardy98: hmm, i think there's a check-then-act race condition where it's possible to upload 2 of the same object in `storePayload` in spite of this check.

`storePayload` runs concurrently in `map`, consider:

  - on payload 1, `objectExists` blocks, we switch to task with payload 2, which sees `objectExists == false`
  - payload 1 also sees `objectExists == false`
  - both `putObject`, not that it matters but it may also be the case that the `putObject` calls finish in a different ordering (i.e. payload 2 then payload 1)

I'm not sure if there's an actual issue with this, I think it'd just overwrite which would probably be fine. Maybe worth noting though.

contrib/external-storage-s3/src/driver.ts
THardy98, line 114: Remind me - what do we do when an operation fails, is it a task failure and then we retry?

Just thinking of the scenario when `store` fails partway through. Subsequent retries should attempt to store only the payloads where `objectExists == false` ?

What about if retries are exhausted and we've partly completed the `store` operation? I guess there isn't much we can do

contrib/external-storage-s3/src/driver.ts
cconstable: I don't believe so but I think it makes sense to do this

contrib/external-storage-s3/src/driver.ts
cconstable: Yes. We fail the task and it must be retried at the task-level. If we are trying store many payloads and one fails, the whole thing fails.

cconstable: Right now, none of the drivers will retry. I think we should have some sensible retry policy to address transient failures. Will chat with the team about this.

contrib/external-storage-s3/README.md
cconstable: Yea I suppose that is better! I originally had this linking to the in-repo docs.

contrib/external-storage-s3/src/driver.ts
cconstable: This has been merged into main

contrib/external-storage-s3/src/driver.ts
cconstable: There is a chance that the same payload may be uploaded twice. Though uploads are idempotent, if I was a user, I would assume we would do some deduping here. Added a small fix for this. Nice catch.

contrib/external-storage-s3/src/__tests__/test-driver.ts
jmaeagle99: Sorry, was thinking of a different structure. I sure hope that a stored payload would be retrieved and deserialized exactly as what was stored!

contrib/external-storage-s3/src/driver.ts
```diff
@@ -0,0 +1,244 @@
...
+    this.client = client;
+    this.bucket = typeof bucket === 'string' ? () => bucket : bucket;
+    this.name = driverName;
+    this.maxPayloadSize = maxPayloadSize;
+  }
+
+  async store(context: StorageDriverStoreContext, payloads: Payload[]): Promise<StorageDriverClaim[]> {
+    const contextSegments = buildContextSegments(context.target);
+    return runAllWithAbortOnError(context.abortSignal, (signal) => {
+      const uploads = new Map<string, Promise<void>>();
```
jmaeagle99, line 146: That's an interesting optimization! Should we do the same for downloads too?

jmaeagle99 (approved):

contrib/external-storage-s3/src/driver.ts
jmaeagle99: Still a chance of this happening across different payload-bearing fields, so you can't dedup that. The check-the-act race is fine because it should be the same content going to the same key. The check is just to avoid having to upload something that already exists and not waste time or network on it.

THardy98 (approved):

contrib/external-storage-s3/src/driver.ts
cconstable: Yea actually I think this would be great for downloads too. I think I'd like to get this PR merged and then do a quick follow-up PR to add that optimization. I'll make some issues to add this to the other SDKs.

PR #2130, opened 2026-06-18
extstore gcs driver

cconstable:
## What changed?

  - Adds the built-in GCS driver for External Storage.
  - `@temporalio/external-storage-gcs`: `GcsStorageDriver` + `GcsStorageDriverClient` interface, no Google SDK
  dependency.
  - `@temporalio/external-storage-gcs-google-sdk`: `GoogleCloudGcsStorageDriverClient`, backed by
  `@google-cloud/storage` (peer dep).
  - `@experimental` and both packages `private` (unpublished)

  ## Why?

- extstore GA

## Checklist

- New tests for both packages.

### DIFF
--- a/contrib/external-storage-gcs-google-sdk/README.md
+++ b/contrib/external-storage-gcs-google-sdk/README.md
@@ -1,0 +1,27 @@
+# @google-cloud/storage Client for the Temporal GCS External Storage Driver
+
+> ⚠️ **This package is experimental and may be subject to change.** ⚠️
+
+`@temporalio/external-storage-gcs-google-sdk` provides a [`@google-cloud/storage`](https://www.npmjs.com/package/@google-cloud/storage)-backed `GcsStorageDriverClient` for [`@temporalio/external-storage-gcs`](../external-storage-gcs).
+
+`@google-cloud/storage` is a peer dependency, so the driver uses the same `Storage` instance (and version) your application already configures.
+
+## Usage
+
+    npm install @temporalio/external-storage-gcs @temporalio/external-storage-gcs-google-sdk @google-cloud/storage
+
+```ts
+import { Storage } from '@google-cloud/storage';
+import { GcsStorageDriver } from '@temporalio/external-storage-gcs';
+import { GoogleCloudGcsStorageDriverClient } from '@temporalio/external-storage-gcs-google-sdk';
+
+const storage = new Storage();
+const driver = new GcsStorageDriver({
+  client: new GoogleCloudGcsStorageDriverClient(storage),
+  bucket: 'my-temporal-payloads',
+});
+```
+
+## Notes
+
+- `@google-cloud/storage` cannot cancel a request once it is in flight, so an `abortSignal` is honored only up front (via `AbortSignal.throwIfAborted()`): a signal that has already fired prevents the request from starting. An in-flight request is always awaited, so no request is left running in the background.
--- a/contrib/external-storage-gcs-google-sdk/package.json
+++ b/contrib/external-storage-gcs-google-sdk/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@temporalio/external-storage-gcs-google-sdk",
+  "version": "1.18.1",
+  "private": true,
+  "description": "@google-cloud/storage client for the Temporal GCS external storage driver",
+  "main": "lib/index.js",
+  "types": "./lib/index.d.ts",
+  "keywords": [
+    "temporal",
+    "workflow",
+    "external storage",
+    "payload",
+    "gcs",
+    "gcp"
+  ],
+  "author": "Temporal Technologies Inc. <sdk@temporal.io>",
+  "license": "MIT",
+  "scripts": {
+    "build": "tsc --build",
+    "test": "ava ./lib/__tests__/test-*.js"
+  },
+  "ava": {
+    "timeout": "60s"
+  },
+  "dependencies": {
+    "@temporalio/external-storage-gcs": "workspace:*"
+  },
+  "peerDependencies": {
+    "@google-cloud/storage": "^7.0.0"
+  },
+  "devDependencies": {
+    "@google-cloud/storage": "^7.0.0",
+    "ava": "^5.3.1"
+  },
+  "engines": {
+    "node": ">= 20.3.0"
+  },
+  "bugs": {
+    "url": "https://github.com/temporalio/sdk-typescript/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/temporalio/sdk-typescript.git",
+    "directory": "contrib/external-storage-gcs-google-sdk"
+  },
+  "homepage": "https://github.com/temporalio/sdk-typescript/tree/main/contrib/external-storage-gcs-google-sdk",
+  "files": [
+    "src",
+    "lib",
+    "!src/__tests__",
+    "!lib/__tests__"
+  ]
+}
--- a/contrib/external-storage-gcs/README.md
+++ b/contrib/external-storage-gcs/README.md
@@ -1,0 +1,142 @@
+# Google Cloud Storage External Storage Driver for the Temporal TypeScript SDK
+
+> ⚠️ **This package is experimental and may be subject to change.** ⚠️
+
+`@temporalio/external-storage-gcs` stores and retrieves Temporal payloads in Google Cloud Storage via the [External Storage](https://docs.temporal.io/external-storage) feature.
+
+This package has no Google Cloud dependency: it defines the driver and the `GcsStorageDriverClient` interface, and you supply the GCS client. Use the companion [`@temporalio/external-storage-gcs-google-sdk`](../external-storage-gcs-google-sdk) package for a [`@google-cloud/storage`](https://www.npmjs.com/package/@google-cloud/storage) `Storage`-backed client, or implement the interface yourself.
+
+## Using the Google Cloud SDK client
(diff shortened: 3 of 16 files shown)


contrib/external-storage-gcs-google-sdk/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@temporalio/external-storage-gcs-google-sdk",
+  "version": "1.18.1",
```
cconstable, line 3: this is locked to whatever `sdk-ts` is regardless of what is put here

contrib/external-storage-gcs-google-sdk/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "private": true,
```
cconstable, line 4: prevents npm publish from exposing these until extstore is released. we'll want to flip this later.

contrib/external-storage-gcs/src/driver.ts
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * Awaits every task, even when one rejects, so no request is left running in the
+ * background, then surfaces the first rejection (or all results when none fail).
+ */
+async function allSettledOrThrow<T>(tasks: Promise<T>[]): Promise<T[]> {
+  const results = await Promise.allSettled(tasks);
+  const rejected = results.find((r): r is PromiseRejectedResult => r.status === 'rejected');
+  if (rejected) throw rejected.reason;
+  return (results as PromiseFulfilledResult<T>[]).map((r) => r.value);
+}
```
cconstable: n.b. 
- S3 driver's `runAllWithAbortOnError` is concurrent and cancels in-flight sibling requests on the first failure (AWS SDK allows for cancelling mid-request). 
- GCS driver's `allSettledOrThrow` is concurrent but doesn't attempt mid-flight cancellation. `@google-cloud/storage` can't cancel an in-flight request. Abort in GCS is only check up-front.

contrib/external-storage-gcs/src/driver.ts
THardy98, line 1: What are your thoughts on extracting shared driver components like `BucketSelector`, client context formatting etc.?

Maybe a bit premature now, but something to consider when/if there are more drivers ?
(not blocking at all, just curious)

contrib/external-storage-gcs/src/driver.ts
```diff
@@ -0,0 +1,225 @@
...
+ */
+export class GcsStorageDriver implements StorageDriver {
+  readonly name: string;
+  readonly type = DRIVER_TYPE;
+  private readonly client: GcsStorageDriverClient;
+  private readonly bucket: BucketSelector;
+  private readonly maxPayloadSize: number;
+
+  constructor(options: GcsStorageDriverOptions) {
+    const { client, bucket, driverName, maxPayloadSize = DEFAULT_MAX_PAYLOAD_SIZE } = options;
```
THardy98: ditto nit here from the S3 PR: maybe we can assign the default value to `driverName` here, instead of later, so defaults are in one spot

contrib/external-storage-gcs/src/driver.ts
```diff
@@ -0,0 +1,225 @@
...
+        `Payload size ${payloadBytes.length} bytes exceeds the configured maxPayloadSize of ${this.maxPayloadSize} bytes`
+      );
+    }
+
+    const hashValue = createHash('sha256').update(payloadBytes).digest('hex');
+    const object = `v0${contextSegments}/d/sha256/${hashValue}`;
+
+    try {
+      await this.client.save(bucket, object, payloadBytes, { abortSignal });
+    } catch (err) {
```
THardy98, line 201: Is there a reason why we don't check for duplicates for GCS? (i.e. vs S3). `objectExists` is not included in the client interface I've noticed.

contrib/external-storage-gcs/src/driver.ts
```diff
@@ -0,0 +1,225 @@
...
+function percentEncodeChar(char: string): string {
+  let encoded = '';
+  for (const byte of new TextEncoder().encode(char)) {
+    encoded += '%' + byte.toString(16).toUpperCase().padStart(2, '0');
+  }
+  return encoded;
+}
+
+function encodeObjectNameSegment(value: string | undefined): string {
+  if (!value) return 'null';
```
THardy98: ditto nit from the S3 PR - maybe `null` (and other sentinels as you see fit) can be `const` values, like `GCS_UNSAFE_SEGMENT` above

contrib/external-storage-gcs/src/__tests__/test-driver.ts
```diff
@@ -0,0 +1,194 @@
...
+  // Per Google's recommendations: space, + = ~ ! ' ( ) are left intact;
+  // / -> %2F and * -> %2A are escaped.
+  t.true(
+    claim!.claimData.object!.startsWith(
+      "v0/ns/payments prod/wt/Capture%2FCharge!%2A'()/wi/order+123=abc/ri/r~1/d/sha256/"
+    )
+  );
+});
+
+test('a target with no identity falls back to a bare digest object name', async (t) => {
```
THardy98, line 113: ditto on adding a test to check that escaped segments like `.`, `..` and empty segments like `null` are created correctly

contrib/external-storage-gcs/README.md
THardy98, line 1: WDYT of having sections for custom GCS client implementations and bucket selectors?

I don't have strong feelings on this, just noticed you included them in the S3 driver README (as well as perms - not sure if that applies here)

contrib/external-storage-gcs/src/driver.ts
THardy98: Thanks for the note, helpful 👍

contrib/external-storage-gcs/src/driver.ts
cconstable: I do think we will eventually have some shared driver utils. I actually split something out already but then then merged it back in since it was a bit premature.

contrib/external-storage-gcs/src/driver.ts
cconstable: I like this

contrib/external-storage-gcs/src/driver.ts
cconstable: Open to better names here too. These functions have the same "await all tasks" shape but they behave a little differently

contrib/external-storage-gcs/src/driver.ts
cconstable: Yep! GCS handles this for us in the actual save call (inside this save call) via `preconditionOpts`:

```
await file.save(data, { preconditionOpts: { ifGenerationMatch: 0 } });
```

contrib/external-storage-gcs/README.md
cconstable: I think that's a good call. Added.

contrib/external-storage-gcs/src/driver.ts
cconstable: I did add the same upload deduping as was added to the S3 driver so we don't spend time + bandwidth uploading unnecessary payloads.

THardy98 (approved):

contrib/external-storage-gcs-google-sdk/README.md
```diff
@@ -0,0 +1,28 @@
...
+const storage = new Storage();
+const driver = new GcsStorageDriver({
+  client: new GoogleCloudGcsStorageDriverClient(storage),
+  bucket: 'my-temporal-payloads',
+});
+```
+
+## Notes
+
+- Uploads use an atomic create-if-absent precondition (`ifGenerationMatch: 0`); a `412 Precondition Failed` from an already-stored object is treated as a successful no-op.
```
jmaeagle99: Isn't this immaterial to someone who uses this as a package? If so, then I don't think we need to describe that here.

jmaeagle99: Only think I can think of is if there is some kind of observability or diagnostic they could attach and see the HTTP 412.

contrib/external-storage-gcs/src/driver.ts
```diff
@@ -0,0 +1,239 @@
...
+  const entries = Object.entries(info);
+  return entries.map(([k, v]) => `, ${k}=${v}`).join('');
+}
+
+/**
+ * Awaits every task, even when one rejects, so no request is left running in the
+ * background, then surfaces the first rejection (or all results when none fail).
+ */
+async function allSettledOrThrow<T>(tasks: Promise<T>[]): Promise<T[]> {
+  const results = await Promise.allSettled(tasks);
```
jmaeagle99: Even though `@google-cloud/storage` may not support in-flight cancellation, we might have a GCS client that does. We shouldn't let the implementation of an interface dictate how we handle the semantics of the interface. I'm thinking we should still do `Promise.all` (which will fail-fast), cancel the abort signal, and then `Promise.allSettled`. We can validate the behavior with unit tests.

jmaeagle99 (approved):

contrib/external-storage-gcs-google-sdk/README.md
cconstable: That's fair. It wasn't obvious behavior to me but then again it's not something folks using this likely care about.

contrib/external-storage-gcs/src/driver.ts
cconstable: Agreed, thanks for catching this.

PR #2131, opened 2026-06-18
AI-278: Rename @temporalio/openai-agents tracing sink to the reserved __temporal_ prefix

xumaple:
Renames the openai-agents worker tracing sink key `agentTracing` → `__temporal_openai_agents_tracing`, routed through the existing `AGENT_TRACING_SINK_NAME` constant, and adds the matching allowlist entry in `packages/common`'s reserved-name guard. The sink key is internal (not exported), so there's no public API change.

This adopts the reserved `__temporal_` sink-naming convention (matching the snake_case style of the other reserved names, e.g. `__temporal_workflow_stream_offset`), which guarantees the key can't collide with a user-registered sink.


contrib/openai-agents/src/common/agent-sink-types.ts
```diff
@@ -31,8 +31,8 @@ export type AgentTracingSinkEvent =
   | { kind: 'span_started'; span: SerializedAgentSpan; forwardToOtel: boolean }
   | { kind: 'span_complete'; span: SerializedAgentSpan; forwardToOtel: boolean };
 
+export const AGENT_TRACING_SINK_NAME = '__temporal_openaiAgentsTracing' as const;
```
brianstrauch: what about __temporal_openai_agents_tracing?

packages/common/src/reserved.ts
```diff
@@ -18,6 +18,8 @@ const INTERNAL_HANDLER_NAME_ALLOWLIST: ReadonlyMap<string, ReservedNameEntityTyp
   ['__temporal_workflow_stream_publish', 'signal'],
   ['__temporal_workflow_stream_poll', 'update'],
   ['__temporal_workflow_stream_offset', 'query'],
+  // @temporalio/openai-agents
+  ['__temporal_openaiAgentsTracing', 'sink'],
```
brianstrauch: Is this really a "handler" name like this map suggests? As far as I can tell, it isn't associated with an update/query/signal

contrib/openai-agents/src/common/agent-sink-types.ts
xumaple: typescript is camelcase, no? so the `__temporal_` is just to make it internal, but I chose the name to follow language convention. meh peanuts

packages/common/src/reserved.ts
xumaple: renamed the map.

contrib/openai-agents/src/common/agent-sink-types.ts
brianstrauch: I'm just looking at the style of the other reserved names

brianstrauch: (for example: __temporal_workflow_stream_offset)

brianstrauch (approved):

chris-olszewski (approved):

PR #2137, opened 2026-06-23
test(ephemeral-server): poll dev server UI until ready

brianstrauch:
## What

The dev server's gRPC frontend and its Web UI start as **independent listeners**. `createLocal()` only waits for the gRPC frontend to become reachable (the readiness loop in sdk-core's `ephemeral_server` polls the gRPC frontend, never the UI). So the UI's HTTP server may not be accepting connections yet when `createLocal()` resolves, and the two UI tests that `fetch` the UI endpoint immediately could intermittently fail with a connection-refused error — observed flaky on the Bun runner ([example failure](https://github.com/temporalio/sdk-typescript/actions/runs/28045112631/job/83021193220)).

## Change

Replace the bare `fetch()` calls in the two UI tests with a bounded `fetchUntilReady()` poll that retries until the UI serves a **2xx** response or a timeout elapses. Requiring a 2xx (rather than just a resolved `fetch`) confirms the UI is actually *serving*, not merely listening — `fetch` resolving only means the socket connected and some response came back.

The gRPC liveness of these tests is unchanged and still asserted via `connection.ensureConnected()`; this only adds proper readiness waiting for the separate UI listener, which is the thing these tests exist to verify.

## Test

- `pnpm --filter @temporalio/test build:ts` + typecheck clean
- prettier / eslint clean
- Both affected tests pass locally:
  - `TestEnvironment sets up dev server with custom port and ui`
  - `TestEnvironment sets up dev server with custom ui port`


packages/test/src/test-ephemeral-server.ts
```diff
@@ -52,6 +52,34 @@ async function runSimpleWorkflow(t: ExecutionContext<Context>, testEnv: TestWork
...
+// The dev server's gRPC frontend and its Web UI start as independent listeners.
+// createLocal() only waits for the gRPC frontend to become reachable (see the
+// readiness loop in sdk-core's ephemeral_server), so the UI's HTTP server may not
+// be accepting connections yet when the call resolves. Until it is, fetch() rejects
+// with a connection-refused error (Node: `TypeError: fetch failed` with
+// `cause.code === 'ECONNREFUSED'`; Bun: `Error` with `code === 'ConnectionRefused'`).
+//
+// A resolved fetch() only means the socket connected and some response came back, so
+// we additionally require a 2xx status: that confirms the UI is actually serving, not
+// merely listening. Poll until that holds, or fail after a bounded timeout.
```
brianstrauch:
```suggestion
```

chris-olszewski (approved):

PR #2140, opened 2026-06-23
Add concurrent payload visitor

cconstable:
## What was changed
- Added a generic payload visitor for the TS SDK that walks over the `WorkflowActivation` / `WorkflowActivationCompletion` proto trees.
- Added a `proto/scripts/gen-payload-visitor.ts` script that generates the visitors at `proto/src/payload-visitor.generated.ts`. Runs automatically when new protos are generated. About ~1k lines of checked in code.

```ts
// two distinct transform
const options = {
  transformPayload: (p: Payload) => p,
  transformPayloads: (ps: Payload[]) => ps,
};

// sequential
await visitWorkflowActivationCompletion(completion, options);

// bounded per visit
await visitWorkflowActivationCompletion(completion, { ...options, limit: limit(10) });

// global bound
const globalLimit = limit(20);
await Promise.all(
  completions.map((c) => visitWorkflowActivationCompletion(c, { ...options, limit: globalLimit }))
);
```

## Why?
Prerequisite for External Storage store/retrieve (and reusable by codecs / payload validation): we need to find and transform every payload in an arbitrary proto message with bounded concurrency.


.github/workflows/conventions.yml
```diff
@@ -51,5 +51,13 @@ jobs:
       - name: Compile all non-rust code
         run: pnpm --recursive --stream --filter '!@temporalio/core-bridge' --filter '!typescript-sdk' run build
 
+      - name: Check generated payload visitor is up to date
+        run: |
+          pnpm run gen:payload-visitor
+          if ! git diff --exit-code -- packages/proto/src/payload-visitor.generated.ts; then
+            echo "::error::payload-visitor.generated.ts is out of date with the protos. Run 'pnpm run gen:payload-visitor' and commit the result."
+            exit 1
+          fi
```
cconstable, line 60: safeguard to make sure generated protos stay up to data

package.json
```diff
@@ -23,8 +23,9 @@
   "scripts": {
     "rebuild": "pnpm run clean && pnpm run build",
     "build": "pnpm --recursive --stream run build",
-    "build:watch": "pnpm run build:protos && tsc --build --watch packages/*/tsconfig.json",
+    "build:watch": "pnpm run build:protos && pnpm run gen:payload-visitor && tsc --build --watch packages/*/tsconfig.json",
```
cconstable: Building the protos package now builds the payload visitor as well. I can't think of a reason why you wouldn't want to keep them in sync but you can still run `build:protos` to just build the protos.

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -0,0 +1,167 @@
...
+   * background) and not-yet-started ones are skipped.
+   */
+  abortSignal?: AbortSignal;
+}
+
+/**
+ * A counting semaphore. `acquire` resolves once a permit is available; `release` returns one,
+ * handing it directly to the longest-waiting acquirer if any.
+ */
+class Semaphore {
```
cconstable: Hand rolling concurrency primitives feels bad but I didn't see any reusable alternatives. I could move this to a different place if folks have opinions.

jmaeagle99 (commented):
Haven't read too closely, but I think a few things need to be changed:
- The visitor shouldn't enforce cardinality. If a callback wants to mutate the arity of an array, it should be able to e.g. codecs.
- There needs to be some kind of context that is passed through while walking and a separate visitor that allows updating/replacing the context on a per-message basis. This would allow things like SerializationContext or StorageDriverStoreContext to change as we iterate commands in a WFT completion.
- There should be settings for skipping search attributes and headers.

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+export interface VisitOptions<Ctx> {
+  payloadTransform: PayloadTransform<Ctx>;
+  deriveContext?: ContextDeriver<Ctx>;
+  /** Context in scope before any message is entered. */
+  initialContext?: Ctx;
+  /** Max payload-transform calls in flight at once, across all sites. Default 1 (sequential). */
+  concurrency?: number;
+  skipHeaders?: boolean;
+  skipSearchAttributes?: boolean;
```
cconstable, line 59: @jmaeagle99 Added these as requested.

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -0,0 +1,193 @@
...
+ * Called for each payload-bearing field in a message.
+ *
+ * @internal
+ * @experimental
+ */
+export type PayloadTransform<Ctx> = (
+  payloads: Payload[],
+  context: Ctx,
+  abortSignal?: AbortSignal
+) => Promise<Payload[]>;
```
cconstable: This was the most simple implementation: a `PayloadTransform` gets `Payload[]` in and give `Payload[]` back out. However, there is a hidden contract here: if a non-repeatable field represents a payload is being transformed here, we must return exactly one payload in the transform. Practically this means that if `payloads.length === 1` then the return must also be of length 1. As @jmaeagle99 pointed out, we can't just not return a payload when the proto message needs one.

Few options:
1. Leave it as it is and document it in the comments.
2. Add separate "transform" mechanism that enforce this rule statically
```ts
export interface PayloadVisitor<Ctx> {
  transformSingle(payload: Payload, context: Ctx): Promise<Payload>;
  transformRepeated(payloads: Payload[], context: Ctx): Promise<Payload[]>;
}
```
3. Keep one `PayloadTransform` but have the type of the input `payloads` be generic and encode the output type information. I think this is elegant and I might reach for this option in other languages but I think it would be a little unwieldy in TS.

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -0,0 +1,193 @@
...
+ * aborts the rest via the `failure` controller.
+ *
+ * @internal
+ * @experimental
+ */
+export function boundPayloadTransform<Ctx>(
+  transform: PayloadTransform<Ctx>,
+  concurrency: number,
+  failure: AbortController
+): (payloads: Payload[], context: Ctx) => Promise<Payload[]> {
```
cconstable: I chose to wrap the user supplied "transform" in the concurrency limiting logic to keep the actual visiting/walking logic synchronous and agnostic to whatever constraints we want to place on concurrent visiting.

packages/common/src/internal-non-workflow/payload-visitor.ts
cconstable: Additionally, is returning a different number of payloads from transforming a repeated field something we want to allow for or encourage? Does it break any assumptions that eventual users of this payload visitor might reasonably have? 

For example, if a codec using this payload visitor returned a different number of payloads and we didn't enforce any rules this could happen:
```ts
let input = ...
let output = decode(encode(input))

input.length === output.length // could be false!
```

Whether that is "our problem" or not is debatable but we can at least prevent that behavior if we wish to.

cconstable: Lastly, I want to point out that for map fields containing payloads they are visited per entry. The visitor receives each map value as a single-payload rather than receiving the whole map batch. This means map keys and cardinality are preserved (since we enforce that one payload in means one payload out). Technically, its OK to "remove" an item from a map but I'm not sure that is something we want to allow in this visitor (maps are represented as repeated fields on the wire).

Just wanted to bring this up in case anyone can think of a reason to allow for changing the structure of a map with the payload visitor.

jmaeagle99: > Technically, its OK to "remove" an item from a map but I'm not sure that is something we want to allow in this visitor (maps are represented as repeated fields on the wire).

We don't have any need to do this as part of the payload visitor.

packages/common/src/internal-non-workflow/payload-visitor.ts
mjameswh: I'd be inclined to move this to `common/src/concurrency-utils.ts` (new file), but no strong opinion.

packages/proto/package.json
```diff
@@ -10,9 +10,10 @@
     "lib/"
   ],
   "scripts": {
-    "build": "pnpm run build:protos && pnpm run build:ts",
+    "build": "pnpm run build:protos && pnpm run gen:payload-visitor && pnpm run build:ts",
     "build:ts": "tsc --build",
-    "build:protos": "tsx ./scripts/compile-proto.ts"
+    "build:protos": "tsx ./scripts/compile-proto.ts",
```
mjameswh: Let's rename that one to `gen:protos`, to be consistent with the name you just introduced. Also, introduce a new `gen` script, that calls `gen:protos` and `gen:payload-visitor`, and use that one from `build` instead of explicitly calling both `gen:xxx`.

package.json
mjameswh: Same as before, let's rename that one to `gen:protos`, to be consistent with the name you just introduced. Also, introduce a new `gen` script, that calls `gen:protos` and `gen:payload-visitor`, and use that one from `build:watch` instead of explicitly calling both `gen:xxx`.

packages/common/src/internal-non-workflow/payload-visitor.ts
cconstable: I like that 👍

packages/proto/scripts/gen-payload-visitor.ts
```diff
@@ -0,0 +1,211 @@
...
+    } else {
+      lines = [
+        `const p = ${access};`,
+        `if (p != null) pending.push(env.transform([p], ctx).then((r) => { ${access} = one(r); }));`,
+      ];
+    }
+  } else {
+    const walk = fnName(message);
+    if (field.map) {
+      lines = [`const m = ${access};`, `if (m) for (const v of Object.values(m)) ${walk}(v, env, ctx, pending);`];
```
mjameswh: nit: Use a single, multiline \`...\` block. That's what we do in many other places (e.g. `reusable-vm.ts`, `bundler.ts`, `create-project/src/index.ts`, etc). If you want to avoid extra leading spaces in the generated output file (not a big deal tbh), use dedent, with the "dedent`...`" syntax (see the examples I named before).

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -0,0 +1,193 @@
...
+      failure.abort(abortSignal.reason);
+    } else {
+      const onAbort = () => failure.abort(abortSignal.reason);
+      abortSignal.addEventListener('abort', onAbort, { once: true });
+      removeListener = () => abortSignal.removeEventListener('abort', onAbort);
+    }
+  }
+
+  const env: WalkEnv<Ctx> = {
+    transform: boundPayloadTransform(payloadTransform, concurrency, failure),
```
mjameswh: I'd suggest moving concurrency control out of this function. The caller can itself call `boundPayloadTransform()` with appropriate settings, and pass the bounded transformer to the visitor.

This would give the caller more control on the scope of what is concurrency controlled, e.g. applying a global concurrency limit per payload store. As it is now, concurrency limit applies per "visit job", but there could be multiple visit jobs being run concurrently (e.g. processing multiple workflow activations), which means the total number of access operations to the payload store really performed concurrently might be multiple times the per-visit-job limit.

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+}
+
+/**
+ * Applies `payloadTransform` to every {@link Payload} in a `WorkflowActivation`, mutating it in place.
+ *
+ * @internal
+ * @experimental
+ */
+export function visitWorkflowActivation<Ctx = void>(
```
mjameswh: Please add the `async` keyword on this function. No need to await the `runVisit`'s promise, just add the keyword.

Without the `async` keyword, the `visitWorkflowActivation` function will be completely missing in stack traces produced by async code downstream of this function, making it harder to diagnose eventual bugs.

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/**
+ * Applies `payloadTransform` to every {@link Payload} in a `WorkflowActivationCompletion`, mutating it
+ * in place.
+ *
+ * @internal
+ * @experimental
+ */
+export function visitWorkflowActivationCompletion<Ctx = void>(
```
mjameswh: same: `async`

packages/common/src/internal-non-workflow/payload-visitor.ts
cconstable: I think this is a good improvement. I settled on exposing a plimit-like interface `limit` that uses the `Semaphore` in this PR under the hood. I think it's a little more ergonomic and composable that way.

```ts
import { limit, sequential } from '@temporalio/common/lib/concurrency/limit';

// sequential (no concurrency)
await visitWorkflowActivationCompletion(completion, options);

// bounded per visitor
await visitWorkflowActivationCompletion(completion, { ...options, limit: limit(5) });

// globally bounded
const global = limit(10);
await Promise.all(
  completions.map((c) => visitWorkflowActivationCompletion(c, { ...options, limit: global }))
);
```

We could even do things like have a global bound and per visitor bounds.

mjameswh (approved):

PR #2141, opened 2026-06-24
feat(workflow): add nondeterministic unsafe.random for read-only contexts

### LINKED ISSUE #2110: [Feature Request] Nondeterministic random generation API for use in query plugins

xumaple:
Adds `workflowInfo().unsafe.random` — a nondeterministic random source (`random()`, `uuid4()`, `fillRandom()`) for read-only contexts (Query handlers, update validators, interceptors), where the seed-derived random stream would otherwise repeat across evaluations.

It mirrors `unsafe.now`: the native `Math.random` is captured before the Workflow sandbox overrides it, and the closures are stripped across the worker-thread boundary and reinstalled on each side without mutating the live Workflow info.

Fixes #2110


packages/workflow/src/random-helpers.ts
```diff
@@ -19,3 +21,11 @@ export function uuid4FromRandom(random: () => number): string {
     4
   )}-${ho(view.getUint32(10), 8)}${ho(view.getUint16(14), 4)}`;
 }
+
+export function createUnsafeRandomSource(random: () => number): UnsafeRandomSource {
+  return {
+    random,
+    uuid4: () => uuid4FromRandom(random),
+    fillRandom: (bytes) => fillWithRandom(random, bytes),
+  };
```
eamsden: When a workflow returns or otherwise serializes `workflowInfo()` with the default JSON payload converter, `unsafe.now` is omitted because it is a function, but this new `unsafe.random` value is a normal enumerable object whose function members stringify to `{}`. The updated `Workflow can read WorkflowInfo` expectation assumes `random` disappears, so serialized `WorkflowInfo` now contains an extra non-functional `unsafe.random: {} `field and that test/client behavior will fail unless the exposed source/property is made non-serializable.

[codex]

Reproducers committed in temporalio/sdk-typescript@7089ca99a5a2ca12f31ac72048defdb2ee717ece

I am reviewing a proposed fix with an SDK teammate, more tomorrow.

eamsden (changes requested):

packages/test/src/test-random-stream-reset.ts
```diff
@@ -103,6 +103,37 @@ test.serial('named random streams are reseeded after workflow reset', async (t)
...
+  t.regex(second.unsafe, UUID_V4_REGEX);
+
+  t.is(first.named, second.named);
+
+  t.true(first.float >= 0 && first.float < 1);
+  t.true(second.float >= 0 && second.float < 1);
+
+  t.is(first.filled.length, 8);
+  t.is(second.filled.length, 8);
+  t.not(first.filled.join(','), second.filled.join(','));
```
chris-olszewski:
```suggestion
  t.notDeepEqual(first.filled.join(','), second.filled.join(','));
```

packages/workflow/src/random-helpers.ts
```diff
@@ -19,3 +21,11 @@ export function uuid4FromRandom(random: () => number): string {
...
+}
```
eamsden, line 34:
```suggestion

export function createUnsafeRandomSource(random: () => number): UnsafeRandomSource {
  const source: UnsafeRandomSource = {
    random,
    uuid4: () => uuid4FromRandom(random),
    fillRandom: (bytes) => fillWithRandom(random, bytes),
  };

  Object.defineProperty(source, 'toJSON', {
    value: () => undefined,
  });

  return source;
}
```

packages/workflow/src/random-helpers.ts
eamsden: See suggested change below: essentially, prior to returning the `UnsafeRandomSource`, override its `toJSON()` method to return `undefined` thus preventing serialization.

eamsden (changes requested):

packages/workflow/src/random-helpers.ts
xumaple: applied

packages/test/src/test-random-stream-reset.ts
xumaple: done

packages/workflow/src/random-helpers.ts
xumaple: applied that change.

eamsden (changes requested):
Most of this looks good but please cherry-pick the reproducers in 7089ca99. The serialization fix seems fragile and non-obvious and those should go in as regression tests.

eamsden (approved):

PR #2142, opened 2026-06-26
feat(opentelemetry): add support for OTEL v2

### LINKED ISSUE #1658: [Feature Request] Add support for opentelemetry v2

chris-olszewski:
## What was changed
Add a new package that supports OTEL v2 interceptors.

Key changes:
 - Package targets v2 of all OTEL packages it consumes
 - Removed spanExporter support
 - Removed deprecated `OpenTelemetryWorkflowClientCallsInterceptor` alias
 - Provide our own `IdGenerator` that consumes a named random stream to make span ID generation deterministic and preventing it from affecting randomness in workflow.

Reviewers should primarily focus on reviewing the second commit as it is the diff between v1 and v2 of this package. The first commit is a straight copy of the v1 package into the v2 directory.

Replacing https://github.com/temporalio/sdk-typescript/pull/1951

## Why?
We are adding a new package instead of bumping the dependency on the old one for a few reasons:
 - We do not want to perform a major semver bump on just one of our packages, it confuses our versioning story.
 - We do not want to attempt to support both OTEL major versions in the same code as that becomes extremely hacky and hard to read.
 - We want to correct mistakes in the first OTEL package: removing deprecated functionality and moving to use named random streams so adding these interceptors to a worker no longer can trigger NDE if attempting to replay a workflow started without them.

Quick note on James comment from previous PR

> Btw, we now have a preloadedModules bundler option, which may help reduce the overhead associated with per-workflow loading of the OpenTelemetry library.

I don't think this is viable here as registering the tracer provider/context manager in `workflow/index.ts` mutates the loaded frozen OTEL module.

## Checklist

1. Closes #1658

2. How was this tested:
Most tests were directly kept. Replay tests had to be dropped as we changed how we used randomness so histories are incompatible. Added history replay tests for this package as regression tests for future changes to this package.

3. Any docs updates needed?
Once this moves out of experimental, docs page should update to suggest using this package instead of v1.


contrib/interceptors-opentelemetry-v2/src/worker/index.ts
```diff
@@ -197,30 +205,17 @@ export class OpenTelemetryNexusOutboundInterceptor implements NexusOutboundCalls
...
-): InjectedSink<OpenTelemetryWorkflowExporter>;
 /**
  * Takes an opentelemetry SpanProcessor and turns it into an injected Workflow span exporter sink.
- *
- * For backward compatibility, passing a `SpanExporter` directly is still supported.
  */
 export function makeWorkflowExporter(
   spanProcessor: SpanProcessor,
   resource: Resource
 ): InjectedSink<OpenTelemetryWorkflowExporter>;
```
THardy98: do we need this overload anymore? the function signatures are the same, just `spanProcessor` and `processor` param names are different

THardy98 (commented):
LGTM but might want to get someone else to look at this as well. The polyfill is going a bit over my head.

Aside from that, IIUC, the changes largely look like:
- using our seeded RNG for consistent histories
- `parentSpanId -> parentSpanContext`

I'm not as clear why the parent span id -> context change happened

THardy98:
Relevant upgrade doc for OTEL v2: https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/upgrade-to-2.x.md

THardy98 (approved):

bweis:
This is awesome! Thank you!

bweis:
> This is awesome! Thank you!

Is there a way that we can publish a minor version bump and release today?  It's serendipitous that I was running into this issue just today.

chris-olszewski:
@bweis This has been released

PR #2143, opened 2026-06-26
Send SANO links in requests

VegetarianOrc:
## What was changed

- Remove filtering of nexus links when sending start worfklow or signal requests.

## Why?

Standalone Nexus operation links are now supported server side.

## Checklist

1. How was this tested:
- Added a test assertions that a started workflow has the link to the SANO and vice versa.


chris-olszewski (approved):
Just confirming: server side fix is fully rolled out on cloud and never made it into OSS release that supported SANO?

VegetarianOrc:
> Just confirming: server side fix is fully rolled out on cloud and never made it into OSS release that supported SANO?

That's correct. There hasn't been an OSS release w/ SANO support yet and the fix is now rolled out in Cloud.

PR #2147, opened 2026-06-29
AI-310: streaming support for @temporalio/openai-agents

xumaple:
Adds streaming to the OpenAI Agents integration by wiring in the @temporalio/workflow-streams primitive, mirroring the Python SDK. An agent running inside a Workflow can stream model events to an external subscriber via a Workflow Stream topic, exposed through `run(agent, input, { stream: true })`.

The streaming model Activity publishes each event live to the configured topic and returns the full event list, which the Workflow yields deterministically on replay rather than polling the live stream. The user hosts the `WorkflowStream`; the topic name is plugin config via `modelParams.streamingTopic`.

### DIFF
--- a/contrib/openai-agents/README.md
+++ b/contrib/openai-agents/README.md
@@ -170,7 +170,18 @@ Pass `{ stream: true }` to `run()` to stream incremental events as the model res
 
 External consumers (UIs, tracing pipelines) observe events as they arrive by hosting a [`WorkflowStream`](../workflow-streams/README.md) in the Workflow and subscribing with `WorkflowStreamClient`. The streaming Activity publishes each event to the topic configured on `modelParams.streamingTopic`. The topic is required for streaming runs; calling `run(agent, input, { stream: true })` without a configured topic throws before any Activity is scheduled.
 
-Configure the topic on the plugin and host a `WorkflowStream` in the Workflow:
+Streaming requires a `streamingTopic`. Set it on the `modelParams` of the plugin registered on the **Client** — the Client interceptor injects the config header into every Workflow-starting call:
+
+```ts
+const plugin = new OpenAIAgentsPlugin({
+  modelProvider: new OpenAIProvider(),
+  modelParams: { streamingTopic: 'events', streamingBatchInterval: '100ms' },
+});
+
+const client = new Client({ connection, plugins: [plugin] });
+```
+
+Workflows started without that Client — by a Schedule or the Temporal UI/CLI — receive no config header. Supply Workflow-authored defaults through the runner's `defaultModelParams`; unlike the header, these can carry a function-form `summary`. Host a `WorkflowStream` in the Workflow to receive the publishes:
 
 ```ts
 import { Agent } from '@openai/agents-core';
@@ -180,7 +191,7 @@ import { WorkflowStream } from '@temporalio/workflow-streams/workflow';
 export async function streamingWorkflow(prompt: string): Promise<string> {
   new WorkflowStream();
   const agent = new Agent({ name: 'Assistant', instructions: '...' });
-  const runner = new TemporalOpenAIRunner();
+  const runner = new TemporalOpenAIRunner({ defaultModelParams: { streamingTopic: 'events' } });
   const result = await runner.run(agent, prompt, { stream: true });
   for await (const event of result) {
     if (event.type === 'raw_model_stream_event' && event.data.type === 'output_text_delta') {
@@ -191,12 +202,7 @@ export async function streamingWorkflow(prompt: string): Promise<string> {
 }
 ```
 
-```ts
-new OpenAIAgentsPlugin({
-  modelProvider,
-  modelParams: { streamingTopic: 'events', streamingBatchInterval: '100ms' },
-});
-```
+The two layers combine per field: the Client's `modelParams` overrides the runner's `defaultModelParams`.
 
 External subscribers receive the unwrapped native model events directly:
 
@@ -207,6 +213,8 @@ for await (const item of WorkflowStreamClient.create(client, workflowId).topic('
   // item.data — native model stream event
 }
 ```
+
+The hosted `WorkflowStream` retains every published event on the Workflow heap and in history until `truncate(upToOffset)` is called. A streaming agent publishes many events per run, so a long-running or high-volume Workflow that never truncates will eventually hit Temporal's Workflow history size limits. Call `truncate(upToOffset)` periodically to keep the heap and history bounded; see [`@temporalio/workflow-streams`](../workflow-streams/README.md) for the offset model and batching.
 
 Streaming is incompatible with `useLocalActivity`, which supports neither Activity heartbeats nor the Workflow Stream signal channel. See [`@temporalio/workflow-streams`](../workflow-streams/README.md) for the publisher/subscriber API and delivery semantics.
 
--- a/contrib/openai-agents/README.md
+++ b/contrib/openai-agents/README.md
@@ -99,7 +99,6 @@ Differences from the upstream runner:
 
 - `runConfig.model` must be a model name string. The Worker's `modelProvider` resolves it inside the model Activity.
 - `signal` is not supported. Use Temporal cancellation APIs, such as `CancellationScope`, to cancel Workflow work.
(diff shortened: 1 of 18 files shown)


chris-olszewski (approved):

contrib/openai-agents/src/common/model-activity-options.ts
```diff
@@ -34,3 +50,10 @@ export const DEFAULT_MODEL_ACTIVITY_OPTIONS: SerializableModelActivityOptions =
   startToCloseTimeout: '60s',
   useLocalActivity: false,
 };
+
+export const STREAMING_TOPIC_NOT_CONFIGURED = {
+  type: 'StreamingTopicNotConfigured',
+  message:
+    'Streaming requires modelParams.streamingTopic to be set on OpenAIAgentsPlugin. ' +
```
DABH: Can we update the readme to ensure we document that clients need to set modelParams / modelParams.streamingTopic on the plugin?  And/or have a snippet in the readme showing how people need to update their client code to use streaming, not just the workflow side?

DABH: Also - workflows started via schedules or the Temporal UI/CLI don't get a header at all so streaming can never be enabled for them, right?  (IIUC `modelParams` reaches the Workflow exclusively through the `__openai_agents_config` header injected at workflow start by the client-side interceptor...)  When you're calling `configureBundler` could you bake in a default `modelParams` to enable that use case when no header is present?  Or would that force streaming to be enabled all the time, which is also not what we want?  Maybe just a limitation to document?

contrib/openai-agents/README.md
```diff
@@ -163,6 +162,54 @@ export async function approvalWorkflow(input: ApprovalInput = {}): Promise<strin
 
 The agent passed to `RunState.fromString` must define the same tool names, handoff graph, and MCP servers as the run that produced the serialized state.
 
+### Streaming
```
DABH, line 165: Is it worth noting somewhere (here?) that the hosted  WorkflowStream log also holds every event on the workflow heap until the user calls `truncate()`, so can basically grow unbounded till it hits some limits (looks like we enforce a 50K-event/50MB history limit - and apparently hard-terminate the workflow if it exceeds those limits)?  Should we add some guidance that users using workflow streams should be conscious of periodically calling `truncate()`?

contrib/openai-agents/src/workflow/activity-backed-model.ts
```diff
@@ -117,10 +125,33 @@ export class ActivityBackedModel implements Model {
     });
   }
 
-  // eslint-disable-next-line require-yield
-  async *getStreamedResponse(_request: ModelRequest): AsyncIterable<StreamEvent> {
-    throw new Error(
-      'Streaming is not supported in Temporal workflows. ' + 'Use non-streaming mode with TemporalOpenAIRunner.'
-    );
+  async *getStreamedResponse(request: ModelRequest): AsyncIterable<StreamEvent> {
```
DABH, line 128: So `getResponse` wraps with `withGenerationSpan` but `getStreamedResponse` does not.  What are your thoughts on this / how intentional was this?  You could do like
```
async *getStreamedResponse(request: ModelRequest): AsyncIterable<StreamEvent> {
  const streamingTopic = this.modelParams.streamingTopic;
  if (streamingTopic === undefined) {
    throw ApplicationFailure.create({ /* unchanged guard */ });
  }

  const events = await withGenerationSpan(async (span) => {
    span.spanData.model = this.modelName;
    const input: InvokeModelStreamActivityInput = { /* unchanged construction */ };
    return this.activities.invokeModelStreamActivity(input);
  });

  for (const event of events) {
    yield fromSerializedStreamEvent(event);
  }
}
```
if you wanted this to wrap with `withGenerationSpan` as well, and then you'd have better parity of semantics?

contrib/openai-agents/src/workflow/activity-backed-model.ts
DABH, line 117: So `getResponse` supports this dynamic `ModelSummaryProvider` branch (the `executeWithOptions({ summary }` below), but `getStreamedResponse` ignores `modelParams.summary`, so streamed model Activities always show the default summary in the UI.  Do we want to fix that for parity?

contrib/openai-agents/src/common/model-activity-options.ts
xumaple: Good catch. Rather than baking defaults into the bundler, I added a defaultModelParams option to the TemporalOpenAIRunner constructor.

xumaple: also added the readme snippets.

contrib/openai-agents/README.md
xumaple: added

contrib/openai-agents/src/workflow/activity-backed-model.ts
xumaple: good catch. fixed

contrib/openai-agents/src/workflow/activity-backed-model.ts
xumaple: good catch. fixed.

contrib/openai-agents/README.md
```diff
@@ -163,6 +162,62 @@ export async function approvalWorkflow(input: ApprovalInput = {}): Promise<strin
...
+import { WorkflowStreamClient } from '@temporalio/workflow-streams/client';
+
+for await (const item of WorkflowStreamClient.create(client, workflowId).topic('events').subscribe()) {
+  // item.data — native model stream event
+}
+```
+
+The hosted `WorkflowStream` retains every published event on the Workflow heap and in history until `truncate(upToOffset)` is called. A streaming agent publishes many events per run, so a long-running or high-volume Workflow that never truncates will eventually hit Temporal's Workflow history size limits. Call `truncate(upToOffset)` periodically to keep the heap and history bounded; see [`@temporalio/workflow-streams`](../workflow-streams/README.md) for the offset model and batching.
+
+Streaming is incompatible with `useLocalActivity`, which supports neither Activity heartbeats nor the Workflow Stream signal channel. See [`@temporalio/workflow-streams`](../workflow-streams/README.md) for the publisher/subscriber API and delivery semantics.
```
DABH, line 219: There's no fail-fast guard for this combination (stream: true + useLocalActivity: true).  Maybe add a new const in `src/common/model-activity-options.ts` like
```
export const STREAMING_LOCAL_ACTIVITY_UNSUPPORTED = {
  type: 'StreamingLocalActivityUnsupported',
  message:
    'Streaming is incompatible with useLocalActivity. Local Activities cannot heartbeat, so a ' +
    'long-running streaming model call cannot extend past its startToCloseTimeout. ' +
    'Set useLocalActivity: false (the default) for streaming runs.',
} as const;
```

and then after the existing topic guard in `runner.ts` (around L173) add
```
import {
  DEFAULT_MODEL_ACTIVITY_OPTIONS,
  STREAMING_TOPIC_NOT_CONFIGURED,
  STREAMING_LOCAL_ACTIVITY_UNSUPPORTED,   // + add
  type ModelActivityOptions,
  type SerializableModelActivityOptions,
} from '../common/model-activity-options';

    // Fail fast before the upstream streaming runner starts, so a missing topic
    // surfaces as a clean error rather than being captured into the streamed result.
    if (options?.stream === true && this.modelParams.streamingTopic === undefined) {
      throw ApplicationFailure.create({
        message: STREAMING_TOPIC_NOT_CONFIGURED.message,
        type: STREAMING_TOPIC_NOT_CONFIGURED.type,
        nonRetryable: true,
      });
    }

    // + Local Activities can't heartbeat, so a long stream would silently ride
    // + startToCloseTimeout and then time out mid-run. Reject the combination up front.
    if (options?.stream === true && this.modelParams.useLocalActivity === true) {
      throw ApplicationFailure.create({
        message: STREAMING_LOCAL_ACTIVITY_UNSUPPORTED.message,
        type: STREAMING_LOCAL_ACTIVITY_UNSUPPORTED.type,
        nonRetryable: true,
      });
    }
```
?  Could add a test for this as well then.

xumaple: done

DABH (approved):

PR #2149, opened 2026-06-30
AI-311: use unsafe.random for openai-agents read-only handler tracing

xumaple:
Replaces the hand-rolled per-call PRNG in the openai-agents tracing layer (`mulberry32`/`fnv1a32` seeded from queryId/updateId) with `workflowInfo().unsafe.random.random()` in the Query handler and update validator, now that #2141 ships a nondeterministic random source safe in read-only contexts. Removes `prng.ts`, the `resolveQueryKey`/`legacy_query` seed-collision workaround, and the explicit `spanId` plumbing; body/signal/update paths keep the replay-seeded stream.

Adds a worker-based test that issues a Query and an update-validation repeatedly under forced replay and asserts the read-only handlers emit distinct, well-formed span IDs.


DABH (approved):
LGTM!

PR #2153, opened 2026-07-01
chore(deps): unpin protobufjs and bump to 7.6.4

chris-olszewski:
## What was changed
Switch back to using a version range for our `protobufjs` dependency in `@temporalio/protobuf`

## Why?
The version was pinned when we started needing to patch `protobufjs` as it tripped over parsing annotations in our proto files. This was overly cautious as we're able to specify a patch over a version range and `pnpm install` will fail if our patch cannot be applied. 

## Checklist

1. Closes N/A

2. How was this tested:
CI

3. Any docs updates needed?
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,3 +49,7 @@ to docs, or any other relevant information.
 - avoid logging `NativeConnection` on worker startup
 
 ## [Unreleased]
+
+### Changed
+
+- protobufjs bumped to ^7.6.4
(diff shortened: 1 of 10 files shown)


THardy98 (approved):

vavsab:
@chris-olszewski @THardy98 When are you going to publish this fix to npm?
https://www.npmjs.com/package/@temporalio/proto

THardy98:
@vavsab we have a release planned for tomorrow, 1.20.0

PR #2160, opened 2026-07-02
test(update): remove flake in first-WFT update integration tests

cconstable:
## What changed
- Some `test-integration-update.ts` tests need their `updates` delivered in the first WFT (bundled with `startWorkflow`), which they set up via a large `startDelay`.
- They relied on explicit timing ( e.g. `setTimeout`) to get updates admitted before the worker polls which is flakey.

## Fix
- Replaces the timing-based waits with the `updateHasBeenAdmitted` + `waitUntil` admission poll.
- For the two-update test, await the first update's admission before sending the second so their order is deterministic.

### DIFF
--- a/packages/test/src/test-integration-update.ts
+++ b/packages/test/src/test-integration-update.ts
@@ -564,12 +564,15 @@ export async function setUpdateHandlerAndExit(): Promise<string> {
 }
 
 test('Update is always delivered', async (t) => {
-  const { createWorker, startWorkflow } = helpers(t);
+  const { createWorker, startWorkflow, updateHasBeenAdmitted } = helpers(t);
   const wfHandle = await startWorkflow(setUpdateHandlerAndExit, { startDelay: '10000 days' });
 
-  wfHandle.executeUpdate(stateMutatingUpdate).catch(() => {
+  const updateId = 'update-id';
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

packages/test/src/test-integration-update.ts
```diff
@@ -579,18 +582,20 @@ test('Update is always delivered', async (t) => {
...
+  const { createWorker, startWorkflow, updateHasBeenAdmitted } = helpers(t);
   const wfHandle = await startWorkflow(workflowWithUpdates, { startDelay: '10000 days' });
 
-  wfHandle.executeUpdate(update, { args: ['1'] }).catch(() => {
+  const firstUpdateId = 'update-1';
+  wfHandle.executeUpdate(update, { args: ['1'], updateId: firstUpdateId }).catch(() => {
     /* ignore */
   });
-  wfHandle.executeUpdate(doneUpdate).catch(() => {
+  await waitUntil(() => updateHasBeenAdmitted(wfHandle, firstUpdateId), 5000);
```
cconstable, line 592: eventually this will timeout at 5s so if something goes terribly wrong it won't cause the suite to hang

PR #2163, opened 2026-07-06
Extend test harness to support envconfig

THardy98:
## What was changed

Add opt-in envconfig support to the shared TypeScript integration-test harness.

When `TEMPORAL_TEST_ENV_CONFIG_SERVER` is truthy, the harness loads the server address, namespace, API key, TLS configuration, and gRPC metadata through `@temporalio/envconfig`. These settings are propagated to client, worker, and additional native connections.

When the opt-in is absent, existing behavior is preserved: `TEMPORAL_SERVICE_ADDRESS` selects the legacy existing-server path, otherwise the harness starts a local server.

This PR does not migrate individual integration-test call sites. Existing tests that use the shared `createTestWorkflowEnvironment` helper can opt in through environment variables; tests that instantiate `TestWorkflowEnvironment` directly remain unchanged. The envconfig factory remains internal SDK test infrastructure and is not added to public `@temporalio/testing` (this can be done in a follow-up PR, but requires removing a dependency cycle between the `envconfig` and `@temporalio/testing` packages).

**Note:** includes a bug fix with the `envconfig` package when reading `TEMPORAL_TLS` (setting to `true` _enables_ TLS, instead of disables TLS).

## Why?

This enables integration tests to target arbitrary Temporal server environments, including local, staging, release, and Temporal Cloud environments.

This is the first SDK implementation of the approach intended to be replicated across Temporal SDKs.

Part of https://github.com/temporalio/features/issues/851

## How was this tested?

  - A focused integration test creates a test environment from an inline envconfig profile and verifies address, namespace, API key, gRPC metadata, and explicit TLS-disabled propagation.
  - A focused envconfig unit test verifies that `TEMPORAL_TLS=true` enables TLS and `TEMPORAL_TLS=false` disables it.
  - The test-helper and test packages build successfully, and the focused integration test passes.

## Documentation

No user-facing documentation changes are required. This PR changes internal integration-test infrastructure.

### DIFF
--- a/packages/envconfig/src/envconfig-toml.ts
+++ b/packages/envconfig/src/envconfig-toml.ts
@@ -233,8 +233,9 @@ function applyProfileEnvVars(profile: TomlClientConfigProfile, envProvider: Reco
 }
 
 function getTLSFromEnvVars(envProvider: Record<string, string | undefined>): TomlClientConfigTLS | undefined {
+  const tlsEnabled = envVarToBool(envProvider['TEMPORAL_TLS']);
   const tlsConfig: TomlClientConfigTLS = filterNullAndUndefined({
-    disabled: envVarToBool(envProvider['TEMPORAL_TLS']),
+    disabled: tlsEnabled === undefined ? undefined : !tlsEnabled,
     client_cert_path: envProvider['TEMPORAL_TLS_CLIENT_CERT_PATH'],
     client_cert_data: envProvider['TEMPORAL_TLS_CLIENT_CERT_DATA'],
     client_key_path: envProvider['TEMPORAL_TLS_CLIENT_KEY_PATH'],
--- a/packages/test-helpers/src/environment.ts
+++ b/packages/test-helpers/src/environment.ts
@@ -1,10 +1,12 @@
 import type { LocalTestWorkflowEnvironmentOptions } from '@temporalio/testing';
 import { workflowInterceptorModules as defaultWorkflowInterceptorModules } from '@temporalio/testing';
+import { loadClientConnectConfig, type LoadClientProfileOptions } from '@temporalio/envconfig';
 import type { BundlerPlugin, WorkflowBundleWithSourceMap, BundleOptions } from '@temporalio/worker';
 import { bundleWorkflowCode, DefaultLogger } from '@temporalio/worker';
 import { defineSearchAttributeKey, SearchAttributeType } from '@temporalio/common/lib/search-attributes';
 import { TestWorkflowEnvironment } from './wrappers';
 import { baseBundlerIgnoreModules } from './bundler';
+import { isSet } from './flags';
 
 export const defaultDynamicConfigOptions = [
   'system.enableActivityEagerExecution=true',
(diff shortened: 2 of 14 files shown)


packages/test-helpers/src/environment.ts
```diff
@@ -76,14 +78,26 @@ export async function createLocalTestEnvironment(
...
+    const { namespace, connectionOptions } = loadClientConnectConfig();
+    const { address, apiKey, metadata, tls } = connectionOptions;
+    env = await TestWorkflowEnvironment.createFromExistingServer({
+      address,
+      namespace,
+      connectionOptions: { apiKey, metadata, tls },
+      client: opts?.client,
+      plugins: opts?.plugins,
+    });
+  } else if (isSet(process.env.TEMPORAL_SERVICE_ADDRESS, false)) {
```
chris-olszewski: Should we just remove this support and go towards the standard env var names now?

packages/envconfig/src/envconfig-toml.ts
```diff
@@ -233,8 +233,9 @@ function applyProfileEnvVars(profile: TomlClientConfigProfile, envProvider: Reco
 }
 
 function getTLSFromEnvVars(envProvider: Record<string, string | undefined>): TomlClientConfigTLS | undefined {
+  const tlsEnabled = envVarToBool(envProvider['TEMPORAL_TLS']);
   const tlsConfig: TomlClientConfigTLS = filterNullAndUndefined({
-    disabled: envVarToBool(envProvider['TEMPORAL_TLS']),
+    disabled: tlsEnabled === undefined ? undefined : !tlsEnabled,
```
chris-olszewski, line 238: So this was flipped this whole time? Please call this out in the changelog as this could easily get missed in release note generation since it is hidden in this test harness PR.

packages/test/src/helpers-integration.ts
```diff
@@ -126,21 +128,25 @@ export function makeTestFunction<C extends Context = Context>(opts: TestFunction
...
+      const workflowBundle = await createTestWorkflowBundle({
         workflowsPath: opts.workflowsPath,
         workflowInterceptorModules: opts.workflowInterceptorModules,
-      }),
-      env,
-    };
+      });
+      return { workflowBundle, env };
+    } catch (err) {
+      await env.teardown();
```
chris-olszewski, line 137: Seems like the correct change, but just confirming this is an unrelated fix to test harness changes.

chris-olszewski (approved):

packages/test-helpers/src/environment.ts
THardy98: Don't want to break this in case people are using it. But not severely opposed

packages/envconfig/src/envconfig-toml.ts
THardy98: > So this was flipped this whole time?

Yup 🙃 

> Please call this out in the changelog as this could easily get missed in release note generation since it is hidden in this test harness PR

👍

packages/test/src/helpers-integration.ts
THardy98: yup 👍

PR #2169, opened 2026-07-07
fix(worker): patch workflow bundles with const/let/var

### LINKED ISSUE #2166: [Bug] No isolation between worklflow modules with webpack >=5.108.0

In a production multi-tenant app, we have a workflow which polls a queue of tasks, and supervises one child workflow per queue task. We have many of these queues, and many instances of this workflow. After deploying a code change, we started to see strange behavior from this workflow:

- Sending a Query to the supervisor workflow which asked it "which child workflows are currently inflight" would return children from OTHER workflows (even from other tenants!).
- The supervisor workflow was getting "stuck"—it would refuse to dispatch any more Activities. The stack-trace Query indicated that the workflow code was parked on an `await` of a module-level Promise, which never got resolved—and where that module-level promise variable should have been `null` by that point in the workflow.
- We use workflow-ID-prefixes to support multitenancy---and saw a worklflow of `TENANT_A:supervisor` with running children of `TENANT_B:procees_item`

We eventually figured out that *state kept in a module-level variable was getting leaked between workflow instances*.

### Describe the bug

The Temporal SDK uses code-patching to implement per-module isolation when `reuseV8Context` is `true`. That patching works like this, in `src/workflow/bundler.ts` `createBundle`:

```
// Replace webpack's module cache with an object injected by the runtime.
// This is the key to reusing a single v8 context.
code = code.replace(
  'var __webpack_module_cache__ = {}',
  'var __webpack_module_cache__ = globalThis.__webpack_module_cache__'
);
```

However, in Webpack version [version `5.108.0`](https://github.com/webpack/webpack/releases#release-v5.108.0), they released [this PR](https://github.com/webpack/webpack/pull/21010), which changes webpack's default behavior from emitting `var` to emitting `const` and/or `let`, if the target environment supports it.

When the SDK's [dependency on `"webpack": "^5.106.2"`](https://github.com/temporalio/sdk-typescript/blob/6397952c8ad42838e9f845255466c0d40a6edc6b/packages/worker/package.json#L38) resolves to a version at or after `5.108.0`, the code emitted by Webpack will NOT contain `var __webpack_module_cache__ = {}`. Instead, it will contain `const __webpack_module_cache__ = {}`---and the `code.replace` statement above will fail silently.

If this patch fails, modules use the global Webpack module cache, INSTEAD of the proxy object injected by the runtime as `globalThis.__webpack_module_cache__`. And when `reuseV8Context: true`, this means that __every workflow execution__ will share the __exact same module context__.

### Minimal Reproduction

https://github.com/wgoodall01/temporal-webpack-isolation-repro

In this repo, we have the following:
- A test Worker setup that uses `bundleWorkflowCode` to create a workflow bundle script
- A workflow that declares a module-level `moduleState`, returns if there's anything in it, and if not, writes its own workflow ID to it before returning. It looks like this:

```typescript
let moduleState: string | undefined;

export async function stateProbe(): Promise<{ leakedFrom: string | null }> {
  const { workflowId } = workflowInfo();
  if (moduleState === undefined) {
    moduleState = workflowId;
    return { leakedFrom: null };
  }
  return { leakedFrom: moduleState };
}
```
- A test harness which will use a single-threaded Worker, with `reuseV8Context: true`, and run two instances of this workflow, `probe-1` and `probe-2`.
- If either workflow returns something other than `{leakedFrom: null}`, the test harness fails.

With the SDK's default dependency on Webpack, we get the following output:

```bash
$ npm install
$ npm run repro
< ...snip, omit worker startup logs... >
webpack version:         5.108.3
module cache in bundle:  ["const __webpack_module_cache__ = {}"]
SDK cache rewrite applied: false

probe-1 observed leaked module state from: null
probe-2 observed leaked module state from: probe-1

❌ ISOLATION BROKEN: workflow probe-2 observed module-level state written by workflow probe-1.
```

If we add a `package.json` override to pin the webpack version before the problematic change:

```
$ node swap-webpack.mjs 5.107.0   # or 5.106.2, the SDK's declared minimum
$ npm run repro
< ...snip, omit worker startup logs... >
webpack version:         5.107.0
module cache in bundle:  ["var __webpack_module_cache__ = globalThis.__webpack_module_cache__"]
SDK cache rewrite applied: true

probe-1 observed leaked module state from: null
probe-2 observed leaked module state from: null

✅ Isolation intact: no module-level state leaked between executions.
```

The repro-repo's [`README.md`](https://github.com/wgoodall01/temporal-webpack-isolation-repro) also has more detail on how this works.

### Environment/Versions

```
$ sw_vers
ProductName:		macOS
ProductVersion:		15.7.7
BuildVersion:		24G720

$ uname -a
Darwin Wheatley4.local 24.6.0 Darwin Kernel Version 24.6.0: Tue Apr 21 20:18:11 PDT 2026; root:xnu-11417.140.69.710.16~1/RELEASE_ARM64_T6020 arm64

$ node --version
v26.4.0

$ cat package.json | jq .dependencies
{
  "@temporalio/client": "1.19.0",
  "@temporalio/testing": "1.19.0",
  "@temporalio/worker": "1.19.0",
  "@temporalio/workflow": "1.19.0",
  "tsx": "^4.22.4"
}
```

### Additional context

I think this fix is as easy as changing the static-string-replacement from `var` to a regex accepting `var|let|const`. I might try to open a small PR to verify that fix locally.

> **LLM disclosure:** All text in this issue is my own writing; the root-cause analysis and minimal reproduction was written mostly by Claude Fable 5, with manual edits for clarity.

wgoodall01:
## What was changed

- In `packages/worker/src/workflow/bundler.ts`, we now use a regexp to match `(var|let|const) __webpack_module_cache__ = \{\}` instead of only `var`. Recent webpack versions emit `const`, and without this change, the patch doesn't happen and workflow isolation breaks.
- We also add a check that this patch happens exactly once. If it does not (for instance, a future webpack update changes the emitted code to `__wp_mod_c__`, breaking our replacement), then we throw an exception and fail the bundling operation.
    - Hopefully this catches any future similar issues early, at Temporal users' build-step, before they start seeing isolation failures in prod.

## Why?

- Without this patch, workflows share the same module scope, instead of hitting the runtime-injected proxy. This means their module-level variables won't be isolated between workflows.

## Note on lockfiles and `minimumReleaseAge`

This repo uses a pnpm lockfile, so that package developers all get the same dependency versions. However, when end-users install the package, their npm/pnpm/yarn/etc will NOT respect the lockfile in this package—instead, they'll resolve the latest versions permitted by the specifiers in the various `package.json`s. Additionally, a `minimumReleaseAge` pnpm config in this repo prevents SDK developers from installing package versions newer than 2 weeks old.

This bug was caused by a recently-released dependency version (at the time of writing, the `webpack` issue causing this bug was still <2weeks old, which falls inside the configured `minimumReleaseAge` window). **Because this repo uses a lockfile/minimumReleaseAge, the tests will NEVER catch issues with newly-released dependency versions**.

It might be worth investigating one of the following options:

- Remove the lockfile and `minimumReleaseAge` configuration
- Set up a regular GitHub Actions test job, which ignores the lockfile/minimumReleaseAge config, installing all packages like an end-user's package manager would, and runs the regular test suite.
    - This is probably the cleanest—because a failure in this job, while the main tests stay passing, would indicate "something is wrong with a recent dependency update"

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/2166

2. How was this tested:
  - Tested against webpack `5.107.2` (which emits `var`) and webpack `5.108.4` (which emits `const`) using a manual version override.

3. ~~Any docs updates needed?~~ Nope

> **LLM disclosure**: This issue's text was hand-written, the small code patch was LLM-generated (Claude Fable 5).

### DIFF
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -229,7 +229,7 @@ exports.importInterceptors = function importInterceptors() {
       // Ignore the "node:" prefix if any.
       const module: string = data.request?.startsWith('node:')
         ? data.request.slice('node:'.length)
-        : (data.request ?? '');
+        : data.request ?? '';
 
       if (moduleMatches(module, disallowedModules) && !moduleMatches(module, this.ignoreModules)) {
         this.foundProblematicModules.add(module);
--- a/packages/worker/src/workflow/bundler.ts
+++ b/packages/worker/src/workflow/bundler.ts
@@ -136,10 +136,21 @@ export class WorkflowCodeBundler {
     let code = memoryFs.readFileSync(bundleFilePath, 'utf8') as string;
     // Replace webpack's module cache with an object injected by the runtime.
     // This is the key to reusing a single v8 context.
-    code = code.replace(
-      'var __webpack_module_cache__ = {}',
-      'var __webpack_module_cache__ = globalThis.__webpack_module_cache__'
-    );
+    // Webpack may emit the declaration as `var`, `let`, or `const` depending on the
+    // configured output environment (webpack >= 5.108.0 defaults to `const`).
+    let cacheDeclarationsPatched = 0;
+    code = code.replace(/(^|\s)(var|let|const) __webpack_module_cache__ = \{\}/gm, (_match, prefix, keyword) => {
+      cacheDeclarationsPatched++;
+      return `${prefix}${keyword} __webpack_module_cache__ = globalThis.__webpack_module_cache__`;
+    });
+    if (cacheDeclarationsPatched !== 1) {
+      throw new Error(
+        `Failed to patch the Workflow bundle: expected to find exactly one __webpack_module_cache__ declaration ` +
+          `emitted by webpack, but found ${cacheDeclarationsPatched}. Without this patch, Workflow isolation ` +
+          `would be broken. This is likely due to a change in webpack output; please report this at ` +
+          `https://github.com/temporalio/sdk-typescript/issues`
+      );
+    }
 
     this.logger.info('Workflow bundle created', { size: `${toMB(code.length)}MB` });
 
@@ -218,7 +229,7 @@ exports.importInterceptors = function importInterceptors() {
       // Ignore the "node:" prefix if any.
       const module: string = data.request?.startsWith('node:')
         ? data.request.slice('node:'.length)
-        : data.request ?? '';
+        : (data.request ?? '');
 
       if (moduleMatches(module, disallowedModules) && !moduleMatches(module, this.ignoreModules)) {
         this.foundProblematicModules.add(module);


tconley1428:
> Remove the lockfile and minimumReleaseAge configuration
Set up a regular GitHub Actions test job, which ignores the lockfile/minimumReleaseAge config, installing all packages like an end-user's package manager would, and runs the regular test suite.
This is probably the cleanest—because a failure in this job, while the main tests stay passing, would indicate "something is wrong with a recent dependency update"

This is not an option due to security concerns around dependency injection.

mjameswh:
Would you mind running `pnpm run lint`?

packages/worker/src/workflow/bundler.ts
```diff
@@ -218,7 +229,7 @@ exports.importInterceptors = function importInterceptors() {
       // Ignore the "node:" prefix if any.
       const module: string = data.request?.startsWith('node:')
         ? data.request.slice('node:'.length)
-        : data.request ?? '';
+        : (data.request ?? '');
```
chris-olszewski: Linter doesn't like this change
```suggestion
        : data.request ?? '';
```

wgoodall01: fixed—we have different prettier versions :)

wgoodall01:
Just fixed the lint issues and rebased against the current main.

mjameswh (approved):

mjameswh:
Thanks @wgoodall01 for the report, investigation and fix. This is highly appreciated.

PR #2183, opened 2026-07-09
fix(langsmith): install workflow-bundle rewrites via compiler.webpack

DABH:
`LangSmithPlugin.configureBundler` installs its Workflow-bundle rewrites (the langsmith `node:*` → `.browser.cjs` redirects and a `DefinePlugin`) via `require('webpack')`. webpack is a dependency of `@temporalio/worker`, not of this plugin, so under a non-hoisting layout such as pnpm `hoist=false`, which `samples-typescript` uses, `require` throws `MODULE_NOT_FOUND`, which the helpers silently swallow. See initial CI failures in temporalio/samples-typescript#480.

This fix installs the rewrites from a single plugin that reads webpack off `compiler.webpack` instead of importing it - no `webpack` dependency, always the exact instance running the bundle, should work under any `node_modules` layout.

Verified by reproducing temporalio/samples-typescript#480 in isolation (pnpm `hoist=false`, webpack 5.107.2, langsmith 0.7.17): fails on `main`, compiles cleanly with this change.

### DIFF
--- a/contrib/langsmith/src/plugin.ts
+++ b/contrib/langsmith/src/plugin.ts
@@ -37,7 +37,7 @@ type WebpackConfiguration = Parameters<NonNullable<BundleOptions['webpackConfigH
 // Bare identifier (not dotted) so DefinePlugin substitutes textually.
 const CONFIG_GLOBAL = '__TEMPORAL_LANGSMITH_CONFIG__';
 
-// Bare builtin name added to `ignoreModules`; see {@link aliasAsyncHooks}.
+// Bare builtin name added to `ignoreModules`; see {@link bundleRewritesPlugin}.
 const ASYNC_HOOKS_MODULE = 'async_hooks';
 
 /**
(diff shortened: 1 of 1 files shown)


contrib/langsmith/src/plugin.ts
```diff
@@ -200,55 +201,62 @@ export class LangSmithPlugin extends SimplePlugin {
...
+  webpack: WebpackApi;
 }
 
 /**
- * Redirect langsmith's node-only CJS utilities to its `.browser.cjs` siblings so
- * the workflow bundle keeps node builtins out of the isolate. Scoped to langsmith
- * so a user's own `fs.cjs` is never rewritten.
+ * A webpack plugin that installs the three Workflow-bundle rewrites LangSmith needs:
+ *
+ *  - redirect `node:async_hooks` to the isolate-safe workflow-interceptor shim.
```
chris-olszewski, line 231: Outside of the scope of this PR, but it might be worthwhile to look into the `AsyncLocalStorage` we provide that is hooked up to workflow lifecycle instead of handrolling a shim: https://github.com/temporalio/sdk-typescript/blob/main/packages/worker/src/workflow/vm-shared.ts#L156

chris-olszewski (approved):

chris-olszewski:
Would also be good to get a changelog entry for this

PR #2186, opened 2026-07-09
AI-231: add SandboxAgent support to the OpenAI Agents plugin

xumaple:
This PR adds experimental `SandboxAgent` support to the OpenAI Agents plugin. Sandbox operations run as Activities, and serialized session state allows any compatible Worker to resume the backend session.

It also adds Worker-side hosted-tool credential injection, allowlisted Worker environment references for sandbox manifests, and sandbox tracing spans. Credentials resolved on the Worker stay out of Workflow history.

Note: this updates the `@openai/agents-core` and `@openai/agents-openai` peer dependencies to `~0.14.3`.

In regards to keeping secrets out of workflow history, things I deliberately did not do:
- Parse model response objects to re-censor any secrets that the model provider accidentally allowed through instead of censoring themselves. This is their responsibility, not ours, and we don't have a good way of doing it safely/comprehensively/extendibly.
- Take away paths that were not forced on the user. As in, if there is one path that does leak secrets into workflow history, but the user can find a workaround that doesn't, then I considered this a non-issue for the time being.
- Any form of parsing/gating to stop a user from intentionally placing secrets where they shouldn't. That is a user's responsibility.


contrib/openai-agents/src/worker/sandbox-provider.ts
```diff
@@ -0,0 +1,492 @@
...
+      },
+
+      [`${n}${SANDBOX_CLIENT_SERIALIZE_SESSION_STATE_SUFFIX}`]: async (
+        input: SandboxSerializeSessionStateInput
+      ): Promise<Record<string, unknown>> => {
+        const session = await this.session(input.state);
+        const providerState = this.client.serializeSessionState
+          ? await this.client.serializeSessionState(session.state, input.options)
+          : await this.providerState(session.state);
+        return { sessionId: input.state.sessionId, providerState };
```
brianstrauch: `client.serializeSessionState()` may mutate SDK envelope fields on `session.state` (for example, snapshot and snapshot-fingerprint fields), but this result returns only the provider state. The Workflow-side session therefore keeps its older envelope. On a later cache miss, `toSdkStateRecord()` spreads `providerState` first and then the stale envelope fields, so the stale values overwrite the newly serialized ones. Snapshot-backed clients can consequently resume an older snapshot after workspace changes. Could this Activity return the refreshed serialized envelope as well, so `TemporalSandboxClient.serializeSessionState()` can update the Workflow-side `state` before returning the provider state to the Agents SDK?

contrib/openai-agents/src/worker/sandbox-provider.ts
```diff
@@ -0,0 +1,492 @@
...
+    if (!this.client.resume) {
+      throw ApplicationFailure.create({
+        message: `Sandbox backend '${this.name}' does not support resume(); session '${payload.sessionId}' cannot be recovered on this Worker.`,
+        type: 'SandboxOperationUnsupported',
+        nonRetryable: true,
+      });
+    }
+    const resume = this.client.resume.bind(this.client);
+    const resumePromise = (async (): Promise<SandboxSession> => {
+      const session = await resume(await this.rehydrateState(payload));
```
brianstrauch, line 223: `resume()` is not sufficient for process-local handles. Sandbox operations are separate Activities and may be dispatched to different Worker processes, so an `exec()` that returns a live process/session ID can run on Worker A while the following `writeStdin()` runs here on Worker B. For `UnixLocalSandboxClient`, the active-process table is in memory and is not reconstructed by `resume()`, even though this adapter forwards `supportsPty() === true`; the follow-up then reports that the session was not found. Please either provide Worker affinity for process-bound operations or report PTY/background-process support only for providers whose handles survive resume on another Worker.

brianstrauch (commented):
Two correctness issues remain around durable sandbox session recovery and process-local handles.

contrib/openai-agents/src/workflow/sandbox-client.ts
```diff
@@ -0,0 +1,187 @@
...
+ * the Worker.
+ *
+ * Create instances via {@link temporalSandboxClient}.
+ */
+export class TemporalSandboxClient implements SandboxClient<SandboxClientOptions, TemporalSandboxSessionState> {
+  readonly backendId: string;
+  private readonly _name: string;
+  private readonly _config: ActivityOptions;
+
+  constructor(name: string, options?: TemporalSandboxClientOptions) {
```
JasonSteving99: I think it probably makes more sense to pass in the base sdk `ActivityOptions` type like the [python impl does](https://github.com/temporalio/sdk-python/blob/dd76ea98550ee43cfb2068f4700feb994b5cb247/temporalio/contrib/openai_agents/sandbox/_temporal_sandbox_client.py#L55) rather than redeclaring a subset of the fields in `TemporalSandboxClientOptions`

contrib/openai-agents/src/worker/sandbox-provider.ts
xumaple: Fixed. The Activity now returns the refreshed envelope, and the Workflow updates the existing state in place before returning the Temporal provider record to the Agents SDK. Added coverage for refreshed and removed fields, Worker-secret references, cache-miss resume, and approval interruption.

contrib/openai-agents/src/workflow/sandbox-client.ts
xumaple: Changed this to accept native `ActivityOptions` directly. Omitted config gets the five-minute start-to-close default; supplied config is passed through as-is.

contrib/openai-agents/src/worker/sandbox-provider.ts
xumaple: Documented this limitation. The plugin cannot tell whether process handles survive `resume` because the current sandbox API does not expose that capability, so behavior is unchanged for now; the long-term fix is an upstream portability signal.

CHANGELOG.md
```diff
@@ -57,6 +59,9 @@ to docs, or any other relevant information.
 
 ### Added
 
+- **Experimental**: `@temporalio/openai-agents` can run OpenAI Agents `SandboxAgent`s as Temporal Workflows. SandboxAgent
```
brianstrauch, line 62:
```suggestion
- **Experimental**: `@temporalio/openai-agents` can run OpenAI Agents `SandboxAgent`s as Temporal Workflows. `SandboxAgent`
```

brianstrauch (approved):

THardy98 (approved):

PR #2192, opened 2026-07-10
feat(ai-sdk): upgrade to Vercel AI SDK v7

brianstrauch:
## What was changed

Breaking upgrade of the experimental `@temporalio/ai-sdk` package from `ai@6` (provider spec V3) to `ai@7` (provider spec V4).

### Dependencies & runtime
- Peers bump to `ai ^7`, `@ai-sdk/provider ^4`, `@ai-sdk/mcp ^2`; dev deps add `@ai-sdk/openai ^4` and `@ai-sdk/otel ^1`.
- `ai@7` is ESM-only and requires Node >= 22. The package `engines` moves to `>= 22.12.0` (first unflagged `require(esm)` in the 22.x line — the CJS build loads `ai` through it). The workspace-wide test run on the Node 20 CI leg **skips this package by design** via the new `scripts/run-tests.mjs` guard.

### V4 provider spec
- `TemporalProvider` / `TemporalLanguageModel` / `TemporalEmbeddingModel` now implement `ProviderV4` / `LanguageModelV4` / `EmbeddingModelV4` (`specificationVersion: 'v4'`); `transcriptionModel` dropped (optional in V4).
- The streaming accumulator passes through the new `reasoning-file`, `custom`, and `tool-approval-request` parts (stream-part and content shapes are identical), explicitly ignores `raw` chunks, and throws a retryable `ApplicationFailure` when a stream emits an `error` and ends without a `finish` part.
- The new V4 `reasoning` effort option and richer `responseFormat` serialize through activity args with no code changes.

### Workflow subpath export
- New workflow-safe entry point `@temporalio/ai-sdk/workflow` exporting `temporalProvider`, the model classes, and `TemporalMCPClient`. Workflow code should import from it instead of the package barrel, whose worker-side re-exports pull bundler-disallowed modules (`@temporalio/activity`, `@temporalio/client`, `crypto`).
- Polyfill installation is now an explicit, idempotent `installPolyfills()` (no import side effects); the workflow entry point calls it at load. The old `@temporalio/ai-sdk/lib/load-polyfills` deep import is no longer exported.
- Import order matters: the workflow entry point must be imported **before** `ai`, because `ai@7` evaluates `class ... extends TransformStream` at module scope.

### MCP
- Stable `createMCPClient` / `MCPClient` names; removed the nonexistent `vercel.ai.validator` schema marker (the `validate: undefined` spread is what `isSchema()` actually requires and is kept, now documented).

### Tests
- Mocks moved to V4; `stepCountIs` → `isStepCount`, `system` → `instructions`, middleware to v7's `LanguageModelMiddleware`.
- Telemetry now uses `@ai-sdk/otel`'s supported `OpenTelemetry` integration; spans follow GenAI semantic conventions (`invoke_agent <modelId>`) and compose with `@temporalio/interceptors-opentelemetry` in a single trace (workflow/activity spans + AI-operation spans).

## Compatibility notes

- **Workflow histories recorded under `ai@6` should be drained on v6 workers before upgrading** (the package is `@experimental`). Payload shapes are additive-compatible, but `ai@7`'s agent loop may change the number/order of activity invocations in multi-step workflows.
- Requires a worker containing the `const`/`let`/`var` workflow-bundle patch (#2169, unreleased): with webpack >= 5.108 and older workers, the module-cache patch silently fails and sandbox polyfills are lost for all workflow runs after the first one per V8 context.

## How it was verified

- Full AVA suite green on Node 24 (mock provider), including MCP schema round-trip and telemetry span assertions; lint/prettier clean.
- End-to-end against a local OpenAI-compatible stub: the `samples-typescript/ai-sdk` sample (updated separately) ran haiku/tools/middleware/mcp workflows twice each on one worker, on both `ai@7.0.2` and `ai@7.0.20`.


contrib/ai-sdk/src/workflow.ts
```diff
@@ -0,0 +1,18 @@
...
+// Workflow code should import from '@temporalio/ai-sdk/workflow' rather than the package
+// root: the root barrel also re-exports the worker-side activities and plugin, whose
+// transitive imports (`@temporalio/activity`, `@temporalio/client`, `crypto`, ...) are
+// disallowed by the workflow bundler.
+import { installPolyfills } from './load-polyfills';
+
+// Install the Web-API polyfills the AI SDK needs inside the sandbox. This runs when the module
+// loads so the globals exist before the `ai` package is evaluated — import this entry point
+// before `ai` in workflow code.
+installPolyfills();
```
DABH: This installs the Web-API polyfills as an import-time side effect, and it must run before `ai` is evaluated (ai@7 defines `class ... extends TransformStream` at module scope). Nothing enforces this: correctness rests on every user placing the `/workflow` import physically above their `ai` import in every workflow file.

In your test you do `// eslint-disable-next-line import/order` to stop the linter reordering them (`__tests__/workflows/ai-sdk.ts:4`) so even a linter could break this, feels fragile.

`@temporalio/openai-agents` already does the fix: `configureBundler` prepends the polyfill module to the webpack entry array (`contrib/openai-agents/src/worker/plugin.ts:134-154`), which the TS reference doc prescribes. So I suggest prepending `load-polyfills` to the bundle entry in `AiSdkPlugin.configureBundler`, which should make import order irrelevant.

contrib/ai-sdk/src/activities.ts
```diff
@@ -180,22 +183,40 @@ export function createActivities(provider: ProviderV3, mcpClientFactories?: McpC
...
             break;
         }
       }
 
+      if (streamError !== undefined && !sawFinish) {
+        throw ApplicationFailure.retryable(
+          `Model stream emitted an error and ended without a finish part: ${streamError}`
+        );
+      }
+
```
DABH: I think there may be one or two potential issues here:

(a) The throw is gated on streamError !== undefined && !sawFinish, so a stream that emits an error part and a finish part returns a normal successful ModelResponse.  The provider-reported error is dropped from the assembled result (it's only on the side channel, which the workflow-side reconstruction in `provider.ts:128-153` never inspects), which means that a genuinely failed stream can surface as success.

(b) When it does throw, it's always `ApplicationFailure.retryable`, regardless of whether the underlying error is deterministic. Suggest fixing so that the code fails whenever an `error` part is seen (or fold it into `finishReason: 'error'`); confirm the retryable default is intentional given `part.error` is untyped. (The current code could be your intent and I think `doGenerate` works similarly right now, so this is more just making sure that this is the expected behavior.)

contrib/ai-sdk/src/__tests__/test-ai-sdk.ts
```diff
@@ -91,15 +90,15 @@ export class TestModel implements LanguageModelV3 {
     return result.value;
   }
 
-  doStream(_options: LanguageModelV3CallOptions): PromiseLike<LanguageModelV3StreamResult> {
+  doStream(_options: LanguageModelV4CallOptions): PromiseLike<LanguageModelV4StreamResult> {
```
DABH, line 93: The streaming accumulator has zero test coverage - `TestModel.doStream` throws 'Streaming not supported.', and nothing else exercises `invokeModelStreaming`.  Can you add a `TestModel` that yields a scripted V4 stream-part sequence including an error-without-finish case, asserting the assembled result and the throw?

contrib/ai-sdk/src/__tests__/test-ai-sdk.ts
DABH, line 1: Can we possibly add a `Replayer` test over a recorded multi-step (tools/MCP) history captured under ai@7?  Currently looks like we have no replayer tests which would be nice to add.

DABH (changes requested):

brianstrauch:
Took all suggestions, thanks @DABH

DABH (approved):

pnpm-workspace.yaml
```diff
@@ -54,6 +54,10 @@ overrides:
   'minimatch@>=7.0.0 <7.4.8': '>=7.4.8'
   serialize-javascript: '>=7.0.3'
   handlebars: '>=4.7.9'
+  # @ai-sdk/otel pins @opentelemetry/api to exactly 1.9.1, which would split the
+  # workspace's OTel dependency graph into instances keyed against both 1.9.0 and
+  # 1.9.1 (structurally incompatible types, e.g. Resource). Keep a single instance.
+  '@ai-sdk/otel>@opentelemetry/api': 1.9.0
```
chris-olszewski, line 60: Outside the scope of this PR, but worth looking into bumping this across the repo.

chris-olszewski (approved):

PR #2193, opened 2026-07-10
Add patch activation callback

Sushisource:
## What was changed

Added an experimental `WorkerOptions.patchActivationCallback` that lets a worker decide whether the first newly encountered, non-replay `patched` call activates a patch.

The callback receives frozen workflow information and the patch ID. Its strict boolean decision is memoized for the workflow run. Declined patches return `false` without recording a marker, while replay, existing history markers, and deprecated patches retain their existing behavior.

Because Workflow code normally executes in worker threads while `patched` is synchronous, the implementation uses a scoped synchronous worker-thread bridge based on `SharedArrayBuffer` and `Atomics`. The callback closure stays on the Worker side and is never serialized or evaluated from source.

This ports the behavior introduced by https://github.com/temporalio/sdk-ruby/pull/481.

## Why?

This allows workflow changes using `patched` to roll out gradually without every new-code worker immediately recording a marker that older workers cannot replay.

## Testing

- Worker, Workflow, and test TypeScript builds
- Focused runtime suite: 7 cases
- Live rolling-worker integration suite: 2 cases
- Focused ESLint and Prettier checks
- Diff and sdk-core submodule cleanliness checks

### DIFF
--- a/packages/worker/src/workflow/patch-activation-callback.ts
+++ b/packages/worker/src/workflow/patch-activation-callback.ts
@@ -1,0 +1,122 @@
+import { IllegalStateError, TypedSearchAttributes, type SearchAttributePair } from '@temporalio/common';
+import type { WorkflowInfo } from '@temporalio/workflow';
+import { createUnsafeRandomSource } from '@temporalio/workflow/lib/random-helpers';
+import type { PatchActivationCallback, PatchActivationInput } from '../worker-options';
+
+// ts-prune-ignore-next (used by the workflow Worker thread entry point)
+export const PATCH_ACTIVATION_CALLBACK_BUFFER_SIZE = 64 * 1024;
+export const PATCH_ACTIVATION_CALLBACK_HEADER_SIZE = 3 * Int32Array.BYTES_PER_ELEMENT;
+
+export const enum PatchActivationCallbackStatus {
+  Pending = 0,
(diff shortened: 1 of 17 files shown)


packages/worker/src/workflow/patch-activation-callback.ts
```diff
@@ -0,0 +1,122 @@
...
+export function completePatchActivationCallback(resultBuffer: SharedArrayBuffer): void {
+  const header = getHeader(resultBuffer);
+  Atomics.store(header, 0, 1);
+  Atomics.notify(header, 0);
+}
+
+// ts-prune-ignore-next (used by the workflow Worker thread entry point)
+export function waitForPatchActivationCallbackResult(resultBuffer: SharedArrayBuffer): boolean {
+  const header = getHeader(resultBuffer);
+  Atomics.wait(header, 0, PatchActivationCallbackStatus.Pending);
```
cconstable, line 112: This is waiting on user-supplied logic right? So what happens if it hangs and never returns? I'm guessing the workflow times out but does that cause other things running on this worker to stall as well? Might be fine but worth checking.

Sushisource: Good point can add a test

cconstable (approved):

PR #2194, opened 2026-07-10
Support Workflow Update as a Nexus Operation

Evanthx:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -1033,7 +1033,7 @@ export class WorkflowClient extends BaseClient {
     const dataConverter = this.dataConverter;
     const context = this.workflowSerializationContext(input.workflowExecution.workflowId!);
     const updateId = input.options?.updateId ?? randomUUID();
-    const internalOptions = (input.options as InternalWorkflowUpdateOptions | undefined)?.[
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[
       InternalWorkflowUpdateOptionsSymbol
     ];
     return {
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -1033,9 +1033,7 @@ export class WorkflowClient extends BaseClient {
     const dataConverter = this.dataConverter;
     const context = this.workflowSerializationContext(input.workflowExecution.workflowId!);
     const updateId = input.options?.updateId ?? randomUUID();
-    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[
-      InternalWorkflowUpdateOptionsSymbol
-    ];
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[InternalWorkflowUpdateOptionsSymbol];
     return {
       namespace: this.options.namespace,
       workflowExecution: input.workflowExecution,
@@ -1108,9 +1106,7 @@ export class WorkflowClient extends BaseClient {
       this.rethrowUpdateGrpcError(err, 'Workflow Update failed', input.workflowExecution);
     }
     await visit(response, walkUpdateWorkflowExecutionResponse, extstoreInboundOptions(externalStorage));
-    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[
-      InternalWorkflowUpdateOptionsSymbol
-    ];
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[InternalWorkflowUpdateOptionsSymbol];
     if (internalOptions != null) {
       // Capture the link the server attached to the Update response so the Nexus helper can add it
       // as a handler link. Older servers leave it unset.
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -1108,7 +1108,7 @@ export class WorkflowClient extends BaseClient {
       this.rethrowUpdateGrpcError(err, 'Workflow Update failed', input.workflowExecution);
     }
     await visit(response, walkUpdateWorkflowExecutionResponse, extstoreInboundOptions(externalStorage));
-    const internalOptions = (input.options as InternalWorkflowUpdateOptions | undefined)?.[
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[
       InternalWorkflowUpdateOptionsSymbol
     ];
     if (internalOptions != null) {
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -111,7 +111,12 @@ import { BaseClient, defaultBaseClientOptions } from './base-client';
 import { mapAsyncIterable } from './iterators-utils';
 import { WorkflowUpdateStage, encodeWorkflowUpdateStage } from './workflow-update-stage';
 import type { InternalWorkflowHandle, InternalWorkflowSignalInput, InternalWorkflowStartOptions } from './internal';
-import { InternalWorkflowSignalOptionsSymbol, InternalWorkflowStartOptionsSymbol } from './internal';
+import {
+  InternalWorkflowSignalOptionsSymbol,
+  InternalWorkflowStartOptionsSymbol,
+  type InternalWorkflowUpdateOptions,
+  InternalWorkflowUpdateOptionsSymbol,
+} from './internal';
 import { adaptWorkflowClientInterceptor } from './interceptor-adapters';
 
 const UpdateWorkflowExecutionLifecycleStage = temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage;
@@ -1028,6 +1033,9 @@ export class WorkflowClient extends BaseClient {
     const dataConverter = this.dataConverter;
     const context = this.workflowSerializationContext(input.workflowExecution.workflowId!);
     const updateId = input.options?.updateId ?? randomUUID();
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions | undefined)?.[
+      InternalWorkflowUpdateOptionsSymbol
+    ];
     return {
       namespace: this.options.namespace,
       workflowExecution: input.workflowExecution,
@@ -1045,6 +1053,9 @@ export class WorkflowClient extends BaseClient {
           name: input.updateName,
           args: { payloads: await encodeToPayloadsWithContext(dataConverter, context, input.args) },
         },
+        requestId: internalOptions?.requestId,
+        completionCallbacks: internalOptions?.completionCallbacks,
+        links: internalOptions?.links,
       },
     };
   }
@@ -1097,6 +1108,17 @@ export class WorkflowClient extends BaseClient {
       this.rethrowUpdateGrpcError(err, 'Workflow Update failed', input.workflowExecution);
     }
     await visit(response, walkUpdateWorkflowExecutionResponse, extstoreInboundOptions(externalStorage));
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions | undefined)?.[
+      InternalWorkflowUpdateOptionsSymbol
+    ];
+    if (internalOptions != null) {
+      // Capture the link the server attached to the Update response so the Nexus helper can add it
+      // as a handler link. Older servers leave it unset.
+      internalOptions.responseLink = response.link ?? undefined;
+      // Capture the terminal outcome (if any) so the Nexus helper can distinguish an already-completed
+      // Update (return a synchronous result) from one that is merely accepted (return async).
+      internalOptions.outcome = response.outcome ?? undefined;
+    }
     return {
       updateId: request.request!.meta!.updateId!,
 
--- a/packages/nexus/src/link-converter.ts
+++ b/packages/nexus/src/link-converter.ts
@@ -5,6 +5,7 @@ import { temporal } from '@temporalio/proto';
 const { EventType } = temporal.api.enums.v1;
 type TemporalLink = temporal.api.common.v1.ILink;
 type WorkflowEventLink = temporal.api.common.v1.Link.IWorkflowEvent;
+type WorkflowLink = temporal.api.common.v1.Link.IWorkflow;
 type NexusOperationLink = temporal.api.common.v1.Link.INexusOperation;
 type EventReference = temporal.api.common.v1.Link.WorkflowEvent.IEventReference;
 type RequestIdReference = temporal.api.common.v1.Link.WorkflowEvent.IRequestIdReference;
@@ -20,6 +21,7 @@ const REQUEST_ID_REFERENCE_TYPE = 'RequestIdReference';
 // fullName isn't part of the generated typed unfortunately.
 const WORKFLOW_EVENT_TYPE: string = (temporal.api.common.v1.Link.WorkflowEvent as any).fullName.slice(1);
 const NEXUS_OPERATION_TYPE: string = (temporal.api.common.v1.Link.NexusOperation as any).fullName.slice(1);
+const WORKFLOW_TYPE: string = (temporal.api.common.v1.Link.Workflow as any).fullName.slice(1);
 
 export function convertTemporalLinkToNexusLink(link: TemporalLink): NexusLink {
   if (link.workflowEvent != null) {
@@ -73,6 +75,48 @@ export function convertWorkflowEventLinkToNexusLink(we: WorkflowEventLink): Nexu
     url,
     type: WORKFLOW_EVENT_TYPE,
   };
+}
+
+/**
+ * Converts a plain Workflow link (as opposed to a {@link WorkflowEventLink}) to a Nexus link.
+ *
+ * Used to point at a Workflow when there is no specific history event to reference, e.g. when an
+ * UpdateWorkflow-backed Nexus operation fails validation and no Update Accepted event exists. The
+ * resulting URL is the workflow history URL without a reference query; the link is distinguished
+ * from a workflow event link by its {@link NexusLink.type}.
+ */
+export function convertWorkflowLinkToNexusLink(wl: WorkflowLink): NexusLink {
+  if (!wl.namespace || !wl.workflowId) {
+    throw new TypeError(
+      `Missing required fields: namespace or workflowId (namespace=${wl.namespace}, workflowId=${wl.workflowId})`
+    );
+  }
+  const url = new URL(
+    `temporal:///namespaces/${encodeURIComponent(wl.namespace)}/workflows/${encodeURIComponent(
+      wl.workflowId
+    )}/${encodeURIComponent(wl.runId ?? '')}/history`
+  );
+
+  return {
+    url,
+    type: WORKFLOW_TYPE,
+  };
+}
+
+/**
+ * Converts a Temporal common Link into a Nexus link, preferring a {@link WorkflowEventLink} when
+ * present and falling back to a plain {@link WorkflowLink}. The fallback safely points at a workflow
+ * for cases where a Nexus operation fails before a history event exists (e.g. UpdateWorkflow
+ * validation failure).
+ */
+export function convertCommonLinkToNexusLink(link: TemporalLink): NexusLink {
+  if (link.workflowEvent != null) {
+    return convertWorkflowEventLinkToNexusLink(link.workflowEvent);
+  }
+  if (link.workflow != null) {
+    return convertWorkflowLinkToNexusLink(link.workflow);
+  }
+  throw new TypeError('Invalid Temporal link - workflow and workflowEvent fields were null');
 }
 
 export function convertNexusOperationLinkToNexusLink(opLink: NexusOperationLink): NexusLink {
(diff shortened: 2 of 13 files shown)


packages/test/src/test-nexus-update-operation.ts
```diff
@@ -0,0 +1,300 @@
...
+    t.is(result.count, 5);
+  });
+});
+
+// An Update that has not yet completed when it is accepted returns asynchronously; its result is
+// delivered to the caller via the Nexus completion callback attached to the Update request. This
+// requires server-side support for delivering Workflow Update completion callbacks to Nexus, which
+// the bundled ephemeral server (Temporal 1.31.2) does not yet provide (workflow-run completion
+// callbacks do work there). Enable once the test server delivers Update completion callbacks.
+test.skip('UpdateWorkflow Nexus operation - pending update completes asynchronously via callback', async (t) => {
```
Quinn-With-Two-Ns: Same comment as dotnet, this should be enabled

packages/nexus/src/workflow-helpers.ts
```diff
@@ -516,6 +737,22 @@ export class TemporalOperationHandler<I, O> implements nexus.OperationHandler<I,
         }
         await this.cancelWorkflowRunHandler(ctx, { workflowId: opToken.wid });
         return;
+      case OperationTokenType.UPDATE_WORKFLOW: {
+        let updateToken;
+        try {
+          updateToken = loadUpdateWorkflowOperationToken(token);
```
Quinn-With-Two-Ns: Why are we reparsing the token, we already have `opToken`?

packages/nexus/src/workflow-helpers.ts
```diff
@@ -469,13 +654,47 @@ export interface CancelWorkflowRunOptions {
...
+export interface CancelWorkflowUpdateOptions {
+  /**
+   * The ID of the workflow whose Update backs the Nexus Operation that is being canceled.
+   */
+  readonly workflowId: string;
+
+  /**
+   * The Run ID extracted from the operation token. May be empty if the run was not pinned.
+   */
+  readonly runId: string;
```
Quinn-With-Two-Ns, line 711: The run ID should be know here since the server returns the run ID on the update response

packages/client/src/workflow-client.ts
```diff
@@ -994,6 +999,9 @@ export class WorkflowClient extends BaseClient {
     const dataConverter = this.dataConverter;
     const context = this.workflowSerializationContext(input.workflowExecution.workflowId!);
     const updateId = input.options?.updateId ?? randomUUID();
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions | undefined)?.[
```
VegetarianOrc: Casting to `InternalWorkflowUpdateOptions` is sufficient here. `internalOptions` stays the same type b/c the value at the `InternalWorkflowUpdateOptionsSymbol` is an optional property.

```suggestion
    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[
```

packages/client/src/workflow-client.ts
```diff
@@ -1048,6 +1059,17 @@ export class WorkflowClient extends BaseClient {
     } catch (err) {
       this.rethrowUpdateGrpcError(err, 'Workflow Update failed', input.workflowExecution);
     }
+    const internalOptions = (input.options as InternalWorkflowUpdateOptions | undefined)?.[
```
VegetarianOrc:
```suggestion
    const internalOptions = (input.options as InternalWorkflowUpdateOptions)[
```

packages/nexus/src/workflow-helpers.ts
```diff
@@ -315,6 +330,49 @@ export const TemporalOperationResult = {
...
+  /**
+   * Run ID of the Workflow to send the Update to. If unset, the Update is sent to the most recent
+   * run of the Workflow.
+   */
+  readonly runId?: string;
+
+  /**
+   * The Update definition (as returned by `defineUpdate`) or the Update name.
+   */
+  readonly update: UpdateDefinition<Ret, Args> | string;
```
VegetarianOrc: I think we should consider threading the `Name` type through to the `UpdateDefinition` like the WorkflowHandle

```ts
  executeUpdate<Ret, Args extends [], Name extends string = string>(
    def: UpdateDefinition<Ret, Args, Name> | string,
    options?: WorkflowUpdateOptions & { args?: Args }
  ): Promise<Ret>;
```

packages/nexus/src/workflow-helpers.ts
```diff
@@ -315,6 +330,49 @@ export const TemporalOperationResult = {
...
+   * Arguments to pass to the Update handler.
+   */
+  readonly args?: Args;
+
+  /**
+   * The Update lifecycle stage to wait for. Only {@link WorkflowUpdateStage.ACCEPTED} is supported
+   * for Nexus operations; any other value is rejected. Defaults to
+   * {@link WorkflowUpdateStage.ACCEPTED}.
+   */
+  readonly waitForStage?: WorkflowUpdateStage;
```
VegetarianOrc: Should we expose this at all if there's only one valid value and we default to it?

packages/nexus/src/link-converter.ts
```diff
@@ -75,6 +77,48 @@ export function convertWorkflowEventLinkToNexusLink(we: WorkflowEventLink): Nexu
...
+export function convertWorkflowLinkToNexusLink(wl: WorkflowLink): NexusLink {
+  if (!wl.namespace || !wl.workflowId) {
+    throw new TypeError(
+      `Missing required fields: namespace or workflowId (namespace=${wl.namespace}, workflowId=${wl.workflowId})`
+    );
+  }
+  const url = new URL(
+    `temporal:///namespaces/${encodeURIComponent(wl.namespace)}/workflows/${encodeURIComponent(
+      wl.workflowId
+    )}/${encodeURIComponent(wl.runId ?? '')}/history`
```
VegetarianOrc: If the `wl.runId` property is coalesced to `''` does the URL still resolve to a valid page on the UI?

packages/nexus/src/link-converter.ts
```diff
@@ -75,6 +77,48 @@ export function convertWorkflowEventLinkToNexusLink(we: WorkflowEventLink): Nexu
...
+ */
+export function convertCommonLinkToNexusLink(link: TemporalLink): NexusLink {
+  if (link.workflowEvent != null) {
+    return convertWorkflowEventLinkToNexusLink(link.workflowEvent);
+  }
+  if (link.workflow != null) {
+    return convertWorkflowLinkToNexusLink(link.workflow);
+  }
+  throw new TypeError('Invalid Temporal link - workflow and workflowEvent fields were null');
+}
```
VegetarianOrc: This function duplicates the above `convertTemporalLinkToNexusLink`. We should move the new case to the existing implementation and remove this one.

packages/nexus/src/token.ts
```diff
@@ -67,6 +91,38 @@ export function generateWorkflowRunOperationToken(namespace: string, workflowId:
...
+  if (!updateId) {
+    throw new TypeError('invalid update workflow token: missing update ID (uid)');
+  }
+
+  const token: UpdateWorkflowOperationToken = {
+    t: OperationTokenType.UPDATE_WORKFLOW,
+    ns: namespace,
+    wid: workflowId,
+    // `rid` is optional, so omit it if not present
+    ...(runId ? { rid: runId } : {}),
```
VegetarianOrc: I think this is sufficient and avoids the spread. Another thing to note is that `runId` is required in this function so this is only catching an empty string. Is that the intended behavior?

```suggestion
    rid: runId ? runId : undefined,
```

packages/test/src/test-nexus-update-operation.ts
```diff
@@ -0,0 +1,300 @@
+/**
+ * Integration tests for UpdateWorkflow-backed Nexus operations
+ * ({@link TemporalNexusClient.updateWorkflow}).
+ *
+ * These mirror the acceptance criteria of the Go SDK's TestNexusUpdateWorkflowOperation suite:
+ * validation failures surface as failed operations, valid updates succeed, updates sharing an
+ * Update ID are idempotent, and updates against a completed workflow fail.
```
VegetarianOrc:
```suggestion
```

packages/test/src/test-nexus-update-operation.ts
```diff
@@ -0,0 +1,300 @@
...
+  let done = false;
+
+  workflow.setHandler(
+    addUpdate,
+    async (amount: number, sleepMs: number): Promise<UpdateAddOutput> => {
+      counter += amount;
+      const snapshot = counter;
+      if (sleepMs > 0) {
+        await workflow.sleep(sleepMs);
+      }
```
VegetarianOrc, line 133: Is this sleep necessary? I'm wondering if there's a potential flake being hidden by this.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -363,6 +421,26 @@ export interface TemporalNexusClient {
...
+   * delivered via the callback. If the Update has already completed by the time it is accepted (for
+   * example a retried request with the same Update ID, or an Update that completes immediately), a
+   * synchronous result is returned instead; if that completed Update failed (e.g. validation
+   * rejection, which is non-retryable), it surfaces as a failed Nexus Operation.
+   *
+   * @experimental Nexus support in Temporal SDK is experimental.
+   */
+  updateWorkflow<Ret, Args extends any[] = []>(
+    options: NexusUpdateWorkflowOptions<Ret, Args>
+  ): Promise<TemporalOperationResult<Ret>>;
```
VegetarianOrc: I think we should put this function onto the Nexus specific workflow handle to match the `signal` method.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -427,6 +505,113 @@ class TemporalNexusClientImpl implements TemporalNexusClient {
...
+    const ctx = this.startOperationContext;
+
+    const waitForStage = options.waitForStage ?? WorkflowUpdateStage.ACCEPTED;
+    if (waitForStage !== WorkflowUpdateStage.ACCEPTED) {
+      // Non-retryable misconfiguration: surface as a failed operation rather than retrying forever.
+      throw new nexus.OperationError(
+        'failed',
+        'Nexus operation Workflow Updates only support the ACCEPTED wait stage for async updates'
+      );
+    }
```
VegetarianOrc: Similar to the above comment, I think we should consider removing this check and just always using ACCEPTED.

packages/nexus/src/workflow-helpers.ts
Evanthx: Good catch, now using opToken

packages/nexus/src/workflow-helpers.ts
Evanthx: Now using handle.workflowRunId

packages/nexus/src/workflow-helpers.ts
Evanthx: Done

packages/nexus/src/workflow-helpers.ts
Evanthx: Good point, removed it

packages/nexus/src/link-converter.ts
Evanthx: Point. RunId is now required

packages/test/src/test-nexus-update-operation.ts
Evanthx: workflow.sleep suspends the update handler after it accepts, that is what forces the async completion callback path. So when it sleeps we get the async test, when it doesn't sleep it's a sync test.

packages/nexus/src/workflow-helpers.ts
Evanthx: Done

packages/nexus/src/workflow-helpers.ts
Evanthx: Moved!

packages/nexus/src/token.ts
```diff
@@ -39,20 +39,44 @@ export interface WorkflowRunOperationToken extends OperationToken {
 }
 
 /**
- * OperationTokenType is used to identify the type of Operation token.
- * Currently, we only have one type of Operation token: WorkflowRun.
+ * An OperationToken that identifies an UpdateWorkflow operation.
+ *
+ * Unlike {@link WorkflowRunOperationToken}, it carries the run and update IDs so the update can be
+ * addressed for cancellation.
```
VegetarianOrc: Nit:
```suggestion
 * It carries the run and update IDs so the update can be
 * addressed for cancellation.
```

packages/nexus/src/workflow-helpers.ts
```diff
@@ -426,23 +489,157 @@ class TemporalNexusClientImpl implements TemporalNexusClient {
...
+}
+
+/**
+ * Tracks whether an async backing operation has already been started for a given Nexus Operation
+ * invocation, keyed by its {@link nexus.StartOperationContext}. The guard is shared between
+ * {@link TemporalNexusClient.startWorkflow} and {@link WorkflowHandle.update} (which lives on the
+ * handle rather than the client) because at most one async operation may back a single handler
+ * invocation, and both capture the same context object.
+ */
+const asyncOperationStartedByContext = new WeakMap<nexus.StartOperationContext, boolean>();
```
VegetarianOrc: I'm not sure this is the right approach here. Specifically I don't love this package level state.

One way we can explore is to refactor where `createWorkflowHandle` is defined so that an implementation exists on `TemporalNexusClientImpl` so any functions defined in there can capture the client and use the previous version of `withAsyncOperationStartReservation`.

Also open to other implementations, but again would like to avoid the the package level state.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -426,23 +489,157 @@ class TemporalNexusClientImpl implements TemporalNexusClient {
...
+      // The Update already reached a terminal outcome (e.g. a retried request with the same Update
+      // ID, or an immediate completion). Return a synchronous result; if it failed (validation
+      // rejection is non-retryable), surface it as a failed Nexus operation.
+      try {
+        const result = (await handle.result()) as Ret;
+        releaseAsyncOperationReservation(ctx);
+        return TemporalOperationResult.sync(result);
+      } catch (err) {
+        releaseAsyncOperationReservation(ctx);
+        throw new nexus.OperationError('failed', (err as Error).message, { cause: err as Error });
```
VegetarianOrc: I was just reviewing this on the Python implementation. In that implementation, we don't release the reservation in this case. I think that might be the right move so that the behavior is more predictable in that the flag is always set after an update. If we do clear the flag, then I _think_ we'd need to expose a way to check if the flag is set or not w/o erroring out, which doesn't sound like a great pattern.

Either way, we should make sure the SDKs are aligned on this detail.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -426,23 +489,157 @@ class TemporalNexusClientImpl implements TemporalNexusClient {
...
+    const wfHandle = client.workflow.getHandle(workflowId, runId) as unknown as {
+      startUpdate(
+        def: UpdateDefinition<Ret, Args> | string,
+        startUpdateOptions: WorkflowUpdateOptions & {
+          args?: Args;
+          waitForStage: typeof WorkflowUpdateStage.ACCEPTED;
+        } & InternalWorkflowUpdateOptions
+      ): Promise<WorkflowUpdateHandle<Ret>>;
+    };
+    const handle = await wfHandle.startUpdate(def, startUpdateOptions);
```
VegetarianOrc: I think we should do our best here to avoid these casts.

VegetarianOrc (changes requested):

packages/nexus/src/workflow-helpers.ts
Evanthx: Python has the same issue with multiple startUpdate overloads. I tried to emulate what is being done there, basically making an internal startUpdate that wraps the public one and hides the casts.

packages/nexus/src/workflow-helpers.ts
Evanthx: If I have this right, Python does release it via BaseException though (in that branch, not in main, in sdk-python/temporalio/nexus/_temporal_client.py, lines 401-402 ) - so while I did clean things up a bit so this matches Python's current behavior, I'm quite open to changing it to not clear the flag. But I figured I'd wait and make sure that's the behavior we wanted across SDKs before I changed that.

packages/nexus/src/workflow-helpers.ts
Evanthx: See how you feel about the change I made to this?

packages/nexus/src/workflow-helpers.ts
```diff
@@ -426,23 +529,108 @@ class TemporalNexusClientImpl implements TemporalNexusClient {
...
-      return await fn();
-    } catch (err) {
-      this.asyncOperationStarted = false;
-      throw err;
+    if (internalOptions.outcome != null) {
+      // The Update already reached a terminal outcome (e.g. a retried request with the same Update
+      // ID, or an immediate completion). Return a synchronous result; if it failed (validation
+      // rejection is non-retryable), surface it as a failed Nexus operation.
+      try {
+        const result = (await handle.result()) as Ret;
```
VegetarianOrc: I don't _think_ this cast is necessary.

VegetarianOrc (approved):
One minor nit, but look okay to me! Let's make sure we get @chris-olszewski or @mjameswh to take a close look as well.

packages/nexus/src/link-converter.ts
Evanthx: Missed this comment, but it is now fixed!

packages/nexus/src/workflow-helpers.ts
Evanthx: It wasn't, removed it

packages/nexus/src/workflow-helpers.ts
```diff
@@ -208,6 +250,13 @@ function createWorkflowHandle<T extends Workflow>(
         pushResponseLink(ctx, internalOptions.responseLink);
       }
     },
+
+    update<Ret, Args extends any[] = [], Name extends string = string>(
+      def: UpdateDefinition<Ret, Args, Name> | string,
+      options?: NexusUpdateWorkflowOptions<Args>
```
chris-olszewski: I think we need to do the whole no-arg vs arg differentiation here. Currently one could call an update that takes no args with args and it would pass type checking. Can probably add a test to verify.
```suggestion
    update<Ret, Args extends [any, ...any[]], Name extends string = string>(
        def: UpdateDefinition<Ret, Args, Name> | string,
        options: NexusUpdateWorkflowOptions<Args> & { readonly args: Args }
      ): Promise<TemporalOperationResult<Ret>>;
    update<Ret, Args extends [] = [], Name extends string = string>(
      def: UpdateDefinition<Ret, Args, Name> | string,
      options?: NexusUpdateWorkflowOptions<Args>
```

packages/nexus/src/workflow-helpers.ts
```diff
@@ -178,10 +210,20 @@ function pushResponseLink(ctx: nexus.StartOperationContext, responseLink: tempor
...
+ */
+type AsyncOperationStartReservation = <T>(fn: () => Promise<T>) => Promise<T>;
+const passThroughReservation: AsyncOperationStartReservation = (fn) => fn();
+
 function createWorkflowHandle<T extends Workflow>(
   ctx: nexus.StartOperationContext,
   workflowId: string,
-  runId?: string
+  runId?: string,
+  reserve: AsyncOperationStartReservation = passThroughReservation
```
chris-olszewski: Just because missing this `reserve` is pretty impactful, I would prefer we remove the default here so callers have to intentionally pass a noop if that is what they want.

IIUC `startWorkflow` should set a `() => throw new nexus.HandlerError('BAD_REQUEST', 'one async operation per operation handler');` as its `reserve` function since the start workflow operation will have already consumed the slot.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -54,6 +65,27 @@ export interface WorkflowHandle<T> {
...
+   * example a retried request with the same Update ID, or an Update that completes immediately), a
+   * synchronous result is returned instead; if that completed Update failed (e.g. validation
+   * rejection, which is non-retryable), it surfaces as a failed Nexus Operation.
+   *
+   * @experimental Nexus support in Temporal SDK is experimental.
+   */
+  update<Ret, Args extends any[] = [], Name extends string = string>(
+    def: UpdateDefinition<Ret, Args, Name> | string,
+    options?: NexusUpdateWorkflowOptions<Args>
+  ): Promise<TemporalOperationResult<Ret>>;
```
chris-olszewski: Might be worth to move `update` to a new interface that extends from `WorkflowHandle` so the handle returned by `startWorkflow` doesn't advertise `.update` as functionality even though it shouldn't be valid.

chris-olszewski (changes requested):
Some concerns around preventing misuse at the type level, but overall I think this makes sense to me.

packages/nexus/src/workflow-helpers.ts
Evanthx: Good idea - done!

packages/nexus/src/workflow-helpers.ts
Evanthx: Default is now gone and it is required

packages/nexus/src/workflow-helpers.ts
Evanthx: Done

THardy98 (commented):
Shape lgtm, appreciate the testing.

chris-olszewski (approved):

Issue #2188, opened 2026-07-09
[Bug] bundleWorkflowCode fails ("found 2 __webpack_module_cache__") when a workflow dependency ships its own webpack runtime


avolkov-dev:
## What are you really trying to do?

Bundle workflow code whose payload converter depends on the [`ejson`](https://www.npmjs.com/package/ejson) npm package — as shown in the official [`ejson` sample](https://github.com/temporalio/samples-typescript/tree/main/ejson).

## Describe the bug

Since the `reuseV8Context` isolation fix from #2170 (shipped in `1.20.1`, `1.19.1`, `1.18.2`, `1.17.5`, `1.16.3`, `1.15.1`, `1.14.2`), `bundleWorkflowCode` throws whenever the workflow bundle transitively includes a dependency that is itself a **pre-bundled webpack build** carrying its own webpack runtime.

`ejson` is one such package: `node_modules/ejson/index.js` is a committed webpack bundle, so it contains its own `var __webpack_module_cache__ = {}`. The workflow bundle therefore ends up with **two** declarations, and the new strict `=== 1` assertion fails.

This is a **regression**: the same code bundled fine before the patched versions listed above.

## Minimal reproduction

1. Check out the official `ejson` sample: https://github.com/temporalio/samples-typescript/tree/main/ejson
2. Bump `@temporalio/*` to `1.20.1` (or any of the patched versions above).
3. `npm start` — i.e. run `src/worker.ts`, which does:

   ```ts
   const worker = await Worker.create({
     workflowsPath: require.resolve('./workflows'),
     taskQueue: 'ejson',
     dataConverter: { payloadConverterPath: require.resolve('./payload-converter') },
   });
   ```

The worker bundles the workflows together with `./payload-converter`, which (via `ejson-payload-converter.ts`) does `import EJSON from 'ejson'`, pulling ejson's pre-bundled runtime into the workflow bundle.

## Actual error

```
Failed to patch the Workflow bundle: expected to find exactly one __webpack_module_cache__
declaration emitted by webpack, but found 2. Without this patch, Workflow isolation would be
broken. This is likely due to a change in webpack output; please report this at
https://github.com/temporalio/sdk-typescript/issues
    at WorkflowCodeBundler.createBundle (@temporalio/worker/lib/workflow/bundler.js:143:19)
    at async bundleWorkflowCode (@temporalio/worker/lib/workflow/bundler.js:350:12)
```

## Root cause

[...]

Issue #2170, opened 2026-07-07
[Bug] Workflow context leak across executions with Webpack >= 5.108.0 and reuseV8Context


mjameswh:
## Summary

We have identified a severe bug in the TypeScript SDK where Workflow isolation can be broken when both of the following are true:

- the Worker is using `reuseV8Context: true`;
- Workflow code is bundled with Webpack `>= 5.108.0`.

In this situation, module-level state in Workflow code may leak across Workflow executions. This can cause one Workflow execution to observe module-level state written by another Workflow execution, resulting in various, severe misbehaviors.

## Affected configuration

Known affected configuration:

- `@temporalio/worker` versions:
  - v1.20.0
  - v1.19.0
  - v1.18.0 and v1.18.1
  - v1.17.0 through v1.17.4
  - v1.16.0 through v1.16.2
  - v1.15.0
  - v1.14.0 and v1.14.1
  - All releases older than v1.14.0
- `reuseV8Context: true`
- Webpack `>= 5.108.0`

## Action Items

All Temporal TypeScript SDK users are advised to update to one of the following releases:

- v1.20.1 (recommended)
- v1.19.1
- v1.18.2
- v1.17.5
- v1.16.3
- v1.15.1
- v1.14.2

Users that are unable to update to v1.14.2 or more recent should force the `webpack` package to `~5.107.0`. Depending on your package manager, this might be done using a dependency override or by adding an explicit dependency from your own package to `'webpack': '~5.107.0'`. Note that pinning `webpack` could have security implications if security vulnerabilities are later uncovered in that version of webpack.

## Other recommendations

- **IMPORTANT**: Users who deployed or refreshed their dependencies in the last 14 days may have poisoned lockfile (i.e. include webpack 5.108.x with an affected version of the Temporal SDK). You are advised to refresh your dependencies and confirm that you have effectively upgraded to a known safe version of the Temporal SDK before redeploying your application.

- Deployments that apply "cooldown periods" (e.g. using PNPM's `minimumReleaseAge`) might face delayed exposure to this issue. You are strongly advised to either add a cooldown period override on your Temporal SDK dependencies _or_ to add a temporary override on `'webpack': '~5.107.0'` until expiration of the cooldown period.

- We do not advise blindly turning off the `reuseV8Context` option. The `reuseV8Context` executor significantly improves the performance and reduces the per-workflow-execution memory footprint. Disabling this option should only be done with appropriate performance monitoring.

## Impact Details

[...]

PR #2195, opened 2026-07-10
Make the Bundler's webpack-module-cache hijack more reliable

mjameswh:
## What was changed

Rewrite the Workflow Bundler's `__webpack_module_cache__` hijack logic (i.e. initializing it as a global variable, which is required by the Reusable V8 Context executor) to be more reliable.

More specifically:

- Replacement is now performed from within a dedicated Webpack plugin (`InjectWorkflowModuleCacheGlobalPlugin`) rather than by the bundler itself on the final generated bundle code. This ensures that the substitution is performed before execution of minimizers and similar that might alter webpack's bootstrap code in a way that would make it impossible to locate using a regex. This addresses the issue reported in [this comment](https://github.com/temporalio/sdk-typescript/issues/2170#issuecomment-4925636742).

- A marker is temporarily injected while generating the `__webpack_require__` function. That marker is used to reliably locate the `__webpack_module_cache__` variable declaration of the top-level module, vs `__webpack_module_cache__` variables that belongs to nested modules that have themselves been prebundled using webpack. This addresses the issue reported in #2188.

- More runtime checks have been added to ensure that eventual regressions, if ever, would not result in silent failure similar to the #2170 incident.

Also:

- Bump webpack to 5.108.4 (and pin it) to ensure our tests correctly exercise the const-cache output that triggered #2170.
- Add bundling tests covering the some of the problematic cases uncovered by this incident.
- Fix `@temporalio/strands-agents` webpack config injection logic to read webpack from `compiler.webpack` similar to the [langsmith fix](#2183).
- Fix ordering of versions in `CHANGELOG.md`, and add missing entries for versions published from release branches.

### DIFF
--- a/contrib/strands/src/plugin.ts
+++ b/contrib/strands/src/plugin.ts
@@ -158,27 +158,58 @@ export class StrandsPlugin extends SimplePlugin {
         (config.output ??= {}).asyncChunks = false;
         (config.optimization ??= {}).splitChunks = false;
 
-        // Reach webpack via the constructor of an existing plugin instance —
-        // the strands plugin doesn't list webpack as a direct dependency.
-        const existing = (config.plugins ?? [])[0] as
-          | { constructor: new (regex: RegExp, replacement: (data: { request: string }) => void) => unknown }
-          | undefined;
-        if (existing) {
-          const NormalModuleReplacementPlugin = existing.constructor;
-          const empty = require.resolve('./empty-module');
-          const ignorePattern =
-            /^(?:node:(?:fs\/promises|os|path|process|stream)|@modelcontextprotocol\/sdk\/client\/(?:stdio|sse|streamableHttp)\.js|eventsource-parser\/stream)$/;
-          // Cast through `unknown` because the strands plugin doesn't list
-          // webpack as a dependency and therefore lacks its types.
-          config.plugins = (config.plugins ?? []).concat(
-            new NormalModuleReplacementPlugin(ignorePattern, (data: { request: string }) => {
-              data.request = empty;
-            }) as unknown as never
-          );
-        }
+        const plugins = [...(config.plugins ?? [])];
+        plugins.push(emptyModuleRewritesPlugin() as (typeof plugins)[number]);
+        const merged = { ...config, plugins };
 
-        return prevHook ? prevHook(config) : config;
+        return prevHook ? prevHook(merged) : merged;
       },
     };
   }
+}
+
+/**
+ * The slice of webpack's runtime API these rewrites use, read off the running
+ * compiler (see {@link emptyModuleRewritesPlugin}).
+ */
+interface WebpackApi {
+  NormalModuleReplacementPlugin: new (
+    resourceRegExp: RegExp,
+    newResource: (resource: { request: string }) => void
+  ) => WebpackPlugin;
+}
+
+interface WebpackPlugin {
+  apply(compiler: WebpackCompiler): void;
+}
+
+interface WebpackCompiler {
+  /** webpack 5 exposes its own runtime API on the compiler. */
+  webpack: WebpackApi;
+}
+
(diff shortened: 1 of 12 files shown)


contrib/strands/src/plugin.ts
```diff
@@ -158,27 +158,58 @@ export class StrandsPlugin extends SimplePlugin {
...
-            /^(?:node:(?:fs\/promises|os|path|process|stream)|@modelcontextprotocol\/sdk\/client\/(?:stdio|sse|streamableHttp)\.js|eventsource-parser\/stream)$/;
-          // Cast through `unknown` because the strands plugin doesn't list
-          // webpack as a dependency and therefore lacks its types.
-          config.plugins = (config.plugins ?? []).concat(
-            new NormalModuleReplacementPlugin(ignorePattern, (data: { request: string }) => {
-              data.request = empty;
-            }) as unknown as never
-          );
-        }
+        const plugins = [...(config.plugins ?? [])];
```
mjameswh, line 161: Changes on this file replicate #2183.

This was required to fix a bug in the existing code uncovered by the rest of this PR resulting in this package's tests failing.

pnpm-workspace.yaml
```diff
@@ -40,6 +40,10 @@ minimumReleaseAgeExclude:
   - '@temporalio/*'
   - nexus-rpc
 
+  # Pin as part of #2170 mitigation to make sure our tests confirm validity
```
chris-olszewski: "pin" makes it seem like this controls which version we use
```suggestion
  # Allow as part of #2170 mitigation to make sure our tests confirm validity
```

packages/worker/src/workflow/bundler/patch-module-cache.ts
```diff
@@ -0,0 +1,173 @@
...
+          `https://github.com/temporalio/sdk-typescript/issues`
+      );
+    }
+
+    // The top-level module-cache declaration is the one immediately preceding the marked
+    // require function. Any other `__webpack_module_cache__ = {}` occurrences belong to
+    // nested, pre-bundled dependencies and appear earlier in the emitted module content, so
+    // the nearest match before the marker is always the real one.
+    let declaration: RegExpExecArray | null = null;
+    MODULE_CACHE_DECLARATION.lastIndex = 0;
```
chris-olszewski, line 97: Is there a reason we need to explicitly set this? If so a comment would be useful

packages/worker/src/workflow/bundler/patch-module-cache.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    const replaced = new sources.ReplaceSource(source);
+
+    // Redirect the top-level declaration to the runtime-injected global.
+    const declStart = declaration.index;
+    const declEndInclusive = declStart + declaration[0].length - 1;
+    replaced.replace(declStart, declEndInclusive, declaration[0].replace('= {}', `= ${MODULE_CACHE_GLOBAL}`));
+
+    // Strip the marker (and its trailing newline, if present) so it never ships.
```
chris-olszewski: Agree with being defensive wrt the trailing newline, but is there a situation we already know of where the newline after the marker might not be present?

chris-olszewski (approved):
Only questions for my own learning

packages/worker/src/workflow/bundler/patch-module-cache.ts
mjameswh: Yes, there is.

The regex is using the `/g` flag so that we can loop over all occurences. When tht flag is in use, the regex stores the position of its last match into the `lastIndex` property. This is where the regex will start its next search. In the case where we're handling a different code (seems very unlikely, but not impossible), then we need to reset that property so the search starts from the beginning of the new code.

That quirk is not widely known, but not uncommon.

PR #2201, opened 2026-07-13
External Storage pipeline integration

### LINKED ISSUE #2202: [Feature Request] Extstore Pipeline Integration
Once the payload visitor work is merged, use it to integrate external storage into the workflow activation and completion pipelines. Some logic in the worker will need to be decoupled to support this.

cconstable:
## What was changed
- Wired extstore store + retrieve into the worker pipeline (workflow, activity, nexus boundaries) behind an internal `LoadedDataConverter.externalStorage` field, using the payload visitor

## Why?
Extstore pipeline integration sets us up for preview release.

## Checklist
Closes #2202

### DIFF
--- a/packages/common/src/internal-non-workflow/external-storage-visitor.ts
+++ b/packages/common/src/internal-non-workflow/external-storage-visitor.ts
@@ -1,67 +1,67 @@
 /**
- * Visits payloads for External Storage and applies payload transformations via
+ * Visits payloads for External Storage and applies payload transformations via                                                                                                                                                             ║
  * {@link ExternalStorageRunner}.
  *
  * @module
  */
-import type { coresdk } from '@temporalio/proto';
-
 import type { ConcurrencyLimit } from '../concurrency/limit';
 import type { ExternalStorage, StorageDriverTargetInfo } from '../converter/extstore';
 import { ExternalStorageRunner } from './external-storage-runner';
-import { visitWorkflowActivation, visitWorkflowActivationCompletion } from './payload-visitor';
+import type { ContextDeriver, VisitOptions } from './payload-visitor';
 
 /**
- * Offload every eligible payload in a `WorkflowActivationCompletion` to external storage,
- * replacing it in place with a reference payload.
- *
- * @internal
- * @experimental
+ * The storage target in scope at a given payload site during a store walk. It starts at
+ * {@link ExternalStorageStoreOptions.initialTarget} and {@link ExternalStorageStoreOptions.deriveContext}
+ * may retarget it per message (e.g. a child-workflow command retargets its payloads at the child).
  */
-export function storeWorkflowActivationCompletion(
-  externalStorage: ExternalStorage,
-  completion: coresdk.workflow_completion.IWorkflowActivationCompletion,
-  {
-    target,
-    limit,
-    abortSignal,
-  }: {
-    /** Identity of the workflow / activity that produced the payloads, used for object naming. */
-    target?: StorageDriverTargetInfo;
-    /** Bounds concurrent transform calls across payload sites. Omit for sequential. */
-    limit?: ConcurrencyLimit;
-    /** Aborts the walk and every in-flight driver call. */
-    abortSignal?: AbortSignal;
-  } = {}
-): Promise<void> {
-  const runner = new ExternalStorageRunner(externalStorage);
-  return visitWorkflowActivationCompletion(completion, {
-    transformPayloads: (payloads, _context, signal) => runner.store(payloads, { target, abortSignal: signal }),
-    transformPayload: (payload, _context, signal) =>
-      runner.store([payload], { target, abortSignal: signal }).then((stored) => stored[0]!),
-    limit,
-    abortSignal,
-  });
+type StoreTarget = StorageDriverTargetInfo | undefined;
+
+/** @internal @experimental */
+export interface ExternalStorageStoreOptions {
+  /** Storage target before any message is entered (the initial enclosing workflow / activity). */
+  initialTarget?: StorageDriverTargetInfo;
+  /** Derives new storage target from the current message. */
+  deriveContext?: ContextDeriver<StoreTarget>;
+  /** Bounds concurrent transform calls across payload sites. Omit for sequential. */
+  limit?: ConcurrencyLimit;
+  /** Aborts the walk and every in-flight driver call. */
+  abortSignal?: AbortSignal;
 }
 
 /**
- * Replace every reference payload in a `WorkflowActivation` with the payload bytes fetched
- * from external storage, in place.
- *
  * @internal
  * @experimental
  */
-export function retrieveWorkflowActivation(
+export function extstoreStoreOptions(
   externalStorage: ExternalStorage,
-  activation: coresdk.workflow_activation.IWorkflowActivation,
-  { limit, abortSignal }: { limit?: ConcurrencyLimit; abortSignal?: AbortSignal } = {}
-): Promise<void> {
+  { initialTarget, deriveContext, limit, abortSignal }: ExternalStorageStoreOptions = {}
+): VisitOptions<StoreTarget> {
   const runner = new ExternalStorageRunner(externalStorage);
-  return visitWorkflowActivation(activation, {
+  return {
+    transformPayloads: (payloads, target, signal) => runner.store(payloads, { target, abortSignal: signal }),
+    transformPayload: (payload, target, signal) =>
+      runner.store([payload], { target, abortSignal: signal }).then((stored) => stored[0]!),
+    deriveContext,
+    initialContext: initialTarget,
+    limit,
+    abortSignal,
+  };
+}
+
+/**
+ * @internal
+ * @experimental
+ */
+export function extstoreRetrieveOptions(
+  externalStorage: ExternalStorage,
+  { limit, abortSignal }: { limit?: ConcurrencyLimit; abortSignal?: AbortSignal } = {}
+): VisitOptions<void> {
+  const runner = new ExternalStorageRunner(externalStorage);
+  return {
     transformPayloads: (payloads, _context, signal) => runner.retrieve(payloads, { abortSignal: signal }),
     transformPayload: (payload, _context, signal) =>
       runner.retrieve([payload], { abortSignal: signal }).then((retrieved) => retrieved[0]!),
     limit,
     abortSignal,
-  });
+  };
 }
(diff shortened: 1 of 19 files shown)


packages/common/src/internal-non-workflow/external-storage-visitor.ts
```diff
@@ -0,0 +1,67 @@
...
+    /** Identity of the workflow / activity that produced the payloads, used for object naming. */
+    target?: StorageDriverTargetInfo;
+    /** Bounds concurrent transform calls across payload sites. Omit for sequential. */
+    limit?: ConcurrencyLimit;
+    /** Aborts the walk and every in-flight driver call. */
+    abortSignal?: AbortSignal;
+  } = {}
+): Promise<void> {
+  const runner = new ExternalStorageRunner(externalStorage);
+  return visitWorkflowActivationCompletion(completion, {
```
jmaeagle99: This needs to provide different storage contexts for each command within the workflow task completion. See https://github.com/temporalio/sdk-python/blob/1.30.0/temporalio/worker/_workflow_instance.py#L2302 in Python.

packages/common/src/internal-non-workflow/external-storage-visitor.ts
```diff
@@ -0,0 +1,67 @@
...
+import { visitWorkflowActivation, visitWorkflowActivationCompletion } from './payload-visitor';
+
+/**
+ * Offload every eligible payload in a `WorkflowActivationCompletion` to external storage,
+ * replacing it in place with a reference payload.
+ *
+ * @internal
+ * @experimental
+ */
+export function storeWorkflowActivationCompletion(
```
jmaeagle99: Need analogous functions for activity activation and completion.

jmaeagle99: Maybe similarly, we need to update the client call sites to store and retrieve as well. Could use the visitor or invoke per-field; my preference is to use the visitor so we don't have to remember to add more specialized calls for new fields.

packages/test/src/test-extstore-visit.ts
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  await storeWorkflowActivationCompletion(externalStorage, completion, { target: WORKFLOW_TARGET });
+
+  const args = completion.successful!.commands![0]!.scheduleActivity!.arguments!;
+  t.true(isReferencePayload(args[0]!));
+  t.deepEqual(args[1], small);
+});
+
+test('store then retrieve round-trips the original payload bytes through a worker message', async (t) => {
```
THardy98: Might be nice to have an integration test that does a round-trip, where we don't manually construct the activation/completion ourselves.

THardy98 (approved):

THardy98 (commented):
LGTM, would prefer some additional test coverage though (noted in comment)

probably prudent to get approval from @jmaeagle99 for ext storage API semantics

packages/test/src/test-extstore-visit.ts
THardy98: Also would be nice to add equivalent cases for activity payloads that exceed the external storage payload threshold (unsure if this was intended to be part of the change or if this PR was scoped to workflows)

THardy98 (commented):
should we add to `CHANGELOG.MD` ?

THardy98 (dismissed):
(changing to requesting changes because I only meant to comment not approve, and I don't want to accidentally merge - sorry for the paranoia)

cconstable:
> should we add to `CHANGELOG.MD` ?

@THardy98 I can add the changelog entry when the actual public interface is committed (hopefully today).

packages/test/src/test-extstore-visit.ts
cconstable: Added a round trip test to make sure payload in == payload out and we get a reference payload in between.

THardy98: Hmm maybe I missed it but I'm not sure I saw a round trip integration test.

I was thinking about a test where we actually run a workflow / activity, running the e2e code path.

packages/common/src/internal-non-workflow/payload-visitor.ts
```diff
@@ -160,3 +164,58 @@ export async function visitWorkflowActivationCompletion<Ctx = void>(
 ): Promise<void> {
   return runVisit(options, (env, context) => walkWorkflowActivationCompletion(root, env, context));
 }
+
```
jmaeagle99: Should have one for visitActivityHeartbeat

packages/common/src/internal-non-workflow/external-storage-visitor.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * @internal
+ * @experimental
+ */
+export function extstoreRetrieveOptions(
+  externalStorage: ExternalStorage,
+  { limit, abortSignal }: { limit?: ConcurrencyLimit; abortSignal?: AbortSignal } = {}
+): VisitOptions<void> {
+  const runner = new ExternalStorageRunner(externalStorage);
+  return {
```
jmaeagle99, line 62: Need to skip search attributes since they are used for searching and must remain their actual value.

packages/worker/src/worker.ts
```diff
@@ -2267,6 +2317,37 @@ async function extractActivityInfo({
...
 
+function workflowCommandStoreTarget(
+  namespace: string
+): (
+  message: object,
+  typeName: string,
+  context: StorageDriverTargetInfo | undefined
+) => StorageDriverTargetInfo | undefined {
+  return (message, typeName, context) => {
+    switch (typeName) {
```
jmaeagle99, line 2333: Missing a case for continueAsNewWorkflowExecution. Looks like we're missing that for Python too.

packages/worker/src/worker.ts
```diff
@@ -2267,6 +2317,37 @@ async function extractActivityInfo({
...
+      case 'coresdk.workflow_commands.StartChildWorkflowExecution': {
+        const command = message as coresdk.workflow_commands.IStartChildWorkflowExecution;
+        return command.workflowId
```
jmaeagle99: We shouldn't need to check the workflow ID property for two reasons:
- In practice, this is always provided, either by the caller or filled by the SDK via `uuid4()` fallback
- Even if it didn't have a value, external storage still works without it, just like a client call to start a workflow.

We should unconditionally create the child workflow target info.

packages/worker/src/worker.ts
```diff
@@ -1238,7 +1247,13 @@ export class Worker {
             return { taskToken, result };
           }),
           filter(<T>(result: T): result is Exclude<T, undefined> => result !== undefined),
-          map((rest) => coresdk.ActivityTaskCompletion.encodeDelimited(rest).finish()),
+          mergeMap(async (rest) => {
+            const { externalStorage } = this.options.loadedDataConverter;
+            if (externalStorage) {
+              await visitActivityTaskCompletion(rest, extstoreStoreOptions(externalStorage));
+            }
```
jmaeagle99, line 1256: For all decode paths, where external storage is not enabled, we'll want to raise an exception with TMPRL1105 and make sure it retryably fails the WFT/AT/NT (because that is correctible by enabling external storage), just like https://github.com/temporalio/sdk-python/blob/1.30.0/temporalio/converter/_data_converter.py#L323. Does not have to be part of this change though.

jmaeagle99 (approved):

packages/worker/src/worker.ts
cconstable: That makes sense. Updating.

packages/test-helpers/src/utilities.ts
```diff
@@ -1,10 +1,44 @@
...
+ */
+export async function assertEventually<Context = unknown>(
+  t: ExecutionContext<Context>,
+  assertion: (t: ExecutionContext<Context>) => void | Promise<void>,
+  timeoutMs = 10000,
+  intervalMs = 100
+): Promise<void> {
+  const endTime = Date.now() + timeoutMs;
+  for (;;) {
+    const attempt = await t.try(async (tt) => {
```
mjameswh, line 30: Never heard of `t.try` before. That's incredibly handy. TIL

packages/test/src/test-runtime-otel.ts
```diff
@@ -139,7 +139,7 @@ test.serial('Exporting OTEL metrics from Core works', async (t) => {
           });
           const req = await Promise.race([
             capturedRequest,
-            await new Promise<undefined>((resolve) => setTimeout(() => resolve(undefined), 2000)),
```
mjameswh, line 142: Oh, that was a nasty bug. The extra `await` means we were systematically waiting the full timeout period, _then_ doing the `Promise.race`.

packages/test/src/test-runtime-otel.ts
```diff
@@ -139,7 +139,7 @@ test.serial('Exporting OTEL metrics from Core works', async (t) => {
...
+            new Promise<undefined>((resolve) => setTimeout(() => resolve(undefined), 10000)),
```
mjameswh, line 142: The test itself has a 10s timeout, and will have started prior to this timer. Bottom line, this 10s timer can't do anything useful. I'd try with 5s.

packages/test/src/test-runtime-otel.ts
```diff
@@ -192,7 +192,7 @@ test.serial('Exporting OTEL metrics using OTLP/HTTP from Core works', async (t)
...
```
mjameswh, line 195: Same

packages/test-helpers/src/utilities.ts
```diff
@@ -1,10 +1,44 @@
...
+ * diagnostics and logs.
+ *
+ * Use this for state that only becomes true asynchronously (e.g. metrics that Core exports to its
+ * Prometheus endpoint on a delay). The check is written once, as the assertion, instead of being
+ * duplicated as both a `waitUntil` predicate and a separate assertion.
+ */
+export async function assertEventually<Context = unknown>(
+  t: ExecutionContext<Context>,
+  assertion: (t: ExecutionContext<Context>) => void | Promise<void>,
+  timeoutMs = 10000,
```
mjameswh, line 25: By default, all tests run with a timeout of 10s. Having this inner timeout that is also 10s means tests will timeout from the outer (test-level) timer, which tends to cause other problems (i.e. ava doesn't properly run test cleanup function, which then prevents shutdown of the process).

This is worse than test simply failing.

cconstable: I will adjust this now.

PR #2208, opened 2026-07-14
:boom: feature: payloads/memo size validation

### LINKED ISSUE #1913: SDK should fail workflow task if payloads size it known to be too large - TypeScript
Parent issue: https://github.com/temporalio/features/issues/723

jmaeagle99:
## What was changed

- Added `PayloadLimitsConfig` (`payloadsWarnSize`, `memoWarnSize`) and wired it onto
  `NativeConnectionOptions.PayloadLimits`; over-threshold values are logged (`[TMPRL1103]`) but
  still sent, `0` disables.
- Added `WorkerOptions. disablePayloadErrorLimit` to opt out of worker error-limit
  enforcement and defer to the server.
- Update `sdk-core` submodule to bring in payloads/memo size enforcement.

## :boom: Breaking Change

Payload size-limit enforcement is now on by default; existing code changes behavior without opting in: an oversized payload is now failed proactively by the worker (as a retryable task failure) before it's sent, rather than being sent and hard-failed by the server as before.

If you use a proxy between the worker and server that alters the size of payloads (e.g.
compression, encryption, external storage), it is advised that you disable size enforcement by
setting `disablePayloadErrorLimit` to `true` on the worker.

## Why?

Prevent server from hard-failing workflows, activities, and operations due to payloads/memo size limit violations. Allows customers to potentially update their workflows, activities, and operations to reduce the oversized payloads/memos.

## Checklist

1. Closes #1913
1. How was this tested: Core library tests and minimal TypeScript tests

## Todo

1. [x] Wait for https://github.com/temporalio/sdk-rust/pull/1363 to be merged
1. [x] Updated sdk-core submodule after https://github.com/temporalio/sdk-rust/pull/1363 is merged.


packages/test/src/test-payload-size-limits.ts
```diff
@@ -0,0 +1,150 @@
...
+    const worker = await Worker.create({
+      connection: env.nativeConnection,
+      namespace: env.namespace,
+      taskQueue: 'payload-size-limits',
+      workflowsPath: require.resolve('./workflows'),
+      activities,
+    });
+
+    // Core repeatedly fails the workflow task (PAYLOADS_TOO_LARGE) for the oversized result, so the
+    // workflow never completes and hits its execution timeout.
```
THardy98, line 73: Is there value in asserting that this:
- does fail repeatedly (i.e. asserting that we get multiple log entries) or,
- produces a _task_ failure at the worker level, in addition to a workflow failure from server?

The log assertion distinguishes this from the `disablePayloadErrorLimit` case (or existing behaviour) but wondering if a task failure assertion makes the expected behaviour more clear. If this comment wasn't here, the test reads as "hard (server) failure + log".

packages/test/src/test-payload-size-limits.ts
```diff
@@ -0,0 +1,150 @@
...
+import { waitUntil } from '@temporalio/test-helpers';
+import { Worker } from './helpers';
+import * as activities from './activities';
+import { payloadSizeLimitsWorkflow } from './workflows/payload-size-limits';
+
+const PAYLOAD_ERROR_LIMIT = 10 * 1024;
+const PAYLOAD_LIMITS_EXTRA_ARGS = [
+  '--dynamic-config-value',
+  `limit.blobSize.error=${PAYLOAD_ERROR_LIMIT}`,
+  // Warn limit must be specified to have the server enforce the error limit.
```
THardy98: 👍 is this doc'd somewhere?

packages/test/src/test-payload-size-limits.ts
THardy98, line 1: Could we add cases for the `memo` path?

packages/worker/src/connection-options.ts
```diff
@@ -203,5 +234,7 @@ export function toNativeClientOptions(options: NativeConnectionOptions): native.
     headers,
     apiKey: options.apiKey ?? null,
     disableErrorCodeMetricTags: options.disableErrorCodeMetricTags ?? false,
+    payloadsWarnSize: options.payloadLimits?.payloadsWarnSize ?? 512 * 1024,
+    memoWarnSize: options.payloadLimits?.memoWarnSize ?? 2 * 1024,
```
THardy98: nit: prefer `const` values, but not blocking

packages/test/src/test-payload-size-limits.ts
jmaeagle99: Good idea! I'll add a memo test for the warn path to validate the configuration knob. More than that is just retesting the same code coverage.

packages/test/src/test-payload-size-limits.ts
jmaeagle99: I think this can be removed at this point with more modern servers. Will try to do so.

packages/test/src/test-payload-size-limits.ts
jmaeagle99: I don't think we should really check that it logs repeatedly. I'm even consider adjusting that behavior to only log once. But checking that a task failure was produced, is finalized in the workflow history, and has the PAYLOADS_TOO_LARGE cause would validate the worker sent the WFT failure.

packages/test/src/test-payload-size-limits.ts
jmaeagle99: Actually, this is still the case, but the comment is inaccurate. I will update it to correctly describe the problem.

packages/test/src/test-payload-size-limits.ts
```diff
@@ -24,7 +24,8 @@ const PAYLOAD_ERROR_LIMIT = 10 * 1024;
 const PAYLOAD_LIMITS_EXTRA_ARGS = [
   '--dynamic-config-value',
   `limit.blobSize.error=${PAYLOAD_ERROR_LIMIT}`,
-  // Warn limit must be specified to have the server enforce the error limit.
+  // The server only enforces the error limit for payloads that also exceed the warn limit, so the
+  // warn limit must be below the error limit.
```
THardy98, line 28: wow that is weird behaviour!

packages/test/src/test-payload-size-limits.ts
```diff
@@ -67,19 +70,31 @@ test.serial('oversized worker completion is failed by core with a [TMPRL1103] er
...
+    // Failure mechanism: worker reported a WorkflowTaskFailed with PAYLOADS_TOO_LARGE.
+    const { events } = await env.client.workflow.getHandle(workflowId).fetchHistory();
+    t.true(
+      (events ?? []).some(
+        (e) =>
+          e.eventType === EventType.EVENT_TYPE_WORKFLOW_TASK_FAILED &&
+          e.workflowTaskFailedEventAttributes?.cause ===
+            WorkflowTaskFailedCause.WORKFLOW_TASK_FAILED_CAUSE_PAYLOADS_TOO_LARGE
+      )
+    );
```
THardy98, line 101: might be worth running this in a "try eventually" test helper to avoid flakes, in case visibility is lagging

THardy98 (approved):
left a small comment, but lgtm

THardy98 (commented):
FWIW - I haven't looked closely at the underlying core changes, but I imagine that shouldn't be necessary

packages/test/src/test-payload-size-limits.ts
jmaeagle99: I don't think it's necessary because the server terminated the workflow due to execution timeout. So server should have recorded that in history already before the worker finishes.

PR #2209, opened 2026-07-14
feat: Event Groups support (experimental)

mjameswh:
## Summary

Add experimental support for *Event Groups*.

**Event Groups** is a new form of Workflow-level metadata that allows for improved visibility into a Workflow execution's history by grouping logically related Events together based on user-defined or system-inferred criteria.

- **Explicit markers**: `createEventGroup(label)` returns a marker attachable to commands via a `eventGroups: [...]` option or an execution scope (`eventGroup.withScope(...)`, propagated through coroutines via `AsyncLocalStorage`). Nested scopes compose; markers dedupe by id.
- **Implicit markers**: automatically wrap signal and update handlers.
- Markers are serialized onto outgoing commands as `eventGroupMarkers`. Explicit label marker ids are derived deterministically as `SHA1(originalExecutionRunId + label)` (sandbox-safe SHA-1 in `sha1.ts`).
- Exposes `WorkflowInfo.originalExecutionRunId` (plumbed core → lang) and bumps the `sdk-core` submodule to the matching build.

## Testing

Event Groups support requires Server v1.32.0-158.0, which has not yet been included in a published CLI release. In the mean time, tests can be run using the following CLI specifier:

```
TESTS_CLI_VERSION=v1.7.4-standalone-nexus-operations
```


packages/test/src/test-event-groups.ts
```diff
@@ -0,0 +1,437 @@
...
+    kind: 'scheduleActivity',
+    eventGroups: [WORKFLOW_START_MARKER, { id: orderId, label: 'order' }],
+  });
+  t.deepEqual(captured[5], {
+    kind: 'scheduleActivity',
+    eventGroups: [WORKFLOW_START_MARKER],
+  });
+});
+
+test('implicit markers wrap workflow start, signal handlers, and update handlers', async (t) => {
```
Sushisource: This one would be easier to read if it asserted the whole event groups array on commands like the others do

packages/workflow/src/event-groups.ts
```diff
@@ -0,0 +1,248 @@
...
+/** Active event group markers in the current execution scope. */
+const activeMarkerScopes: ALS<ActiveMarkerScopes> = new AsyncLocalStorage();
+
+interface ActiveMarkerScopes {
+  /**
+   * The active implicit event group marker.
+   *
+   * There's only one active implicit event group marker scope at any point in time, corresponding
+   * to the inbound workflow event that triggered execution of the current handler (i.e. one of
+   * WorkflowExecutionStarted, WorkflowExecutionSignaled or WorkflowExecutionUpdated).
```
Sushisource:
```suggestion
   * There's at most one active implicit event group marker scope at any point in time, corresponding
   * to the inbound workflow event that triggered execution of the current handler (i.e. one of
   * WorkflowExecutionStarted, WorkflowExecutionSignaled or WorkflowExecutionUpdated).
```

What is the condition under which it can be undefined?

packages/workflow/src/event-groups.ts
```diff
@@ -0,0 +1,248 @@
...
+ * option, or implicitly via {@link EventGroupMarker.withScope}, to indicate that those commands
+ * belong to the same logical group.
+ *
+ * @param label a user-visible label for the group, surfaced in the UI / CLI
+ *     This label will be codec-encoded using the worker's configured Payload Codecs.
+ *
+ *     Note that if no `id` is provided, the `id` will be derived from the `label` using a
+ *     deterministic hash function. Given short and predictable labels, brute-forcing the `id`'s
+ *     hashed value may be computationally feasible, thus allowing for recovery of the `label`
+ *     value. Thus, it is recommended to avoid including highly sensitive informations in
```
Sushisource:
```suggestion
 *     value. Thus, it is recommended to avoid including highly sensitive information in
```

packages/workflow/src/sha1.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * Minimal, dependency-free SHA-1 implementation.
+ *
+ * Event Group label identifiers are derived deterministically from the label text (see the Event
+ * Groups design). That derivation runs inside the Workflow sandbox, where Node's `crypto` module is
+ * not available, so we cannot use `crypto.createHash('sha1')`. This implementation relies only on
```
Sushisource: Don't think we really need to refer to event groups in here

packages/workflow/src/workflow.ts
```diff
@@ -462,6 +466,7 @@ function startChildWorkflowExecutionNextHandler({
         options?.staticDetails,
         context
       ),
+      eventGroupMarkers: eventGroupMarkersToProto(options.eventGroups, context),
     });
     activator.completions.childWorkflowStart.set(seq, {
       resolve,
```
Sushisource, line 507: I think some places `activator.pushCommand` is called might be missing group propagation. EX: signal external workflow https://github.com/temporalio/sdk-typescript/blob/58a1fe62d253bd88c8f522497c23698a82ffe322/packages/workflow/src/workflow.ts#L519-L535 and search attrib upsert: https://github.com/temporalio/sdk-typescript/blob/58a1fe62d253bd88c8f522497c23698a82ffe322/packages/workflow/src/workflow.ts#L1590-L1598

packages/common/src/index.ts
```diff
@@ -7,7 +7,6 @@
 import * as encoding from './encoding';
 import * as helpers from './type-helpers';
 
-export * from './activity-options';
```
Sushisource, line 10: Are we sure there's no re-export here we lost that could be breaking?

packages/workflow/src/sha1.ts
```diff
@@ -0,0 +1,122 @@
...
+ * plain arithmetic and string operations and is therefore safe to run in the deterministic Workflow
+ * environment.
+ *
+ * SHA-1 is used here for preimage resistance (making the identifier a one-way function of the label),
+ * not collision resistance; see the design's security rationale.
```
Sushisource: What design? LLM comment probably referencing something that doesn't exist any more.

Sushisource (commented):
Overall making sense to me. I think the only important comment I have is making sure all commands are actually covered, I think a few are missing right now.

packages/common/src/index.ts
mjameswh: Reexports of `ActivityOptions`, `LocalActivityOptions` and `ActivityCancellationType` have been moved to the "Deprecated APIs"  section below, so not breaking.

Internal helper functions `encodeActivityCancellationType` and `decodeActivityCancellationType` were also exported, so removing now is technically breaking, but there's no absolutely no reason why a user would have ever referred to those functions. I'm comfortable not calling this out explicitly.

packages/workflow/src/event-groups.ts
mjameswh: > What is the condition under which it can be undefined?

Queries and Update validators. I'll call this out explicitly.

packages/workflow/src/sha1.ts
mjameswh: Indeed. Fixed.

Sushisource:
I forgot to mention in last review, but, we discussed this a little - I'm not sure about the value of the implicit main routine group. Feels like it's not really providing any information that can't be easily inferred. I'm inclined to just drop it unless there's some value you think is there that I'm not seeing.

mjameswh:
> I forgot to mention in last review, but, we discussed this a little - I'm not sure about the value of the implicit main routine group. Feels like it's not really providing any information that can't be easily inferred.

Yeah, I had been considering removing it even before our discussion. I will.

mjameswh:
The following table details the status of various command-producing workflow APIs in regard to capture of event groups.

```
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| Workflow API                                                      | Command                                       | Coverage        | Comment                                            |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| Timers                                                            |                                               |                 |                                                    |
|    `sleep()`                                                      | `startTimer`                                  | Scoped + direct |                                                    |
|    `condition(fn, timeout, opts)`                                 | `startTimer` (timeout timer)                  | Scoped + direct |                                                    |
|    `setTimeout()` (JS mock)                                       | `startTimer` (delegates to `sleep`)           |   Scoped only   |                                                    |
|    `CancellationScope.withTimeout()`                              | `startTimer` (delegates to `sleep`)           |   Scoped only   |                                                    |
| Timer Cancellations                                               |                                               |                 |                                                    |
|     Cancellation of parent scope                                  | `cancelTimer`                                 |   Scoped only   |                                                    |
|    `clearTimeout()` (JS mock)                                     | `cancelTimer`                                 |   Scoped only   |                                                    |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| Activities / LA                                                   |                                               |                 |                                                    |
|    `proxyActivities()` / `.executeWithOptions()`                  | `scheduleActivity`                            | Scoped + direct |                                                    |
|    `proxyLocalActivities()` / `.executeWithOptions()`             | `scheduleLocalActivity` / `RecordMarker`      | Scoped + direct |                                                    |
| Activity / LA Cancellations                                       |                                               |                 |                                                    |
|     Cancellation of parent scope (Activity)                       | `requestCancelActivity`                       |   Scoped only   |                                                    |
|     Cancellation of parent scope (Local Activity)                 | `requestCancelLocalActivity` / `RecordMarker` |   Scoped only   |                                                    |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| Nexus Operations                                                  |                                               |                 |                                                    |
|    `startOperation()` / `executeOperation()`                      | `scheduleNexusOperation`                      | Scoped + direct |                                                    |
| Nexus Operation Cancellations                                     |                                               |                 |                                                    |
|     Cancellation of parent scope                                  | `requestCancelNexusOperation`                 |   Scoped only   |                                                    |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| Child Workflows                                                   |                                               |                 |                                                    |
|    `startChild()` / `executeChild()`                              | `startChildWorkflowExecution`                 | Scoped + direct |                                                    |
|    `startOperation()` / `executeOperation()` (Nexus Client)       | `scheduleNexusOperation`                      | Scoped + direct |                                                    |
| Child Workflows Cancellations                                     |                                               |                 |                                                    |
|     Cancellation of parent scope                                  | `cancelChildWorkflowExecution`                |   Scoped only   |                                                    |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| External Workflow Interractions                                   |                                               |                 |                                                    |
|     External / Child handle `.signal()`                           | `signalExternalWorkflowExecution`             |   Scoped only   |                                                    |
|     External signal cancellation                                  | `cancelSignalWorkflow`                        |   Scoped only   |                                                    |
|     External handle `.cancel()`                                   | `requestCancelExternalWorkflowExecution`      |   Scoped only   |                                                    |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| Workflow State and Properties manipulations                       |                                               |                 |                                                    |
|    `patched()` / `deprecatePatch()`                               | `setPatchMarker` → `recordMarker`             |   Scoped only   |                                                    |
|    `upsertSearchAttributes()`                                     | `upsertWorkflowSearchAttributes`              |   Scoped only   |                                                    |
|    `upsertMemo()`                                                 | `modifyWorkflowProperties`                    |   Scoped only   |                                                    |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
| Terminal Commands                                                 |                                               |                 |                                                    |
|    `makeContinueAsNewFunc()`                                      | `continueAsNewWorkflowExecution`              | Scoped + direct |                                                    |
|    `continueAsNew()`                                              | `continueAsNewWorkflowExecution`              |   Scoped only   |                                                    |
|     Workflow returns                                              | `completeWorkflowExecution`                   |      None       | Issued after user code returns; no scope is active |
|     Workflow throws                                               | `failWorkflowExecution`                       |      None       | Issued after user code returns; no scope is active |
|     Workflow cancelled                                            | `cancelWorkflowExecution`                     |      None       | Issued after user code returns; no scope is active |
| ----------------------------------------------------------------- | --------------------------------------------- | --------------- | -------------------------------------------------- |
```

Notes:

- **Terminal commands.** `continueAsNew` is the one terminal command supporting direct attachment.
  The other three set no markers; since they are issued by the SDK after user code has returned,
  no ambient scope is active either.

chris-olszewski (approved):
Implementation looks solid. Haven't gone through the entirety of the tests still.

packages/workflow/src/internals.ts
```diff
@@ -1156,11 +1159,20 @@ export class Activator implements ActivationHandler {
     this.inProgressSignals.set(signalExecutionNum, { name: signalName, unfinishedPolicy });
     const execute = composeInterceptors(interceptors, 'handleSignal', this.signalWorkflowNextHandler.bind(this));
     const context = this.workflowSerializationContext();
-    execute({
-      args: arrayFromPayloads(this.payloadConverter, activation.input, context, signalHandler?.typeInfo?.inputTypes),
-      signalName,
-      headers: headers ?? {},
-    })
+    const implicitMarker = createInboundEventMarker(activation.originatingEventId);
```
Sushisource, line 1162: This will throw if `originatingEventId` isn't set we fail WFTs here, and I see a bunch of work was done to make sure it is set, but, I'm wondering if just to be safe we shouldn't fail WFT but instead just not create the implicit marker.

Sushisource (approved):
Looking good! Excited to ship this.

packages/workflow/src/internals.ts
mjameswh: Fixed.

Now returning a Stub Implicit EG Maker in that case. That preserves the active EG bookkeeping logic as well as the caller's code path (i.e. they still call `withScope()` without having to know about the anomaly), but then skips proto serialization of the stub entirely.

PR #2214, opened 2026-07-15
feat(proto): auto-discover payload-visitor roots from coresdk

cconstable:
## What was changed
- Replace the hand-maintained visitor path list in gen-payload-visitor.ts with auto generated roots

## Why?
- Avoids the need to add new payload bearing types to the visitor generator manually.

## Checklist
- Added tests.

### DIFF
--- a/packages/proto/scripts/gen-payload-visitor.ts
+++ b/packages/proto/scripts/gen-payload-visitor.ts
@@ -3,18 +3,16 @@ import { writeFileSync } from 'node:fs';
 import * as protobuf from 'protobufjs';
 import * as prettier from 'prettier';
 
-// Generates `packages/proto/src/payload-visitor.generated.ts`: a synchronous walk of the
-// WorkflowActivation / WorkflowActivationCompletion message trees that invokes a payload visitor at
-// every Payload-bearing field, threading a per-message context. Runs in proto's build (and via
-// `pnpm gen:payload-visitor`).
+// Generates `packages/proto/src/payload-visitor.generated.ts`: a synchronous walk of the worker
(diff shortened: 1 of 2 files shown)


packages/proto/scripts/gen-payload-visitor.ts
```diff
@@ -3,18 +3,16 @@ import { writeFileSync } from 'node:fs';
...
 
 const PAYLOAD = 'temporal.api.common.v1.Payload';
 const ANY = 'google.protobuf.Any';
 
-const ROOTS = [
-  { type: 'coresdk.workflow_activation.WorkflowActivation', entry: 'walkWorkflowActivation' },
-  { type: 'coresdk.workflow_completion.WorkflowActivationCompletion', entry: 'walkWorkflowActivationCompletion' },
-] as const;
+// Namespaces scanned for root messages.
+const ROOT_NAMESPACES = ['coresdk'] as const;
```
cconstable, line 15: Generate everything under `coresdk` instead of explicit types now

jmaeagle99: Are all of the messages that are handled lang side in the `coresdk` namespace? I think the generated code looks correct for task processing. Unsure whether we'll use the visitor for client calls or not; in Go we used the visitor, in Python we piggy-backed onto the existing encode/decode calls.

jmaeagle99 (approved):

packages/proto/scripts/gen-payload-visitor.ts
cconstable: AFAICT, yes.

PR #2216, opened 2026-07-15
extstore/replay and heartbeat metadata

cconstable:
## What was changed
- enable safe worker replay (no extstore.store)
- add driver types to worker heartbeat

## Why?
- external storage release

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -1493,7 +1493,8 @@ export class Worker {
       try {
         const unencodedCompletion = await workflow.workflow.activate(decodedActivation);
         const encodedCompletion = await workflowCodecRunner.encodeCompletion(unencodedCompletion);
-        if (externalStorage) {
+        // Skip extstore.store on replay: the completion is discarded and its payloads were already offloaded on the original run.
+        if (externalStorage && !this.isReplayWorker) {
           const namespace = workflowCodecRunner.workflowContext.namespace;
           await visit(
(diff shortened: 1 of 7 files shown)


THardy98 (approved):

packages/worker/src/worker.ts
```diff
@@ -1463,7 +1463,8 @@ export class Worker {
       try {
         const unencodedCompletion = await workflow.workflow.activate(decodedActivation);
         const encodedCompletion = await workflowCodecRunner.encodeCompletion(unencodedCompletion);
-        if (externalStorage) {
+        // skip extstore.store for replay
+        if (externalStorage && !this.isReplayWorker) {
```
jmaeagle99, line 1497: Would love to have a test to validate this behavior, but won't block on it. And should apply this to the other SDKs as well.

jmaeagle99 (approved):

packages/worker/src/worker.ts
cconstable: I think the logic checks out here. I believe Go does this too. Will verify before release though.

jmaeagle99: Probably also want to use `&& !activation.isReplaying` in addition to the the `!this.isReplayWorker`.

jmaeagle99: Queries are potentially complicated here too. So maybe what you have is a good start and then we can pare it down more and more.

PR #2217, opened 2026-07-15
AI-338: drive langsmith workflow tracing from the SDK-provided AsyncLocalStorage

xumaple:
Replaces the hand-rolled synchronous context provider in the LangSmith workflow interceptor with the AsyncLocalStorage the worker injects onto the workflow-sandbox global, so trace context propagates across awaits and Promise.all fan-out and stays deterministic under replay. The node:async_hooks bundle rewrite now targets a small shim that re-exports that global.


chris-olszewski (approved):
Nice

PR #2219, opened 2026-07-16
Enable re-use of resource controllers

### LINKED ISSUE #2159: [Feature Request] Share a single resource-based tuner / ResourceController across multiple Workers in one process

There is currently no way to share a single resource-based tuner (i.e. one `ResourceController` + one system-resource sampler + one PID loop) across **multiple `Worker`s running in the same process**.

Our deployment is a modular monolith: a single Node process hosts ~25 services, each polling its own task queue, so we create ~25 `Worker`s in one process. With resource-based tuning this behaves poorly:

- Each `Worker.create` builds its **own** tuner. Tracing it through the stack:
  - `@temporalio/worker` `asNativeTuner` only emits **declarative config** (`{ type: 'resource-based', tunerOptions: { targetCpuUsage, targetMemoryUsage }, minimumSlots, maximumSlots, rampThrottle }`) — there is no controller object to reuse.
  - The Neon bridge (`core-bridge/src/worker.rs`, `WorkerTuner::into_core_config`) builds a **fresh `Arc<TunerHolder>` per worker**.
  - sdk-core (`worker/tuner.rs`, `build_tuner_holder`) does `resource_based_options.map(ResourceBasedTuner::new_from_options)`, and `ResourceBasedTuner::new_from_options` calls `ResourceController::new_with_sysinfo(opts, Arc::new(RealSysInfo::new()))` — a **new `RealSysInfo` (its own background sampling thread) and its own PID controller per worker**.

Consequences for N workers in one process:

1. **No true global resource cap.** Each of the N controllers independently targets e.g. 60% CPU / 75% memory of the **same** container. They all read container-wide usage but never coordinate slot grants, so under broad load the aggregate can overshoot the intended target, and under narrow load one queue can starve at its per-worker floor while the container sits far below target.
2. **N redundant sampler threads**, each re-measuring the same cgroup CPU/memory.
3. The official docs even state this as expected: *"A Worker Tuner instance exists per-Worker."*

Expose a **constructible, shareable resource controller** in `@temporalio/worker` that can be passed to the tuners of multiple `Worker`s, so a single controller + single sampler + single PID loop governs slot allocation across every worker in the process.

This already exists in the **Go SDK**, which is the precedent: the composite-tuner example constructs a controller explicitly and shares it across suppliers:

```go
controller := worker.NewResourceController(options)
actSS, _ := worker.NewResourceBasedSlotSupplier(controller, ...)
laSS,  _ := worker.NewResourceBasedSlotSupplier(controller, ...)
```

A TS equivalent might look like:

```ts
import { ResourceBasedController } from '@temporalio/worker';

// one instance for the whole process
const controller = ResourceBasedController.create({ targetCpuUsage: 0.6, targetMemoryUsage: 0.75 });

// passed into each Worker's tuner
const tuner = {
  workflowTaskSlotSupplier: { type: 'resource-based', controller, minimumSlots: 2, maximumSlots: 12 },
  activityTaskSlotSupplier: { type: 'resource-based', controller, minimumSlots: 1, maximumSlots: 24 },
  // ...
};
```

sdk-core already models this internally (`ResourceController` is an `Arc`, and `TunerBuilder` can accept a shared `sys_info`) — the gap is that the TS/Neon layer only accepts declarative options and constructs everything fresh per worker, with no way to inject a shared controller handle.

### Additional context

- **Workaround today:** a shared `CustomSlotSupplier` singleton (added in #1553) passed to every worker gives a process-wide budget, but it forfeits the built-in CPU/memory sampling (you must re-implement it in JS) and runs `reserveSlot` on the event loop on the hot path.
- This is distinct from the autotuning umbrella (temporalio/features#334); that tracks *whether* autotuning exists per SDK, whereas this is specifically about **sharing one resource controller across multiple Workers in a single process**.
- Version: `@temporalio/worker@1.16.2` (core-bridge/sdk-core 1.16.2).

Sushisource:
## What was changed

- Added a public `ResourceBasedController.create()` API whose controller can be shared by resource-based tuners across multiple Workers in the same process.
- Added controller-backed tuner configuration for both the high-level resource-based tuner and individual resource-based slot suppliers. `controller` and `tunerOptions` are mutually exclusive, and all resource-based suppliers within one Worker must resolve to the same controller or equivalent options.
- Preserved the existing declarative `tunerOptions` API. Equal declarative options within a Worker continue to produce one Worker-scoped controller.
- Updated the Core bridge to pass an opaque shared controller handle through to Core and advanced the sdk-core submodule to include [temporalio/sdk-core#1407](https://github.com/temporalio/sdk-core/pull/1407).
- Added coverage for two Workers executing workflows with one shared controller, as well as conflicting options and conflicting controller configurations.

Example:

```ts
const controller = ResourceBasedController.create({
  targetCpuUsage: 0.6,
  targetMemoryUsage: 0.75,
});

  // ...
  tuner: {
    controller,
    activityTaskSlotOptions: {
      minimumSlots: 1,
      maximumSlots: 24,
    },
  },
});
```

## Why?

Previously, each Worker using resource-based tuning constructed its own resource sampler and PID controller. Processes hosting multiple Workers therefore could not coordinate slot allocation against one process-wide CPU and memory target, and created redundant sampler threads.

Sharing a `ResourceBasedController` lets all participating Workers use one sampler and controller so their aggregate slot allocation is governed by the same resource targets.

## Checklist

1. Closes https://github.com/temporalio/sdk-typescript/issues/2159

2. How was this tested:
   - TypeScript build for the test workspace
   - ESLint and Prettier checks for the changed TypeScript sources
   - Core bridge formatting, Clippy, and unit tests
   - Local integration run with two Workers on separate task queues executing workflows through one shared controller

3. Any docs updates needed?
   - No standalone docs update is required; the new public types and controller API include inline API documentation.


packages/worker/src/worker-tuner.ts
```diff
@@ -74,6 +93,25 @@ export interface ResourceBasedTunerOptions {
...
+ *
+ * Share one instance across resource-based tuners to enforce a single resource target across
+ * multiple Workers in the same process.
+ */
+export class ResourceBasedController {
+  private constructor(private readonly nativeController: native.ResourceBasedController) {}
+
+  /** Create a resource controller with the given target resource usage. */
+  public static create(options: ResourceBasedTunerOptions): ResourceBasedController {
+    return new ResourceBasedController(native.newResourceBasedController(options));
```
mjameswh, line 107: - Out of curiosity, why the early initialization of the native counterpart?
- What's the expected lifecycle of that object? Notably, would it be a problem if the native controller object is created before the Runtime, or if it survives after the Runtime's destruction?

Sushisource: Yeah, reasonable point, we could delay initialization

Sushisource: Looks like it's not a problem if it's started outside the runtime. It just creates its own thread and happily can do its own thing. If it's never used, the thread it started will terminate at some point after it gets GCd

mjameswh (approved):

PR #2223, opened 2026-07-21
feat(activity): reuse MCP client connections in @temporalio/ai-sdk (idle-timeout pooling)

### LINKED ISSUE #2222: @temporalio/ai-sdk: reuse MCP client connections across tool calls (idle-timeout pooling, matching strands-agents)
Title: `@temporalio/ai-sdk`: reuse MCP client connections across tool calls (idle-timeout pooling, matching strands-agents)

In `contrib/ai-sdk/src/activities.ts`, `activitiesForName(name, mcpClientFactory)` creates a brand new MCP client on every single `listToolsActivity`/`callToolActivity` invocation:

```ts
const mcpClient = await mcpClientFactory(args.clientArgs)
try {
  // ... call the tool ...
} finally {
  await mcpClient.close()
}
```

For a multi-turn agent loop — the normal usage pattern, where an LLM calls the same MCP server's tools repeatedly across steps — this means the underlying MCP transport (a stdio subprocess spawn, an SSE/HTTP handshake, an OAuth token exchange) is torn down and rebuilt on every single tool call and every `listTools` refresh. For stdio-based or OAuth-secured MCP servers in particular, this adds real, avoidable latency and resource churn to every agent step.

Notably, this repo has already solved this exact problem twice, just not in this package:
- `@temporalio/strands-agents` (`contrib/strands/src/plugin.ts`, `contrib/strands/src/temporal-mcp-client.ts`) already ships `StrandsPluginOptions.mcpConnectionIdleTimeout` (default 5 minutes), keeping one lazily-opened MCP connection alive across activity calls and resetting the idle timer on reuse.
- `@temporalio/openai-agents` (`contrib/openai-agents/src/worker/stateful-mcp-provider.ts`) goes further with a `StatefulMCPServerProvider` that keeps one persistent MCP session alive for an entire workflow run.

`@temporalio/ai-sdk` is the newest, most actively-developed of the three contrib packages (it just had its `ai@7` provider-spec upgrade merged in #2192), and it's the one left without either pattern.

Port the lighter, already-proven `strands-agents` idle-timeout pattern into `@temporalio/ai-sdk`:

1. In `contrib/ai-sdk/src/activities.ts`, change `activitiesForName` to hold a `Map<string, { client: MCPClient; timer: NodeJS.Timeout }>` keyed by MCP server name, lazily creating the client via `mcpClientFactory` on first use and reusing it across subsequent `listTools`/`callTool` calls within the same worker process, instead of closing it in every `finally`.
2. Add an `mcpConnectionIdleTimeout` option to `createActivities(provider, mcpClientFactories, options?)`, mirroring `StrandsPluginOptions.mcpConnectionIdleTimeout` (a `Duration`, default ~5 minutes); reset the idle timer on each reuse; evict/close the connection on idle expiry or worker shutdown.
3. Keep today's stateless behavior available as an opt-out (e.g. `mcpConnectionIdleTimeout: 0`) for transports/servers that don't tolerate a reused/concurrent session.
4. Add tests mirroring the existing `test-ai-sdk.ts` MCP tests, verifying the client factory is called once and reused across repeated `tools()` calls, and that the connection is evicted after the idle window.

This is scoped to a lifecycle/caching change inside one existing file plus one new option — not a rewrite of the MCP integration, and not an attempt to unify all three contrib packages' MCP handling into a shared library (a larger, separate ask).

**Describe alternatives you've considered**

- Leaving connection lifecycle as-is and relying on users to configure long-lived MCP servers themselves — doesn't help stdio-subprocess-based servers, where "long-lived" isn't an option if the SDK itself closes the client after every call.
- Adopting the heavier `openai-agents`-style dedicated-Worker-per-run session model instead of the `strands-agents` idle-timeout model — possible, but a bigger design/risk surface for a first pass; happy to expand scope if maintainers prefer that direction.

**Additional context**

Not aware of any existing issue or PR covering this — searched for connection pooling/reuse, MCP session reuse, and reconnect-overhead related terms with no hits; the only PRs touching `contrib/ai-sdk/src/activities.ts` are the `ai@7` upgrade (#2192) and an unrelated schema fix. Happy to contribute a PR for this if the team is open to the approach (or to the heavier session-model alternative, if preferred).

wankhede04:
## What was changed

`contrib/ai-sdk/src/activities.ts`'s `listToolsActivity`/`callToolActivity` created a brand new MCP client via `mcpClientFactory` on every single invocation and closed it in a `finally` block. This ports the idle-timeout connection pooling pattern already shipped in `@temporalio/strands-agents` (`StrandsPluginOptions.mcpConnectionIdleTimeout`):

- A worker-process cache keyed by `${server name}:${clientArgs}` reuses one MCP client connection across repeated `listTools`/`callTool` activity invocations, resetting an idle timer on every reuse.
- A new `mcpConnectionIdleTimeout` option on `createActivities(provider, mcpClientFactories, options?)` and `AiSdkPluginOptions`, defaulting to 5 minutes (matching strands).
- Passing `mcpConnectionIdleTimeout: 0` opts out and restores the original create-then-close-per-call behavior, for MCP servers/transports that don't tolerate a reused or concurrent session.
- `AiSdkPlugin`'s `runContext` now evicts (closes) any cached connections for its registered MCP server names on worker teardown, mirroring `StrandsPlugin`.
- Also includes an unrelated one-line fix in `packages/common/src/__tests__/test-payload-visitor.ts`, found while verifying this change: it called the removed `visitWorkflowActivation` export directly instead of `visit(activation, walkWorkflowActivation, ...)` (the pattern every other test in the file uses), which broke `tsc --build` on `main` as of commit 4862844e. Happy to split that into its own PR if preferred — flagging it here since I needed it fixed locally just to get a clean build/typecheck for this change.

## Why?

In a multi-turn agent loop — the normal usage pattern where an LLM calls the same MCP server's tools repeatedly across steps — the underlying MCP transport (stdio subprocess spawn, SSE/HTTP handshake, OAuth token exchange) was being torn down and rebuilt on every single tool call and every `listTools` refresh. This adds avoidable latency and resource churn to every agent step, especially for stdio-based or OAuth-secured MCP servers.

## How tested

- `pnpm -C contrib/ai-sdk build` (tsc --build) passes cleanly.
- `pnpm exec eslint contrib/ai-sdk/src` and `pnpm exec prettier --check contrib/ai-sdk/src` both pass cleanly.
- Added unit tests in `contrib/ai-sdk/src/__tests__/test-ai-sdk.ts` (mocked `mcpClientFactory`, no live MCP server needed) covering:
  - the client factory is called once and the same connection is reused across repeated `tools()`/`callTool` invocations within the idle window
  - an idle connection is closed and evicted once the idle timeout elapses, and the next call reconnects via a fresh factory call
  - `mcpConnectionIdleTimeout: 0` opts out and restores the original create-then-close-per-call behavior
  - updated the existing `callToolActivity awaits tool.execute before closing MCP client` test to opt out (`mcpConnectionIdleTimeout: 0`), since its assertion only holds under the non-pooled path
- Was not able to run the AVA suite end-to-end in my local environment (this package's tests require Node >= 22.12 plus a prebuilt `@temporalio/core-bridge` native addon, which I could not build locally — no Rust toolchain / no prebuilt binary for this dev commit). The new tests were designed to run standalone against `createActivities` directly (no Worker/Temporal dev server needed) so a maintainer or CI can validate them directly.

## Risks

- Scoped to connection lifecycle only inside `activities.ts` and the new plugin option — no change to tool-calling semantics, schema handling, or the `ai@7` provider surface.
- The opt-out (`mcpConnectionIdleTimeout: 0`) preserves today's exact behavior for any MCP server/transport that can't tolerate a reused or concurrent session.

## Checklist

- [x] Closes #2222
- [x] Tests added/updated
- [ ] Docs needed? Not beyond the JSDoc added on the new option (no user-facing docs site changes identified for this package)

Closes #2222


brianstrauch (changes requested):
Thanks for this — the port of the strands connection-pooling pattern is clean, the opt-out is well thought out, and `tsc --build` is green locally. Two connection-lifecycle issues I'd like resolved before merging, since both are default-on behavior (pooling is enabled by default) and both are ways a *single* failure/timeout turns into a broken session for calls that shouldn't have been affected.

### 1. Idle timer can close a connection while a call is still in flight

In `getMcpConnection` (`contrib/ai-sdk/src/activities.ts`), `resetMcpIdleTimer` is armed at **call start**, before the client is awaited and the tool runs, and it's never reset when the call completes. So a connection's idle window is measured from when a call *began*, not from when it last *finished*.

With the default `idleMs` of 5 min and the activities' `startToCloseTimeout` of 10 min, a `callTool`/`tools()` that runs longer than 5 minutes has its underlying transport `close()`d by the eviction timer mid-request, and the in-flight `execute` rejects spuriously. Long-running MCP tools are exactly the case Temporal activities are meant to cover.

Suggested fix: reset the idle timer in a `finally` after the call resolves, or track an in-flight refcount per key and skip eviction while it's `> 0`. (This is inherited from the strands prior-art, so worth fixing there too, but let's not propagate it.)

### 2. Error-path eviction closes a connection shared by a concurrent activity

Pooling means two concurrent activities for the same `${name}:${clientArgs}` key now share one `MCPClient` (pre-PR each call had a fully isolated client). In `callToolActivity`/`listToolsActivity`, the `catch { await evictMcpConnection(...) }` closes that shared session on any error — including a plain tool-level failure — while a sibling activity may still be mid-`execute` on the same session. One real failure fans out into spurious failures for unrelated concurrent callers.

This is the hazard `mcpConnectionIdleTimeout: 0` is designed for, but since pooling is default-on, users get concurrent session sharing (and this failure mode) without opting in. I'd like:
- eviction restricted to connection-/transport-level errors rather than any tool error, or an in-flight refcount before `close()`

wankhede04:
Thanks for the detailed review! Both issues are addressed, pushed as two separate commits so each maps to a review point:

**1. Idle timer armed at call start (4ad6c846)**

Went with the refcount option: `acquireMcpConnection` no longer arms the idle timer — it just clears any pending timer and increments an in-flight `refCount`. The timer is only armed in `releaseMcpConnection`, and only once `refCount` drops back to 0. So a connection's idle window is now measured from when it was *last actually used*, not from call start, and a long-running call can never be evicted mid-flight regardless of how long it takes.

**2. Error-path eviction closing a connection shared by a concurrent activity (573f36d2)**

Also went with the refcount approach here, since `@ai-sdk/mcp` doesn't expose a clean way to distinguish transport-level errors from tool-execution errors. The `catch` no longer calls `evictMcpConnection` directly — it flags the connection as `broken` and defers the decision to `releaseMcpConnection`, which only evicts once `refCount` reaches 0. So a tool-level failure in one call can no longer tear down a session a concurrent sibling is still mid-`execute()` on; the connection is still dropped and reconnected on the next call once it's safe to do so.

Both fixes share the same underlying refcount mechanism in `acquireMcpConnection`/`releaseMcpConnection`, guarded by entry identity (not just the cache key) so a release can never disturb a different connection that got cached under the same key in the meantime.

Added two new tests covering each fix directly (a long-running call surviving a short idle timeout, and a tool-level error on one concurrent call not disrupting a sibling sharing the same connection), plus reconfirmed the existing reuse/idle-eviction/opt-out tests still pass unchanged.

brianstrauch (approved):
Thank you!

wankhede04:
@brianstrauch Thanks for approving this! Since both review points are addressed, would you be able to merge it whenever convenient?

brianstrauch:
@wankhede04 Happy to do so once the merge conflict is resolved! (If I remember correctly, this test file is unrelated to your PR and has been fixed upstream)

wankhede04:
Rebased onto latest `main` — the merge conflict in `test-payload-visitor.ts` is resolved (upstream's fix was equivalent, so `d071e85` was dropped as redundant). Build/lint/format all pass. Ready to merge whenever convenient!

brianstrauch:
@wankhede04 CI is looking for a brief changelog entry, and we might have introduced a flaky test here as well.

wankhede04:
@brianstrauch Fixed both — added the missing CHANGELOG.md entry and fixed the failing test's mock (it needed `jsonSchema()` wrapping, not actually flaky). Pushed in 262809d. Could you approve/re-run the CI workflow when you get a chance? Thanks!

CHANGELOG.md
```diff
@@ -19,6 +19,16 @@ to docs, or any other relevant information.
...
 
+### Added
+
+- `@temporalio/ai-sdk`: `listToolsActivity`/`callToolActivity` now reuse a single MCP client connection
+  across repeated invocations for the same server instead of creating and closing one on every call,
+  porting the idle-timeout connection pooling pattern already shipped in `@temporalio/strands-agents`.
+  Configure the idle window via the new `mcpConnectionIdleTimeout` option on `createActivities` and
+  `AiSdkPluginOptions` (defaults to 5 minutes); pass `mcpConnectionIdleTimeout: 0` to opt out and restore
+  the original create-then-close-per-call behavior for MCP servers/transports that don't tolerate a
+  reused or concurrent session.
```
brianstrauch:
```suggestion
- `@temporalio/ai-sdk`: `listToolsActivity`/`callToolActivity` now reuse a single MCP client connection
  across repeated invocations for the same server instead of creating and closing one on every call.
  Configure the idle window via the new `mcpConnectionIdleTimeout` option on `createActivities` and
  `AiSdkPluginOptions` (defaults to 5 minutes); pass `mcpConnectionIdleTimeout: 0` to opt out and restore
  the original behavior for MCP servers/transports that don't tolerate a
  reused or concurrent session.
```

brianstrauch:
@copilot resolve the merge conflicts in this pull request

chris-olszewski (approved):

PR #2228, opened 2026-07-21
feat(common): add external storage config to DataConverter

jmaeagle99:
## What was changed

Add external storage config to `DataConverter`.

## Why?

Allow end-to-end testing of external storage capabilities.

## Checklist

1. How was this tested: New integration test
1. Any docs updates needed? No


jmaeagle99:
Need to add changelog entry

cconstable (approved):
Looks good but we should probably drop the `@internal` on the public facing interfaces for extstore now so they show in the generated docs.

packages/test/src/test-integration-extstore.ts
```diff
@@ -0,0 +1,52 @@
...
+import { ExternalStorage } from '@temporalio/common';
+import { decodeReferencePayload, isReferencePayload } from '@temporalio/common/lib/internal-non-workflow';
+import * as activities from './activities';
+import { makeFakeDriver } from './extstore-fake-driver';
+import { helpers, makeTestFunction } from './helpers-integration';
+import { externalStorageOffload } from './workflows';
+
+const test = makeTestFunction({ workflowsPath: require.resolve('./workflows') });
+
+test('large activity result is offloaded to external storage and retrieved once', async (t) => {
```
chris-olszewski, line 13: Are there other places external storage is leveraged outside of activities currently that we should be testing?

chris-olszewski (approved):

packages/test/src/test-integration-extstore.ts
jmaeagle99: I can do everything that is within workflows and activities for now. Maybe something involving Nexus as well. We don't have external storage in the client at this time, so can't test starting workflows/activities or reading results that were externally stored.

PR #2232, opened 2026-07-22
feat(extstore): add extstore to client pipelines. add integration tests.

cconstable:
## What was changed
Integrate external storage into the remaining client paths. 

- **workflow-client**: retrieve on `describe`, `list`
- **schedule-client**: store on `create`/`update`; retrieve on `describe`/`list`
- **async-completion-client**: store on `complete`/`fail`/`reportCancellation`/`heartbeat` (token + by-id)
- **activity-client**: store on `start`; retrieve on `poll`/`describe`/`list`
- **nexus-client**: store on `start`; retrieve on `poll`/`describe`/`list`

The rest (`start`, `signalWithStart`, etc are already in main).

## Checklist
- Integration round-trips against dev server: workflow start-arg + result, query result, workflow/schedule memo via describe, and async-completion store.

### DIFF
--- a/packages/client/src/activity-client.ts
+++ b/packages/client/src/activity-client.ts
@@ -267,7 +267,18 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
       const req = await this.buildStartActivityExecutionRequest(input);
       const externalStorage = this.dataConverter.externalStorage;
       if (externalStorage) {
-        await visit(req, walkStartActivityExecutionRequest, extstoreStoreOptions(externalStorage));
+        await visit(
+          req,
+          walkStartActivityExecutionRequest,
+          extstoreStoreOptions(externalStorage, {
+            initialTarget: {
+              kind: 'activity',
+              namespace: this.options.namespace,
+              id: input.options.id,
+              type: input.activityType,
+            },
+          })
+        );
       }
       const resp = await this.workflowService.startActivityExecution(req);
       return this.createHandle(input.options.id, resp.runId);
--- a/packages/client/src/activity-client.ts
+++ b/packages/client/src/activity-client.ts
@@ -30,6 +30,13 @@ import {
   decodeOptionalFailureToOptionalError,
   encodeToPayloads,
   encodeUserMetadata,
+  extstoreRetrieveOptions,
+  extstoreStoreOptions,
+  visit,
+  walkDescribeActivityExecutionResponse,
+  walkListActivityExecutionsResponse,
+  walkPollActivityExecutionResponse,
+  walkStartActivityExecutionRequest,
 } from '@temporalio/common/lib/internal-non-workflow';
 import { temporal } from '@temporalio/proto';
 import type { Replace } from '@temporalio/common/lib/type-helpers';
@@ -257,9 +264,12 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
     validateActivityOptions(input.options);
 
     try {
-      const resp = await this.workflowService.startActivityExecution(
-        await this.buildStartActivityExecutionRequest(input)
-      );
+      const req = await this.buildStartActivityExecutionRequest(input);
+      const externalStorage = this.dataConverter.externalStorage;
+      if (externalStorage) {
+        await visit(req, walkStartActivityExecutionRequest, extstoreStoreOptions(externalStorage));
+      }
+      const resp = await this.workflowService.startActivityExecution(req);
       return this.createHandle(input.options.id, resp.runId);
     } catch (err) {
       if (isGrpcServiceError(err) && err.code === grpcStatus.ALREADY_EXISTS) {
@@ -328,6 +338,10 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
 
       try {
         const resp = await this.workflowService.pollActivityExecution(req);
+        const externalStorage = this.dataConverter.externalStorage;
+        if (externalStorage) {
+          await visit(resp, walkPollActivityExecutionResponse, extstoreRetrieveOptions(externalStorage));
+        }
         if (resp.outcome?.result) {
           const [result] = await decodeArrayFromPayloads(this.dataConverter, resp.outcome.result.payloads ?? []);
           return result;
@@ -362,6 +376,10 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
         activityId: input.activityId,
         runId: input.activityRunId || undefined,
       });
+      const externalStorage = this.dataConverter.externalStorage;
+      if (externalStorage) {
+        await visit(resp, walkDescribeActivityExecutionResponse, extstoreRetrieveOptions(externalStorage));
+      }
       return buildActivityDescription(resp.info!, this.dataConverter);
     } catch (err) {
       this.rethrowGrpcError(err, 'Failed to describe activity');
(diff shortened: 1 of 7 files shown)


cconstable:
Fixing integration tests...

packages/client/src/async-completion-client.ts
```diff
@@ -155,21 +169,30 @@ export class AsyncCompletionClient extends BaseClient {
...
-        await this.workflowService.respondActivityTaskCompleted({
+        const req: temporal.api.workflowservice.v1.IRespondActivityTaskCompletedRequest = {
           identity: this.options.identity,
           namespace: this.options.namespace,
           taskToken: taskTokenOrFullActivityId,
           result: { payloads },
-        });
+        };
+        if (externalStorage) {
+          await visit(req, walkRespondActivityTaskCompletedRequest, extstoreStoreOptions(externalStorage));
```
jmaeagle99: Each of these visit calls needs to set the initialContext to the appropriate storage target info. Tests should validate that the storage target info was correctly provided to the store calls.

packages/client/src/activity-client.ts
```diff
@@ -257,9 +264,12 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
     try {
-      const resp = await this.workflowService.startActivityExecution(
-        await this.buildStartActivityExecutionRequest(input)
-      );
+      const req = await this.buildStartActivityExecutionRequest(input);
+      const externalStorage = this.dataConverter.externalStorage;
+      if (externalStorage) {
+        await visit(req, walkStartActivityExecutionRequest, extstoreStoreOptions(externalStorage));
+      }
+      const resp = await this.workflowService.startActivityExecution(req);
```
mjameswh, line 283: Out of curiosity, was there any specific reason not to do the external storage transformations inside the existing encode/decode methods (i.e. `this.buildStartActivityExecutionRequest()` in this case, but same pattern everywhere below)?

To be clear, I'm not asking you to do that. Only thinking ahead what this code could look like if we were to uniformize payload conversion/storage/codec across everywhere.

jmaeagle99: I think we'll want to go the other way and unify transform application through visitors. We don't have to special case field visiting with individual calls. And system nexus operation envelopes already require visiting; more of that will need to be handled when envelopes with payloads are returned from server as results and when we add client support.

jmaeagle99: That being said, we probably want to still keep the codec visit separate from the storage visit since we want storage to be able to run concurrently.

mjameswh: > I think we'll want to go the other way and unify transform application through visitors.

Definitely. Unifying to visitors is what I have in mind.

My question is simply whether there are non-obvious constraints that impose differentiating the storage vs codec at this specific level, or if that is a purely stylistic decision (which I'd be totally ok).

packages/client/src/async-completion-client.ts
cconstable: Adding

packages/client/src/activity-client.ts
cconstable: +1 to eventually unifying to visitors.
This is purely stylistic. Open to change though.

packages/client/src/activity-client.ts
```diff
@@ -362,6 +387,10 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
         activityId: input.activityId,
         runId: input.activityRunId || undefined,
       });
+      const externalStorage = this.dataConverter.externalStorage;
+      if (externalStorage) {
+        await visit(resp, walkDescribeActivityExecutionResponse, extstoreRetrieveOptions(externalStorage));
```
jmaeagle99, line 392: This should probably be folded into the promises within buildActivityDescription so that the caller doesn't pay for retrieval unless they explicitly access the involved fields. This can probably be updated in another change, unless you think it's easy to fix now. Might require more `walk*` variants.

packages/client/src/async-completion-client.ts
```diff
@@ -155,21 +180,39 @@ export class AsyncCompletionClient extends BaseClient {
       this.serializationContextFor(taskTokenOrFullActivityId, options),
       [result]
     );
+    const externalStorage = this.dataConverter.externalStorage;
+    const target = this.storageTargetFor(taskTokenOrFullActivityId);
```
jmaeagle99: nit: maybe move these to inside the `if (externalStorage)` block so it's not created unconditionally.

packages/client/src/async-completion-client.ts
jmaeagle99, line 407: `handleError` is going to obfuscate the error by raising `ActivityCompletionError('Unexpected failure')`. Maybe we can move the request creation and external storage calls outside of the try...catch. Maybe something we can defer and fix later.

packages/client/src/activity-client.ts
cconstable: I would prefer to defer this to make sure I understand the tradeoffs of adding new walk variants (and I don't want it to delay this release). Maybe there is a more elegant way to solve the lazy vs eager walking in a consistent way across the sdk too.

packages/client/src/async-completion-client.ts
cconstable: I fixed this in `handleError`... it rethrows non-gRPC errors as-is instead of masking them as `ActivityCompletionError`

packages/client/src/async-completion-client.ts
```diff
@@ -120,6 +134,17 @@ export class AsyncCompletionClient extends BaseClient {
       : activityContextFromFullActivityId(this.options.namespace, taskTokenOrFullActivityId);
   }
 
+  /**
+   * Storage target for offloaded payloads. By-ID operations carry the Activity ID; by-token operations
+   * only know the namespace (the token is opaque), so the target is namespace-scoped and the driver falls
+   * back to content-addressed naming.
+   */
+  protected storageTargetFor(taskTokenOrFullActivityId: Uint8Array | FullActivityId): StorageDriverActivityInfo {
```
jmaeagle99: This should be more like:

```typescript
protected storageTargetFor(taskTokenOrFullActivityId: Uint8Array | FullActivityId): StorageDriverTargetInfo {
  const namespace = this.options.namespace;
  if (taskTokenOrFullActivityId instanceof Uint8Array) {
    return { kind: 'workflow', namespace };
  }
  const { workflowId, runId, activityId } = taskTokenOrFullActivityId;
  if (workflowId != null) {
    return { kind: 'workflow', namespace, id: workflowId, ...(runId != null ? { runId } : {}) };
  }
  return { kind: 'activity', namespace, id: activityId, ...(runId != null ? { runId } : {}) };
}
```

Workflow activities should store their results within the workflow key-space target. Standalone activities should store their results within the activity key-space target.

If it's just a token, we don't really know, but it's more likely to be a workflow activity. Otherwise, we can pivot on the existence of the workflowId to determine which key-space target to use.

packages/client/src/async-completion-client.ts
jmaeagle99: @chris-olszewski @mjameswh what do you think of this?

jmaeagle99: Decided to revert this for now and will address in a separate change.

jmaeagle99 (approved):

packages/client/src/async-completion-client.ts
chris-olszewski: If we go that route we should make sure to update the doc comment for `ActivityCompletionError`:
```
/**
 * Thrown by {@link AsyncCompletionClient} when trying to complete or heartbeat
 * an Activity for any reason apart from {@link ActivityNotFoundError}.
 */
```
as that change in rethrowing non-gRPC errors could affect all users, I would prefer we not do that without some additional consideration. I think deferring addressing the error obfuscation is okay.

PR #2241, opened 2026-07-22
feat(exstore): add tmprl1105 error handling for when reference payloads are encountered without extstore configured.

cconstable:
## What was changed

When an inbound payload is an external-storage reference but no ExternalStorage is configured to resolve it, we now fail with `ExternalStorageNotConfiguredError` (TMPRL1105).

- New `ExternalStorageNotConfiguredError` (common/src/errors.ts).
- New extstoreInboundOptions(externalStorage?): the single inbound decision point
- Wired unconditionally at every inbound boundary: worker (workflow / activity / nexus tasks) and client (getHistory / query / update / multi-op / poll-update).
- Moved the activity/nexus inbound pass out of the poll loops into the task handlers, so a throw fails the task instead of crashing the poller.

## Why

- Trade-off: the inbound walk runs even when extstore is off (a read-only pass over an already-decoded message), in exchange for failing loudly instead of mis-decoding a reference.

## Checklist

- Testing: 46 extstore unit tests pass, incl. inbound retrieve-vs-throw dispatch; build + lint clean.

### DIFF
--- a/packages/common/src/internal-non-workflow/external-storage-visitor.ts
+++ b/packages/common/src/internal-non-workflow/external-storage-visitor.ts
@@ -53,11 +53,7 @@ export function extstoreStoreOptions(
   };
 }
 
-/**
- * @internal
- * @experimental
- */
-export function extstoreRetrieveOptions(
+function extstoreRetrieveOptions(
   externalStorage: ExternalStorage,
   { limit, abortSignal }: { limit?: ConcurrencyLimit; abortSignal?: AbortSignal } = {}
 ): VisitOptions<void> {
@@ -73,14 +69,11 @@ export function extstoreRetrieveOptions(
 }
 
 /**
- * Options for a detection-only inbound walk used when no {@link ExternalStorage} is configured.
- * The walk leaves every payload untouched but throws {@link ExternalStorageNotConfiguredError}
- * on the first reference payload it encounters.
- *
- * @internal
- * @experimental
+ * Detection-only inbound walk used when no {@link ExternalStorage} is configured: leaves every
+ * payload untouched but throws {@link ExternalStorageNotConfiguredError} on the first reference
+ * payload.
  */
-export function extstoreDetectReferencesOptions(): VisitOptions<void> {
+function extstoreDetectReferencesOptions(): VisitOptions<void> {
   const assertNoReference = (payloads: Payload[]): Payload[] => {
     if (payloads.some(isReferencePayload)) {
       throw new ExternalStorageNotConfiguredError();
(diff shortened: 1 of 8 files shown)


packages/worker/src/worker.ts
```diff
@@ -1297,7 +1298,7 @@ export class Worker {
                 throw new IllegalStateError(`Got empty task for task variant with token: ${base64TaskToken}`);
               }
               const requestDeadline = task.requestDeadline != null ? tsToDate(task.requestDeadline) : undefined;
-              return await this.handleNexusRunTask(task.task, base64TaskToken, protobufEncodedTask, requestDeadline);
+              return await this.handleNexusRunTask(task, base64TaskToken, protobufEncodedTask, requestDeadline);
```
cconstable: The inbound extstore pass now runs inside `handleNexusRunTask` and that pass needs to walk the payloads with `walkNexusTask` to check for reference payloads and that walk is rooted at the whole `NexusTask` (`coresdk.nexus.INexusTask`) not the inner `PollNexusTaskQueueResponse` which we were previously passing.

packages/worker/src/worker.ts
```diff
@@ -1380,18 +1375,20 @@ export class Worker {
   }
 
   private async handleNexusRunTask(
-    task: temporal.api.workflowservice.v1.IPollNexusTaskQueueResponse,
+    nexusTask: coresdk.nexus.INexusTask,
     base64TaskToken: string,
     protobufEncodedTask: ArrayBuffer,
     requestDeadline: Date | undefined
   ) {
+    const task = nexusTask.task!; // Caller guarantees the 'task' variant is populated.
```
jmaeagle99: Maybe just throw nexus.HandlerError just like a few line below to be absolutely sure?

packages/common/src/internal-non-workflow/external-storage-visitor.ts
```diff
@@ -68,3 +71,37 @@ export function extstoreRetrieveOptions(
...
+
+/**
+ * Options for a detection-only inbound walk used when no {@link ExternalStorage} is configured.
+ * The walk leaves every payload untouched but throws {@link ExternalStorageNotConfiguredError}
+ * on the first reference payload it encounters.
+ *
+ * @internal
+ * @experimental
+ */
+export function extstoreDetectReferencesOptions(): VisitOptions<void> {
```
jmaeagle99: Let's not export this one and update the tests to use extstoreInboundOptions and pass undefined. That way we are actually testing the consumed API instead of the assumed internal implementation in the test.

packages/common/src/internal-non-workflow/external-storage-visitor.ts
jmaeagle99: Maybe don't export this anymore if there are no other callers.

jmaeagle99 (approved):

packages/worker/src/worker.ts
cconstable: 👍

jmaeagle99 (approved):

cconstable:
Bad rebase. Fixing.

PR #2248, opened 2026-07-23
feat(workflow): support TypeInfo for workflow transitions

THardy98:
## What was changed

Added experimental `TypeInfo` support to Child Workflow calls and continue-as-new. Child Workflows resolve definition-supplied type information or accept it when called by workflow type string; their inputs are encoded and results decoded with it. Same-type continue-as-new reuses the current Workflow's input type information, while transitions to another Workflow type accept explicit input type information.

This PR builds on the Workflow Client `TypeInfo` support merged in #2252.

## Why?

Without this change, type-aware conversion stops at the initial Workflow run. Application values passed to Child Workflows or later runs cannot use the transfer conversion established for Client-started Workflows.

Conversions without `TypeInfo` are unchanged. No rollout or manual steps are required.

## Checklist

1. Closes: N/A

2. How was this tested: Workflow and Test builds; 13 focused `TypeInfo` integration tests; ESLint; Prettier; and `ts-prune`.


packages/workflow/src/workflow.ts
```diff
@@ -1071,13 +1086,19 @@ export function makeContinueAsNewFunc<F extends Workflow>(
...
+  }
+
   return (...args: Parameters<F>): Promise<never> => {
     const context = currentWorkflowSerializationContext(info);
     const fn = composeInterceptors(activator.interceptors.outbound, 'continueAsNew', async (input) => {
       const { headers, args, options } = input;
       throw new ContinueAsNew({
         workflowType: options.workflowType,
-        arguments: toPayloadsWithContext(activator.payloadConverter, context, args),
+        arguments: toPayloadsWithContext(activator.payloadConverter, context, args, resolvedTypeInfo?.inputTypes),
```
cconstable: shouldn't this be the type info be read from the `options` object? that way if an interceptor modifies it, we will use the modified value.

cconstable (changes requested):
1 question, but lgtm

packages/workflow/src/workflow.ts
cconstable: It's a little hard to understand with all the nested closures but I think `requiredOptions` is what is passed to the interceptors and it has a `typeInfo` field which means someone writing an interceptor might think they can change that but once the interceptor chain has finished it looks like we don't actually use the potentially modified type info. I don't have enough context to know if that is a bug or desired behavior.

THardy98: oh good catch, i missed this too with the nested closure

At least for the sake of consistency this is worth doing (interceptors can modify type info in the case of workflows/child workflows already). One plausible use case for supporting this is if an interceptor modifies the calling workflow (i.e. `workflowType` is modified by the interceptor, in which case, you'd likely want to provide the corresponding type info)

packages/workflow/src/workflow.ts
```diff
@@ -979,15 +987,22 @@ export async function executeChild<T extends Workflow>(
     'Workflow.executeChild(...) may only be used from a Workflow Execution. Consider using Client.workflow.execute(...) instead.'
   );
   const optionsWithDefaults = addDefaultWorkflowOptions(options ?? ({} as any));
-  const workflowType = extractWorkflowType(workflowTypeOrFunc);
+  const { type: workflowType, typeInfo } = extractWorkflowTypeAndConfig(
+    workflowTypeOrFunc,
+    optionsWithDefaults.typeInfo
+  );
```
cconstable, line 995: Is this logic correct?

1. inside `extractWorkflowTypeAndConfig`, if `workflowTypeOrFunc` is a function and the `typeInfo` exists, this function throws a `TypeError`
2. a `TypeError` means a failed workflow task which means the workflow task may be retried (and would fail again in the same way)
3. one of the overloaded `executeChild` functions takes a `workflowFunc: T,` and an `options: WithWorkflowArgs<T, ChildWorkflowOptions>` and eventually reaches this path
4. you could potentially call that overload, pass in a `typeInfo`, get to this line, and then have a `TypeError` thrown which would cause the workflow task to fail and it to retry until the retry policy stops it.

If that checks out, I think we could use `Omit<ChildWorkflowOptions, 'typeInfo'>` to force the overload that takes a function to not accept a `typeInfo`. That would make this a compile time error rather than a runtime one.

THardy98: Yeah - I think that checks out. This is a nice improvement, I'll add it here for child workflows and open a separate PR to make this consistent across the other workflow start APIs.

Appreciate the find!

packages/workflow/src/workflow.ts
THardy98: Addressed in https://github.com/temporalio/sdk-typescript/pull/2248/commits/d7456c1ff6a98adf7fab3cbdcf917f592871e43e

packages/workflow/src/workflow.ts
THardy98: Addressed in https://github.com/temporalio/sdk-typescript/pull/2248/commits/4659fff2143fa3a289ecb11243c81337374f9078

cconstable (approved):
lgtm!

PR #2249, opened 2026-07-23
fix(common): preserve zero optional durations

### LINKED ISSUE #2233: msOptionalToTs() drops Duration=0 (falsy guard bug) — silently ignores nextRetryDelay:0 on ApplicationFailure
## Bug

`msOptionalToTs()` in `packages/common/src/time.ts` (line 71) uses a falsy truthiness guard on a `Duration` parameter that can legally be the number `0`:

```ts
export function msOptionalToTs(str: Duration | undefined | null): Timestamp | undefined {
  return str ? msToTs(str) : undefined;  // ← 0 is falsy → silently drops valid zero Duration
}
```

`Duration` includes `number`, and `0` is a valid duration meaning "zero milliseconds / no delay". Because `0` is falsy in JavaScript, `msOptionalToTs(0)` returns `undefined` instead of a zero `Timestamp`.

## Impact

The most impactful call site is `packages/common/src/converter/failure-converter.ts` line 355:

```ts
nextRetryDelay: msOptionalToTs(err.nextRetryDelay),
```

If an `ApplicationFailure` is thrown with `nextRetryDelay: 0` (intent: retry this activity immediately with no backoff), `msOptionalToTs(0)` returns `undefined`, the proto field is omitted, and the Temporal server falls back to normal exponential backoff — silently ignoring the user's override.

## Inconsistency with sibling function

The sibling function `msOptionalToNumber` (line 74) already uses the correct explicit check:

```ts
export function msOptionalToNumber(val: Duration | undefined): number | undefined {
  if (val === undefined) return undefined;  // ← explicit check, handles 0 correctly
  return msToNumber(val);
}
```

## Suggested fix

```ts
export function msOptionalToTs(str: Duration | undefined | null): Timestamp | undefined {
  return str != null ? msToTs(str) : undefined;
}
```

This matches the intent (skip only `null`/`undefined`) and handles `0` correctly.

sidsri14:
## What changed

Updated `msOptionalToTs` to treat only `null` and `undefined` as absent values, so a numeric `0` duration is preserved and converted to a zero protobuf timestamp.

Added a unit test covering `msOptionalToTs(0)`.

Fixes #2233.

## Validation

- `npx --yes pnpm@10.27.0 --filter @temporalio/proto build`
- `npx --yes pnpm@10.27.0 --filter @temporalio/common build`
- `npx --yes pnpm@10.27.0 --filter @temporalio/common test`
- `npx --yes pnpm@10.27.0 exec eslint packages/common/src/time.ts packages/common/src/__tests__/test-time.ts`
- `npx --yes pnpm@10.27.0 prettier --check packages/common/src/time.ts packages/common/src/__tests__/test-time.ts`


mjameswh:
@sidsri14 May you please [sign the CLA](https://github.com/temporalio/sdk-typescript/pull/2249#issuecomment-5060581161)?

I can't have a look at your contribution until the CLA is cleared.

mjameswh (approved):

PR #2250, opened 2026-07-23
feat(common): add TypeInfo-aware payload conversion

THardy98:
## What was changed

Added the experimental `TypeInfo` and `TransferTypeConverter` APIs so applications can map domain objects to serialization-friendly transfer values.

The new `toPayloadWithTypeInfo` and `fromPayloadWithTypeInfo` helpers apply transfer conversion around the configured `PayloadConverter`. Existing positional payload and codec helpers accept optional TypeInfo and apply it by argument position.

## Why?

Application-to-transfer conversion is independent of the payload format. Keeping it outside payload converters allows JSON and other general-purpose converters to remain unaware of application classes while giving SDK boundaries one shared implementation of the conversion order.

Existing calls without TypeInfo are unchanged. This API is experimental and requires no rollout or manual steps.

## Checklist

1. Closes: N/A

2. How was this tested: Common build and test suite; focused JSON transfer conversion coverage; ESLint; Prettier; and `ts-prune`.

### DIFF
--- a/packages/common/src/converter/payload-converter.ts
+++ b/packages/common/src/converter/payload-converter.ts
@@ -1,6 +1,7 @@
 import { decode, encode } from '../encoding';
 import { PayloadConverterError, ValueError } from '../errors';
 import type { Payload } from '../interfaces';
+import type { TypeInfo } from '../type-info';
 import type { SerializationContext } from './serialization-context';
 import { encodingKeys, encodingTypes, METADATA_ENCODING_KEY } from './types';
 
@@ -29,6 +30,38 @@ export interface PayloadConverter {
 }
 
 /**
+ * Apply TypeInfo transfer conversion before converting a value to a Payload.
+ *
+ * @experimental
+ */
+export function toPayloadWithTypeInfo<T>(
+  converter: PayloadConverter,
+  value: T,
+  context: SerializationContext | undefined,
+  typeInfo: TypeInfo<T> | undefined
+): Payload {
+  const transferValue = typeInfo?.transferTypeConverter ? typeInfo.transferTypeConverter.toTransferType(value) : value;
+  return converter.toPayload(transferValue, context);
+}
+
+/**
+ * Convert a Payload and then apply TypeInfo transfer conversion to the result.
+ *
+ * @experimental
+ */
+export function fromPayloadWithTypeInfo<T>(
+  converter: PayloadConverter,
+  payload: Payload,
+  context: SerializationContext | undefined,
+  typeInfo: TypeInfo<T> | undefined
+): T {
+  const transferValue = converter.fromPayload<unknown>(payload, context);
+  return typeInfo?.transferTypeConverter
+    ? typeInfo.transferTypeConverter.fromTransferType(transferValue)
+    : (transferValue as T);
+}
+
+/**
  * Implements conversion of a list of values.
  *
  * @param converter
@@ -49,13 +82,14 @@ export function toPayloads(converter: PayloadConverter, ...values: unknown[]): P
 export function toPayloadsWithContext(
   converter: PayloadConverter,
   context: SerializationContext | undefined,
-  values: unknown[]
+  values: unknown[],
+  typeInfo?: readonly TypeInfo[]
 ): Payload[] | undefined {
   if (values.length === 0) {
     return undefined;
   }
 
-  return values.map((value) => converter.toPayload(value, context));
+  return values.map((value, index) => toPayloadWithTypeInfo(converter, value, context, typeInfo?.[index]));
 }
 
 /**
@@ -101,7 +135,8 @@ export function fromPayloadsAtIndex<T>(
   converter: PayloadConverter,
   index: number,
   payloads?: Payload[] | null,
-  context?: SerializationContext
+  context?: SerializationContext,
+  typeInfo?: TypeInfo<T>
 ): T {
   // To make adding arguments a backwards compatible change
   if (payloads === undefined || payloads === null || index >= payloads.length) {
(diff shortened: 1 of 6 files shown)


packages/common/src/converter/payload-converter.ts
```diff
@@ -28,6 +29,26 @@ export interface PayloadConverter {
   fromPayload<T>(payload: Payload, context?: SerializationContext): T;
 }
 
+function toPayloadWithTypeInfo(
+  converter: PayloadConverter,
+  value: unknown,
+  context: SerializationContext | undefined,
+  typeInfo: TypeInfo | undefined
+): Payload {
+  const intermediate = typeInfo?.mapper ? typeInfo.mapper.toIntermediate(value) : value;
```
tconley1428: Is it easy to validate that this is always used in TS. It seems awfully easy to accidentally use the converter more directly and skip the mapper.

packages/common/src/type-info.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Type information used to map an application value to an intermediate form appropriate for serialization.
+ *
+ * @experimental
+ */
+export interface TypeInfo<T = unknown> {
+  mapper?: TypeMapper<T>;
+}
+
+/** @experimental */
+export interface TypeMapper<T> {
```
tconley1428: I think we should mimic the terminology from https://github.com/temporalio/sdk-python/pull/1655

packages/common/src/converter/payload-converter.ts
THardy98: How do you mean?
I've plugged it into the existing flow when we decode/encode payloads in `codec-helpers`

If you want validation like an integration test with a workflow, the workflow API changes I put after this

packages/common/src/type-info.ts
THardy98: done

packages/common/src/converter/payload-converter.ts
THardy98: Or were you thinking of a wrapper so `TypeInfo` handling isn’t duplicated across call sites? We could wrap the loaded `PayloadConverter` once in an internal `TypeInfoAwarePayloadConverter`. SDK conversion paths would pass that wrapper the applicable `TypeInfo`, and the wrapper would exclusively own transfer conversion and converter-hint forwarding. Is that what you had in mind?

tconley1428: That's what I did in Python, we need to ensure in happens in every use of the payload converters, which largely isn't in codec-helpers as far as I know. If there's another way to ensure some level of safety in TS, go for it. But that's what I was pushing on. Even if we go through all the points where it is used and make it use these methods, what about when we add some new thing which uses a payload converter? Will it do the same?

THardy98: see: `packages/common/src/converter/type-info-aware-payload-converter.ts`
can follow commit-by-commit for clear logical boundaries

this felt like the best we could do without proliferating a large change

packages/common/src/converter/payload-converter.ts
```diff
@@ -49,12 +51,17 @@ export function toPayloads(converter: PayloadConverter, ...values: unknown[]): P
...
-  values: unknown[]
+  values: unknown[],
+  typeInfo?: readonly TypeInfo[]
 ): Payload[] | undefined {
   if (values.length === 0) {
     return undefined;
   }
 
+  if (typeInfo !== undefined) {
+    const typeInfoAwareConverter = TypeInfoAwarePayloadConverter.wrap(converter);
```
tconley1428: I don't think the converter provides much benefit if we are still just wrapping in this API, it is the same as just using a function. It would only make sense to me if it wraps when the payloadconverter is first produced/provided by the user, if that is a narrower point at which to insert it. Possibly not

THardy98: yeah - i had a similar thought, but i really don't see a better way to about it

could remove

THardy98: removed

tconley1428 (approved):

packages/common/src/index.ts
```diff
@@ -31,6 +31,7 @@ export * from './priority';
 export * from './metrics';
 export * from './retry-policy';
 export type { Timestamp, Duration, StringValue } from './time';
+export { TransferTypeConverter, TypeInfo } from './type-info';
```
chris-olszewski: Not that it is impactful, but I think there is only type information to be reexported.

```suggestion
export type { TransferTypeConverter, TypeInfo } from './type-info';
```

packages/common/src/converter/payload-converter.ts
```diff
@@ -49,13 +82,14 @@ export function toPayloads(converter: PayloadConverter, ...values: unknown[]): P
...
-  values: unknown[]
+  values: unknown[],
+  typeInfo?: readonly TypeInfo[]
 ): Payload[] | undefined {
   if (values.length === 0) {
     return undefined;
   }
 
-  return values.map((value) => converter.toPayload(value, context));
+  return values.map((value, index) => toPayloadWithTypeInfo(converter, value, context, typeInfo?.[index]));
```
chris-olszewski, line 92: Slightly concerned about failure mode if there are more values that type infos. There won't be any real signal until someone attempts to call/use a rich type on the receiving end. There are probably valid reasons to have "mixed use", so throwing on value/typeInfo length mismatch when the latter is present is probably overkill.

(Nothing to change, just thinking out loud)

chris-olszewski (approved):

packages/common/src/converter/payload-converter.ts
THardy98: Yeah - I can definitely see scenarios where there is a rich type value (i.e. class) as param 1 and maybe a primitive (string/number) as param 2. I didn't want to validate on that for the exact reason you provided. Might be worth validating that values >= type info though, but I'm not really sure if that buys much

PR #2251, opened 2026-07-23
feat(common): add TypeInfo converter hints

THardy98:
## What was changed

Extended `TypeInfo<T, D>` with an optional `ConverterHint<D>` and added optional hint parameters to payload-converter serialization methods. Payload converters may expose the experimental `validateConverterHint` method, and `CompositePayloadConverter` uses it to select or validate the converter that receives a hint.

The TypeInfo conversion helpers now forward hints after applying transfer conversion. This is PR 2 of the stack and depends on #2250; Nexus operation integration follows separately.

## Why?

Transfer conversion is sufficient for formats such as JSON, but some payload converters require format-specific runtime information. Protobuf deserialization, for example, needs the message type in addition to the decoded bytes.

Conversions without a hint retain their existing best-effort behavior. This API is experimental and requires no rollout or manual steps.

## Checklist

1. Closes: N/A

2. How was this tested: Common build and all 38 Common tests; focused JSON transfer conversion and Protobuf converter-hint coverage; ESLint; Prettier; and `ts-prune`.

### DIFF
--- a/packages/common/src/converter/payload-converter.ts
+++ b/packages/common/src/converter/payload-converter.ts
@@ -50,7 +50,7 @@ export function toPayloadWithTypeInfo<T>(
   typeInfo: TypeInfo<T, unknown> | undefined
 ): Payload {
   const transferValue = typeInfo?.transferTypeConverter ? typeInfo.transferTypeConverter.toTransferType(value) : value;
-  return converter.toPayload(transferValue, context, typeInfo?.hint);
+  return converter.toPayload(transferValue, context, typeInfo?.converterHint);
 }
 
 /**
@@ -64,7 +64,7 @@ export function fromPayloadWithTypeInfo<T>(
   context: SerializationContext | undefined,
   typeInfo: TypeInfo<T, unknown> | undefined
 ): T {
-  const transferValue = converter.fromPayload<unknown>(payload, context, typeInfo?.hint);
+  const transferValue = converter.fromPayload<unknown>(payload, context, typeInfo?.converterHint);
   return typeInfo?.transferTypeConverter
     ? typeInfo.transferTypeConverter.fromTransferType(transferValue)
     : (transferValue as T);
--- a/packages/common/src/converter/payload-converter.ts
+++ b/packages/common/src/converter/payload-converter.ts
@@ -1,7 +1,7 @@
 import { decode, encode } from '../encoding';
 import { PayloadConverterError, ValueError } from '../errors';
 import type { Payload } from '../interfaces';
-import type { TypeInfo } from '../type-info';
+import type { ConverterHint, TypeInfo } from '../type-info';
 import type { SerializationContext } from './serialization-context';
 import { encodingKeys, encodingTypes, METADATA_ENCODING_KEY } from './types';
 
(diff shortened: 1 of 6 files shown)


packages/common/src/type-info.ts
```diff
@@ -1,17 +1,55 @@
...
+   * This transformation runs outside the payload converter and should not depend on its serialization format.
+   */
   transferTypeConverter?: TransferTypeConverter<T>;
+
+  /**
+   * Metadata forwarded unchanged to the payload converter.
+   *
+   * Use this when conversion requires format-specific runtime information, such as a Protobuf message type.
+   */
+  converterHint?: ConverterHint<D>;
```
tconley1428: I think we should clarify the name of this since it is specifically for payload converters, not transfer type converters defined right above.

THardy98: Renamed to `payloadConverterHint`

packages/common/src/type-info.ts
```diff
@@ -1,17 +1,55 @@
...
+ *
+ * On encoding, transfer type conversion runs before payload conversion. On decoding, payload conversion runs before
+ * transfer type conversion. Either mechanism may be used independently or they may be combined.
+ *
+ * SDK conversion helpers apply this information when supplied. Calling a {@link PayloadConverter} directly does not.
  *
  * @experimental
  */
-export interface TypeInfo<T = unknown> {
+export interface TypeInfo<T = unknown, D = T> {
```
chris-olszewski, line 19: Should `D` be passed as a generic to `TransferTypeConverter` so the mapping function typings line up with the type hint?

packages/common/src/converter/payload-converter.ts
```diff
@@ -21,12 +21,21 @@ export interface PayloadConverter {
...
    * Should throw {@link ValueError} if unable to convert.
    */
-  toPayload<T>(value: T, context?: SerializationContext): Payload;
+  toPayload<T>(value: T, context?: SerializationContext, hint?: ConverterHint): Payload;
 
   /**
    * Converts a {@link Payload} back to a value.
    */
-  fromPayload<T>(payload: Payload, context?: SerializationContext): T;
+  fromPayload<T>(payload: Payload, context?: SerializationContext, hint?: ConverterHint): T;
```
chris-olszewski: Should we restrict converter hints to ones that are branded with the expected type? I can't think of a reason we would want to be more permissive here.
```suggestion
  fromPayload<T>(payload: Payload, context?: SerializationContext, hint?: ConverterHint<T>): T;
```

chris-olszewski (commented):
It would be nice to have a test that exercises a `TypeInfo` with both a transfer type converter and a converter hint.

packages/common/src/converter/payload-converter.ts
THardy98: Yup - added in last commit

packages/common/src/type-info.ts
THardy98: Yeah I think so - i've added in the last commit.

chris-olszewski (approved):

PR #2252, opened 2026-07-23
feat(workflow): support TypeInfo for client-started workflows

THardy98:
## What was changed

Added `PayloadTypeInfo` and `defineWorkflowOptions` metadata for client-started Workflows. Workflow functions use definition metadata, while string Workflow types can supply metadata at the call site; start, execute, result, detached-handle, signal-with-start, and update-with-start paths carry it through conversion.

This is PR 3 of the stack and depends on #2251. Workflow-internal transitions follow in #2248.

## Why?

Workflow input TypeInfo must be available before arguments are decoded in the Workflow isolate. Definition-time metadata provides it for typed function references while preserving a call-site option for string Workflow types.

Workflows without TypeInfo are unchanged. Providing call-site metadata with a Workflow function is rejected to avoid conflicting sources; no rollout or manual steps are required.

## Checklist

1. Closes: N/A

2. How was this tested: Common, Client, Worker, Workflow, and test package builds; ESLint; Prettier; 10 focused unit tests; and eight mapper-only JSON integration scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ to docs, or any other relevant information.
 
 ## [Unreleased]
 
+### Added
+
+- **Experimental**: Workflow Clients can now use `TypeInfo` to encode Workflow inputs and decode Workflow results.
+
 ### Changed
 
 - Nexus is now generally available (GA) for calling Nexus Operations from Workflows and handling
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -24,7 +24,8 @@ to docs, or any other relevant information.
 ### Added
 
 - **Experimental**: Added `TypeInfo` and `TransferTypeConverter` to `@temporalio/common` for converting
```
chris-olszewski, line 35: I think this might need a new section now that 1.22 has been released

chris-olszewski (approved):

PR #2254, opened 2026-07-23
fix(strands): ignore @aws-sdk/client-s3 in bundler

brianstrauch:
The S3-backed context-offloader vended plugin dynamically imports
@aws-sdk/client-s3 worker-side; add it to the ignore list so workflow
bundling does not attempt to resolve it.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


chris-olszewski (commented):
Please add a `CHANGELOG.md` entry for this change

CHANGELOG.md
```diff
@@ -64,6 +64,12 @@ to docs, or any other relevant information.
 - `temporal_worker_task_slots_used` no longer counts reserved but unused task slots as in use.
 - When worker heartbeats are enabled, host CPU and memory are sampled at the configured heartbeat interval rather than every 100ms.
 
+### Fixed
```
chris-olszewski, line 67: This is under the wrong section, we need it to be under `## Unreleased` which will get swapped to `1.21.1` when we cut the release.

brianstrauch: It _is_ under the unreleased section, but it looks like there's no 1.21.0 section. Is that intended?

contrib/strands/src/plugin.ts
```diff
@@ -145,6 +148,7 @@ export class StrandsPlugin extends SimplePlugin {
       '@temporalio/activity',
       '@temporalio/client',
       '@aws-sdk/client-bedrock-runtime',
+      '@aws-sdk/client-s3',
```
chris-olszewski, line 151: Can we bump the version of the strands SDK we test against so we can verify the fix?

brianstrauch: The bundling error is present for all versions of @strands-agents/sdk. Looking into if it's possible to write a test that verifies that bundling succeeds...

chris-olszewski: Nevermind, this is something different. Bumping version wouldn't have caught this.

CHANGELOG.md
chris-olszewski, line 20: Add the changelog entry here

chris-olszewski (approved):

PR #2259, opened 2026-07-24
fix(strands): don't evict an MCP connection with a call in flight

brianstrauch:
## What was wrong

The worker-process MCP connection cache in `contrib/strands/src/temporal-mcp-client.ts` armed its idle timer inside `getConnection` — that is, at the **start** of each call — and nothing extended it while the call ran. So the window measured time since the last call *started*, not since the last call *finished*.

A `callTool` or `listTools` activity that outlived `mcpConnectionIdleTimeout` (5 minutes by default, against a 10-minute default `startToCloseTimeout`) had `client.disconnect()` run underneath it, rejecting the in-flight request. Each retry was cut off the same way, so an MCP tool slower than the idle window could never succeed. A concurrent call holding the same client also continued against a closed transport.

## The fix

Track in-flight uses per connection, matching what the Python SDK already does in `temporalio/contrib/strands/_temporal_mcp_client.py`:

- `acquireConnection()` increments `inflight` and **clears** any pending idle timer.
- `releaseConnection()`, called from a `finally` in both activities, re-arms the timer only once `inflight` is 0 **and** the record is still the cached one — an already-evicted record must not schedule a timer that would later kill its healthy replacement.
- `_evictConnection()` removes the entry from the cache before awaiting `disconnect()`, so a call arriving during teardown opens a fresh connection instead of joining the dying one.

The two parallel maps (`CONNECTIONS` + `IDLE_TIMERS`) collapse into one `ConnectionRecord { client, inflight, idleTimer }`.

## Test plan

New test `mcpConnectionIdleTimeout does not evict a connection with a call in flight`: a 500ms tool call against a 50ms idle window, asserting no `disconnect()` lands while the call is running. It fails on `main` ("idle timer disconnected a connection with a call in flight") and passes with the fix.

Full strands suite passes (16/16), including the existing `mcpConnectionIdleTimeout evicts the cached connection while the worker runs` test — idle eviction still happens between activities, just not during one.


contrib/strands/src/temporal-mcp-client.ts
```diff
@@ -174,68 +175,107 @@ export function buildListToolsActivity(
...
+    clearTimeout(record.idleTimer);
+    record.idleTimer = undefined;
   }
-  resetIdleTimer(server, idleMs);
-  return entry;
+  return record;
+}
+
+/** Mark a call done; arm idle eviction once no calls remain in flight. */
+function releaseConnection(server: string, record: ConnectionRecord, idleMs: number): void {
```
patbeqo, line 264: Do we need to pass in the record? Could we fetch it from `CONNECTIONS` as the source of truth?

brianstrauch: The record can get evicted from the map during the `catch`, so since this is being called from `finally`, it may already be gone.

patbeqo: Makes sense, also covers the evicted and replaced case.

patbeqo (approved):

chris-olszewski (approved):

PR #2260, opened 2026-07-24
extstore/surface exstore errors in client

cconstable:
## What was changed

- Added new ExternalStorageError base type and a small set of high-level external storage errors
- Client surfaces that raise errors now properly handle extstore errors

## Why?

Previously, surfacing external storage errors from client logic meant that callers would see the new extstore error type which could break compatibility. This PR patches the error types that are thrown to include a `cause` where upstream errors can be put. This means we can move the extstore calls back inside the try/catch blocks, catch them meaningfully, and still throw consistent error types.

## Checklist
- Added tests

### DIFF
--- a/packages/common/src/errors.ts
+++ b/packages/common/src/errors.ts
@@ -111,8 +111,8 @@ export class ExternalStorageDriverError extends ExternalStorageError {
 }
 
 /**
- * [TMPRL-1105] Thrown when external storage is asked to use a driver that is not registered 
- * on the configured `ExternalStorage`. 
+ * [TMPRL-1105] Thrown when external storage is asked to use a driver that is not registered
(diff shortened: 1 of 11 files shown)


packages/common/src/errors.ts
```diff
@@ -74,15 +74,56 @@ export class NamespaceNotFoundError extends Error {
...
+  public readonly cause?: unknown;
+
+  constructor(message: string, opts?: { cause?: unknown }) {
+    super(message);
+    this.cause = opts?.cause;
+  }
+}
+
+/**
+ * [TMPRL-1105] Thrown when external storage is asked to use a driver that is not registered
```
cconstable: Attributed this to the wrong error. Will fix.

THardy98 (approved):

PR #2267, opened 2026-07-28
fix(workflow-streams): re-derive log offset after parking in onPoll

### LINKED ISSUE #2263: [Bug] workflow-streams: parked onPoll serves from a stale log index after truncate(), silently skipping events
## Describe the bug

`WorkflowStream.onPoll` (in `@temporalio/workflow-streams`) computes its log index from `baseOffset` once, **before** parking on `condition()`:

```js
async onPoll(input) {
    let logOffset = input.from_offset - this.baseOffset;
    if (logOffset < 0) {
        // ("from the beginning" fast-forward / TruncatedOffset validation — pre-wait only)
    }
    await condition(() => this.log.length > logOffset || this.draining);
    const allNew = this.log.slice(logOffset);
```

If `truncate(upToOffset)` runs while that poll is parked, the splice shifts every log index and advances `baseOffset`, but the parked handler still holds the pre-truncate `logOffset`. On wake it slices from the wrong position: entries are **silently skipped** — no `TruncatedOffset` failure — and the returned `next_offset` then jumps past them, so the subscriber can never recover the lost entries.

Concrete interleaving (all normal operation):

1. Log = `[A@0]`, `baseOffset = 0`. A subscriber has consumed `A`; its poll for `from_offset = 1` parks (`logOffset = 1`, waiting for `log.length > 1`).
2. The workflow, within one workflow task, publishes `B` (log `[A, B]`), calls `truncate(1)` (log `[B@1]`, `baseOffset = 1`), and publishes `C` (log `[B@1, C@2]`).
3. The parked condition evaluates: `log.length (2) > logOffset (1)` → wakes. `allNew = log.slice(1)` = `[C]`. **`B` is skipped**, and the item offsets (`baseOffset + logOffset + i` = 2) make the response look continuous.

If the poll instead *arrives* after the truncate, the pre-wait validation handles it correctly — only the parked-across-truncate interleaving corrupts the read, which is why this surfaces as a load-dependent flake.

We hit it using the "keep the terminal event" pattern (`truncate(headOffset - 1)` when starting each new work item): a subscriber parked across that boundary woke up to the *next* item's first event and never received the prior item's **terminal** event, permanently hanging the consumer, while a fast subscriber on the same run saw every event.

## Minimal reproduction

Workflow: publish `A`; await a signal; on signal publish `B`, `truncate(headOffset - 1)`, publish `C`.
Client: `subscribe(topic, 0, { pollCooldown: '10 milliseconds' })`, consume `A`, let the next poll park, then send the signal. The parked poll yields only `C` — `B` is lost with no error. Reproduces with `TestWorkflowEnvironment.createLocal()`; the signal ordering forces the publish → truncate → publish to land in one workflow task while the poll is parked.

## Suggested fix

Derive the index inside the condition and re-validate after the wait, so a truncation during the park either serves from the correctly recomputed index or fails the update with `TruncatedOffset` (which `subscribe()` already handles by restarting from the current base):

```js
async onPoll(input) {
    const truncatedFailure = () => ApplicationFailure.create({
        message: `Requested offset ${input.from_offset} has been truncated. Current base offset is ${this.baseOffset}.`,
        type: 'TruncatedOffset',
        nonRetryable: true,
    });
    if (input.from_offset !== 0 && input.from_offset - this.baseOffset < 0) {
        throw truncatedFailure();
    }
    // Index must be derived at evaluation time: truncate() splices the log and
    // advances baseOffset while polls are parked.
    await condition(
        () => this.log.length > Math.max(input.from_offset - this.baseOffset, 0) || this.draining
    );
    let logOffset = input.from_offset - this.baseOffset;
    if (logOffset < 0) {
        if (input.from_offset !== 0) throw truncatedFailure(); // truncated while parked
        logOffset = 0;
    }
    const allNew = this.log.slice(logOffset);
    ...
```

We are running exactly this as a package patch; it eliminated the skipped events (and the flaky hang) under the same workload. Happy to open a PR.

## Environment

- `@temporalio/workflow-streams` 1.19.1 (reproduced); the `onPoll` code in the latest **1.21.1** is byte-identical (verified from the published tarball), so all released versions appear affected
- `@temporalio/*` SDK 1.19.1, against both Temporal Cloud and the local dev server (`TestWorkflowEnvironment.createLocal()`)
- Node 22.x / 24.x, macOS + Linux

Bathula-Adiseshu:
onPoll captured logOffset from baseOffset before awaiting condition(...). If truncate() ran while the poll was parked, the log was shifted and baseOffset advanced, but the parked handler continued using the stale offset and silently skipped events.

This change re-derives the log offset after the wait and re-validates truncation so that a poll either:
- reads from the correct position after truncate(), or
- fails with TruncatedOffset if the requested position was removed while waiting.

Fixes #2263

## What was changed

- Changed WorkflowStream.onPoll() to calculate the effective log index after the poll wait instead of relying on the value captured before condition().
- Added re-validation after waiting to handle truncation races.
- Added a regression test covering the case where a poll is parked while publish → truncate → publish happens.

## Why?

A parked poll could wake up after truncate() and use a stale log index. This caused events to be silently skipped because the underlying log array had shifted while the poll was waiting.

## Checklist

1. Closes #2263

2. How was this tested:
- Added a regression test reproducing the parked onPoll + truncate interleaving.
- Verified TypeScript checks introduce no new errors compared to baseline.
- Verified ESLint passes.
- Integration test requires Temporal test server binary via TestWorkflowEnvironment.createLocal(); this could not be downloaded in the current environment, but the scenario was validated through the standalone reproduction.

3. Any docs updates needed?
No docs updates needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ to docs, or any other relevant information.
 ### Fixed
 
 - strands: Declare `zod` as a peer dependency.
+- workflow-streams: `WorkflowStream.onPoll` no longer serves a stale log index for a poll that
+  was parked across a `truncate()` call, which could silently skip events.
 
 ## [1.21.1] - 2026-07-23
 
--- a/contrib/workflow-streams/src/__tests__/test-workflow-streams.ts
+++ b/contrib/workflow-streams/src/__tests__/test-workflow-streams.ts
@@ -19,6 +19,7 @@ import {
 } from '../client';
 import {
   type WorkflowStreamState,
+  decodePayloadWire,
   encodePayloadWire,
   workflowStreamOffsetQuery,
   workflowStreamPublishSignal,
@@ -39,6 +40,7 @@ import {
   forceFlushWorkflow,
   publisherSequencesQuery,
   truncateUpdate,
+  truncateWhileParkedWorkflow,
   truncateWorkflow,
   ttlTestWorkflow,
   workflowSidePublishWorkflow,
@@ -247,6 +249,40 @@ test('poll_truncated_offset_returns_application_failure', async (t) => {
     const after = await collectItems(handle, undefined, 3, 2);
     t.is(after.length, 2);
     t.is(after[0]!.offset, 3);
+
+    await handle.signal('close');
+  });
+});
+
+test('poll_parked_across_truncate_sees_all_events', async (t) => {
+  // Regression for #2263: a poll parked on from_offset=1 must not miss 'B'
+  // when the workflow publishes 'B', truncates, and publishes 'C' in a
+  // single task while the poll is parked.
+  const { createWorker, startWorkflow } = helpers(t);
+  const { env } = t.context;
+  const worker = await createWorker();
+  await worker.runUntil(async () => {
+    const handle = await startWorkflow(truncateWhileParkedWorkflow, { args: [] });
+    const rawHandle = env.client.workflow.getHandle(handle.workflowId);
+
+    // Consume 'A' so the next poll starts from_offset=1.
+    const first = await rawHandle.executeUpdate<PollResult, [PollInput]>(workflowStreamPollUpdate, {
+      args: [{ topics: [], from_offset: 0 }],
+    });
+    t.is(first.items.length, 1);
+    t.is(payloadString(decodePayloadWire(first.items[0]!.data)), 'A');
+
+    // Park from_offset=1, then trigger the publish/truncate/publish race.
+    const parkedPoll = rawHandle.executeUpdate<PollResult, [PollInput]>(workflowStreamPollUpdate, {
+      args: [{ topics: [], from_offset: 1 }],
+    });
+    await handle.signal('triggerContinue');
+
+    const result = await parkedPoll;
+    t.deepEqual(
+      result.items.map((i) => payloadString(decodePayloadWire(i.data))),
+      ['B', 'C']
+    );
 
     await handle.signal('close');
   });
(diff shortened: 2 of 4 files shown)


brianstrauch:
Thank you! Taking a look.

brianstrauch:
@Bathula-Adiseshu please fix both issues flagged by Copilot and sign our CLA. We'll get this merged in shortly after!

Bathula-Adiseshu:
@brianstrauch  Thanks for the review! Addressed both points:
Updated the regression test to use startUpdate(..., { waitForStage: 'ACCEPTED' })  so the poll is deterministically parked before triggering the truncate race.
Updated the `condition()` predicate to wake when truncation moves baseOffset past from_offset, including cases where the log becomes empty, so it returns TruncatedOffset instead of hanging.

Bathula-Adiseshu:
@brianstrauch I had previously signed the CLA for the Temporal Python SDK repository, but the CLA check for this TypeScript SDK PR is still showing as pending. I have checked the CLA page t shows that I have already agreed to the CLA for repositories/organizations, but it may need to be associated with this repository. Could you please help verify the CLA status?

brianstrauch:
I wonder if the CLA check is not passing because you're using two GitHub accounts: @Bathula-Adiseshu and @adiseshu-bathula. Make sure you've signed for both. We also need a brief sentence added to the changelog and a fix for a few other jobs, thanks!

Bathula-Adiseshu:
@brianstrauch  Thanks for the feedback! I've addressed the requested changes:
Added the changelog entry.
Fixed the payload decoding in the regression test.
Fixed the CI-breaking type issue.
Rebased onto the latest `main` to keep the history clean.
The latest commits are authored as `Bathula-Adiseshu `. The PR should be ready for another review. Thanks!

brianstrauch (approved):
Thank you!

chris-olszewski (approved):

Issue #2151, opened 2026-06-30
[Feature Request] Move off Node 20 (EOL)


cconstable:
### Is your feature request related to a problem? Please describe.

We pin node >20.3.0 for the sdk packages but [Node 20 hit EOL in April](https://nodejs.org/en/about/previous-releases) and is no longer receiving updates.

### Describe the solution you'd like

- Pin an actively maintained node version (22 or 24)
- Need to evaluate impact

PR #2270, opened 2026-07-29
Cleanup pass for CI logging

cconstable:
CI test steps print 14k+ lines even on success and failures are hard to find. This PR routes all test output through a wrapper that shows only what's actionable and pipes the rest to a file.

## What was changed

**On the console:** 
one line per suite (`✓ @temporalio/test  1298 passed (312s)`), failures printed live with their diagnostic, a 30s heartbeat so long suites aren't silent. 

**On the run summary page:** 
a pass/fail/duration table plus a failures section. 

**Archived:** 
the full per-package logs as an artifact.

### Changes
- `scripts/ava-ci.mjs` (new) — `ava --tap` wrapper; per-suite summary, full log to `.test-results/<pkg>.log`
- `scripts/ci-test-summary.mjs` (new) — aggregates results into `$GITHUB_STEP_SUMMARY`
- `scripts/utils.ts` — `spawnNpxLogged` (streams to log, tail-on-failure) + `ciLogPath`
- `scripts/publish-to-verdaccio.ts`, `scripts/init-from-verdaccio.ts` — redirect ~6k lines of publish/install noise
- `packages/*` + `contrib/*` `package.json`, `contrib/ai-sdk` runner — `test` scripts call the wrapper
- `.github/workflows/ci.yml` — `TEST_RESULTS_DIR`, summary step, artifact upload
- `.gitignore` — ignore `.test-results/`

Bun's path is intentionally left as-is.

## Why?
CI test steps print 14k+ lines even on success and failures are hard to find.


cconstable:
patching to consolidate summary output and make the summary page more concise + actionable...

scripts/ava-ci.ts
```diff
@@ -0,0 +1,250 @@
+// Quiet, CI-friendly wrapper around `ava`.
```
chris-olszewski, line 1: I would love to avoid introducing untyped JS back into the codebase even in scripts. ~~Could we get away with just type stripping?
https://nodejs.org/docs/latest-v22.x/api/typescript.html#typescript-features~~ Nevermind, I see it was just added in a later version of Node 22 so blocked on #2151

scripts/ci-run-summary.ts
```diff
@@ -0,0 +1,204 @@
+// Produces ONE consolidated CI summary answering "did this PR pass, and if not,
```
chris-olszewski, line 1: Same as above, would love for this to be typed

chris-olszewski (changes requested):
Very nice changes. Navigating and finding the full logs is a little awkward, but they are still available which I think is the big thing.

Requesting to type the new scripts, we already typecheck the `scripts` package and generally use `tsx` to execute `.ts` files. It is already a root dependency for the workspace so it should be available in all the packages.

scripts/ava-ci.ts
cconstable: I could just have the CI summary job use node 22

.github/workflows/ci.yml
```diff
@@ -335,3 +363,51 @@ jobs:
...
+      - docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout code'
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Setup Node
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
+        with:
+          node-version: '22'
```
chris-olszewski: If you want to future proof this so we don't have to bump it in 6 months you could try out Node 24 or 26

package.json
```diff
@@ -32,7 +32,7 @@
     "ci-stress": "node ./packages/test/lib/load/run-all-stress-ci-scenarios.js",
     "ci-nightly": "node ./packages/test/lib/load/run-all-nightly-scenarios.js",
     "lint": "eslint --no-error-on-unmatched-pattern --fix && prettier --write . && pnpm --recursive --stream run lint",
-    "lint:check": "eslint --no-error-on-unmatched-pattern && prettier --end-of-line auto --check .  && pnpm --recursive --stream run lint:check",
+    "lint:check": "eslint --no-error-on-unmatched-pattern && prettier --end-of-line auto --check .  && tsc -p scripts/tsconfig.json && pnpm --recursive --stream run lint:check",
```
chris-olszewski: Why is this `tsc` call needed? `pnpm run build` should be performing typechecking for us

chris-olszewski (approved):

package.json
cconstable: Not needed. Thanks! Fixed.

.github/workflows/ci.yml
cconstable: Bumped to 24

PR #2279, opened 2026-08-03
Run integration tests in an isolated Cloud namespace

THardy98:
## What was changed

**TLDR**: basically a port of temporalio/sdk-python#1731.

- Adds initial `Cloud Integration Tests` CI job (creates an isolated mTLS Temporal Cloud namespace for the GHA run), executes one existing integration test through the `envconfig`-backed test harness, and deletes the namespace afterwards

- A small TypeScript CLI under `.github/scripts` manages the namespace creation/polling/deletion (basically matches Python implementation)

Currently just running `Uses default activity if no matching activity exists` in cloud, with the intention to expand it.

## Why?

The integration suite needs an isolated Cloud target before it can run reliably in CI. A namespace owned by each run prevents parallel CI jobs from sharing workflow state and gives every run an explicit cleanup boundary.

## Checklist

1. Part of temporalio/features#851.
2. How was this tested: CI job

3. Any docs updates needed?

   No. This changes internal test infrastructure only.

### DIFF
--- a/.github/scripts/cloud-namespace.ts
+++ b/.github/scripts/cloud-namespace.ts
@@ -1,0 +1,110 @@
+import { randomUUID } from 'node:crypto';
+import { appendFile, readFile } from 'node:fs/promises';
+import { CloudOperationsClient, CloudOperationsConnection } from '../../packages/cloud';
+import { temporal } from '../../packages/proto';
+
+const OPERATION_TIMEOUT_MS = 10 * 60 * 1_000;
+const OperationState = temporal.api.cloud.operation.v1.AsyncOperation.State;
+
+function requiredEnv(name: string): string {
+  const value = process.env[name];
+  if (!value) throw new Error(`Missing required environment variable ${name}`);
+  return value;
+}
+
+async function waitForOperation(
+  client: CloudOperationsClient,
+  initialOperation: temporal.api.cloud.operation.v1.IAsyncOperation
+): Promise<void> {
+  const operationId = initialOperation.id;
+  if (!operationId) throw new Error('Cloud operation response did not include an ID');
+
+  const deadline = Date.now() + OPERATION_TIMEOUT_MS;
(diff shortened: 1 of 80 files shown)


.github/workflows/ci.yml
```diff
@@ -319,6 +319,91 @@ jobs:
...
+        run: pnpm exec tsx .github/scripts/cloud-namespace.ts create
+        env:
+          TEMPORAL_CLIENT_CLOUD_API_KEY: ${{ secrets.TEMPORAL_CLIENT_CLOUD_API_KEY }}
+          TEMPORAL_CLIENT_CLOUD_API_VERSION: v0.19.1
+
+      - name: Run tests against Temporal Cloud
+        run: |
+          pnpm --dir packages/test exec ava --tap \
+            --match 'Uses default activity if no matching activity exists' \
+            ./lib/test-default-activity.js | tee cloud-integration.tap
```
THardy98, line 383: just picking a single test case for now

.github/scripts/cloud-namespace.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+async function create(): Promise<void> {
+  await withCloudClient(async (client) => {
+    const namespaceName = `sdk-typescript-ci-${requiredEnv('GITHUB_RUN_ID')}-${requiredEnv('GITHUB_RUN_ATTEMPT')}`;
+    const result = await client.cloudService.createNamespace({
+      asyncOperationId: randomUUID(),
+      spec: {
+        name: namespaceName,
+        regions: ['aws-ca-central-1'],
```
chris-olszewski, line 68: Should this be populated via an env var? I know this can change and when it does we need to update an env var in the GHA YAML.

chris-olszewski (approved):
One small q about if region should be fetched from GHA so we only need to update one place when our cell moves.

PR #1734, opened 2025-06-27
Map AggegateError and Error.cause to ApplicationFailureInfo

mdouglass:
## What was changed

Change how `ensureApplicationFailure` and the `DefaultFailureConverter` handle `Error.cause` and `AggregateError.errors`.

## Why?

- `ensureApplicationFailure` was not providing `Error.cause` to `ApplicationFailure.create` which meant that cause information was not being built

- `AggegateError.errors` wasn't being handled in either `ensureApplicationFailure` or the `DefaultFailureConverter`. It is not possible to map that directly to an `IFailure`, so that information is added to then `ApplicationFailureInfo.details` field.

## Checklist

1. Closes #1675 

2. How was this tested:
I was using an activity that looks like this:
```typescript
export async function failHard( ): Promise<void> {
  throw new AggregateError(
    [
      new Error('Analyze workflow gets all the love.', {
        cause: new Error('Analyze workflow is the money machine.'),
      }),
      new AggregateError(
        [
          new Error('Everyone knows acquire does the hard work'),
          new Error('Everyone can see what acquire does'),
        ],
        'Acquire workflow gets all the praise.',
      ),
    ],
    'Scrape workflow quits!',
    {
      cause: new AggregateError(
        [
          new Error('Scrape workflow is a drama queen'),
          new Error('Did I mention, scrape workflow is a drama queen?'),
        ],
        'Scrape workflow is a drama queen x2.',
      ),
    },
  )
}
```

Previously, this was all the information available in the Temporal web site for the error:
```json
{
  "type": "workflowExecutionFailedEventAttributes",
  "failure":
    {
      "message": "Activity task failed",
      "cause":
        {
          "message": "Scrape workflow quits!",
          "source": "TypeScriptSDK",
          "stackTrace": "AggregateError: Scrape workflow quits!\n    at getLastScrape (file:///Users/matthew/adology/adology-backend/packages/ad-scraper/src/activities/last-scrape.ts:16:9)",
          "applicationFailureInfo": { "type": "AggregateError" },
        },
      "activityFailureInfo":
        {
          "scheduledEventId": "5",
          "startedEventId": "6",
          "identity": "714@camilla.local",
          "activityType": { "name": "getLastScrape" },
          "activityId": "1",
          "retryState": "RETRY_STATE_MAXIMUM_ATTEMPTS_REACHED",
        },
    },
  "retryState": "RETRY_STATE_RETRY_POLICY_NOT_SET",
  "workflowTaskCompletedEventId": "10",
}

```

[...]

PR #2292, opened 2026-08-04
fix(common): preserve Error.cause in activity failures

### LINKED ISSUE #2291: [Bug] Activity errors lose Error.cause during conversion

Preserve the native `Error.cause` chain when an Activity throws a non-`ApplicationFailure` error.

The Activity worker first calls `ensureApplicationFailure(error)`, then sends the result through the configured failure converter. `DefaultFailureConverter` already serializes `TemporalFailure.cause`, but `ensureApplicationFailure` currently copies only the error's message, type, and stack. As a result, the cause is lost before serialization.

This affects errors from libraries that wrap their underlying error with `Error.cause`. For example, a database wrapper error may retain the SQL query in its own message while the actionable PostgreSQL error exists only in `cause`.

### Minimal reproduction

```ts
export async function activity(): Promise<void> {
  const databaseError = new Error('connection terminated unexpectedly');
  throw new Error('query failed', { cause: databaseError });
}
```

The resulting Activity failure contains the outer `ApplicationFailure("query failed")`, but not the nested `"connection terminated unexpectedly"` failure.

### Expected behavior

When converting a non-`ApplicationFailure` `Error`, `ensureApplicationFailure` should preserve `error.cause` when it is an `Error`. The existing failure converter will then serialize the nested failure using the standard Temporal failure `cause` field.

This is intentionally narrower than #1734: it does not map `AggregateError.errors`, repurpose `details`, or change protobuf schemas.

mohan-garimella:
## What changed

- Preserve a native `Error.cause` when `ensureApplicationFailure` converts an Activity error.
- Add one focused common-package test covering conversion and serialization of the nested failure.

## Why

Activity execution calls `ensureApplicationFailure` before the configured failure converter. The default converter already serializes `TemporalFailure.cause`, but the conversion step discarded the cause first. Libraries such as database clients commonly use `Error.cause` for the actionable underlying error, so Temporal history retained only the wrapper message.

This is intentionally narrower than #1734. It does not map `AggregateError.errors`, use `ApplicationFailure.details`, or change the failure protobuf.

Closes #2291

## Verification

- `pnpm -F @temporalio/proto run build`
- `pnpm -F @temporalio/common run build`
- `pnpm -F @temporalio/common run test` — 39 passed
- Prettier check on changed files
- ESLint on changed files


THardy98:
Thanks @mohan-garimella for the contribution! 

The approach is this PR is largely correct, but I've superceded it with https://github.com/temporalio/sdk-typescript/pull/2376. It just covers our bases a bit more with our `isError` helper to capture other errors outside of `instanceOf Error` and a bit more test coverage.

I'll be closing this PR because of that, but really appreciate the contribution!

PR #2294, opened 2026-08-05
feat(extstore): run store and retrieve operations concurrently (default 3, configurable)

### LINKED ISSUE #2275: [Bug] External storage does not allow for concurrent storage operations
The external storage implementation uses the payload visitor to ensure that all payloads are visited within a proto message graph. By default, the visit and payload transform is serial. The external storage visitors are not specifying a custom `limit` callback on the `ExternalStorageStoreOptions`, unlike other Temporal SDKs that support concurrent external store and retrieve options. The numeric limit should be exposed as a configurable option on the worker and have a default value of 3, just like the other SDKs.

VihaanAgarwal:
## What was changed

External storage store and retrieve calls now run concurrently during a payload walk, capped at 3 by default. The cap is configurable via a new `ExternalStorage.maxConcurrentOperations` option, validated as a positive integer.

## Why

The payload visitor already supports a concurrency limit, but the extstore option builders never passed one, so every driver call ran sequentially. Closes #2275.

## How

`extstoreStoreOptions` and `extstoreRetrieveOptions` derive their default `limit` from `ExternalStorage.maxConcurrentOperations`, so the cap applies at every store and retrieve site (worker and clients). An explicit `limit` still overrides. The knob lives on `ExternalStorage` rather than worker options so client-side walks get the same behavior and the extstore config stays in one place.

## Testing

Added unit tests that assert peak driver-call concurrency for store and retrieve walks (3 by default, configurable), plus config validation tests. The extstore unit and integration suites pass against a local dev server.

### DIFF
--- a/packages/common/src/internal-non-workflow/external-storage-visitor.ts
+++ b/packages/common/src/internal-non-workflow/external-storage-visitor.ts
@@ -4,7 +4,7 @@
  *
  * @module
  */
-import type { ConcurrencyLimit } from '../concurrency/limit';
+import { limit as concurrencyLimit, type ConcurrencyLimit } from '../concurrency/limit';
 import type { ExternalStorage, StorageDriverTargetInfo } from '../converter/extstore';
 import { ExternalStorageNotConfiguredError } from '../errors';
 import type { Payload } from '../interfaces';
@@ -25,7 +25,10 @@ export interface ExternalStorageStoreOptions {
   initialTarget?: StorageDriverTargetInfo;
   /** Derives new storage target from the current message. */
   deriveContext?: ContextDeriver<StoreTarget>;
-  /** Bounds concurrent transform calls across payload sites. Omit for sequential. */
+  /**
+   * Bounds concurrent transform calls across payload sites. Omit to derive one from
+   * {@link ExternalStorage.maxConcurrentOperations}.
+   */
   limit?: ConcurrencyLimit;
   /** Aborts the walk and every in-flight driver call. */
   abortSignal?: AbortSignal;
@@ -37,7 +40,12 @@ export interface ExternalStorageStoreOptions {
  */
 export function extstoreStoreOptions(
   externalStorage: ExternalStorage,
-  { initialTarget, deriveContext, limit, abortSignal }: ExternalStorageStoreOptions = {}
+  {
+    initialTarget,
+    deriveContext,
+    limit = concurrencyLimit(externalStorage.maxConcurrentOperations),
+    abortSignal,
+  }: ExternalStorageStoreOptions = {}
 ): VisitOptions<StoreTarget> {
   const runner = new ExternalStorageRunner(externalStorage);
   return {
@@ -55,7 +63,10 @@ export function extstoreStoreOptions(
 
 function extstoreRetrieveOptions(
   externalStorage: ExternalStorage,
-  { limit, abortSignal }: { limit?: ConcurrencyLimit; abortSignal?: AbortSignal } = {}
+  {
+    limit = concurrencyLimit(externalStorage.maxConcurrentOperations),
+    abortSignal,
+  }: { limit?: ConcurrencyLimit; abortSignal?: AbortSignal } = {}
 ): VisitOptions<void> {
   const runner = new ExternalStorageRunner(externalStorage);
   return {
--- a/packages/common/src/converter/extstore.ts
+++ b/packages/common/src/converter/extstore.ts
@@ -104,6 +104,9 @@ export type StorageDriverSelector = (context: StorageDriverStoreContext, payload
 /** Default {@link ExternalStorage.payloadSizeThreshold}: 256 KiB. */
 const DEFAULT_PAYLOAD_SIZE_THRESHOLD = 256 * 1024;
 
+/** Default {@link ExternalStorage.maxConcurrentOperations}: 3. */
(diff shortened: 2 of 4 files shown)


packages/common/src/internal-non-workflow/external-storage-visitor.ts
```diff
@@ -37,7 +40,12 @@ export interface ExternalStorageStoreOptions {
  */
 export function extstoreStoreOptions(
   externalStorage: ExternalStorage,
-  { initialTarget, deriveContext, limit, abortSignal }: ExternalStorageStoreOptions = {}
+  {
+    initialTarget,
+    deriveContext,
+    limit = concurrencyLimit(externalStorage.maxConcurrentOperations),
```
mjameswh, line 46: I don't think that's correct. The concurrency set on `externalStorage` should be shared across all callers, not independently per visitors.

For example, if I set a limit of four on my `externalStorage` that's because I don't want to ever have more than four pending requests to the store. As it is written now, if I have three tasks being decoded concurrently, each task visitor gets it's own concurrency limit, potentially resulting in a total of 12 concurrent requests.

mjameswh: Also, I'd argue that if a concurrency limit is set both on the store and on the visitor, then should both apply; i.e. the latter doesn't simply override the former.

mjameswh: In any case, I'd recommend to wait for @jmaeagle99 to chime in before changing your PR. He's the lead on the external storage initiative, so I'd be interested to hear his opinion. He's out today, but should be back tomorrow.

VihaanAgarwal:
Good point on the sharing semantics, a per-walk limit does multiply under concurrent task decode. Moving the semaphore to the ExternalStorage instance so all walks share one budget, and composing an explicitly passed visitor limit with it instead of overriding both sound right to me. I'll hold off on changes until jmaeagle99 has a chance to weigh in.

cconstable:
I think we've got a few definitions of what a "concurrency limit" means in the context of external storage and should reconcile them.

- Limiting the payload visitor is something the SDK can control but due to the way payload visiting works a single payload site visit could send multiple payloads at once to external storage i.e. in that scenario, `limit = 1` could hypothetically result in an unbounded number of underlying concurrent external storage operations. Payload visitor limiting could be done per either activation/message or globally.
- We could attempt to limit _actual_ external storage operations (which feels more intuitive to me) either per activation/message we visit or globally. That may require threading through some context into the drivers themselves.

Also for reference, the Python SDK limits external storage by limiting the payload visiting per activation/message walk and not globally. The limit is applied at the visitor level and not the actual external storage operation level.

VihaanAgarwal:
Agree these need reconciling. One thing I noticed while sketching the shared version locally: if the semaphore lives on the ExternalStorage instance but is acquired per payload operation (around the actual driver call, not per visit site), it acts as a limit on real storage operations without threading any context into the drivers, since every payload op already funnels through that path. The fan-out case then can't exceed the budget either: one site visiting N payloads takes N acquisitions.

That layering would give the instance-level budget the semantics mjameswh described (never more than X in-flight requests to the store), and a per-walk visitor limit could still compose on top for parity with how the Python SDK scopes it. Whether that's the right shape is exactly the jmaeagle99 question, so I'll keep holding off on changes until you all settle direction. Happy to implement whichever way it lands.

cconstable:
I did some investigating and arrived at a similar conclusion: `ExternalStorage` feels like the intuitive place to put concurrency limits. It's also confusing not having there because what happens for clients? Go hardcodes this to 0 which ends up meaning "sequential" and in Python we don't have a visitor on the client path so it's all sequential.

> For example, if I set a limit of four on my externalStorage that's because I don't want to ever have more than four pending requests to the store.

Unfortunately that is not how the other SDK implementations work. A limit of 1 can still mean unbounded concurrent requests to the store. In Go and Python, the limit only limits concurrent payload traversal. Each "visit" can call an unbounded number of drivers with an unbounded number of payloads resulting in an unbounded number of "requests" to external storage.

I see three paths forward:

1. Per-message and per-driver limits. I still think the only way to get a "reasonable limit" (something I can have confidence is doing what I think it's doing) is to make each driver responsible for limiting one "unit" of work and then having a separate limit per-message so one noisy messages/activations doesn't stop the rest from progressing. This would mean we would need to pass some context into the drivers that they would use to limit themselves.
2. Keep per-visit limits but configure them on external storage. This allows clients to be configured as well but doesn't address the ambiguity in what the word "limit" means.
3. Keep things as they are. Clients remain unconfigurable and the "limit" is on the worker options and it's meaning is obtuse.

VihaanAgarwal:
Path 1 is the semantics a user reading `limit: 4` expects: the number bounds requests to the store, full stop. Path 2 fixes client configurability cheaply and doesn't preclude 1 later, but the word keeps its wrong meaning.

On the mechanics of 1: if the acquire wraps each driver payload call inside ExternalStorage (every payload op already funnels through there), the store-request bound holds without changing driver signatures or threading context into community drivers. That covers the per-driver half. The per-message half stays at the visitor, as today, so one noisy activation can't starve the rest.

Happy to carry either in this PR. Which way do you want it?

VihaanAgarwal:
#2424 landed per-message and total concurrency limits for extstore, built into the limiter that gets passed to each driver. That covers what this PR was doing and goes further, so I'm closing it.

Issue #2227, opened 2026-07-21
[Bug] workflow cache size can grow and crash worker processes with memory error


millerick:
### What are you really trying to do?

Our production system uses a workflow worker to run about 100 distinct `workflow_type`s.  We use `zod` schemas to validate inputs and outputs to workflows and activities.  Many of these are defined at the top level of files, all of which appear to be instantiated and have their own allocation in memory for each workflow that is in the cache.  We want our production system to be able to survive bursts of workflows starting without the workers crashing.

### Describe the bug

The workflow cache size can grow and crash the process with an out of memory error.  This can be achieved with the default workflow cache settings.  Our production workers use a 6GB memory request/JS heap size and would regularly see crashes due to exhausting memory as the cache size grew.  It is particularly noticeable if we received a large burst of new workflows since those would all start around the same time and crashes would happen before the workflows would have a chance to complete--occasionally resulting in all workers crashing and ending up in a crash loop backoff--preventing our system from continuing to process workflows until someone manually scaled up the number of workers to ensure that the average number of workflows per worker was low enough.

These crashes stopped when we explicitly capped the size of the workflow cache to 60 workflows.  Eventually this number will need to be lowered as our codebase grows larger.  I don't want to have to manually tune and tweak this value every few months.  The workflow cache should be safely evicted from if it begins to get to the size where the process could crash as a result.  I believe that when the cache grows too large, instead of crashing, old workflows should be evicted from the cache--particularly if the user has not changed the size of the cache themselves.

### Minimal Reproduction

https://github.com/millerick/temporal-worker-crash-reproduction reproduces the issue

Follow the instructions in the README and a crash will happen in the worker
```
2026-07-21T18:26:14.282Z [ERROR] Worker failed {
  sdkComponent: 'worker',
  taskQueue: 'crash-repro',
  error: UnexpectedError: Workflow Worker Thread exited prematurely: Error [ERR_WORKER_OUT_OF_MEMORY]: Worker terminated due to reaching memory limit: JS heap out of memory
      at Worker.<anonymous> (/System/Volumes/Data/src/temporal-crash-reproduction/node_modules/@temporalio/worker/lib/workflow/threaded-vm.js:68:30)
      at Worker.emit (node:events:518:28)
      at [kOnExit] (node:internal/worker:313:12)
      at Worker.<computed>.onexit (node:internal/worker:229:20)
      at Worker.callbackTrampoline (node:internal/async_hooks:130:17) {
[...]

PR #2302, opened 2026-08-10
Add maxWorkflowThreadHeapMiB to prevent OOM crashes from workflow cache

sachinsharma3191:
## Summary

Adds a new `WorkerOptions.maxWorkflowThreadHeapMiB` option that sets `resourceLimits.maxOldGenerationSizeMb` on workflow worker threads created in `threaded-vm.ts`.

**Problem:** The workflow cache is count-based LRU (`maxCachedWorkflows`) with no memory awareness. When per-workflow memory footprint is large (e.g. heavy `zod` schemas or top-level module state), the cache can exhaust the V8 heap, causing `ERR_WORKER_OUT_OF_MEMORY` — a fatal, unrecoverable crash that kills the entire worker process.

**Solution:** By setting `resourceLimits` on the `node:worker_threads` Worker constructor, V8 heap exhaustion in a workflow thread becomes a catchable error instead of a process crash. The existing `UnexpectedError` handling path already reports the failure and initiates graceful shutdown.

### Changes

- **`worker-options.ts`**: New `maxWorkflowThreadHeapMiB?: number` option with documentation. Validation rejects non-positive values.
- **`worker.ts`**: Passes the option through to `ThreadedVMWorkflowCreator.create()`.
- **`threaded-vm.ts`**: 
  - Accepts `maxOldGenerationSizeMb` and passes it as `resourceLimits` to `new NodeWorker()`
  - Detects OOM errors specifically and logs an actionable message suggesting `maxCachedWorkflows` or `maxWorkflowThreadHeapMiB` tuning

### Usage

```typescript
  taskQueue: 'my-task-queue',
  maxWorkflowThreadHeapMiB: 512, // cap each thread to 512 MiB
});
```

### Design note

This is the minimal, self-contained fix that prevents the fatal crash. A full memory-aware cache (feeding V8 heap stats back to the Rust core's eviction decisions) would require cross-layer changes and is tracked separately. This PR gives users an immediate escape hatch.

Ref #2227

## Test plan

- [x] Syntax and structural correctness verified (TypeScript parses clean)
- [x] `resourceLimits` is only set when `maxWorkflowThreadHeapMiB` is provided (default behavior unchanged)
- [x] Validation: `maxWorkflowThreadHeapMiB <= 0` throws `TypeError`
- [x] OOM error detection: specific log message with tuning guidance
- [ ] Manual: Reproduce OOM with [millerick's repro](https://github.com/millerick/temporal-worker-crash-reproduction), verify worker survives with `maxWorkflowThreadHeapMiB` set

### DIFF
--- a/packages/test/src/test-threaded-vm.ts
+++ b/packages/test/src/test-threaded-vm.ts
@@ -1,0 +1,73 @@
+/**
+ * Unit tests for {@link WorkerThreadClient}'s handling of Workflow Worker Thread `error` events,
+ * in particular its ability to detect out-of-memory (OOM) errors and log a more actionable
+ * message pointing users at `maxCachedWorkflows` / `maxWorkflowThreadHeapMiB`.
+ */
+import { EventEmitter } from 'node:events';
+import test from 'ava';
+import type { LogEntry } from '@temporalio/worker';
+import { DefaultLogger } from '@temporalio/worker';
+import { WorkerThreadClient } from '@temporalio/worker/lib/workflow/threaded-vm';
+
+/**
+ * A minimal stand-in for a `node:worker_threads` `Worker`. `WorkerThreadClient` only ever
+ * registers listeners on it (via `.on`) in its constructor, so a plain `EventEmitter` is
+ * sufficient to drive its `error` handling logic without spinning up a real thread.
+ */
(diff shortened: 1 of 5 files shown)


mjameswh:
Poor API additions. Please do not introduce new APIs without prior discussion with the team.

PR #2303, opened 2026-08-10
chore(proto)!: bump protobufjs to v8 and protobufjs-cli to v2

mjameswh:
## Summary

Upgrades `protobufjs` 7.6.4 -> 8.7.1 and `protobufjs-cli` 1.x -> 2.6.1.

This major version bump resolves multiple issues we had previously, allowing cleaning up various workarounds from our code base:

- Drop our local pnpm patch on `protobufjs`. The parser bug has been fixed upstream (protobufjs/protobuf.js#2379).
- `proto3-json-serializer` is replaced by `protobufjs/ext/protojson`. `proto3-json-serializer` has no `protobufjs` 8 support and no prospect of it.
- `protobufjs`'s extension is spec-compliant out of the box, removing the need for `fixPayloads`, `fixBuffers`, and the `Buffer` global shimming inside the sandbox.
- `replaceBuffers` had to stay, but is now skipped outright if `Buffer` is absent (e.g. in the Workflow sandbox).
- The webpack `fs` shim for `protobufjs` is gone, since v8's `src/util/fs.js` uses `require(/* webpackIgnore: true */ "fs")`. That comment only takes effect with `module.parser.javascript.commonjsMagicComments`, which is now enabled in the bundler.
- `protobufjs` now properly defines its dependency to `long`, so injection of that dependency (in `protos/root.js`) goes away.

This has some visible consequences on generated protobuf types, which are exposed as part of `@temporalio/proto`. See details in `CHANGELOG.md`.


.github/workflows/ci.yml
```diff
@@ -332,7 +332,7 @@ jobs:
       typescript-repo-path: ${{github.event.pull_request.head.repo.full_name}}
       version: ${{github.event.pull_request.head.ref}}
       version-is-repo-ref: true
-      features-repo-ref: main
+      features-repo-ref: bump-protobufjs
```
chris-olszewski, line 421: Do we want a TODO explaining when this can go back to `main`?

CHANGELOG.md
```diff
@@ -19,6 +19,43 @@ to docs, or any other relevant information.
...
+  - Applications that use the Protobuf Payload Converters with their own `.proto` files will need
+    to upgrade their own `protobufjs` dependency to v8. A root built with v7 still works for binary
+    payloads, but JSON-encoded payloads will fail at runtime with `TypeError: type must be a Type`.
+
+  - The constructor on generated message types is no longer publicly accessible, which means that
+    `new SomeMessage()` no longer compiles. Use `SomeMessage.create({ ... })` instead. Use of the
+    constructor was never officially supported, and would have previously resulted in a runtime error.
+
+  - The `getTypeUrl` method on generated message types is no longer declared, though it remains
+    callable at runtime.
```
chris-olszewski: Do we want to advertise that it is still callable? If it is no longer declared could it be removed upstream without it being considered a breaking change?

packages/test/src/test-nexus-update-operation.ts
```diff
@@ -32,8 +32,8 @@ const test = makeTestFunction({
   workflowEnvironmentOpts: {
     server: {
       // The default dev server (Temporal 1.31.2) does not deliver Workflow Update completion
-      // callbacks to Nexus. Pin a server build that does (server 1.32.0-157.0).
-      executable: { type: 'cached-download', version: 'v1.7.2-standalone-nexus-operations' },
+      // callbacks to Nexus. Pin a server build that does (server 1.32.0-158.0).
+      executable: { type: 'cached-download', version: 'v1.7.4-standalone-nexus-operations' },
```
chris-olszewski: Is this releated to the protobuf bump?

packages/test/src/test-nexus-workflow-caller.cloud-unavailable.ts
```diff
@@ -596,9 +596,7 @@ test('getClient is available in handler context - caller workflow', async (t) =>
     nexusServices: [
       nexus.serviceHandler(getClientService, {
         async getClientOp() {
-          const systemInfo = await temporalnexus
-            .getClient()
-            .connection.workflowService.getSystemInfo({ namespace: 'default' });
+          const systemInfo = await temporalnexus.getClient().connection.workflowService.getSystemInfo({});
```
chris-olszewski, line 599: Is this releated to the protobuf bump?

packages/test/src/test-proto-utils.ts
```diff
@@ -16,10 +16,8 @@ test('cancel-fake-progress-replay', async (t) => {
   const hist = historyFromJSON(histJSON);
   t.deepEqual(hist.events?.[15].eventId, new Long(16));
   t.is(hist.events?.[15].eventType, EventType.EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED);
-  t.is(
-    hist.events?.[0].workflowExecutionStartedEventAttributes?.initiator,
-    ContinueAsNewInitiator.CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED
-  );
+  const initiator = hist.events?.[0].workflowExecutionStartedEventAttributes?.initiator;
+  t.true(initiator === undefined || initiator === ContinueAsNewInitiator.CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED);
```
chris-olszewski, line 20: What are the situations where it is `undefined` vs being a set value? Does the assertion need to handle both?

chris-olszewski (approved):

packages/test/src/test-nexus-workflow-caller.cloud-unavailable.ts
mjameswh: It's a latent bug that the protobufjs bump exposed.

There's actually no `namespace` property on `IGetSystemInfoRequest`, so the former was technically incorrect, but TSC was not reporting this as an error given the previous definition of that type. It is now, so I had to remove the extraneous property.

packages/test/src/test-proto-utils.ts
mjameswh: With protobufjs v8, fields no longer fills default values for unset fields. That's in line with how protobuf works in other language.

In this case, `CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED` is the default value for the `initiator`. So in v7, protobufjs guaranteed that `initiator` would be `CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED` if not explicitly set in the serialized message. In v8, it would be `undefined` in that case. But it could be `CONTINUE_AS_NEW_INITIATOR_UNSPECIFIED` if the value was specifically set in the message.

packages/test/src/test-nexus-update-operation.ts
mjameswh: Good catch, that's contamination from another work stream. I'm dropping the change.

CHANGELOG.md
mjameswh: I removed the claim that it is still callable.

PR #2305, opened 2026-08-10
Upgrade to node 22.

cconstable:
## What was changed
Bumped minimum Node version to 22. 

## Why?
- Node 20 has reached end-of-life.
- Allows usage of more modern node apis without warnings or errors.

## Checklist
- [ ] Update CI checks?


cconstable:
Met internally to discuss. We are OK officially supporting past-EOL node versions providing that support does not impede development.

PR #2309, opened 2026-08-11
fix: don't apply plugin workflowModules when a workflowBundle is used

### LINKED ISSUE #2278: [Bug] OpenTelemetryPlugin + prebuilt workflowBundle always logs "Ignoring WorkerOptions.interceptors.workflowModules" at worker startup
## What are you really trying to do?

Use `OpenTelemetryPlugin` (from `@temporalio/interceptors-opentelemetry-v2`) with a prebuilt workflow bundle, following the guidance in the `interceptors-opentelemetry` sample: pass the plugin to `Worker.create` **and** to `bundleWorkflowCode`.

## Describe the bug

With that documented setup, every worker boot logs:

```
Ignoring WorkerOptions.interceptors.workflowModules because WorkerOptions.workflowBundle is set.
To use workflow interceptors with a workflowBundle, pass them in the call to bundleWorkflowCode.
```

even though the interceptor module *was* passed to `bundleWorkflowCode` and is present in the bundle.

Cause: `SimplePlugin.configureWorker` appends `workerInterceptors.workflowModules` unconditionally -- it resolves `options.workflowBundle` on the adjacent line but never consults it -- and the worker then warns whenever non-default `workflowModules` coexist with a `workflowBundle`. So the two features contradict each other in exactly the configuration the sample prescribes for production (prebuilt bundles).

Note the warning fires whether or not the bundle actually contains the module, so it also cannot serve as a drift detector; it is unconditional noise in bundle mode.

The `interceptors-opentelemetry` sample only exercises `workflowsPath` (the prebundling advice is a comment), which is likely why this combination goes unnoticed.

## Minimal Reproduction

```ts
const plugin = new OpenTelemetryPlugin({ resource, spanProcessor });

// build step
const { code } = await bundleWorkflowCode({ workflowsPath: require.resolve('./workflows'), plugins: [plugin] });

// worker boot
await Worker.create({ workflowBundle: { code }, taskQueue: 'q', plugins: [new OpenTelemetryPlugin({ resource, spanProcessor })] });
// -> WARN logged at startup
```

## Expected behavior

No warning in the documented configuration. A possible fix: `SimplePlugin.configureWorker` skips appending `workflowModules` when `options.workflowBundle` is set (module paths cannot be applied to a prebuilt bundle anyway, and `configureBundler` is the hook that matters for bundling).

## Environment/Versions

- @temporalio/worker 1.20.2, @temporalio/interceptors-opentelemetry-v2 1.20.2 (same logic present in 1.21.1 sources)
- Node 26, Linux/macOS

alexrosenfeld10:
## What was changed

`SimplePlugin.configureWorker` and `configureReplayWorker` no longer inject the plugin's `workerInterceptors.workflowModules` into worker options when the resolved options contain a `workflowBundle`. User-provided `interceptors.workflowModules` are passed through untouched, so the worker's existing warning still fires for explicit user misconfiguration.

## Why?

Fixes #2278.

`SimplePlugin` appended `workflowModules` unconditionally, so any plugin that ships workflow interceptor modules (e.g. `OpenTelemetryPlugin`) triggered this on every worker startup when used with a prebuilt bundle:

```
Ignoring WorkerOptions.interceptors.workflowModules because WorkerOptions.workflowBundle is set.
To use workflow interceptors with a workflowBundle, pass them in the call to bundleWorkflowCode.
```

This happened even in the documented setup where the plugin was also passed to `bundleWorkflowCode` (which is what actually includes the interceptors in the bundle, via `configureBundler`). Module paths cannot be applied to a prebuilt bundle, so injecting them into worker options serves no purpose in that configuration.

## Checklist

1. Closes #2278

2. How was this tested:

- New test in `packages/test/src/test-plugins.ts` covering: modules still appended with `workflowsPath`; not injected when a `workflowBundle` is passed in options or provided by the plugin itself; user-provided modules passed through; same behavior for `configureReplayWorker`. Test was written first and observed failing against the previous behavior.
- End-to-end check reproducing the issue scenario: `bundleWorkflowCode` with `OpenTelemetryPlugin` (from `contrib/interceptors-opentelemetry-v2`) followed by `Worker.create` with the prebuilt bundle and the plugin, capturing runtime logs. Before the fix the warning is logged; after the fix it is not, and the bundle contains the OTel workflow interceptors.
- Full `test-plugins.js` suite passes locally.

3. Any docs updates needed?

- Updated the `workerInterceptors` option doc in `SimplePluginOptions` to describe the behavior.


packages/plugin/src/plugin.ts
```diff
@@ -104,14 +106,15 @@ export class SimplePlugin
...
     return {
       ...options,
       dataConverter: resolveDataConverter(options.dataConverter, this.options.dataConverter),
       activities: resolveAppendObjectParameter(options.activities, this.options.activities),
       nexusServices: resolveAppendParameter(options.nexusServices, this.options.nexusServices),
       workflowsPath: resolveParameter(options.workflowsPath, this.options.workflowsPath),
-      workflowBundle: resolveParameter(options.workflowBundle, this.options.workflowBundle),
-      interceptors: resolveWorkerInterceptors(options.interceptors, this.options.workerInterceptors),
+      workflowBundle,
+      interceptors: resolveWorkerInterceptorsForOptions(options, workflowBundle, this.options.workerInterceptors),
```
mjameswh: 1. Let's reverse the gate condition: only include `workflowModules` interceptors if there's a `workflowsPath`; that means that a worker with no `workflowBundle` and no `workflowsPath` (i.e. non-workflow worker) should not receive any `workflowModules`.
2. Please follow the same arg structure/order as other `resolveXxx` methods; the gate condition should be last and should be a boolean (maybe `includeWorkflowModules`?).
3. I'd argue the extra intermediary resolve function brings no value, yet makes the logic harder to process mentally.

```suggestion
      interceptors: resolveWorkerInterceptors(options.interceptors, this.options.workerInterceptors, resolvedWorkflowsPath != null),
```

alexrosenfeld10: Reworked in 9c9958d1: gate reversed to workflowsPath presence (so non-workflow workers get no workflowModules either), passed as a trailing boolean on resolveWorkerInterceptors, and the intermediary resolve function is removed.

One note on the implementation: since resolveParameterWithResolution returns the parameter directly when existing is undefined (the common case where the user passes no interceptors), the gate strips workflowModules from the object-form parameter up front rather than inside the merge callback. Function-form parameters keep full control of the result, consistent with the other resolvers.

alexrosenfeld10:
@mjameswh  anything else needed here? Thanks in advance.

mjameswh:
Sounds good, thanks!

One last thing, would yoo mind adding a short entry in `CHANGELOG.md`, under the `[unreleased]/changed` section?

alexrosenfeld10:
@mjameswh good to go here?

mjameswh (approved):

PR #2316, opened 2026-08-13
feat(extstore): expose concurrent payload visit limit for external storage.

### LINKED ISSUE #2275: [Bug] External storage does not allow for concurrent storage operations

cconstable:
## What was changed
- Added `maxConcurrentVisits` to `ExternalStorage` constructor for limiting concurrent payload visits.
- This diverges from Go and Python which expose the limit on their worker options. I think a) external storage is the more natural place for this option and b) having the option on external storage allows us to also configure this limit for clients (which is currently not possible in go and python, they are hardcoded to 1). 

## Why?
- This was previously not configurable.

## Checklist
- Closes https://github.com/temporalio/sdk-typescript/issues/2275

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ to docs, or any other relevant information.
 - **Experimental**: `TemporalOperationHandler` can now use Standalone Activities as asynchronous
   Nexus Operation backing executions through `TemporalNexusClient.startActivity` and
   `typedActivity`.
+- **Experimental**: Added `ExternalStorage.maxConcurrentVisits` which limits the number of in-flight payload visits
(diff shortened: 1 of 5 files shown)


cconstable:
exploratory work. does not need review.

PR #2321, opened 2026-08-15
feat(google-adk-agents): export failure-type constants, narrow the ADK peer range

xumaple:
Follow-up to #2120, addressing review comments from @DABH, @brianstrauch and @chris-olszewski.

Workflow code had to hand-type this plugin's `ApplicationFailure.type` strings to match on a failure. `src/error-types.ts` now owns them and `/workflow` exports the four a caller can act on; the two that only fire when the plugin's API is used outside a Workflow stay internal. The README's error-types table becomes TSDoc on the four public constants, and the plugin's own docstrings lose what documented ADK's internals or restated the code beneath them.

A `<name>-callTool` Activity opened two MCP sessions — against a stdio server, two subprocesses — to make one call: the Activity listed the server's tools to resolve the name, costing one session, then invoked the tool in another. Where the factory returns `MCPConnectionParams` it now opens one session and issues one `tools/call`. A factory returning a `BaseToolset` still pays both, because `BaseToolset` offers no way to get a tool handle without listing first.

Note: without that lookup the plugin no longer pre-validates the tool name, so `GoogleAdkMCPToolNotFound` never fires on that route. An `McpServer` answers an unknown tool with a successful result carrying `isError: true`, so the Activity completes and the model is handed the error as a tool result; a server that puts a JSON-RPC error on the wire fails it retryably instead, and unboundedly, since the plugin sets no retry policy. Documented rather than fixed — classifying JSON-RPC codes as non-retryable would be new behavior.

The `@google/adk` peer range narrows to `>=1.5.0 <1.6.0`. @DABH proposed `>=1.4.0 <1.6.0`, but that admitted 1.5.0 while the suite only ever ran 1.4.0; the suite now runs 1.5.0. `@google/genai` rises to `^2.9.0`, which ADK has required since 1.4.0. Also registers the package in `packages/meta` so it reaches the docs site.


chris-olszewski (approved):
Approving the `meta` changes

contrib/google-adk-agents/README.md
```diff
@@ -248,20 +245,18 @@ Cautions:
...
   not register the model/MCP Activities. If composing plugins, place
   observability and governance plugins before this one.
 - Model calls use Temporal retries. The plugin disables nested GenAI SDK retries
   for model requests and honors `retry-after` where available.
-- Heartbeats are sent only when the Activity options include `heartbeatTimeout`.
+- Set `heartbeatTimeout` in the Activity options to have a stalled model or MCP
+  call detected before `startToCloseTimeout` expires. The plugin then heartbeats
+  every such call at half that timeout, and a streaming model call on each chunk as
+  well. Without it there is no heartbeat, and only `startToCloseTimeout` ends a
+  stalled call.
```
DABH: Same issue as in #2120; suggested wording:
```suggestion
- Set `heartbeatTimeout` in the Activity options to have a stalled model or MCP
  call detected before `startToCloseTimeout` expires. The plugin then heartbeats
  every such call at half that timeout; a streaming model call also heartbeats on
  each chunk regardless. Without a `heartbeatTimeout`, nothing detects a stall —
  only `startToCloseTimeout` ends a stalled call.
```

contrib/google-adk-agents/src/index.ts
```diff
@@ -3,17 +3,12 @@
...
- * the `@temporalio/google-adk-agents/workflow` entry point; test helpers in
- * `@temporalio/google-adk-agents/testing`.
+ * Worker-side API for the Google ADK Temporal plugin.
  */
 
 /**
  * @experimental The Google ADK plugin is an experimental feature; APIs may change without notice.
  */
-
 export { GoogleAdkPlugin } from './plugin';
```
DABH: As TSDoc, this block attaches to the first export statement only. Folding the sentence into the header comment makes it module-level for source and docs readers — and `GoogleAdkPlugin` itself already carries `@experimental` on its declaration (`plugin.ts:454`), so nothing is lost.
```suggestion
 * Worker-side API for the Google ADK Temporal plugin.
 *
 * @experimental The Google ADK plugin is an experimental feature; APIs may change without notice.
 */

export { GoogleAdkPlugin } from './plugin';
```

DABH (commented):
Suggestions: two doc nits - heartbeat wording and `@experimental` placement - plus a `GoogleAdkMCPError` / `GoogleAdkModelError` type split. (The nine `GoogleAdkMCPError` suggestions (on `error-types.ts`, `activities.ts`, `workflow.ts`, and the two test files) are one coherent change: apply all or none.)

contrib/google-adk-agents/src/workflow.ts
```diff
@@ -4,16 +4,17 @@
  * SPDX-License-Identifier: MIT
  *
  * Workflow-safe API for the Google ADK Temporal plugin — importable from
- * Workflow code running in the V8 sandbox. The worker-side `GoogleAdkPlugin`
- * lives in the main `@temporalio/google-adk-agents` entry point; test helpers
- * in `@temporalio/google-adk-agents/testing`.
+ * Workflow code running in the V8 sandbox.
  */
```
DABH: Same as `index.ts`: attach the experimental note to the module header so it covers every export, not just `TemporalModel`. Pairs with the suggestion below that removes the misplaced block. (Longer term, the `TemporalModel` / `TemporalMCPToolset` / `activityAsTool` declarations could each carry `@experimental` the way `GoogleAdkPlugin` does.)
```suggestion
 * Workflow-safe API for the Google ADK Temporal plugin — importable from
 * Workflow code running in the V8 sandbox.
 *
 * @experimental The Google ADK plugin is an experimental feature; APIs may change without notice.
 */
```

contrib/google-adk-agents/src/workflow.ts
```diff
@@ -4,16 +4,17 @@
...
 
 // Side-effect import: install Workflow-sandbox polyfills (gated internally on
 // `inWorkflowContext()`) before any ADK code runs in the bundle.
 // eslint-disable-next-line import/no-unassigned-import
 import './load-polyfills';
 
+/**
+ * @experimental The Google ADK plugin is an experimental feature; APIs may change without notice.
+ */
 export { TemporalModel } from './model';
```
DABH: Pairs with the header suggestion above: as written this TSDoc attaches only to the `TemporalModel` re-export — `TemporalMCPToolset`, `activityAsTool`, and the failure-type constants don't get the tag.
```suggestion
export { TemporalModel } from './model';
```

contrib/google-adk-agents/src/error-types.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Inside an `LlmAgent` run most of these never reach a `catch` around the runner: a
+ * failed model call becomes an error event and a failed tool call an `{ error }`
+ * tool response fed back to the model, and neither fails the Workflow, so decide
+ * whether such a run succeeded from the events the runner yields. Those events carry an
+ * `errorCode` of ADK's own rather than this type, so do not try to match it there. Tool
+ * discovery is the exception: `listTools` runs while ADK is still building the request,
+ * outside that handling, so its failure propagates and fails the Workflow.
+ */
+export const MODEL_ERROR_FAILURE_TYPE = 'GoogleAdkModelError';
```
DABH: Since #2120 hasn't shipped, suggest we consider giving MCP failures their own type instead of documenting the "despite the name" comment.  These are two very different failure sources (status-classified model errors vs. status-less MCP errors that retry unboundedly).

Apply all or none of this and the companion suggestions...
```suggestion
/**
 * Error type when the `<name>-callTool` Activity finds no tool by the requested name
 * in the `BaseToolset` its factory returned. Non-retryable, and raised inside that
 * Activity, so it arrives wrapped in an `ActivityFailure`: match it through the
 * `.cause` chain, never against the caught error's own `.type`. Inside an `LlmAgent`
 * run ADK absorbs it instead — see {@link MCP_ERROR_FAILURE_TYPE}.
 *
 * A factory returning `MCPConnectionParams` never raises it: the plugin calls the tool
 * without resolving the name first, so the caller gets whatever the server answers.
 */
export const MCP_TOOL_NOT_FOUND_FAILURE_TYPE = 'GoogleAdkMCPToolNotFound';

/**
 * Error type for a failed model call. Raised inside an Activity, so it arrives
 * wrapped in an `ActivityFailure`: match it through the `.cause` chain.
 *
 * A failure that carries an HTTP status has it appended as `.<status>`, for example
 * `GoogleAdkModelError.429`, so match with `startsWith`, not `===`; such a failure is
 * retryable for 408, 409, 429 and 5xx and non-retryable otherwise. One carrying no
 * status is the bare type and retryable. An `x-should-retry` response header overrides
 * either verdict.
 *
 * Inside an `LlmAgent` run these never reach a `catch` around the runner: a failed
 * model call becomes an error event rather than failing the Workflow, so decide
 * whether such a run succeeded from the events the runner yields. Those events carry
 * an `errorCode` of ADK's own rather than this type, so do not try to match it there.
 */
export const MODEL_ERROR_FAILURE_TYPE = 'GoogleAdkModelError';

/**
 * Error type for a failed MCP `listTools` or `callTool` call. Raised inside an
 * Activity, so it arrives wrapped in an `ActivityFailure`: match it through the
 * `.cause` chain. An HTTP status, when present, is appended and classified exactly
 * as for {@link MODEL_ERROR_FAILURE_TYPE} — but an MCP failure rarely carries one,
 * so a rejected token or an unreachable server classifies as retryable rather than
 * failing fast, and the plugin sets no retry policy: the call retries indefinitely
 * until the caller bounds it with
 * `TemporalMCPToolsetOptions.activity.retry.maximumAttempts`.
 *
 * Inside an `LlmAgent` run a failed tool call becomes an `{ error }` tool response
 * fed back to the model and does not fail the Workflow. Tool discovery is the
 * exception: `listTools` runs while ADK is still building the request, outside that
 * handling, so its failure propagates and fails the Workflow.
 */
export const MCP_ERROR_FAILURE_TYPE = 'GoogleAdkMCPError';
```

contrib/google-adk-agents/src/activities.ts
```diff
@@ -13,26 +13,27 @@
...
   type LlmResponse,
+  type MCPConnectionParams,
 } from '@google/adk';
 import type { FunctionDeclaration } from '@google/genai';
 import type { Duration } from '@temporalio/common';
 import { ApplicationFailure } from '@temporalio/common';
 import { Context as ActivityContext } from '@temporalio/activity';
 import { WorkflowStreamClient } from '@temporalio/workflow-streams/client';
 
+import { MCP_TOOL_NOT_FOUND_FAILURE_TYPE, MODEL_ERROR_FAILURE_TYPE } from './error-types';
```
DABH: Companion to the `GoogleAdkMCPError` suggestion in `error-types.ts`.
```suggestion
import { MCP_ERROR_FAILURE_TYPE, MCP_TOOL_NOT_FOUND_FAILURE_TYPE, MODEL_ERROR_FAILURE_TYPE } from './error-types';
```

contrib/google-adk-agents/src/activities.ts
```diff
@@ -133,7 +130,7 @@ function mcpActivitiesForName(
         throw toApplicationFailure(err);
```
DABH: Companion: MCP route → MCP type.
```suggestion
        throw toApplicationFailure(err, MCP_ERROR_FAILURE_TYPE);
```

contrib/google-adk-agents/src/activities.ts
```diff
@@ -143,37 +140,51 @@ function mcpActivitiesForName(
...
         }
         // The MCP tool reads `toolContext.abortSignal`; supply the Activity's
         // cancellation signal so a cancelled Workflow aborts the call.
-        const toolContext = {
-          abortSignal: ActivityContext.current().cancellationSignal,
-        } as unknown as AdkToolContext;
+        const toolContext = { abortSignal } as unknown as AdkToolContext;
         return await tool.runAsync({ args: args.args, toolContext });
       } catch (err) {
         throw toApplicationFailure(err);
```
DABH: Companion: MCP route → MCP type (covers both the one-session path and the `BaseToolset` path; `MCP_TOOL_NOT_FOUND` still passes through unchanged as an `ApplicationFailure`).
```suggestion
        throw toApplicationFailure(err, MCP_ERROR_FAILURE_TYPE);
```

contrib/google-adk-agents/src/activities.ts
```diff
@@ -224,13 +235,7 @@ function startAdaptiveHeartbeat(): () => void {
...
 
-/**
- * Maps a GenAI/MCP error into Temporal's retry contract: 408/409/429/5xx (or
- * `x-should-retry: true`) → retryable, other 4xx (or `x-should-retry: false`) →
- * non-retryable; `retry-after[-ms]` → `nextRetryDelay`. Existing
- * `ApplicationFailure`s pass through unchanged.
- * @internal
- */
+/** @internal */
 export function toApplicationFailure(err: unknown): ApplicationFailure {
```
DABH: Companion: parameterize the base type; model activities keep the default.
```suggestion
export function toApplicationFailure(err: unknown, baseType: string = MODEL_ERROR_FAILURE_TYPE): ApplicationFailure {
```

contrib/google-adk-agents/src/activities.ts
```diff
@@ -250,7 +255,7 @@ export function toApplicationFailure(err: unknown): ApplicationFailure {
 
   return ApplicationFailure.create({
     message,
-    type: status !== undefined ? `GoogleAdkModelError.${status}` : 'GoogleAdkModelError',
+    type: status !== undefined ? `${MODEL_ERROR_FAILURE_TYPE}.${status}` : MODEL_ERROR_FAILURE_TYPE,
```
DABH: Companion: build the type from the caller's base.
```suggestion
    type: status !== undefined ? `${baseType}.${status}` : baseType,
```

contrib/google-adk-agents/src/workflow.ts
```diff
@@ -22,3 +23,10 @@ export type { TemporalMCPToolsetOptions, MCPToolsetFactory } from './mcp';
 
 export { activityAsTool } from './tools';
 export type { ActivityAsToolOptions } from './tools';
+
+export {
+  MCP_TOOL_NOT_FOUND_FAILURE_TYPE,
```
DABH, line 30: Companion: export the new constant.
```suggestion
  MCP_ERROR_FAILURE_TYPE,
  MCP_TOOL_NOT_FOUND_FAILURE_TYPE,
```

contrib/google-adk-agents/src/__tests__/published-artifact.test.ts
```diff
@@ -18,11 +18,18 @@ test('cjsRequireExposesPublicExports', (t) => {
...
   const main = require('@temporalio/google-adk-agents');
   t.is(typeof main.GoogleAdkPlugin, 'function');
+  t.false('MODEL_ERROR_FAILURE_TYPE' in main);
 
   const workflow = require('@temporalio/google-adk-agents/workflow');
   t.is(typeof workflow.TemporalModel, 'function');
   t.is(typeof workflow.TemporalMCPToolset, 'function');
   t.is(typeof workflow.activityAsTool, 'function');
+  t.is(workflow.MODEL_ERROR_FAILURE_TYPE, 'GoogleAdkModelError');
+  t.is(workflow.MCP_TOOL_NOT_FOUND_FAILURE_TYPE, 'GoogleAdkMCPToolNotFound');
```
DABH, line 28: Companion: pin the new wire string in the published surface.
```suggestion
  t.is(workflow.MCP_TOOL_NOT_FOUND_FAILURE_TYPE, 'GoogleAdkMCPToolNotFound');
  t.is(workflow.MCP_ERROR_FAILURE_TYPE, 'GoogleAdkMCPError');
```

contrib/google-adk-agents/src/__tests__/errors.test.ts
```diff
@@ -56,6 +56,12 @@ test('shouldRetryFalseHeaderForcesNonRetryableOnRetryableStatus', (t) => {
   t.is(failure.nonRetryable, true);
 });
 
+test('shouldRetryFalseHeaderForcesNonRetryableOnUndefinedStatus', (t) => {
+  const failure = toApplicationFailure({ headers: { 'x-should-retry': 'false' } });
+  t.is(failure.type, 'GoogleAdkModelError');
+  t.is(failure.nonRetryable, true);
+});
```
DABH, line 64: Companion: cover the parameterized base type.
```suggestion
});

test('mcpBaseTypeSetsMCPErrorTypeAndStaysRetryable', (t) => {
  const failure = toApplicationFailure(new Error('connection refused'), 'GoogleAdkMCPError');
  t.is(failure.type, 'GoogleAdkMCPError');
  t.is(failure.nonRetryable, false);
});
```

DABH (approved):
Approved once suggestions are processed

contrib/google-adk-agents/README.md
xumaple: done

contrib/google-adk-agents/src/index.ts
xumaple: done

contrib/google-adk-agents/src/workflow.ts
xumaple: done

contrib/google-adk-agents/src/error-types.ts
xumaple: done

contrib/google-adk-agents/src/__tests__/published-artifact.test.ts
xumaple: done

contrib/google-adk-agents/src/__tests__/errors.test.ts
xumaple: done

PR #2323, opened 2026-08-17
feat(workflow): support TypeInfo for string Signals

THardy98:
## What was changed

String-named Signal calls can now provide input `TypeInfo` through `signalWithOptions` on Client, child Workflow, and external Workflow handles. Signal-with-start options can provide the same metadata through `signalTypeInfo`.

Both new call-site paths reuse the existing Signal interceptor and serialization flow. Definition-based Signals remain unchanged, and signal-with-start rejects explicit call-site metadata when a Signal definition is supplied.

This builds on the definition-supplied Signal TypeInfo support merged in #2318.

## Why?

When only a Signal name is available, the SDK cannot obtain conversion metadata from a Signal definition. The existing variadic `signal` method also cannot safely distinguish user arguments from an options object, so string call sites need an explicit options-based API.

This API is experimental. Existing Signal calls retain their current behavior, with no rollout or manual steps required.

## Checklist

1. Closes: N/A

2. How was this tested: Common, Client, Workflow, and integration package builds; focused ESLint and Prettier checks; `ts-prune`; and all 17 focused TypeInfo integration tests, including Client, signal-with-start, external Workflow, child Workflow, and definition/call-site conflict coverage.

### DIFF
--- a/packages/workflow/src/workflow.ts
+++ b/packages/workflow/src/workflow.ts
@@ -8,6 +8,7 @@ import type {
   SearchAttributeValue,
   SignalDefinition,
   SignalDefinitionOptions,
+  SignalTypeInfo,
   UntypedActivities,
   UpdateDefinition,
   WithWorkflowArgs,
@@ -19,6 +20,7 @@ import type {
   SearchAttributeUpdatePair,
   WorkflowDefinitionOptionsOrGetter,
   WorkflowDefinitionConfig,
+  WorkflowSignalOptions,
 } from '@temporalio/common';
 import {
   compileRetryPolicy,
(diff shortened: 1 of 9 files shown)


packages/workflow/src/workflow.ts
```diff
@@ -901,28 +909,38 @@ export async function startChild<T extends Workflow>(
...
+      'signalWorkflow',
+      signalWorkflowNextHandler
+    )({
+      seq: activator.nextSeqs.signalWorkflow++,
+      signalName,
+      args,
+      typeInfo,
+      target: {
+        type: 'child',
+        childWorkflowId: optionsWithDefaults.workflowId,
```
cconstable: small nit: it looks like `workflowOptions` are the options derived from `optionsWithDefaults` that the function should be using... at least that is how I'm reading it.

cconstable (approved):

packages/workflow/src/workflow.ts
THardy98: Addressed in https://github.com/temporalio/sdk-typescript/pull/2323/commits/f9ed507a6bc3c7d55035cefb8b1742c38eabc211

PR #2325, opened 2026-08-17
docs: add the langsmith package and the openai-agents Workflow and OTel subpaths to the API reference

xumaple:
`packages/meta` is the single TypeDoc entry point for typescript.temporal.io — `packages/docs/docusaurus.config.js` points `entryPoints` at `../meta/src/index.ts`, and nothing else enumerates packages. A package absent from it has no published API reference at all.

`@temporalio/langsmith` was never added: #2099 introduced the package across 33 files without touching `packages/meta`. Separately, only the `@temporalio/openai-agents` root was re-exported, so its `/workflow` and `/otel` surfaces — `TemporalOpenAIRunner`, `activityAsTool`, `agentAsTool`, `WorkflowSafeMemorySession`, `createTracerProvider`, `TemporalIdGenerator`, among others — have no pages either, even though the package README directs users to import from both. This is the same catch-up as #1781 (nexus) and #2125 (workflow-streams + openai-agents); nothing in CI asserts that every package appears in meta, which is why it recurs.

`@temporalio/openai-agents/workflow-interceptor` stays out: it exports only the `interceptors()` factory that the bundler resolves by path, not an API anyone calls — the same shape as langsmith's `@internal` `./workflow-interceptors`. `@temporalio/ai-sdk` needs nothing, since its root already re-exports `./provider` and `./mcp`, covering its workflow surface.

Note: this appends to the same two files as #2321, which adds google-adk to meta. Whichever merges second gets an append-only conflict in `src/index.ts` and an alphabetical insert in `package.json`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,11 @@ to docs, or any other relevant information.
 - `@temporalio/ai-sdk` now requires `ai@>=7.0.59` as a peer dependency, up from `7.0.0`, since
   earlier releases threw a `TypeError` on import in runtimes without a global `fetch`.
 
+### Fixed
+
+- The [API reference](https://typescript.temporal.io) now documents `@temporalio/langsmith` and the
+  Workflow-side and OpenTelemetry subpaths of `@temporalio/openai-agents`.
+
 ## [1.22.0] - 2026-08-05
 
 ### Added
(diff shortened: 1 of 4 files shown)


chris-olszewski (approved):

CHANGELOG.md
```diff
@@ -36,6 +36,11 @@ to docs, or any other relevant information.
 - `@temporalio/ai-sdk` now requires `ai@>=7.0.59` as a peer dependency, up from `7.0.0`, since
   earlier releases threw a `TypeError` on import in runtimes without a global `fetch`.
 
+### Fixed
+
+- The [API reference](https://typescript.temporal.io) now documents `@temporalio/langsmith` and the
+  Workflow-side and OpenTelemetry subpaths of `@temporalio/openai-agents`.
```
chris-olszewski: Unsure we need a changelog for this?

brianstrauch (approved):
+1 once CHANGELOG is removed

PR #2327, opened 2026-08-18
fix(client): reject call-site TypeInfo for Workflow functions

THardy98:
## What was changed

Workflow Client APIs that accept either a Workflow function or a string name now reject call-site `TypeInfo` at compile time when a function is supplied. This covers start, execute, signal-with-start, and Update-with-Start operations.

String-named calls continue accepting call-site metadata. Runtime validation remains in place for JavaScript and dynamically typed calls. Existing Nexus and Test Helper APIs retain their current broad signatures through localized internal compatibility adapters; Nexus public parity is deferred to the Nexus TypeInfo PR.

## Why?

Workflow functions already carry definition-supplied TypeInfo. Supplying call-site TypeInfo with the same function is ambiguous and previously failed only at runtime.

The overloads now make the definition-versus-string resolution rule visible to TypeScript users without changing runtime behavior or requiring rollout steps.

## Checklist

1. Closes: N/A

2. How was this tested: Common, Client, Nexus compatibility, Test Helpers, AI SDK, and Test TypeScript builds; focused Client compile-time tests; ESLint; Prettier; `ts-prune`; and commit lint.

3. Any docs updates needed? No; the experimental API documentation already describes definition-supplied and string call-site TypeInfo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,8 @@ to docs, or any other relevant information.
 
 ### Changed
 
+- **Experimental**: Workflow Client APIs now reject call-site `TypeInfo` at compile time when a Workflow function is
+  supplied. Define TypeInfo on the Workflow function or use a string Workflow type for call-site metadata.
 - Nexus is now generally available (GA) for calling Nexus Operations from Workflows and handling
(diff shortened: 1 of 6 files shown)


THardy98:
not needed

PR #2329, opened 2026-08-18
Use new payload visitor for workflow activations and completions

tconley1428:
## Summary
- encode workflow activations and completions through the shared payload visitor
- preserve workflow, activity, child workflow, and external workflow serialization contexts
- keep workflow headers and search attributes outside payload codec processing
- add tests
- remove dead code in codec-helpers

### DIFF
--- a/packages/common/src/internal-non-workflow/codec-helpers.ts
+++ b/packages/common/src/internal-non-workflow/codec-helpers.ts
@@ -45,7 +45,7 @@ export async function encode(
 }
 
 /** Run {@link PayloadCodec.decode} on `payloads` */
-export async function decodeOptional(
+async function decodeOptional(
   codecs: PayloadCodec[],
   payloads: Payload[] | null | undefined,
   context?: SerializationContext
@@ -160,17 +160,6 @@ export async function decodeOptionalFailureToOptionalError(
   return failure
     ? failureConverter.failureToError(await decodeFailure(payloadCodecs, failure, context), payloadConverter, context)
     : undefined;
-}
-
-export async function decodeOptionalMap(
-  codecs: PayloadCodec[],
-  payloads: Record<string, Payload> | null | undefined,
-  context?: SerializationContext
-): Promise<Record<string, DecodedPayload> | null | undefined> {
-  if (payloads == null) return payloads;
-  return Object.fromEntries(
-    await Promise.all(Object.entries(payloads).map(async ([k, v]) => [k, (await decode(codecs, [v], context))[0]]))
-  );
 }
 
 /**
@@ -374,28 +363,6 @@ export async function decodeFailure(
         }
       : undefined,
   };
-}
-
-/**
- * Return a new {@link ProtoFailure} with `codec.encode()` run on all the {@link Payload}s.
- */
-export async function decodeOptionalFailure(
-  codecs: PayloadCodec[],
-  failure: ProtoFailure | null | undefined,
-  context?: SerializationContext
-): Promise<DecodedProtoFailure | null | undefined> {
-  if (failure == null) return failure;
-  return await decodeFailure(codecs, failure, context);
-}
-
-/**
- * Mark all values in the map as decoded.
- * Use this for headers, which we don't encode.
- */
-export function noopDecodeMap<K extends string>(
-  map: Record<K, Payload> | null | undefined
-): Record<K, DecodedPayload> | null | undefined {
-  return map as Record<K, DecodedPayload> | null | undefined;
 }
 
 export async function encodeUserMetadata(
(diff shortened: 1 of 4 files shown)


packages/worker/src/workflow-codec-runner.ts
```diff
@@ -6,17 +6,14 @@ import type {
...
   decodeOptionalSingle,
-  encodeMap,
-  encodeOptional,
-  encodeOptionalFailure,
-  encodeOptionalSingle,
   noopDecodeMap,
-  noopEncodeMap,
-  noopEncodeSearchAttrs,
+  visit,
+  walkWorkflowActivationCompletion,
```
cconstable, line 13: I think some of these are now unused (this file was the last caller). Would like to remove the dead code.

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -369,245 +352,74 @@ export class WorkflowCodecRunner {
...
+            }
+            return context;
           }
-        : null,
-      successful: completion.successful
-        ? {
-            ...completion.successful,
-            commands: completion.successful.commands
-              ? await Promise.all(
-                  completion.successful.commands.map(async (command) => {
```
cconstable, line 407: We were previously using `Promise.all` here to concurrently resolve a list of promises (one for each command that needs to be run through a codec) and each of those promises was eagerly started.

In the new visitor implementation, we call `visit` which walks and processes each payload serially. To match the old behavior I think we'll want to pass in a high concurrent visit limit. I don't think we have an "unbounded" option yet... having a high limit is probably better anyways.

```ts
import { limit } from '@temporalio/common/lib/concurrency/limit';

// on the WorkflowCodecRunner class. its created fresh for every completion
private readonly codecEncodeLimit = limit(MAX_CONCURRENT_CODEC_ENCODES);

// passed into the visitor options
limit: this.codecEncodeLimit,
```

something like that

cconstable (changes requested):

packages/worker/src/workflow-codec-runner.ts
cconstable: I spend more time typing this than it would have taken to actually make the fix. i'll push this. 😄

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -369,245 +361,74 @@ export class WorkflowCodecRunner {
...
+    await visit<coresdk.workflow_completion.IWorkflowActivationCompletion, SerializationContext>(
+      encodedCompletion,
+      walkWorkflowActivationCompletion,
+      {
+        transformPayload: async (payload, context) => (await encode(this.codecs, [payload], context))[0]!,
+        transformPayloads: (payloads, context) => encode(this.codecs, payloads, context),
+        initialContext: this.workflowContext,
+        skipHeaders: true,
+        skipSearchAttributes: true,
+        limit: this.codecEncodeLimit,
```
cconstable: previous walk was unbounded. originally this PR was serial. added some reasonably high limit

packages/common/src/internal-non-workflow/codec-helpers.ts
```diff
@@ -43,16 +43,6 @@ export async function encode(
...
-/** Run {@link PayloadCodec.encode} on `payloads` */
-export async function encodeOptional(
-  codecs: PayloadCodec[],
-  payloads: Payload[] | null | undefined,
-  context?: SerializationContext
-): Promise<EncodedPayload[] | null | undefined> {
-  if (payloads == null) return payloads;
-  return await encode(codecs, payloads, context);
-}
-
```
cconstable, line 56: removing some dead code in this file as workflow-codec-runner was last caller

cconstable (approved):

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -88,279 +91,56 @@ export class WorkflowCodecRunner {
     };
   }
 
-  private async encodeUserMetadata(
```
cconstable, line 91: payload visitor handles walking the `summary` and `details` sites now so this is no longer needed

packages/worker/src/workflow-codec-runner.ts
```diff
@@ -88,279 +91,56 @@ export class WorkflowCodecRunner {
...
-    };
-  }
-
   /**
    * Run codec.decode on the Payloads in the Activation message.
    */
   public async decodeActivation<T extends coresdk.workflow_activation.IWorkflowActivation>(
     activation: T
   ): Promise<Decoded<T>> {
-    return coresdk.workflow_activation.WorkflowActivation.fromObject(<
```
cconstable, line 133: most of this PR is just deleted code. the new workflow-codec-runner is much smaller and it might be helpful to review that first without all the noisy diffs.

THardy98 (approved):

PR #2332, opened 2026-08-19
feat(workflow): add TypeInfo to Queries

THardy98:
## What was changed

Added TypeInfo to Query definitions and introduced `queryWithOptions` for string-named Queries. Query arguments and results now use the resolved TypeInfo across Client conversion, interceptors, and Workflow handlers, including when an inbound interceptor retargets a Query to another handler.

Reviewing commit-by-commit separates the definition, call-site, test, and changelog changes.

## Why?

JSON conversion otherwise loses application runtime types at the Query boundary. Definitions can now carry their conversion contract, while string callers can provide the same information explicitly.

Existing Query calls without TypeInfo retain their current behavior. There is no rollout step or manual action required.

## Checklist

1. Closes: N/A

2. How was this tested: Proto, Common, Client, Workflow, and Test builds; all 31 TypeInfo tests; ESLint; Prettier; `ts-prune`; and `git diff --check`.

### DIFF
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -1847,12 +1847,6 @@ export class WorkflowClient extends BaseClient {
         queryName: string,
         options: WorkflowQueryOptions<Args>
       ): Promise<Ret> {
-        if (typeof queryName !== 'string') {
-          throw new TypeError(
-            'Query TypeInfo can only be provided when querying by name. ' +
-              'Use defineQuery(..., { typeInfo }) on the Query definition instead.'
-          );
-        }
         return await _query(queryName, options.args ?? [], options.typeInfo);
       },
     };
--- a/packages/client/src/workflow-client.ts
+++ b/packages/client/src/workflow-client.ts
@@ -1037,7 +1037,9 @@ export class WorkflowClient extends BaseClient {
       execution: input.workflowExecution,
       query: {
         queryType: input.queryType,
-        queryArgs: { payloads: await encodeToPayloadsWithContext(dataConverter, context, input.args) },
+        queryArgs: {
+          payloads: await encodeToPayloadsWithContext(dataConverter, context, input.args, input.typeInfo?.inputTypes),
+        },
         header: { fields: input.headers },
       },
     };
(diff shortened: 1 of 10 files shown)


packages/client/src/workflow-client.ts
```diff
@@ -1816,16 +1837,23 @@ export class WorkflowClient extends BaseClient {
...
-          headers: {},
-        }) as Promise<Ret>;
+          typeof def === 'string' ? undefined : def.typeInfo
+        );
+      },
+      async queryWithOptions<Ret, Args extends any[]>(
+        queryName: string,
+        options: WorkflowQueryOptions<Args>
+      ): Promise<Ret> {
+        if (typeof queryName !== 'string') {
```
chris-olszewski: Is this just here to be extra defensive of someone doing a cast of `myQuery = defineQuery(...) as string`?

packages/test/src/test-type-info.ts
```diff
@@ -293,6 +296,113 @@ test('Update-with-Start carries Workflow input and output TypeInfo from the defi
...
+test('Client Query interceptor receives definition TypeInfo', async (t) => {
+  const h = configurableHelpers(t, t.context.workflowBundle, t.context.env);
+  const client = new Client({
+    connection: t.context.env.client.connection,
+    namespace: t.context.env.client.options.namespace,
+    interceptors: {
+      workflow: [
+        {
+          async query(input, next) {
+            t.is(input.typeInfo, orderQueryTypeInfo);
```
chris-olszewski: Not that it is incredibly important, but do want to support that interceptor receive the same type info object as the one provided to `defineQuery`? Not sure why a user would want to modify type info from an interceptor

chris-olszewski (approved):
LGTM, one question around the passing of the exact same type info object around. It might not be incredibly clear that if they rewrite fields on `input.typeInfo` instead of fully redefining it that other callers would observe the change. Unsure if changing a typeInfo partially would ever make sense so 🤷

packages/client/src/workflow-client.ts
THardy98: Ah this is a mistake.
This should be in the `query` method above in the `else` conditional. Will move

THardy98: to get around some casting, but it's not worth keeping, removed

PR #2334, opened 2026-08-19
feat(nexus): apply TypeInfo in Workflow callers and handlers

THardy98:
## What was changed

Added experimental `TypeInfo` support to Nexus operation definitions used by Workflow callers and Worker handlers. Operation input and output metadata now flows through Workflow outbound interceptors and Worker handler conversion, including synchronous results and asynchronous Workflow-backed operations.

Workflow-backed asynchronous handlers must configure matching TypeInfo on the backing Workflow because that Workflow produces the eventual result payload. Calls without TypeInfo retain their existing best-effort conversion behavior.

## Why?

Without operation TypeInfo, Nexus serialization boundaries lose application runtime types such as class instances and `bigint` fields. Generated Nexus APIs need the operation definition to preserve those types across Workflow callers and handlers.

The stable `nexus-rpc@^0.0.3` dependency is inherited from `main`; this PR contains no dependency or lockfile changes. No rollout or manual steps are required.

## Checklist

1. Closes: N/A

2. How was this tested: Frozen install against stable `nexus-rpc@0.0.3`; Common, Workflow, Worker, Nexus, and Test builds; 4 focused Common TypeInfo tests; 19 Nexus Workflow/handler tests; ESLint; Prettier; `ts-prune`; and `git diff --check`.

3. Any docs updates needed? No; documentation will accompany the completed experimental TypeInfo API.

### DIFF
--- a/packages/workflow/src/nexus.ts
+++ b/packages/workflow/src/nexus.ts
@@ -1,4 +1,5 @@
 import type * as nexus from 'nexus-rpc';
+import { toPayloadWithTypeInfo } from '@temporalio/common';
 import { msOptionalToTs } from '@temporalio/common/lib/time';
 import { userMetadataToPayload } from '@temporalio/common/lib/user-metadata';
 import { makeProtoEnumConverters } from '@temporalio/common/lib/internal-workflow/enums-helpers';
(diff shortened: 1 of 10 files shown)


packages/workflow/src/nexus.ts
```diff
@@ -127,12 +128,15 @@ export function createNexusServiceClient<T extends nexus.ServiceDefinition>(
...
-            (options.service.operations[operation]?.name as string)
-          : operation.name;
+      let operationDefinition: nexus.OperationDefinition<any, any> | undefined;
+      if (typeof operation === 'string') {
+        operationDefinition = options.service.operations[operation];
+      } else {
+        operationDefinition = operation;
+      }
+      // Casting as string preserves the existing behavior for an unknown operation property name:
+      // the resulting command produces a NexusOperationFailure.NOT_FOUND.
```
chris-olszewski: I think the previous comment is more helpful. This one reads as if this is some compatibility behavior, but reading the other one, I think this is behavior we will want to keep around.

chris-olszewski (approved):
Super minor comment critique

PR #2341, opened 2026-08-20
fix(google-adk-agents): surface the model failures ADK absorbs

xumaple:
A Workflow whose model Activity had permanently failed was completing successfully. `LlmAgent.runAndHandleError` catches any `Error` escaping a model call, turns it into an event carrying `errorCode`/`errorMessage` and no content, and lets the run finish, so nothing ever reaches Workflow code to fail on. A cancelled in-flight model call went the same way.

`TemporalModel` now records such a failure against the inbound frame it was thrown in (the main function, or the Signal or Update handler that ran the turn), and an interceptor re-raises it when that frame returns. Recording is keyed on ADK's `adk_agent_name` request label, so a request without that label still just throws at the call site where the caller can catch it. A consumer who recovers in an ADK `onModelErrorCallback` calls `markModelFailureHandled(error)`, the only public API addition, to drop that one failure and keep their recovery.

The guarantee is scoped to one rule: await an agent turn in the frame that started it. An Update handler's failed turn rejects that Update rather than the Workflow, a Signal handler that awaits its turn fails the execution, and an operator cancel still ends CANCELLED, while a `CancellationScope.withTimeout` the consumer imposed is their own decision and leaves the Workflow to complete on its fallback. A pending failure also wins over `continueAsNew()`, which would otherwise let the chain reach a successful terminal state. One recorded after the main function returned never surfaces, and that case is documented rather than fixed, since such a Workflow is completing with unfinished handlers and Temporal already warns about it (TMPRL1102).

Note: the interceptor module is appended last, hence innermost, so the re-raise rejects through the outer interceptors rather than past them (a composed observability interceptor must not close its span OK on a Workflow that is about to fail). That position depends on plugin order, so `configureBundler`'s TSDoc now says to list `GoogleAdkPlugin` last in `plugins`.

### DIFF
--- a/contrib/google-adk-agents/src/__tests__/absorbed-failure.test.ts
+++ b/contrib/google-adk-agents/src/__tests__/absorbed-failure.test.ts
@@ -1,8 +1,4 @@
 /**
- * @license
- * Copyright 2025 Temporal Technologies Inc.
- * SPDX-License-Identifier: MIT
- *
  * Terminal-outcome tests for model failures that `@google/adk` absorbs. Because the
  * absorbed failure lets the run finish normally, every case here asserts the
  * *execution's* status, not just what the Workflow returned: a failed model call must
--- a/contrib/google-adk-agents/src/__tests__/absorbed-failure.test.ts
+++ b/contrib/google-adk-agents/src/__tests__/absorbed-failure.test.ts
@@ -1,0 +1,342 @@
+/**
+ * @license
+ * Copyright 2025 Temporal Technologies Inc.
+ * SPDX-License-Identifier: MIT
+ *
+ * Terminal-outcome tests for model failures that `@google/adk` absorbs. Because the
+ * absorbed failure lets the run finish normally, every case here asserts the
+ * *execution's* status, not just what the Workflow returned: a failed model call must
+ * FAIL the execution, a cancel must end it CANCELLED, and a failure the caller
+ * handled must leave it COMPLETED.
+ */
+
+import test from 'ava';
+
+import { ApplicationFailure, TimeoutFailure } from '@temporalio/common';
+
+import { GoogleAdkPlugin } from '../index';
+import {
+  countScheduledActivities,
+  defaultTestProvider,
+  findInCauseChain,
+  setupTestEnv,
+  uid,
+  waitForScheduledActivities,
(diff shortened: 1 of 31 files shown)


contrib/google-adk-agents/src/workflow.ts
```diff
@@ -19,6 +15,8 @@ import './load-polyfills';
 export { TemporalModel } from './model';
 export type { TemporalModelOptions } from './model';
 
+export { markModelFailureHandled } from './absorbed-failure';
```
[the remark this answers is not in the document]
xumaple: package is unreleased. not touching changelog.

brianstrauch, line 18: If this is a new export, it should probably be covered in the README

brianstrauch: copilot had the same suggestion :)

contrib/google-adk-agents/src/__tests__/absorbed-failure.test.ts
```diff
@@ -0,0 +1,338 @@
...
+
+function adkPlugin(): GoogleAdkPlugin {
+  return new GoogleAdkPlugin({ modelProvider: defaultTestProvider() });
+}
+
+test.serial('modelFailureThroughTheRunnerFailsTheWorkflow', async (t) => {
+  const env = getEnv();
+  const taskQueue = uid('adk-agent-fail');
+  const workflowId = uid('wf-agent-fail');
+  await withWorker(env, { taskQueue, plugins: [adkPlugin()] }, async () => {
```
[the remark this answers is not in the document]
xumaple: added everywhere

contrib/google-adk-agents/src/workflow.ts
xumaple: added :)

brianstrauch: Did we consider making an upstream change to avoid introducing this, by the way?

brianstrauch (approved):

contrib/google-adk-agents/src/workflow.ts
xumaple: it's coming!

PR #2343, opened 2026-08-20
feat(activity): add TypeInfo support for activities

THardy98:
## What was changed

Added experimental `TypeInfo` support across the complete Activity serialization surface.

Activity implementations can attach shared input and output metadata with `setActivityOptions`. Workflow and Local Activity proxies accept the same typed metadata map, and outbound interceptors may replace the selected metadata before conversion. Workers apply definition metadata when decoding arguments and encoding results, including default Activity handlers.

Standalone Activity Clients apply TypeInfo to start and execute calls, retain output metadata on returned handles, and accept output metadata when reconstructing detached handles. Async completion accepts output TypeInfo for task-token and full-ID completion without exposing ignored serialization-context options.

Calls without TypeInfo retain their existing behavior. Existing Activity registration and positional `getHandle(activityId, runId?)` calls remain unchanged; no rollout or manual steps are required.

## Why?

Without TypeInfo, Activity serialization boundaries lose application runtime types such as class instances and `bigint` fields. Activities require metadata on both the caller and Worker sides because Workflow isolates can import Activity types but not their implementations.

## Checklist

1. Closes: N/A

2. How was this tested: Common, Activity, Client, Workflow, Worker, Testing, and Test TypeScript builds; changed-file ESLint and Prettier; `ts-prune`; and `git diff --check`. The focused TypeInfo suite passed all 43 tests, and all 6 standalone/async Activity TypeInfo tests passed. The full standalone Activity file passed 23 tests and encountered one unrelated existing start-delay timing assertion.

3. Any docs updates needed? No; documentation will accompany the completed experimental TypeInfo API.

### DIFF
--- a/packages/activity/src/index.ts
+++ b/packages/activity/src/index.ts
@@ -117,6 +117,8 @@
 
 import { AsyncLocalStorage } from 'node:async_hooks';
 import type {
+  ActivityDefinitionOptions,
+  ActivityFunction,
   Logger,
   Duration,
   LogLevel,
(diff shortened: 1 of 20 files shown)


packages/common/src/activity-definition-options.ts
```diff
@@ -0,0 +1,40 @@
...
+type ActivityKeys<A> = {
+  [K in keyof A & string]: A[K] extends ActivityFunction ? K : never;
+}[keyof A & string];
+
+/**
+ * Type information keyed only by Activity-valued properties of `A`.
+ *
+ * @experimental
+ */
+export type ActivityTypeInfoMap<A> = Partial<Record<ActivityKeys<A>, PayloadTypeInfo>>;
```
chris-olszewski, line 23: Poked around and tried to get a stronger type check here than just that the names match up. I might be worth investigating in the future, but it requires enough changes to `TypeInfo` and friends that it should just be done in a later PR if we want that functionality.

packages/activity/src/index.ts
```diff
@@ -133,14 +135,30 @@ import type { ActivityCancellationDetailsHolder } from '@temporalio/common/lib/a
...
+ *
+ * Type information is used by a Worker to decode the Activity arguments and encode its result.
+ *
+ * @experimental
+ */
+export function setActivityOptions<Args extends any[], ReturnType>(
+  options: ActivityDefinitionOptions,
+  fn: ActivityFunction<Args, ReturnType>
+): void {
+  Object.assign(fn, { activityDefinitionOptions: options });
```
chris-olszewski, line 159: I don't love mutating the activity functions for providing type hints, but can't come up with a better solution. Probably just something to address when we get to bundler/`proxyActivity` changes.

chris-olszewski (approved):

PR #2344, opened 2026-08-20
feat(client): add TypeInfo to standalone Activities

THardy98:
## What was changed

Standalone Activity Clients now accept experimental `TypeInfo` when starting or executing an Activity. The client applies input information before sending arguments, retains output information on the returned handle, and accepts output information when reconstructing a handle by ID. Activity Client interceptors can replace the metadata before the request is encoded.

## Why?

Activity TypeInfo already covers Workflow Activity proxies and Worker handlers. Standalone Activities need the same conversion boundary so rich application values can be preserved when an Activity is started directly from a Client.

Existing calls and positional `getHandle(activityId, runId?)` remain unchanged. This adds an options-object `getHandle` overload for result decoding; no rollout or manual steps are required.

This draft is stacked on the Workflow/Worker Activity TypeInfo PR.

## Checklist

1. Closes: N/A

2. How was this tested: Client build, Prettier, and `git diff --check`. Focused standalone Activity integration validation remains pending the parent draft's test fix.

3. Any docs updates needed? No; the experimental API is covered by TSDoc and the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ to docs, or any other relevant information.
 
 ### Added
 
+- **Experimental**: Standalone Activity Clients can use `TypeInfo` to convert Activity inputs and results.
 - **Experimental**: Activities can use `TypeInfo` to convert inputs and results when called through Workflow Activity
   proxies and handled by Workers.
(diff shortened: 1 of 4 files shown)


THardy98:
Superseded by the consolidated Activity TypeInfo PR #2343.

PR #2345, opened 2026-08-20
feat(client): add TypeInfo to async Activity completion

THardy98:
## What was changed

Async Activity completion can now use experimental `TypeInfo` to encode a successful result. `complete` accepts an optional output TypeInfo for both task-token and full-Activity-ID calls; failure, cancellation, and heartbeat APIs remain unchanged.

This completes the asynchronous-completion portion of the Activity TypeInfo stack. A Worker or Activity client can now return a rich value, such as a class containing a `bigint`, after an Activity elects asynchronous completion.

This draft is stacked on the standalone Activity Client TypeInfo PR.

## Why?

Activity TypeInfo already covers normal Workflow and Standalone Activity results. Asynchronous completion is a separate Client serialization boundary, so without this change the later completion result bypasses the output conversion supplied to the calling Activity API.

## Checklist

1. Closes: N/A

2. How was this tested: Client build, Prettier, and `git diff --check`. Focused task-token and full-ID runtime coverage is included but remains pending the parent draft's test fix.

3. Any docs updates needed? No; this is an experimental API and the CHANGELOG records it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ to docs, or any other relevant information.
 - **Experimental**: Standalone Activity Clients can use `TypeInfo` to convert Activity inputs and results.
 - **Experimental**: Activities can use `TypeInfo` to convert inputs and results when called through Workflow Activity
   proxies and handled by Workers.
+- **Experimental**: Async Activity completion can use `TypeInfo` to convert successful Activity results.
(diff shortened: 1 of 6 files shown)


THardy98:
Superseded by the consolidated Activity TypeInfo PR #2343.

PR #2349, opened 2026-08-20
fix(bun): support Bun 1.4.0

### LINKED ISSUE #2348: [Feature Request] Investigate Bun 1.4.0 support

This [PR](https://github.com/temporalio/sdk-typescript/pull/2347) pins Bun to 1.3.14 after seeing Bun CI jobs fail. We should look into the CI failures to determine what their cause is.

Bun 1.4.0 working :)

### Additional context

Some CI candidates to look into:
https://github.com/temporalio/sdk-typescript/actions/runs/32398811282
https://github.com/temporalio/sdk-typescript/actions/runs/32397726015

Not much differentiating the failures.

chris-olszewski:
## What was changed
- For almost all of our `isBun` checks, move them to `isBunPre1_4` and gate them.
- There is new weirdness around numeric props on VM context. Was able to change our approach in a way that it works for both 1.3 and 1.4.
- Run *all* packages' tests using Bun, not just the `test` package.
- Add deprecation warning when starting a worker using Bun 1.4 or earlier

## Why?
Bun 1.4.0 introduced some changes that broke our previous workarounds. 

## Checklist

1. Closes #2348 

2. How was this tested:
Enabled running against latest Bun in CI

3. Any docs updates needed?
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,8 +239,6 @@ jobs:
 
       - name: Run Tests (Bun)
         if: matrix.node == 'bun'
-        # Run every workspace package's test script, as in the Node path above.
-        # AVA_RUNTIME=bun makes each AVA wrapper launch its tests under Bun.
         run: pnpm run test
         env:
           AVA_RUNTIME: bun
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,11 +239,9 @@ jobs:
 
       - name: Run Tests (Bun)
         if: matrix.node == 'bun'
-        working-directory: packages/test
-        # Same wrapper as the Node path, but AVA_RUNTIME=bun makes it launch ava
-        # under Bun (mirrors `bun run -b ava`), so Bun is still exercised. Bun runs
-        # the TypeScript wrapper directly (no tsx needed).
-        run: bun ../../scripts/ava-ci.ts ./lib/test-*.js
+        # Run every workspace package's test script, as in the Node path above.
+        # AVA_RUNTIME=bun makes each AVA wrapper launch its tests under Bun.
+        run: pnpm run test
         env:
           AVA_RUNTIME: bun
           RUN_INTEGRATION_TESTS: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,7 +171,7 @@ jobs:
         uses: ./.github/actions/setup-bun-deps
         with:
           platform: ${{ matrix.platform }}
-          version: latest
+          bun-version: latest
 
       # On Windows, the 'runner.temp' variable uses backslashes as path separators, but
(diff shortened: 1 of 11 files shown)


contrib/langsmith/src/__tests__/test-query-filter.ts
```diff
@@ -31,7 +31,13 @@ test('internal-query filtering: traces user queries but not Temporal-internal qu
         workflowId: `qf-${Date.now()}`,
       });
       await handle.query(workflows.myQuery);
-      await handle.query('__stack_trace');
+      if ('bun' in process.versions) {
```
chris-olszewski, line 34: Test was never getting run via Bun in CI

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -50,59 +50,49 @@ function generateNodeCallIntoScopeScript(): string {
...
-// stale values for numeric properties after modification. We must read/write numeric
-// properties directly.
-function generateBunCallIntoScopeScript(): string {
+/**
+ * Bun has inconsistent handling of numeric properties on the VM-context. In
+ * Bun 1.3.x, reads from the global in the VM may not see numeric properties present on the VM context
+ * object. Bun 1.4.0 fixes those reads, but writes and deletes from inside the VM
+ * global still do not update the VM context.
+ *
+ * To work around this we replace the `globalThis` value exposed inside the VM with a proxy.
```
mjameswh, line 59: If Bun supports `globalThis` being a proxy, then it should be possible to completely avoid the swaps in/out to the properties bag, which would be significantly simpler and more efficient, assuming it works fine.

.github/workflows/ci.yml
```diff
@@ -238,11 +239,7 @@ jobs:
 
       - name: Run Tests (Bun)
         if: matrix.node == 'bun'
-        working-directory: packages/test
-        # Same wrapper as the Node path, but AVA_RUNTIME=bun makes it launch ava
-        # under Bun (mirrors `bun run -b ava`), so Bun is still exercised. Bun runs
-        # the TypeScript wrapper directly (no tsx needed).
-        run: bun ../../scripts/ava-ci.ts ./lib/test-*.js
+        run: pnpm run test
```
THardy98, line 242: noice

contrib/langsmith/src/__tests__/test-query-filter.ts
THardy98: i'm confused why the assertion in the Bun case is a failure, but not otherwise

THardy98 (approved):

contrib/langsmith/src/__tests__/test-query-filter.ts
chris-olszewski: Bun doesn't provide the APIs we use in Node to add stack trace query. We reject the query that the query isn't enabled unlike Node where it is supported.

PR #2354, opened 2026-08-25
test: adapt integration tests for Temporal Cloud

akramcodez:
## What was changed
Adapted four integration tests to be fully compatible with Temporal Cloud testing environments by removing their `.cloud-pending` extensions and injecting dynamic configurations:
- `test-integration-extstore.ts`: Updated `Client` instantiation to inject the configured `namespace` (via `t.context.env.client.options.namespace`).
- `test-integration-split-one.ts`: Injected `config.env.client.options.namespace` into the `WorkflowInfo` payload assertions.
- `test-bridge.ts`: Replaced hard-coded targets with `process.env.TEMPORAL_ADDRESS` and `process.env.TEMPORAL_NAMESPACE`.
- `test-sinks.ts`: Injected Cloud environment variables into all raw `NativeConnection`, `WorkflowClient`, and `Worker.create` instances, without modifying the underlying test architecture.

## Why?
These integration tests previously failed against Temporal Cloud because they hard-coded the `default` namespace and/or assumed a local server running on `127.0.0.1:7233`. Temporal Cloud uses a dynamically randomized namespace, so these tests were updated to dynamically obtain the active test environment configurations. 

This completes 4 of the remaining files marked by `pnpm cloud:inventory` under "Needs Cloud adaptation".

## Checklist

1. Addresses #2326 

2. How was this tested:
- **Local Testing**: Ran `pnpm --filter @temporalio/test run test:integration` to ensure the local ephemeral test suites still pass.
- **Inventory Check**: Verified `pnpm cloud:inventory` drops the "Needs Cloud adaptation" count from 31 down to 27, and correctly marks these 4 files as Cloud candidates.

3. Any docs updates needed?
No docs updates needed.

PR #2363, opened 2026-08-27
test: reduce Cloud integration test flakiness and cap harness timeout

DABH:
Two robustness fixes for the flaky Cloud Integration Tests job on `main`, both test/CI-only:

- `test-integration-update.ts` — bump `UPDATE_ADMITTED_WAIT_MS` 20s → 60s. Those `waitUntil(updateHasBeenAdmitted, …)` polls were sized for a localhost dev server; Cloud round-trip latency intermittently blows the 20s budget, which is the dominant recurring Cloud failure (the same `coherent promise completion ordering` test passes on Cloud in other runs).
- `run-cloud-tests.ts` — run ava in its own process group and hard-kill the whole tree, both on a 15m wall-clock cap and on exit. When a test wedges, ava's 60s inactivity timeout fires but leaked workers stay alive holding the inherited `tee` stdout, so `pnpm test:cloud | tee` never gets EOF and the step hangs to its 20m cap (surfaces as the `Channel has been shut down` cascade). So a wedged run is now force-killed at a 15m wall-clock cap instead of hanging out the step's full 20m, and leaked workers get reaped on exit so they can't hold the pipe open. (Not a fast-fail: a mid-suite channel death still cascades until the cap — `--fail-fast` would cut that, at the cost of only surfacing the first failure per run.)

Deliberately leaving the `integration-split-one › child-workflow-invoke` test that actually wedged on current HEAD to its owners — #2354 enabled it on Cloud.

Verified `tsc -p scripts/tsconfig.json` and `prettier --check` clean on both. Also confirmed on this PR's own Cloud run that the 15m cap fires and kills the tree. Test-infra only; please apply `skip-changelog`.


THardy98 (approved):
agh, sorry you had to deal with this, thanks!

PR #2365, opened 2026-08-28
💥 feat(extstore): pass a dedicated select context to the driver selector

jmaeagle99:
## What was changed

Pass a dedicated `StorageDriverSelectContext` type to the external storage driver selector instead of reusing `StorageDriverStoreContext`.

## Why?

Allow selector and store contexts to evolve independently.


packages/test/src/test-extstore-runner.ts
```diff
@@ -165,6 +166,29 @@ test('store aborts sibling drivers on first failure', async (t) => {
   t.true(driverBAborted);
 });
 
+test('selector receives a select context carrying the store target', async (t) => {
```
cconstable, line 169: This tests says "selector receives a select context" but it actually doesn't check that.

cconstable (commented):
The changes lgtm but I think the test is broken.

cconstable (approved):

packages/test/src/test-extstore-runner.ts
cconstable: nvm me. I misread this. I think it's fine 👍

PR #2366, opened 2026-08-28
feat(client): infer detached handle TypeInfo from definitions

VegetarianOrc:
## What was changed

Workflow, standalone Activity, and standalone Nexus Operation handles accept their runtime definition through handle options. The definition and `typeInfo` options are mutually exclusive. At runtime, explicit `typeInfo` is used over any provided definition.

Additionally the types for the various getHandle implementations have been tightened to prevent specifying a mismatched type arg and output type. 

## Why?

Previously, you were allowed to specify a type arg for a workflow, activity, or nexus op but the type info did not enforce that the return type of the definition matched the output type of the type info.

```typescript
  client.activity.getHandleWithOptions<Receipt>(activityId, {
    typeInfo: {
      outputType: outputTypeForInvoice,
    },
  });
```

Accepting the runtime definitions provides a convenient way to get a handle for a definition without separately specifying the type info. This helps with the scenario where the return type requires a typeInfo to be specified to properly decode. Now instead of you can call `client.nexus.getHandle(operationId, {operation: myOperation})` to get a `NexusOperationHandle<NexusOperationHandleResult<typeof myOperation>>` instead of specifying 
```typescript
client.nexus.getHandle<typeof myOperation>(operationId, {
    typeInfo: { outputType: myOperation.outputType }
})
```

## Checklist

1. How was this tested:

   - Added compile time tests to confirm type behavior.
   - Added integration coverage for Workflow, standalone Activity, and standalone Nexus Operation result conversion.


packages/client/src/activity-client.ts
```diff
@@ -178,16 +179,32 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
    *
    * @param activityId ID of the Activity.
    * @param options Options identifying the Activity run and describing its result type.
    * @returns Handle to the specified Activity execution.
    *
    * @experimental Standalone Activities are experimental. APIs may be subject to change.
    */
-  getHandleWithOptions<R = any>(activityId: string, options: GetActivityHandleOptions): ActivityHandle<R> {
-    return this.createHandle(activityId, options.runId, options.typeInfo?.outputType);
+  getHandleWithOptions<A extends ActivityFunction>(
```
bergundy, line 190: You don't need a separate function named `WithOptions`, you can overload `getHandle` directly to accept either `runId?: string` or `options: ...`.

bergundy (commented):
I'm concerned that all of the static `getHandle<T>` methods are foot guns now that we've introduced runtime type information. Maybe we should remove them if they're experimental and deprecate ones that are stable?

packages/client/src/activity-client.ts
VegetarianOrc: Yeah, the activity client has them separate. We can consider deprecating or removing for sure, but I didn't necessarily want to include that change in this attempt to get some extra type safety.

VegetarianOrc:
> I'm concerned that all of the static `getHandle<T>` methods are foot guns now that we've introduced runtime type information. Maybe we should remove them if they're experimental and deprecate ones that are stable?

Would your suggestion be to expose only the options that accept the runtime definitions of the workflow/activity/operation?

packages/client/src/activity-client.ts
THardy98: Yeah - maybe oversight on my part. I'm in favour of reducing the API surface. 

If you want to keep this change tight, we should track with an issue and get to it while the API is still experimental.

packages/client/src/activity-client.ts
```diff
@@ -178,16 +179,32 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
+    activityId: string,
+    options: ActivityHandleDefinitionOptions<A>
+  ): ActivityHandle<Awaited<ReturnType<A>>>;
+  getHandleWithOptions<R = any>(
+    activityId: string,
+    options: Replace<GetActivityHandleOptions<R>, { activity?: never }>
```
THardy98, line 196: Do we need the `Replace` here? `activity` is already typed as `activity?: never`

Also applies to the other overloads (i.e. workflow, nexus)

packages/client/src/helpers.ts
```diff
@@ -18,6 +18,10 @@ import type {
   WorkflowExecutionStatusName,
 } from './types';
 
+export type OutputTypeInfo<T> = {
```
THardy98, line 21: I think `type-info.ts` is probably a better home for this than in the `client` package

THardy98 (commented):
Looks reasonable to me. Feel free to ping me when you're RFR for a final look / approval

bergundy:
> > I'm concerned that all of the static `getHandle<T>` methods are foot guns now that we've introduced runtime type information. Maybe we should remove them if they're experimental and deprecate ones that are stable?
> 
> Would your suggestion be to expose only the options that accept the runtime definitions of the workflow/activity/operation?

Yeah. That would be better IMHO.

packages/client/src/activity-client.ts
bergundy: Instead of these complex types, it would be better to `Simplify` them.

```ts
type Simplify<T> = {
  [K in keyof T]: T[K];
} & {};
```

PR #2368, opened 2026-08-31
Align PayloadValidationError conversion behavior

bergundy:
## Summary

- handle exact non-retryable PayloadValidationError failures according to the client, workflow, activity, signal, update, query, and Nexus conversion boundary
- preserve Workflow Task replay, activity-output retryability, signal isolation and metrics, and retryable Nexus output semantics
- add focused unit and server-backed regression coverage for converters and codecs

## Testing

- pnpm build
- pnpm lint:check
- 6 common helper tests
- 30 focused client, worker, activity, and Nexus unit tests
- 19 workflow/activity server-backed tests
- 11 Nexus server-backed tests


packages/client/src/activity-client.ts
```diff
@@ -362,7 +366,7 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
       idReusePolicy: encodeActivityIdReusePolicy(input.options.idReusePolicy),
       idConflictPolicy: encodeActivityIdConflictPolicy(input.options.idConflictPolicy),
       searchAttributes,
-      header: { fields: input.headers },
+      header: { fields: await encodeMap(this.dataConverter.payloadCodecs, input.headers) },
```
bergundy: This should be reverted, it's an unrelated change.

PR #2369, opened 2026-09-01
Nexus: raw ActivityClient starts inherit request ID and links

tekkaya:
## Summary

`TemporalNexusClient.startActivity()` guards the one Activity start that can complete a Nexus operation, but a synchronous handler that starts additional Activities must call the raw `ActivityClient`. Those starts previously omitted the inbound Nexus request ID and links, so a Nexus-task redelivery could start a duplicate Activity instead of resolving to its original run.

Raw Activity starts now inherit the handler’s request ID and inbound links. Completion callbacks remain limited to the guarded start, since only that start can complete the operation.

Aligns with the [sdk-go PR #2633](https://github.com/temporalio/sdk-go/pull/2633) and [sdk-java PR #3048](https://github.com/temporalio/sdk-java/pull/3048).

## Changes

- Add an internal ambient Nexus Activity-start context for the duration of a `startOperation` handler invocation.
- Make raw `ActivityClient` starts inherit the inbound Nexus request ID and links.
- Forward server-provided Activity response links to the operation’s outbound links.
- Leave guarded-start completion-callback behavior unchanged.

## Testing

- Added a redelivery test showing that a raw Activity start runs exactly once after a retryable handler failure.
- Added link-propagation coverage for standalone Nexus and Workflow callers.
- Existing multiple-start and retry coverage continues to pass.

### DIFF
--- a/packages/nexus/src/workflow-helpers.ts
+++ b/packages/nexus/src/workflow-helpers.ts
@@ -19,7 +19,6 @@ import type {
   WorkflowSignalWithStartOptions as ClientWorkflowSignalWithStartOptions,
 } from '@temporalio/client';
 import { WorkflowUpdateStage, type WorkflowUpdateOptions } from '@temporalio/client';
-import { type temporal } from '@temporalio/proto';
 import type {
   InternalActivityStartOptions,
   InternalWorkflowHandle,
@@ -35,7 +34,6 @@ import {
   InternalWorkflowStartOptionsSymbol,
   InternalWorkflowUpdateOptionsSymbol,
 } from '@temporalio/client/lib/internal';
-import { convertNexusLinkToTemporalLink, convertTemporalLinkToNexusLink } from './link-converter';
 import {
   assertActivityOperationToken,
   assertUpdateWorkflowOperationToken,
@@ -51,10 +49,10 @@ import {
 import {
   getClient,
   getHandlerContext,
-  log,
   type TemporalCancelOperationContext,
   type TemporalStartOperationContext,
 } from './context';
+import { pushResponseLink, requestLinksToTemporalLinks } from './operation-links';
 
 declare const isNexusWorkflowHandle: unique symbol;
 declare const workflowResultType: unique symbol;
@@ -210,40 +208,6 @@ export async function startWorkflow<T extends Workflow>(
   // The Workflow run just started is this operation's async backing operation, so the handle's
   // `update()` must not be able to start another one.
   return createWorkflowHandle(ctx, handle.workflowId, handle.firstExecutionRunId, alreadyBackedReservation);
-}
-
-/**
- * Converts the request links carried on the operation start context into Temporal links so
- * they can be forwarded onto an outgoing Workflow RPC (signal, signalWithStart, start). Links that
- * fail to convert are logged and dropped.
- */
-function requestLinksToTemporalLinks(ctx: nexus.StartOperationContext): temporal.api.common.v1.ILink[] {
-  const links = Array<temporal.api.common.v1.ILink>();
-  if (ctx.inboundLinks?.length > 0) {
-    for (const l of ctx.inboundLinks) {
-      try {
-        links.push(convertNexusLinkToTemporalLink(l));
-      } catch (error) {
-        log.warn('failed to convert Nexus link to Workflow event link', { error });
-      }
-    }
-  }
-  return links;
-}
-
-/**
- * Pushes a response link returned by an outbound Workflow RPC onto the operation's outbound links so
- * the Nexus task handler attaches it to the StartOperationResponse, linking the caller Workflow's
- * NexusOperation history event back to the callee Workflow's event. Callers only invoke this when the
- * server returned a response link; older servers (or CHASM signal response links disabled) leave it
- * unset, in which case there is nothing to push.
- */
-function pushResponseLink(ctx: nexus.StartOperationContext, responseLink: temporal.api.common.v1.ILink) {
-  try {
-    ctx.outboundLinks.push(convertTemporalLinkToNexusLink(responseLink));
-  } catch (error) {
-    log.warn('failed to convert temporal link to Nexus link', { error });
-  }
 }
 
 /**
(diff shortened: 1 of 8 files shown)


tekkaya:
Note: Cloud Integ Tests seems like failing since PR #2367 is merged

packages/client/src/nexus-activity-start-context.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Ambient context set by `@temporalio/worker` for the duration of a Nexus `startOperation` handler
+ * invocation, giving `ActivityClient` access to Nexus request data for raw Activity starts made
+ * during that invocation.
+ *
+ * @internal
+ * @hidden
+ */
+export interface NexusActivityStartContext {
```
VegetarianOrc: This shouldn't be specific to Activity start. This is the context around a StartNexusOperation task and should reflect that rather than being activity specific.

Let's also move it to `packages/nexus/context.ts`

packages/client/src/nexus-activity-start-context.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  /** Inbound links of the Nexus request currently being handled. */
+  readonly links: temporal.api.common.v1.ILink[];
+
+  /** Records a response link returned by an Activity start made during this invocation. */
+  pushResponseLink(link: temporal.api.common.v1.ILink): void;
+}
+
+// Make it safe to use this module with multiple installed copies of @temporalio/client, mirroring
+// the pattern used by packages/nexus/src/context.ts for its own AsyncLocalStorage.
```
VegetarianOrc: Let's make sure to prune out the generated 'thoughts' that spell out things like "mirroring the pattern used by ..."

packages/client/src/activity-client.ts
```diff
@@ -338,11 +344,16 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
       : undefined;
 
     const internalOptions = (input.options as InternalActivityStartOptions)[InternalActivityStartOptionsSymbol];
+    const nexusActivityStartContext = getNexusActivityStartContext();
+
+    // Raw starts inherit the handler's inbound Nexus links. Guarded starts retain their explicit wiring.
```
VegetarianOrc:
```suggestion
```

packages/client/src/activity-client.ts
```diff
@@ -338,11 +344,16 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
+    const ambientLinks = internalOptions ? undefined : nexusActivityStartContext?.links;
 
     return {
       namespace: this.options.namespace,
       identity: this.options.identity,
-      requestId: internalOptions?.requestId ?? randomUUID(),
+      // Reuse the handler request ID to deduplicate starts after Nexus-task redelivery.
```
VegetarianOrc:
```suggestion
```

VegetarianOrc (changes requested):

packages/client/src/nexus-activity-start-context.ts
tekkaya: `StartNexusOperationTaskContext` or `NexusOperationTaskContext` or `NexusStartOperationTaskContext` any other suggestions for the naming? I am leaning towards `NexusStartOperationTaskContext` to align with `NexusStartOperationInput` existing `NexusCancelOperationInput`types

tekkaya: > Let's also move it to /packages/nexus/context.ts

`nexus` package already depends on `client`, so moving this into packages/nexus/context.ts and having `ActivityClient` read it back would create a real circular package dependency. Kept it in `client`, exported only through internal.ts, for that reason.

VegetarianOrc: Thinking about the name made me realize we actually already have a type with these exact fields that is passed to all handlers: [`TemporalStartOperationContext`](https://github.com/temporalio/sdk-typescript/blob/d61c018b572e70433aded334bb36aaed9cf35682/packages/nexus/src/context.ts#L67). This is the type that should be used I think.

packages/client/src/activity-client.ts
VegetarianOrc, line 1: Let's take a look at making these changes via an internal [`ActivityClientInterceptor`](https://github.com/temporalio/sdk-typescript/blob/66effbd2d40ef1e7228afde0bbfd8bc99bad59fe/packages/client/src/interceptors.ts#L376) that is appended at the end of the interceptor chain (so user interceptors run first). That way we can encapsulate this Nexus context specific changes away from this general client implementation. The interceptor can add or modify the internal options and then hopefully we won't need any changes here.

tekkaya: That makes a lot of sense. I'll go with that direction.

tekkaya: Updated ptal @VegetarianOrc

packages/nexus/src/workflow-helpers.ts
```diff
@@ -610,13 +614,50 @@ class TemporalNexusClientImpl implements TemporalNexusClient {
...
-    return getClient();
+    if (this._client == null) {
+      const base = getClient();
+      const activity = new ActivityClient({
+        ...base.activity.options,
+        interceptors: [...(base.activity.options.interceptors ?? []), this.nexusActivityStartInterceptor],
+      });
+      this._client = Object.assign(Object.create(base), { activity }) as Client;
+    }
+    return this._client;
```
VegetarianOrc: Rather than applying the interceptor here, I think we should apply the interceptor on client creation. Doing it here allows users to avoid it by calling `getClient()` directly. The interceptor should no-op if not in a Nexus context so it should be safe to invoke each time. We'll also need to apply the same type of interceptor to the various workflow integration points (start, signal, update, etc). If you'd like to keep this PR focused only on Activities, that's fine let's just make sure we create an issue to track.

tekkaya: I would like to keep this focused on Nexus + SAA and can create an issue to track Nexus + start,signal,update

VegetarianOrc (commented):
I think this looks good to me now, ty for all the updates!

Would love @chris-olszewski or @mjameswh to take a final look before merging.

chris-olszewski (approved):
The interceptor is a nice way to provide this behavior. Nonblocking, but a nice follow up would be asserting how this is displayed in the stack trace. (Along with injecting this interceptor at creation as was mentioned by @VegetarianOrc)

PR #2375, opened 2026-09-02
feat(common): add bidirectional RawValue TypeInfo

THardy98:
## What was changed

Added the experimental `rawValueTypeInfo` export so `RawValue` bypasses payload converters during decoding, after payload codecs have run. Encoding is unchanged because `RawValue` already bypasses payload converters in that direction.

The existing integration test now applies the same TypeInfo at the Workflow and Activity boundaries and verifies that both constructed and direct-payload `RawValue` instances survive the full client → Workflow → Activity → Workflow → client round trip.

## Why?

`RawValue` conversion was previously one-way. Callers could send an already-encoded payload without invoking the configured payload converter, but receivers had no runtime type information telling them to reconstruct a `RawValue`. The wrapper was therefore lost during decoding.

## Checklist

1. How was this tested:

   - Common TypeScript build and focused TypeInfo unit tests (5 passing).
   - Integration TypeScript build and focused Workflow/Activity round-trip test (1 passing).
   - Focused ESLint and Prettier checks.

2. Any docs updates needed?

   No external documentation changes are needed. The new experimental API is documented in `CHANGELOG.md`.

### DIFF
--- a/packages/common/src/converter/payload-converter.ts
+++ b/packages/common/src/converter/payload-converter.ts
@@ -49,6 +49,12 @@ export function toPayloadWithTypeInfo<T, D = T>(
   context: SerializationContext | undefined,
   typeInfo: TypeInfo<T, D> | undefined
 ): Payload {
+  if (isRawValueTypeInfo(typeInfo)) {
+    if (!(value instanceof RawValue)) {
+      throw new ValueError('RawValue TypeInfo requires a RawValue value');
+    }
+    return value.payload;
+  }
   const transferValue = typeInfo?.transferTypeConverter ? typeInfo.transferTypeConverter.toTransferType(value) : value;
   return converter.toPayload(transferValue, context, typeInfo?.payloadConverterHint);
 }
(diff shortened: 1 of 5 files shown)


packages/test/src/test-integration-workflow-info.ts
```diff
@@ -180,37 +175,25 @@ test.serial('can register search attributes to dev server', async (t) => {
...
-        return rv;
-      },
-    },
+    activities: { rawValueActivity },
   });
 
   await worker.runUntil(async () => {
-    const testValue = 'test';
-    const rawValue = new RawValue(testValue);
-    const rawValuePayload = RawValue.fromPayload({
```
chris-olszewski, line 202: Does `RawValue.fromPayload` have some special property that we needed to test before?

packages/common/src/converter/payload-converter.ts
```diff
@@ -265,6 +296,8 @@ export class CompositePayloadConverter implements PayloadConverter {
    */
   public toPayload<T>(value: T, context?: SerializationContext, hint?: ConverterHint<T>): Payload {
     if (value instanceof RawValue) {
+      // TypeInfo-aware calls bypass payload conversion in toPayloadWithTypeInfo. Keep this fallback for direct and
+      // legacy untyped calls; removing it would silently re-encode RawValue and may change Workflow replay payloads.
```
chris-olszewski, line 300: Super minor, but are we intending every call to be typed? Will untyped calls be a legacy pattern?

chris-olszewski (approved):

packages/test/src/test-integration-workflow-info.ts
THardy98: Nah the test was weird before. It would continuously recreate the RawValue after it was decoded so this was a way to make comparisons

packages/common/src/converter/payload-converter.ts
THardy98: I mean I hope it's used in cases where it applies - to preserve rich types, which I imagine probably applies to a lot of use cases. But if you don't need it, don't use it 🤷

PR #2376, opened 2026-09-02
fix(common): preserve Error.cause in
  activity failures

### LINKED ISSUE #2291: [Bug] Activity errors lose Error.cause during conversion

THardy98:
## What changed

Activities convert non-`ApplicationFailure` errors before passing them to the configured failure converter. That conversion copied the error's message, type, and stack but discarded `Error.cause`, so the underlying cause chain never reached Workflow history.

This change preserves error-like causes using the SDK's realm-safe `isError` guard. The existing failure converter then serializes the chain through the standard failure `cause` field. Non-error causes remain omitted, and no protobuf or failure-converter behavior changes.

## Behavior

Before this change, an Activity that threw `new Error('query failed', { cause })` produced an `ApplicationFailure` containing only `query failed`. After this change, the serialized failure also contains the nested cause chain, including causes created in another JavaScript realm.

## Verification

- Built `@temporalio/common` and its dependencies.
- Ran the focused common failure suite: 5 tests passed.
- Ran focused ESLint and Prettier checks.
- Ran `git diff --check`.

Closes #2291.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,7 @@ to docs, or any other relevant information.
 
 ### Fixed
 
+- Activity errors converted to `ApplicationFailure` now preserve native `Error.cause` chains in serialized failures.
 - Local Activities now fall back to a registered `default` activity when the requested type is not
   registered, matching non-local Activity dispatch. Previously the Workflow Task failed immediately
   with `ReferenceError` even if `default` was registered.
--- a/packages/common/src/__tests__/test-failure.ts
+++ b/packages/common/src/__tests__/test-failure.ts
@@ -1,5 +1,48 @@
+import vm from 'vm';
 import test from 'ava';
-import { ApplicationFailure, createPayloadValidationError, defaultFailureConverter, defaultPayloadConverter } from '..';
+import {
+  ApplicationFailure,
+  createPayloadValidationError,
+  defaultFailureConverter,
+  defaultPayloadConverter,
+  ensureApplicationFailure,
+} from '..';
+
+test('ensureApplicationFailure preserves Error.cause through serialization', (t) => {
+  const cause = new Error('connection terminated');
+  const error = new Error('query failed', { cause });
+
+  const applicationFailure = ensureApplicationFailure(error);
+  const failure = defaultFailureConverter.errorToFailure(applicationFailure, defaultPayloadConverter);
+
+  t.is(applicationFailure.cause, cause);
+  t.is(failure.cause?.message, cause.message);
+});
+
+test('ensureApplicationFailure preserves cross-realm Error.cause chains', (t) => {
+  const cause = vm.runInNewContext(
+    `new Error('connection terminated', { cause: new Error('connection refused') })`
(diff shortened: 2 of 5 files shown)


mjameswh (approved):

ethanresnick (commented):
How are you guys thinking about whether the thrown error's cause should get copied onto `ApplicationFailure.cause` vs whether the thrown error itself should be the ApplicationFailure's cause?

For my team, our thrown errors occasionally have extra custom properties that we serialize (usually into the proto's `details`), but that requires the whole thrown error to reach the failure converter, not just its cause.

(Basically, same distinction + use case I was mentioning [here](https://github.com/temporalio/sdk-typescript/issues/2291#issuecomment-5436491419).)

THardy98:
> How are you guys thinking about whether the thrown error's cause should get copied onto `ApplicationFailure.cause` vs whether the thrown error itself should be the ApplicationFailure's cause?
> 
> For my team, our thrown errors occasionally have extra custom properties that we serialize (usually into the proto's `details`), but that requires the whole thrown error to reach the failure converter, not just its cause.
> 
> (Basically, same distinction + use case I was mentioning [here](https://github.com/temporalio/sdk-typescript/issues/2291#issuecomment-5436491419).)

That’s a legitimate use case. Today the Activity path calls `ensureApplicationFailure` before the configured converter, so custom properties on the original thrown error are lost.

However, making _that_ error the `ApplicationFailure`’s cause would duplicate the outer failure and shift the real cause one level deeper. I going to preserve this change as is, it maps the intended change tightly, access to the original thrown error may work better as a change to the failure converter API.

ethanresnick:
>  I going to preserve this change as is... access to the original thrown error may work better as a change to the failure converter API.

👍 makes sense, and access to the original error in the failure converter api would be _wonderful_.

PR #2378, opened 2026-09-03
Retain failed Workflows until Core eviction

bergundy:
## Summary

- Retain the Workflow state when activation processing fails.
- Return `close: false` so Core remains responsible for eviction and disposal.
- Add a regression test that verifies the Workflow is disposed only after a `removeFromCache` activation.

## Why

The activation failure path already documents that Workflow disposal must wait for Core eviction, but it returned `state: undefined` and `close: true`. When an error occurred after the Workflow had been created, such as while encoding an outgoing Nexus input, Core could still send a cache-eviction activation after the worker had dropped the state. That violated the lifecycle contract and could trigger a Workflow execution-context disposal error.

Keeping the state and reporting the Workflow as open makes the implementation match the intended ownership model: Core is the sole driver of Workflow eviction.

### DIFF
--- a/packages/worker/src/worker.ts
+++ b/packages/worker/src/worker.ts
@@ -189,6 +189,8 @@ interface WorkflowWithLogAttributes {
   info: WorkflowInfo;
 }
 
+class WorkflowDisposeError extends UnexpectedError {}
+
 function addBuildIdIfMissing(options: CompiledWorkerOptions, bundleCode?: string): CompiledWorkerOptionsWithBuildId {
   const bid = options.buildId;
   if (bid != null) {
@@ -1532,7 +1534,11 @@ export class Worker {
       activation.jobs = jobs;
       if (jobs.length === 0) {
         this.logger.trace('Disposing workflow', workflow ? workflow.logAttributes : { runId: activation.runId });
-        await workflow?.workflow.dispose();
+        try {
+          await workflow?.workflow.dispose();
+        } catch (cause) {
+          throw new WorkflowDisposeError('Failed to dispose Workflow during eviction', cause);
+        }
         if (!close) {
           throw new IllegalStateError('Got a Workflow activation with no jobs');
         }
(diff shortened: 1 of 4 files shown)


packages/worker/src/worker.ts
```diff
@@ -1641,7 +1641,7 @@ export class Worker {
 
       // We do not dispose of the Workflow yet, wait to be evicted from Core.
       // This is done to simplify the Workflow lifecycle so Core is the sole driver.
-      return { state: undefined, output: { close: true, completion } };
+      return { state: workflow, output: { close: false, completion } };
```
THardy98, line 1660: What happens in the case where either dispose or the eviction itself errors?

(are these even possible states?)

bergundy: Good question!

bergundy: Addressed in a follow up commit.

packages/worker/src/worker.ts
```diff
@@ -1632,6 +1638,16 @@ export class Worker {
...
       });
 
+      if (error instanceof WorkflowDisposeError) {
+        const completion = synthetic
+          ? undefined
+          : coresdk.workflow_completion.WorkflowActivationCompletion.encodeDelimited({
+              runId: activation.runId,
+              successful: {},
+            }).finish();
+        return { state: undefined, output: { close: true, completion } };
```
THardy98, line 1648: Ah so on dispose or eviction error - we return what we had formerly, drop the workflow from TS state

THardy98 (approved):
looks reasonable to me, maybe worth getting another set of eyes on this as there are probably some scenarios/states that i'm not considering or don't know about

mjameswh (approved):

packages/worker/src/worker.ts
```diff
@@ -1632,6 +1638,16 @@ export class Worker {
         workflowExists: workflow !== undefined,
       });
 
+      if (error instanceof WorkflowDisposeError) {
```
mjameswh, line 1641: Given that `WorkflowDisposeError` extends `UnexpectedError`, we'll have already entered the `if` block at lines 1627-1631 and trigger a full worker shutdown. It seems to me like that's basically what you're trying to avoid, no?

mjameswh (changes requested):

packages/worker/src/worker.ts
bergundy: Yes, we should initiate shutdown if dispose fails. That is an unexpected error but we should also close the stream in this case and not wait for core to evict the workflow.

packages/worker/src/worker.ts
bergundy: Yeah, because we dispose as part of eviction already.

mjameswh (approved):

Issue #2326, opened 2026-08-17
[Feature Request] Adapt remaining integration tests to run against Temporal Cloud


THardy98:
### Is your feature request related to a problem? Please describe.

This issue tracks the tests that still require adaptation before they can run against Temporal Cloud.

`pnpm cloud:inventory` on `main` currently reports 68 Cloud candidates and 31 files under **Needs Cloud adaptation**.

Some of the required adaptations include:

- Using the configured environment instead of implicit localhost connections, for example `test-native-connection.cloud-pending.ts`.
- Using the environment namespace instead of hard-coded `default` values, for example `test-serialization-context.cloud-pending.ts`.
- Provisioning required custom search attributes, for example `test-integration-custom-search-attributes.cloud-pending.ts`.
- Polling for Cloud visibility consistency instead of relying on fixed delays, for example `test-integration-workflow-info.cloud-pending.ts`.

### Describe the solution you'd like

Track the remaining adaptations here and land them as small PRs grouped by a shared root cause. The repository inventory is the authoritative, self-updating list:

```sh
pnpm cloud:inventory
```

For each file or coherent group:

1. Identify why it cannot use the configured test environment.
2. Adapt it without weakening the test's intent.
3. Remove `.cloud-pending` from the filename.
4. Run the normal integration tests and `pnpm test:cloud`.
5. Link the implementation PR from this issue.

### Progress

- [ ] #2380 adapts five Worker smoke suites; its inventory reports 73 Cloud candidates and 26 pending files.
- [ ] #2384 adapts the custom payload codec/converter suites; its independent inventory reports 70 candidates and 29 pending files.
- [ ] #2385 adapts the bundler and NYC coverage suites; its independent inventory reports 70 candidates and 29 pending files.
- [ ] #2387 adapts three workflow-behavior suites; its independent inventory reports 71 candidates and 28 pending files.
- [ ] #2389 adapts the NativeConnection lifecycle suite; its independent inventory reports 69 candidates and 30 pending files.
- [ ] #2391 consolidates the ActivityInfo retry-policy case into an existing Cloud candidate; its independent inventory reports 68 candidates and 30 pending files.

After all six PRs merge, the cumulative inventory will be 81 Cloud candidates and 17 pending files.

This issue is complete when `pnpm cloud:inventory` reports no files under **Needs Cloud adaptation**, the Cloud integration job passes, and normal integration coverage remains intact.

### Additional context

- This is part of temporalio/features#851.
[...]

PR #2384, opened 2026-09-03
test: adapt custom data converter tests for Temporal Cloud

THardy98:
## What was changed

The custom payload codec and converter suites now bind their server-backed Workers and Workflow Clients to the configured test environment. Each Worker/client pair shares the same connection, namespace, data converter, and task queue.

The mixed converter file keeps its existing unit tests, while its server-backed cases now register unconditionally and select or provision their environment directly. On this independent branch, the inventory changes from 68 Cloud candidates and 31 pending files to 70 candidates and 29 pending files.

## Why?

The server-backed cases previously used implicit localhost and default-namespace configuration even though their converter behavior is portable. The new wiring allows them to run against Temporal Cloud while preserving every test case and assertion, without relying on the legacy selective integration-test gate.

## Checklist

1. Related issue: #2326. This PR contributes to the tracker but does not close it.

2. How was this tested:
   - `pnpm build`
   - `pnpm lint:check`
   - Focused local AVA run: 10 tests passed
   - Focused run with `RUN_INTEGRATION_TESTS=false`: all 10 tests passed
   - `pnpm cloud:inventory`: 70 candidates, 29 pending
   - Full CI is tracked by this draft PR's checks

3. Any docs updates needed?
   - No. This changes test infrastructure only.

### DIFF
--- a/packages/test/src/test-custom-payload-converter.ts
+++ b/packages/test/src/test-custom-payload-converter.ts
@@ -6,7 +6,7 @@ import { toPayloads } from '@temporalio/common';
 import { coresdk } from '@temporalio/proto';
 import { ProtoActivityResult } from '../protos/root';
 import { protoActivity } from './activities';
-import { cleanOptionalStackTrace, RUN_INTEGRATION_TESTS, Worker } from './helpers';
+import { cleanOptionalStackTrace, Worker } from './helpers';
 import { defaultOptions, isolateFreeWorker } from './mock-native-worker';
 import { messageInstance, payloadConverter } from './payload-converters/proto-payload-converter';
 import { createTestWorkflowEnvironment } from './helpers-integration';
@@ -28,7 +28,7 @@ function compareCompletion(
   );
 }
 
-if (RUN_INTEGRATION_TESTS) {
+function registerIntegrationTests(): void {
   test('Client and Worker work with provided dataConverter', async (t) => {
     const env = await createTestWorkflowEnvironment();
     t.teardown(() => env.teardown());
@@ -90,6 +90,8 @@ if (RUN_INTEGRATION_TESTS) {
     );
   });
 }
+
+registerIntegrationTests();
 
 test('Worker throws on invalid payloadConverterPath', async (t) => {
   t.throws(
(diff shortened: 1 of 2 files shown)


THardy98:
Diff looks big but it's because of the formatting after removing the `RUN_INTEGRATION_TESTS` gate.

packages/test/src/test-custom-payload-converter.ts
```diff
@@ -27,52 +28,66 @@ function compareCompletion(
...
-if (RUN_INTEGRATION_TESTS) {
-  test('Client and Worker work with provided dataConverter', async (t) => {
-    const dataConverter = { payloadConverterPath: require.resolve('./payload-converters/proto-payload-converter') };
-    const taskQueue = 'test-custom-payload-converter';
-    const worker = await Worker.create({
-      ...defaultOptions,
-      workflowsPath: require.resolve('./workflows/protobufs'),
+test('Client and Worker work with provided dataConverter', async (t) => {
+  const env = await createTestWorkflowEnvironment();
+  t.teardown(() => env.teardown());
```
chris-olszewski, line 33: Is there a reason these tests weren't converted to us `makeTestFunction` so we get the env creation/teardown automatically? If so a comment explaining why we can't use that here would be helpful.

packages/test/src/test-custom-payload-converter.ts
```diff
@@ -27,52 +28,66 @@ function compareCompletion(
...
   });
+});
 
-  test('fromPayload throws on Client when receiving result from client.execute()', async (t) => {
-    const worker = await Worker.create({
-      ...defaultOptions,
-    });
+test('fromPayload throws on Client when receiving result from client.execute()', async (t) => {
+  const env = await createTestWorkflowEnvironment();
+  t.teardown(() => env.teardown());
```
chris-olszewski, line 62: Is there a reason these tests weren't converted to us `makeTestFunction` so we get the env creation/teardown automatically? If so a comment explaining why we can't use that here would be helpful.

chris-olszewski (approved):
Overall looks good to me. Q around 2 tests that don't use the standard helpers

Issue #1271, opened 2023-10-26
[Feature Request] Add minimal support to link core-bridge against musl


mjameswh:
### Is your feature request related to a problem? Please describe.

Some TS SDK users have strong requirement of only using alpine-based Docker images. This is a problem, as the `alpine` distribution is based on the `musl` C library (rather than the GNU's `libc` found in most other distributions), which isn't supported at the moment by the `@temporalio/core-bridge` package. Is is therefore not possible to run a TS SDK Worker in an alpine-based container.

The team is hesitant about shipping a `musl`-based native library out of the box, as this would increase the download size for all of our users.

In theory, self-building the native library would be a reasonably simple solution, since `core-bridge` is designed to build the native lib if it can't find the binary file corresponding to the current target environment. However, the TS logic that identify the current target can't recognize the fact that it is running on a `musl`-based system (node itself provides no information on that regard).

### Describe the solution you'd like

- [ ] Modify `core-bridge`'s target environment detection logic to recognize that it is running in a `musl`-based system, and to consequently return the proper build target for those environments.

  With this change, the `@temporalio/core-bridge` will correctly identify that it is running in a `musl` environment, and therefore try to load the `musl`-based native library. Assuming that this library can't be found (ie. because we don't ship the musl), and that the proper go toolchain is installed, then the package would automatically build the native library. This build operation would only happen the first time the application execute, and the built artifact could be copied over. It would therefore be reasonably easy to include this build step as part of a multi-stage Dockerfile.

- [ ] Add some `musl`-based environment to the CI test matrix

mjameswh:
Marking this issue as Good First Contribution — specifically the first part: "adding detection logic to recognize `musl` environment". That requires modifying [this file](https://github.com/temporalio/sdk-typescript/blob/main/packages/core-bridge/common.js). Can probably find hints of a `musl` build by looking into `process` or `os`, or maybe `process.report.getReport()` (IIRC, that last one provides the list of loaded libraries).

Adding `musl` to the CI test matrix might be best left to official maintainers.

korbin:
https://github.com/angellist/sdk-typescript/commit/c34ba641d05541647dfd0aa94bdd81be6f1a0593

I don't have time to pull through the CI-related changes to make this work, but, this does seem to work without issue on an Alpine system with:

``` 
apk add --no-cache build-base cmake protoc protobuf-dev rustup
rustup-init -y

export RUSTFLAGS="-C target-feature=-crt-static"
[...]

PR #2388, opened 2026-09-03
feat(core): ship musl (Alpine) prebuilt binaries and detect musl at runtime

### LINKED ISSUE #1621: [Feature Request] Include musl-based shared library in package

See https://github.com/temporalio/features/issues/594

VictorCodesseira:
## What was changed

- `packages/core-bridge/common.js`: adds `x86_64-unknown-linux-musl` and
  `aarch64-unknown-linux-musl` to the target list. On Linux, `getPrebuiltTargetName()` resolves
  the musl prebuild when the process report carries no `glibcVersionCompiler` — absent on musl
  builds, present on glibc builds even when statically linked — memoized, and falling back to
  glibc on runtimes without report support.
  Unit tests cover the glibc, musl, absent-report, and non-Linux paths.
- `release.yml`: adds `linux-x64-musl` / `linux-arm-musl` to the release build matrix, compiled
  natively in an `alpine:3.20` container (deliberately old, like manylinux_2_24: older libc,
  wider compatibility) with `RUSTFLAGS="-C target-feature=-crt-static -C strip=symbols"` —
  crt-static off because Node cannot `dlopen` a static build, stripped to keep package growth
  to a few MB per target (~7 MB each per the measurement in #1271).
- `ci.yml`: adds both musl targets to the debug compile matrix using the same native-Alpine
  recipe (cross-compiling from a glibc host links the cdylib through the host cc and produces a
  glibc-linked artifact). `linux-x64-musl` also joins the integration-tests matrix, running in a
  `node:{20,24}-alpine` container so a musl Node loads the musl addon; `setup-node-deps` gained
  an `install-node` input since actions/setup-node only installs glibc builds. `linux-arm-musl`
  is compile-only: the GitHub runner cannot execute JS actions in arm64 Alpine containers
  (its injected Alpine Node ships for x64 only). Bun on musl is excluded for now.
- Each musl compile row (debug and release) smoke-loads its built binary in a `node:22-alpine`
  container on the arch that built it, so an artifact that compiles but cannot `dlopen` fails
  the build — this is the only load check for arm64, which cannot run the integration suite.

One deliberate asymmetry worth deciding on: the musl artifacts are stripped while the glibc
ones keep their symbol table (`[profile.release]` sets no `strip`), so a native crash on Alpine
shows bare addresses where Debian shows function names. If uniform behavior is preferred, I can
move `strip = "symbols"` into `[profile.release]` so every target shrinks and behaves the same.

## Why?

Closes #1621 (TypeScript part of temporalio/features#594). Running on Alpine currently requires
compiling core-bridge from source with the full Rust toolchain, which is painful in Docker
builds. #1271 marked the detection half of this as a good first contribution.

## Checklist

1. How was this tested:

   - `packages/test/src/test-native-libc-detection.ts` unit-tests `getPrebuiltTargetName()` for
     glibc, musl, absent-report (the glibc fallback), and non-Linux platforms.
   - Both musl compile rows smoke-load the built artifact on Alpine, x64 and arm64.
   - The integration-tests Alpine container path needs a CI run on this PR to shake out; known
     assumptions: the Temporal CLI's static Go binary runs under musl, and the cold Docker
     builds (no Rust cache) fit the raised 30-minute compile timeout.

2. Any docs updates needed?

   Once released, the "Do not use Alpine" section at
   https://docs.temporal.io/develop/typescript/workers/run-worker-process#do-not-use-alpine
   (and the "musl-based images are not supported" note in the same page's Docker section)
   should be updated. That lives in temporalio/documentation — happy to follow up there.

### DIFF
--- a/packages/core-bridge/common.js
+++ b/packages/core-bridge/common.js
@@ -14,6 +14,8 @@ const targets = [
   'aarch64-apple-darwin',
   'x86_64-unknown-linux-gnu',
   'aarch64-unknown-linux-gnu',
+  'x86_64-unknown-linux-musl',
+  'aarch64-unknown-linux-musl',
   // TODO: this is not supported on macos
   'x86_64-pc-windows-msvc',
   'x86_64-pc-windows-gnu',
@@ -29,6 +31,19 @@ class PrebuildError extends Error {
   }
 }
 
+let muslSystem;
+
+// glibc reports its runtime version in the process report header; a Linux build without it is
+// musl-based (e.g. Alpine). Same detection as the detect-libc package. On runtimes compiled
+// without report support, assume glibc.
+function isMuslSystem() {
+  if (muslSystem === undefined) {
+    const header = process.report?.getReport?.()?.header;
+    muslSystem = header != null && header.glibcVersionRuntime == null;
+  }
+  return muslSystem;
+}
+
 function getPrebuiltTargetName() {
   const arch = archAlias[os.arch()];
   if (arch === undefined) {
@@ -37,6 +52,9 @@ function getPrebuiltTargetName() {
   const platform = platformMapping[os.platform()];
   if (platform === undefined) {
     throw new PrebuildError(`No prebuilt module for platform ${os.platform()}`);
+  }
+  if (os.platform() === 'linux' && isMuslSystem()) {
+    return `${arch}-unknown-linux-musl`;
   }
   return `${arch}-${platform}`;
 }
--- a/.github/actions/setup-node-deps/action.yml
+++ b/.github/actions/setup-node-deps/action.yml
@@ -13,6 +13,13 @@ inputs:
     description: 'Whether to save the NPM cache at the end'
     required: false
     default: 'false'
+  install-node:
+    description: >-
+      Whether to install Node via actions/setup-node. Set to false when the job container already
+      provides the required Node build (e.g. musl-based images: setup-node only installs glibc
+      builds from the official dist).
+    required: false
+    default: 'true'
 
 outputs:
   cache-dir:
@@ -32,6 +39,7 @@ runs:
         run_install: false
 
     - name: Install Node
+      if: inputs.install-node == 'true'
       uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
       with:
         node-version: ${{ inputs.node-version }}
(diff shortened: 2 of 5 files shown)


packages/core-bridge/common.js
```diff
@@ -29,6 +31,19 @@ class PrebuildError extends Error {
...
 
+let muslSystem;
+
+// glibc reports its runtime version in the process report header; a Linux build without it is
+// musl-based (e.g. Alpine). Same detection as the detect-libc package. On runtimes compiled
+// without report support, assume glibc.
+function isMuslSystem() {
+  if (muslSystem === undefined) {
+    const header = process.report?.getReport?.()?.header;
+    muslSystem = header != null && header.glibcVersionRuntime == null;
```
mjameswh: I believe checking at `glibcVersionRuntime` is not sufficient because it would not catch the case of a statically linked glibc build.

We should look at `glibcVersionCompiler` instead.

packages/core-bridge/common.js
```diff
@@ -29,6 +31,19 @@ class PrebuildError extends Error {
   }
 }
 
+let muslSystem;
+
+// glibc reports its runtime version in the process report header; a Linux build without it is
+// musl-based (e.g. Alpine). Same detection as the detect-libc package. On runtimes compiled
```
mjameswh: Please remove the statement about "same detection as detect-libc". That's not exactly true and is kind of misleading: `detect-libc` performs several other checks (e.g. inspect the ELF binary, look at the file system, etc) before finally looking at the `process.report` API.

Are all of these checks pertinent for us? I really don't think so, but I would avoid making it sounds like we're following the same detection logic, where we're really just using a subset of their logic.

mjameswh:
Thanks for this contribution!

I will need to dig a bit deeper in this PR. It touches a few sensitive aspects of the SDK (e.g. new build targets, modifies several of our CI and release workflows, modifies the bridge's `common.js` file which runs at the very start of the SDK's load, etc).

I'll try to review tomorrow.

packages/core-bridge/common.js
VictorCodesseira: You're right, I've fixed this to use the right detector.

packages/core-bridge/common.js
VictorCodesseira: I've removed both the comment and the claim in the PR body.

PR #2392, opened 2026-09-03
feat(worker): report external storage metrics on workflow task completions

### LINKED ISSUE #2066: SDK should log when WFT completion takes a long time - TypeScript
Parent Issue: https://github.com/temporalio/features/issues/827

jmaeagle99:
## What was changed

Accumulate storage driver duration metrics on workflow tasks and report them into core.

## Why?

Allow core to log storage driver durations when workflow tasks have a duration above the warning threshold.

## Checklist

1. Closes #2066

2. How was this tested: Unit tests

3. Any docs updates needed? No

### DIFF
--- a/packages/common/src/internal-non-workflow/external-storage-runner.ts
+++ b/packages/common/src/internal-non-workflow/external-storage-runner.ts
@@ -3,6 +3,8 @@
  *
  * @module
  */
+import { performance } from 'node:perf_hooks';
+
 import { temporal } from '@temporalio/proto';
 
 import { StorageDriverClaim } from '../converter/extstore';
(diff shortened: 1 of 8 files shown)


packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -154,8 +168,12 @@ export class ExternalStorageRunner {
         );
       }
       for (const [j, retrievedPayload] of retrieved.entries()) {
-        const item = group.items[j]!;
-        result[item.index] = retrievedPayload;
+        result[group.items[j]!.index] = retrievedPayload;
```
THardy98: nit: think the old way is a bit easier to read, but no strong opinion, not blocking

packages/common/src/internal-non-workflow/external-storage-runner.ts
```diff
@@ -154,8 +168,12 @@ export class ExternalStorageRunner {
...
+      }
+      // Sizing a retrieved payload re-encodes it, so only pay that cost when metrics are wanted.
+      if (metrics) {
+        const sizeBytes = retrieved.reduce((sum, p) => sum + payloadProtoSize(p), 0);
+        metrics.record(group.driver.name, group.items.length, sizeBytes, startMs, performance.now());
       }
```
THardy98: is this significant enough that users should be made aware?

THardy98 (approved):

cconstable (approved):

PR #2394, opened 2026-09-04
feat(bundler): Add SDK version tag on generated Workflow bundles

mjameswh:
## What was changed

- Add an SDK version metadata in generated Workflow bundles.
- Worker now refuses to load and execute a bundle that has been produced by a different version of the SDK.

## Why?

- Running a Workflow Worker with a bundle produced by a different version of the SDK has never been supported.
- A Workflow bundles includes source files from the SDK packages that were used to produce that bundle. Those source files use and provide internal APIs from/to SDKs packages of the Worker. Those internal APIs may change in incompatible ways at any time, including between patch-level releases of the SDK.
- Use of incompatible workflow bundles is a recurring cause of bugs in production, yet extremely difficult to diagnose.

### DIFF
--- a/packages/worker/src/workflow/bundle-metadata.ts
+++ b/packages/worker/src/workflow/bundle-metadata.ts
@@ -1,0 +1,65 @@
+const WORKFLOW_BUNDLE_ANNOTATION_PREFIX = '/* @temporalio/workflow-bundle ';
+const WORKFLOW_BUNDLE_ANNOTATION_SUFFIX = ' */';
+
+interface WorkflowBundleMetadata {
+  sdkVersion: string;
+}
+
+export function makeWorkflowBundleVersionAnnotation(sdkVersion: string): string {
+  return `${WORKFLOW_BUNDLE_ANNOTATION_PREFIX}${JSON.stringify({ sdkVersion })}${WORKFLOW_BUNDLE_ANNOTATION_SUFFIX}`;
+}
+
+export function getWorkflowBundleSdkVersion(code: string): string {
+  const firstLineEnd = code.indexOf('\n');
+  const firstLine = code.slice(0, firstLineEnd === -1 ? code.length : firstLineEnd).replace(/\r$/, '');
+
+  if (!firstLine.startsWith(WORKFLOW_BUNDLE_ANNOTATION_PREFIX)) {
+    throw new TypeError(
+      `The provided Workflow Bundle does not contain a Temporal SDK version annotation. ` +
+        `Make sure the bundle was generated with the same version of '@temporalio/worker' that is being used to ` +
+        `create the Worker.`
+    );
+  }
+
+  if (!firstLine.endsWith(WORKFLOW_BUNDLE_ANNOTATION_SUFFIX)) {
+    throw new TypeError(
+      `The provided Workflow Bundle contains an invalid Temporal SDK version annotation. ` +
+        `Rebuild the bundle with the same version of '@temporalio/worker' that is being used to create the Worker.`
+    );
+  }
+
+  const serializedMetadata = firstLine.slice(
+    WORKFLOW_BUNDLE_ANNOTATION_PREFIX.length,
+    -WORKFLOW_BUNDLE_ANNOTATION_SUFFIX.length
+  );
+  let metadata: WorkflowBundleMetadata;
+  try {
+    metadata = JSON.parse(serializedMetadata);
+  } catch (_err) {
+    throw new TypeError(
+      `The provided Workflow Bundle contains an invalid Temporal SDK version annotation. ` +
+        `Rebuild the bundle with the same version of '@temporalio/worker' that is being used to create the Worker.`
+    );
+  }
+
+  if (typeof metadata !== 'object' || metadata === null || typeof metadata.sdkVersion !== 'string') {
+    throw new TypeError(
+      `The provided Workflow Bundle contains an invalid Temporal SDK version annotation. ` +
+        `Rebuild the bundle with the same version of '@temporalio/worker' that is being used to create the Worker.`
+    );
+  }
+
+  return metadata.sdkVersion;
+}
+
+export function assertWorkflowBundleSdkVersion(code: string, expectedSdkVersion: string): void {
+  const actualSdkVersion = getWorkflowBundleSdkVersion(code);
+  if (actualSdkVersion !== expectedSdkVersion) {
+    throw new TypeError(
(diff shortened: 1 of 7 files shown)


packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -267,29 +267,6 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
     registeredActivityNames: Set<string>,
     patchActivationCallback?: WorkflowPatchActivationCallback
   ): Promise<InstanceType<T>> {
-    // The Reusable V8 Context executor needs to take control of Webpack's require
```
mjameswh, line 270: This assertion was added in #2195 as a sanity checks against eventual regressions, but it turned out to be inneffective at detecting the case where a boggus SDK+Webpack pair is used to produce a workflow bundle, which is then passed to a patched SDK worker.

The "same SDK version" assertion provides a much more robust validation going forward.

packages/worker/src/workflow/bundler.ts
```diff
@@ -240,6 +242,14 @@ exports.importInterceptors = function importInterceptors() {
         },
       },
       plugins: [
+        new BannerPlugin({
+          banner: makeWorkflowBundleVersionAnnotation(pkg.version),
+          raw: true,
+          entryOnly: true,
+          // Add the annotation after minimizers have run so it cannot be stripped, but before
+          // the inline source map is generated so its line offsets remain correct.
+          stage: Compilation.PROCESS_ASSETS_STAGE_DEV_TOOLING - 1,
```
chris-olszewski, line 251: Is this the recommended pattern for ensuring we get run right before any source maps are generated?

packages/worker/src/workflow/bundle-metadata.ts
```diff
@@ -0,0 +1,63 @@
...
+        `Make sure the bundle was generated with the same version of '@temporalio/worker' that is being used to ` +
+        `create the Worker.`
+    );
+  }
+
+  if (!firstLine.endsWith(WORKFLOW_BUNDLE_ANNOTATION_SUFFIX)) {
+    throw invalidWorkflowBundleVersionAnnotation();
+  }
+
+  const serializedMetadata = firstLine.slice(
```
chris-olszewski, line 35: Assuming serializing an object compared to just the version is for future extension. Do you have any additional metadata you think would be beneficial to track?

chris-olszewski (approved):
This is a nice improvement. Curious if you have additional ideas of metadata we should include in the bundle.

packages/worker/src/workflow/bundle-metadata.ts
mjameswh: Yeah, that's for future extension. Nothing else that I would add immediately.

One thing that could be pertinent to add in the future would be a list of plugins, sinks and payload/failure converters included at build time, as incoherencies on those between the Bundler's and Worker's config is very likely to result into late or silent errors. But every one of those pose some non-trivial challenges, either in capturing or reporting/comparing them, so that's thought for the future.

packages/worker/src/workflow/bundler.ts
mjameswh: @copilot Please validate whether this is the correct stage. First consider what our functional needs really are, don't just assume the comments are correct. Then look at Webpack stage hooks constants to determine what stage is the most appropriate for our precise needs.

packages/worker/src/workflow/reusable-vm.ts
```diff
@@ -267,29 +267,6 @@ export class ReusableVMWorkflowCreator implements WorkflowCreator {
...
-      throw new Error(
-        `The provided Workflow Bundle is not compatible with the Reusable V8 Context executor: it does ` +
-          `not route its module cache through 'globalThis.__webpack_module_cache__', which is required to ` +
-          `keep module state isolated across Workflow executions. Make sure the bundle was produced by a ` +
-          `compatible version of the Temporal SDK, or disable the Reusable V8 Context executor by setting ` +
-          `'WorkerOptions.reuseV8Context' to false.`
-      );
-    }
-
     const script = new vm.Script(workflowBundle.code, { filename: workflowBundle.filename });
```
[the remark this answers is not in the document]
mjameswh: Dismiss.

This assertion was added in #2195 as a sanity checks against eventual regressions, but it turned out to be ineffective at detecting the case where a bogus SDK+Webpack pair is used to produce a workflow bundle, which is then passed to a patched SDK worker. This check simply look for the `globalThis.__webpack_module_cache__` string in the bundle's source code, but can't assert it is on a code path that will effectively be executed. That string could notably be contained in a comment!

> only by a forgeable version comment

The possibility of forging the metadata comment is not a concern. Both of these checks (the version tag and the presence of `this.__webpack_module_cache__`) are meant to protect users against easy, unintended mishandling mistakes. They are not a security feature.

Note also that the `globalThis.__webpack_module_cache__` check is even easier to bypass than the metadata header: a user may simply add a comment in their workflow code containing `globalThis.__webpack_module_cache__`!

packages/worker/src/workflow/bundler.ts
mjameswh: @copilot Please validate whether this is the correct stage. First consider what our functional needs really are, don't just assume the comments are correct. Then look at Webpack stage hooks constants to determine what stage is the most appropriate for our precise needs.

mjameswh: Oh well, looks like copilot don't want to be summoned.

> Is this the recommended pattern for ensuring we get run right before any source maps are generated?

Mostly.

I mean, there's not exactly a webpack blessed, "recommended pattern" for this. What they have is simply a [list of stages constants](https://github.com/webpack/webpack/blob/69adee615b758639f3499d54c0b1d62f8328c6f2/lib/Compilation.js#L6664), which correspond to numbers in steps of at least 100 between each others. Those stages correspond to common webpack features. So a plugin that exactly fits the described use case of a stage should use that constant. Plugins that do not exactly fit a use case should place themselves in-between those steps where they make sense.

That's what we do here. We want to position ourselves between these two contiguous stages:

```
/**
 * Optimize the size of existing assets, e. g. by minimizing or omitting whitespace.
 */
Compilation.PROCESS_ASSETS_STAGE_OPTIMIZE_SIZE = 400;

/**
 * Add development tooling to assets, e. g. by extracting a SourceMap.
 */
Compilation.PROCESS_ASSETS_STAGE_DEV_TOOLING = 500;
```

PR #2395, opened 2026-09-04
feat(activity): add ActivityDescribeOptions for fetching payload fields

maciejdudko:
## What was changed

Add `ActivityDescribeOptions` for `ActivityHandle.describe` that enable fetching optional payload fields.

## Why?

Implements support for new API feature.

## Checklist

1. How was this tested:

New tests in `test-standalone-activities.cloud-pending.ts`

### DIFF
--- a/packages/client/src/activity-client.ts
+++ b/packages/client/src/activity-client.ts
@@ -27,6 +27,7 @@ import {
   searchAttributePayloadConverter,
 } from '@temporalio/common/lib/converter/payload-search-attributes';
 import {
+  decodeArrayFromPayloads,
   decodeFromPayloadsAtIndex,
   decodeOptionalFailureToOptionalError,
   encodeToPayloadsWithContext,
@@ -243,11 +244,17 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
         });
       },
 
-      async describe(): Promise<ActivityExecutionDescription> {
+      async describe(options?: ActivityDescribeOptions): Promise<ActivityExecutionDescription> {
         return await this.client.interceptedHandlers.describe({
           activityId: this.activityId,
           activityRunId: this.runId ?? '',
           headers: {},
+          options: {
+            includeInput: options?.includeInput || false,
+            includeOutcome: options?.includeOutcome || false,
+            includeHeartbeatDetails: options?.includeHeartbeatDetails || false,
+            includeLastFailure: options?.includeLastFailure || false,
+          },
         });
       },
 
@@ -422,15 +429,26 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
       throw new TypeError('activityId is required');
     }
 
+    function hasInfo(resp: any): resp is { info: object } {
+      return !!resp.info;
+    }
+
     try {
       const resp = await this.workflowService.describeActivityExecution({
         namespace: this.options.namespace,
         activityId: input.activityId,
         runId: input.activityRunId || undefined,
+        includeInput: input.options.includeInput,
+        includeOutcome: input.options.includeInput,
+        includeHeartbeatDetails: input.options.includeInput,
+        includeLastFailure: input.options.includeLastFailure,
       });
+      if (!hasInfo(resp)) {
+        throw new ServiceError('Missing info in describeActivityExecution response');
+      }
       const externalStorage = this.dataConverter.externalStorage;
       await visit(resp, walkDescribeActivityExecutionResponse, extstoreInboundOptions(externalStorage));
-      return buildActivityDescription(resp.info!, resp.callbacks, this.dataConverter);
+      return buildActivityDescription(resp, this.dataConverter);
     } catch (err) {
       this.rethrowGrpcError(err, 'Failed to describe activity');
     }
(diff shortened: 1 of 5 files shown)


packages/client/src/activity-client.ts
```diff
@@ -422,15 +429,26 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
+    }
+
     try {
       const resp = await this.workflowService.describeActivityExecution({
         namespace: this.options.namespace,
         activityId: input.activityId,
         runId: input.activityRunId || undefined,
+        includeInput: input.options.includeInput,
+        includeOutcome: input.options.includeInput,
+        includeHeartbeatDetails: input.options.includeInput,
```
cconstable: these should be fixed

packages/client/src/types.ts
```diff
@@ -261,6 +261,7 @@ export interface ActivityExecutionInfo {
   typedSearchAttributes: TypedSearchAttributes;
   taskQueue: string;
   executionDurationMs?: number;
+  executionTime?: Date;
```
cconstable, line 264: is this ever used or set anywhere?

packages/client/src/types.ts
```diff
@@ -288,9 +289,20 @@ export interface ActivityExecutionDescription extends ActivityExecutionInfo {
   lastDeploymentVersion?: WorkerDeploymentVersion;
   priority: Priority;
   canceledReason?: string;
+  startDelayMs?: number;
+  totalHeartbeatCount?: Long;
```
cconstable: I think this should be `number` since `Long` is an object and is less ergonomic to use. AFAICT, most things we expose get converted to numbers anyways via toNumber().

cconstable (commented):
Left a few comments

packages/client/src/activity-client.ts
```diff
@@ -662,6 +683,18 @@ export interface GetActivityHandleOptions {
...
+/**
+ * Options for {@link ActivityHandle.describe}.
+ *
+ * @experimental Standalone Activities are experimental. APIs may be subject to change.
+ */
+export interface ActivityDescribeOptions {
+  includeInput?: boolean;
+  includeOutcome?: boolean;
+  includeHeartbeatDetails?: boolean;
+  includeLastFailure?: boolean;
```
THardy98, line 709: What was the reason behind gating each field independently? (i.e. vs `includeDetails` for example)

packages/client/src/activity-client.ts
```diff
@@ -662,6 +683,18 @@ export interface GetActivityHandleOptions {
...
 }
 
+/**
+ * Options for {@link ActivityHandle.describe}.
+ *
+ * @experimental Standalone Activities are experimental. APIs may be subject to change.
+ */
+export interface ActivityDescribeOptions {
+  includeInput?: boolean;
+  includeOutcome?: boolean;
```
THardy98, line 701: I wonder if this is ambiguous given that what we're returning is:
```
getResult
getOutcomeFailure
```

I don't know if we use the term `Outcome` throughout our user-facing API, `Result` is more common (though I understand that `Outcome` is used as an outer shape for outcome `Result` and `Failure`).

It's a bit odd that we can get a result (which may have failed) and also get the outcome's failure. My thinking is that it may make more sense for the result to be either the successful result, or the failure.

packages/client/src/types.ts
```diff
@@ -288,9 +289,20 @@ export interface ActivityExecutionDescription extends ActivityExecutionInfo {
...
+  hasLastFailure: boolean;
+  hasInput: boolean;
+  hasResult: boolean;
+  hasOutcomeFailure: boolean;
 
   getHeartbeatDetails<T = any>(): Promise<T | undefined>;
   getLastFailure(): Promise<Error | undefined>;
+  getInput<T extends any[] = any[]>(): Promise<T | undefined>;
+  getResult<T = any>(): Promise<T | undefined>;
+  getOutcomeFailure(): Promise<Error | undefined>;
```
THardy98, line 347: I don't know if it's obvious to users what the difference is between `getLastFailure` and `getOutcomeFailure`.

It may be worth documenting when each are populated (and how it's possible for `getLastFailure` to have a failure but `getOutcomeFailure` may not)

packages/client/src/activity-client.ts
THardy98: Is it possible for both the outcome result _and_ the outcome failure to be populated ?

cconstable (approved):

cconstable:
lgtm pending @THardy98's feedback

packages/client/src/activity-client.ts
maciejdudko: These fields contain arbitrarily large payloads that are usually not needed. These flags are what the API currently supports.

packages/client/src/activity-client.ts
```diff
@@ -689,9 +689,21 @@ export interface GetActivityHandleOptions {
  * @experimental Standalone Activities are experimental. APIs may be subject to change.
  */
 export interface ActivityDescribeOptions {
+  /**
+   * Include activity input in the response if available.
+   */
   includeInput?: boolean;
+  /**
+   * Include activity result or outcome failure in the response if available.
```
THardy98:
```
/** Include the activity's outcome. Populates _either_ the activity result or the outcome failure in the response, never both. */
```

THardy98 (approved):
left a nit, not blocking

PR #2396, opened 2026-09-04
feat(workflow): add signal with start

tconley1428:
## Summary
- add experimental Workflow signal-with-start bindings
- add dedicated generic and generated operation-specific System Nexus interception points
- preserve JSON/protobuf outer envelopes across the isolate boundary and use protobuf binary on the server wire
- apply the serialization context generated for each System Nexus operation to its nested payloads

Depends on https://github.com/temporalio/nexgen/pull/181 for generator support. The sdk-core WIT directive is a separate upstream dependency and intentionally is not included in this PR.

## Validation
- TypeScript compile, ESLint, Prettier
- workflow internals tests
- workflow codec-runner tests


packages/worker/src/system-nexus-operations.ts
```diff
@@ -0,0 +1,192 @@
...
+  const context = metadataContext ?? workflowContext;
+  const properties = defaultPayloadConverter.fromPayload(payload) as Record<string, unknown>;
+  normalizePayloadBytes(properties);
+  const message = requestMessageType(service, operation).create(properties) as Record<string, unknown>;
+  await visit(message, walkPayloadsInMessage, {
+    initialContext: context,
+    transformPayload: async (value, valueContext) => (await encode(codecs, [value], valueContext))[0]!,
+    transformPayloads: (values, valueContext) => encode(codecs, values, valueContext),
+    skipSearchAttributes: true,
+  });
```
cconstable: Should this be skipping headers as well? if not, are they always decoded?

packages/worker/src/system-nexus-operations.ts
```diff
@@ -0,0 +1,188 @@
...
```
cconstable: Should this be skipping headers as well? I don't think anything automatically decodes them

tconley1428: Hmm, you appear to be right that we skip headers in the encoding pass, which is actually a bug that I fixed in Python a year ago.

packages/common/src/worker-deployments.ts
```diff
@@ -50,6 +50,30 @@ export const [encodeVersioningBehavior, decodeVersioningBehavior] = makeProtoEnu
...
 export type VersioningOverride = PinnedVersioningOverride | 'AUTO_UPGRADE';
 
+/** @internal */
+export function versioningOverrideToProto(
+  versioningOverride: VersioningOverride | undefined
+): temporal.api.workflow.v1.IVersioningOverride | undefined {
+  if (versioningOverride == null) {
+    return undefined;
+  }
+  // TODO: Remove deprecated field assignments when versioning is non-experimental.
```
THardy98, line 60: may as well remove this, worker versioning is GA

packages/worker/src/system-nexus-operations.ts
```diff
@@ -0,0 +1,176 @@
...
+    const payload = value as Payload;
+    if (payload.data != null) payload.data = bytesFromJson(payload.data) as Uint8Array;
+    for (const [key, item] of Object.entries(payload.metadata ?? {})) {
+      if (item != null) payload.metadata![key] = bytesFromJson(item) as Uint8Array;
+    }
+  }
+  for (const item of Array.isArray(value) ? value : Object.values(value)) normalizePayloadBytes(item, seen);
+}
+
+function bytesFromJson(value: unknown): unknown {
```
THardy98, line 164: is `unknown` the best we can do? the only callers pass in `payload.data` and `item` from `payload.metadata` entries, both of which I thought were `Uint8Array`

packages/worker/src/system-nexus-operations.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (!isSerializationContext(context)) {
+      throw new TypeError('invalid System Nexus serialization context metadata');
+    }
+    return context;
+  } catch {
+    throw new TypeError('invalid System Nexus serialization context metadata');
+  }
+}
+
+function isSerializationContext(value: unknown): value is SerializationContext {
```
THardy98, line 95: feels like this helper belongs in `serialization-context.ts`

THardy98 (commented):
small initial thoughts, still reading through the `workflow` package changes

packages/workflow/src/nexus/system/generated/models.ts
```diff
@@ -0,0 +1,481 @@
...
+  SignalArgs extends any[] = SignalValue extends common.SignalDefinition<infer Args, any> ? Args : never,
+>(
+  model: SignalWithStartWorkflowRequest<WorkflowFn, SignalValue, SignalArgs> | null | undefined
+): temporal.api.workflowservice.v1.ISignalWithStartWorkflowExecutionRequest | undefined {
+  if (model == null) {
+    return undefined;
+  }
+  return {
+    workflowType: workflowTypeToProto(requiredField(model.workflow, 'SignalWithStartWorkflowRequest', 'workflow')),
+    input: requestArgsToPayloads(model.args),
```
THardy98: We need to apply `TypeInfo` from the workflow definition here

packages/workflow/src/nexus/system/generated/models.ts
```diff
@@ -0,0 +1,481 @@
...
+    workflowId: requiredField(model.id, 'SignalWithStartWorkflowRequest', 'id'),
+    taskQueue: taskQueueToProto(requiredField(model.taskQueue, 'SignalWithStartWorkflowRequest', 'taskQueue')),
+    signalName: signalFunctionName(requiredField(model.signal, 'SignalWithStartWorkflowRequest', 'signal')),
+    signalInput: requestArgsToPayloads(model.signalArgs),
```
THardy98: Similarly, we need to apply `TypeInfo` from the signal definition here

packages/test/src/test-system-nexus-codec-runner.ts
THardy98, line 1: Could we add a case that checks the context is freed when the nexus operation fails?

THardy98 (commented):
Couple more comments, with a bit more significance

packages/worker/src/system-nexus-operations.ts
tconley1428: Probably. Good call to improve.

packages/test/src/test-system-nexus-codec-runner.ts
tconley1428: What do you mean by that? Like the memory?

packages/common/src/worker-deployments.ts
tconley1428: I didn't really understand it, but if you say so, you the boss

packages/worker/src/system-nexus-operations.ts
tconley1428: Nope, it uses it recursively on unknown types

packages/test/src/test-system-nexus-codec-runner.ts
tconley1428: Assuming you meant that it was removed from context, added a confirmation of that.

packages/workflow/src/nexus/system/generated/registry.ts
```diff
@@ -0,0 +1,18 @@
+// Generated by nexgen v0.2.5. DO NOT EDIT!
+
+import { signalWithStartWorkflowSerializationContext } from './support';
+
+interface OperationRegistryEntry<Input = unknown> {
+  readonly service: string;
+  readonly operation: string;
+  /** Context for nested payloads, determined by the operation. */
+  readonly serializationContext?: (input: Input) => import('@temporalio/common').SerializationContext;
```
tconley1428: Going to fix this import.

packages/worker/src/system-nexus-operations.ts
THardy98: What values could `value` be in `normalizePayloadBytes`? It broadly looks like it expects `Payload` or an array of payloads, which should allow us to narrow, but I can't really tell.

packages/worker/src/system-nexus-operations.ts
THardy98: did we not want to move this ?

packages/test/src/test-system-nexus-codec-runner.ts
THardy98: > What do you mean by that? Like the memory?

Yeah, do we cleanup: `nexus op seq -> serialization context` correctly even if the nexus op isn't on the happy path. Or do we leak

packages/worker/src/system-nexus-operations.ts
tconley1428: Huh, I thought I did. Must not have.

packages/worker/src/system-nexus-operations.ts
tconley1428: `normalizePayloadBytes` takes a bytes, then decodes it into an object, but we don't really know what the type of that is, so it is `Record<string, unknown>`. Each of the fields in that may be some representation of `UInt8Array` that needs to be turned back into that type so protobufjs works creating the real message type. If there's a better way to preserve `UInt8Array` using the default json serializer as it exits the isolate, by all means let me know.

packages/workflow/src/nexus/system/generated/models.ts
```diff
@@ -0,0 +1,485 @@
...
+          : {
+              /**
+               * Arguments for signal.
+               */
+              signalArgs?: SignalArgs | Readonly<SignalArgs>;
+            }))
+    )
+>;
+
+export function signalWithStartWorkflowRequestFromProto<
```
chris-olszewski: There are a ton of casts in this function and I'm not sure any of them are necessary.

tconley1428: Removed most of them. Two were needed for function arguments.

THardy98 (approved):

packages/common/src/converter/serialization-context.ts
```diff
@@ -56,3 +56,19 @@ export interface ActivitySerializationContext {
  * @experimental Serialization context is an experimental feature and may change.
  */
 export type SerializationContext = WorkflowSerializationContext | ActivitySerializationContext;
+
+/** @internal */
+// Consumed by the worker through common/lib/converter/serialization-context.
+// ts-prune-ignore-next
+export function isSerializationContext(value: unknown): value is SerializationContext {
```
chris-olszewski, line 63: Not something to be addressed in this PR, but I wonder if we should adopt an actual validation library. This sucks...

packages/common/src/index.ts
```diff
@@ -15,8 +15,8 @@ export * from './converter/extstore';
 export * from './converter/failure-converter';
 export * from './converter/payload-codec';
 export * from './converter/payload-converter';
-export type {
-  ActivitySerializationContext,
+export {
```
chris-olszewski: Not sure I understand this change, `SerializationContext` and `WorkflowSerializationContext` are both just interfaces. There's no code to share other than the type.

packages/workflow/src/internals.ts
```diff
@@ -178,6 +180,8 @@ interface ScopedWorkflowRandomSource {
  * - Call user-defined functions, including any form of interceptor.
  */
 export class Activator implements ActivationHandler {
+  /** Converter temporarily scoped while generated System Nexus models are constructed. */
```
chris-olszewski: I don't understand this comment placement

tconley1428: Looks like it was left behind in a change, deleting it.

packages/common/src/index.ts
tconley1428: Good call.

chris-olszewski (approved):

PR #2400, opened 2026-09-05
feat(activity): Standalone Activities operator commands

### LINKED ISSUE #2061: [Feature Request] Implement operator commands for Standalone Activities

See https://github.com/temporalio/features/issues/822

maciejdudko:
## What was changed

Added operator commands `pause`, `unpause`, `updateOptions` and `restoreOriginalOptions` to `ActivityHandle`.

## Why?

Feature request: https://github.com/temporalio/features/issues/822

## Checklist

1. Closes #2061 

2. How was this tested:
New tests in `test-standalone-activities.cloud-pending.ts`

### DIFF
--- a/packages/client/src/activity-client.ts
+++ b/packages/client/src/activity-client.ts
@@ -19,7 +19,7 @@ import {
   decompileRetryPolicy,
 } from '@temporalio/common';
 import type { Duration } from '@temporalio/common/lib/time';
-import { msOptionalToTs, msToNumber, optionalTsToDate, optionalTsToMs } from '@temporalio/common/lib/time';
+import { msOptionalToTs, msToNumber, optionalTsToDate, optionalTsToNonZeroMs } from '@temporalio/common/lib/time';
 import { composeInterceptors } from '@temporalio/common/lib/interceptors';
 import {
   decodeTypedSearchAttributes,
@@ -48,8 +48,12 @@ import type {
   ActivityDescribeInput,
   ActivityGetResultInput,
   ActivityListInput,
+  ActivityPauseInput,
+  ActivityRestoreOriginalOptionsInput,
   ActivityStartInput,
   ActivityTerminateInput,
+  ActivityUnpauseInput,
+  ActivityUpdateOptionsInput,
 } from './interceptors';
 import type { AsyncCompletionClientOptions } from './async-completion-client';
 import { AsyncCompletionClient } from './async-completion-client';
@@ -110,6 +114,14 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
       terminate: composeInterceptors(interceptors, 'terminate', this.terminateHandler.bind(this)),
       list: composeInterceptors(interceptors, 'list', this.listHandler.bind(this)),
       count: composeInterceptors(interceptors, 'count', this.countHandler.bind(this)),
+      pause: composeInterceptors(interceptors, 'pause', this.pauseHandler.bind(this)),
+      unpause: composeInterceptors(interceptors, 'unpause', this.unpauseHandler.bind(this)),
+      updateOptions: composeInterceptors(interceptors, 'updateOptions', this.updateOptionsHandler.bind(this)),
+      restoreOriginalOptions: composeInterceptors(
+        interceptors,
+        'restoreOriginalOptions',
+        this.restoreOriginalOptionsHandler.bind(this)
+      ),
     };
   }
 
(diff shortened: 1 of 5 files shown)


packages/client/src/activity-client.ts
```diff
@@ -564,6 +729,26 @@ export interface ActivityHandle<R = any> {
...
+  /**
+   * Unpauses the Activity execution if it was previously paused.
+   */
+  unpause(options?: ActivityUnpauseOptions): Promise<void>;
+  /**
+   * Updates activity options of a running activity execution. See documentation for {@link ActivityOptionsUpdate}.
+   *
+   * Returns current options after applying the update.
+   */
+  updateOptions(options?: ActivityOptionsUpdate): Promise<ActivityOptionsUpdate>;
```
cconstable: this says the options are optional but line 302 requires them

packages/test/src/test-standalone-activities.cloud-pending.ts
```diff
@@ -863,4 +864,104 @@ if (RUN_INTEGRATION_TESTS) {
...
+      tt.not(desc.retryPolicy.initialInterval, originalDuration);
+      tt.is(desc.retryPolicy.maximumInterval, updatedDuration);
+      tt.is(desc.priority.fairnessKey, undefined);
+      // TODO: uncomment when added
+      // tt.is(desc.startDelay, undefined);
+    });
+
+    const originalOptions = await handle.restoreOriginalOptions();
+    t.is(originalOptions.taskQueue, 'original-task-queue');
+    t.is(updatedOptions.scheduleToCloseTimeout, undefined);
```
cconstable: should `updatedOptions` be `originalOptions` here?

packages/client/src/activity-client.ts
```diff
@@ -517,6 +564,124 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
+        namespace: this.options.namespace,
+        activityId: input.activityId,
+        runId: input.activityRunId || undefined,
+        identity: this.options.identity,
+        requestId: randomUUID(),
+        restoreOriginal: true,
+      });
+      return activityOptionsUpdateFromProto(resp.activityOptions);
+    } catch (err) {
+      this.rethrowGrpcError(err, 'Failed to request activity cancellation');
```
cconstable: I don't think this is the correct error message... maybe something like "failed to restore activity options?"

packages/client/src/activity-client.ts
```diff
@@ -517,6 +564,124 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
+          heartbeatTimeout: msOptionalToTs(input.options.heartbeatTimeout),
+          retryPolicy: input.options.retry ? compileRetryPolicy(input.options.retry) : undefined,
+          priority: input.options.priority ? compilePriority(input.options.priority) : undefined,
+          startDelay: msOptionalToTs(input.options.startDelay),
+        },
+        updateMask: { paths },
+      });
+      return activityOptionsUpdateFromProto(resp.activityOptions);
+    } catch (err) {
+      this.rethrowGrpcError(err, 'Failed to request activity cancellation');
```
cconstable: I don't think this is the correct error message... maybe something like "failed to update activity options?"

packages/client/src/activity-client.ts
```diff
@@ -662,6 +847,65 @@ export interface GetActivityHandleOptions {
...
+ *
+ * If a field is assigned non-null value, the option will be set to that value.
+ * If a field is explicitly assigned null, the option will be cleared.
+ * If a field is undefined, the option will be left unchanged.
+ *
+ * In a return value, currently unset fields are undefined.
+ *
+ * @experimental Standalone Activities are experimental. APIs may be subject to change.
+ */
+export interface ActivityOptionsUpdate {
```
cconstable, line 902: This is also the return type for the calls that update the options which makes things a little unclear to me. Does an undefined in the return type mean the option was left unchanged or that it is actually undefined? I think the doc comment might be misleading for this when it is being used as a return type. How do you feel about separating the input and return type into two different types?

cconstable (commented):
left some comments

packages/common/src/time.ts
```diff
@@ -114,3 +114,12 @@ export function optionalDateToTs(date: Date | null | undefined): Timestamp | und
   }
   return msToTs(date.getTime());
 }
+
+/**
+ * Lossy conversion function from Timestamp to number due to possible overflow.
+ * If ts is zero, null or undefined returns undefined.
+ */
+export function optionalTsToNonZeroMs(ts?: Timestamp | null): number | undefined {
+  const ms = optionalTsToMs(ts);
```
THardy98: Hmm it's possible for `optionalTsToMs` to legitimately return 0. This will change existing callers to return `undefined` now.

packages/client/src/activity-client.ts
```diff
@@ -517,6 +564,124 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
+  }
+
+  protected async updateOptionsHandler(input: ActivityUpdateOptionsInput): Promise<ActivityOptionsUpdateResult> {
+    if (!input.activityId) {
+      throw new TypeError('activityId is required');
+    }
+
+    const paths = [];
+    // != null because taskQueue can't be unset
+    if (input.options.taskQueue != null) {
```
THardy98, line 614: Could we construct each proto value in the same conditional that adds its mask path? Keeping the path and value adjacent would make their relationship clearer and help avoid drift.

i.e.
```
if (input.options.taskQueue != null) {
    paths.push('task_queue.name');
    activityOptions.taskQueue = { name: input.options.taskQueue };
  }
```

packages/client/src/activity-client.ts
```diff
@@ -608,31 +608,54 @@ export class ActivityClient extends AsyncCompletionClient implements TypedActivi
...
     // != null because taskQueue can't be unset
     if (input.options.taskQueue != null) {
       paths.push('task_queue.name');
+      activityOptions.taskQueue = { name: input.options.taskQueue };
     }
     if (input.options.scheduleToCloseTimeout !== undefined) {
       paths.push('schedule_to_close_timeout');
+      if (input.options.scheduleToCloseTimeout != null) {
+        activityOptions.scheduleToCloseTimeout = msOptionalToTs(input.options.scheduleToCloseTimeout);
+      }
```
THardy98, line 622: This is duplicative, why not just check for `!= null` one time ? Can you even specify `null` for these fields?

Ditto for the other added conditionals

maciejdudko: Yes, you can specify null, this is how you unset the field (as opposed to leaving it unchanged). Explained in `ActivityOptionsUpdate` docs.

THardy98: ah, duh. Thanks 👍

THardy98 (approved):
LGTM, subject to @cconstable comments

cconstable (approved):

PR #2413, opened 2026-09-10
fix(activity): Use serialization context in Standalone Activities

maciejdudko:
## What was changed

Change `ActivityClient` to use `ActivitySerializationContext` when de/serializing Standalone Activity payloads.
## Why?

Allows using context-aware data converters with Standalone Activities.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@ to docs, or any other relevant information.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


THardy98:
Could we add integration tests that check that serialization context is propagated correctly for SAA on both happy/failure paths?

THardy98 (approved):
Would like to see one on the failure path as well if possible

PR #2414, opened 2026-09-11
fix(ai-sdk): polyfill base64 globals in workflows

### LINKED ISSUE #2408: [Bug] @temporalio/ai-sdk missing atob/btoa polyfills

I want to use the vercel AI SDK to have image inputs from tool call results.

I'm building a coding agent on `@temporalio/ai-sdk`. One of the agent's tools is
a `read` tool that returns images (screenshots, PNG files) as an image content
part so the model can look at them. Text results work fine, but Image results break.

### Describe the bug

When a tool result contains a file/image part, the Workflow Task fails with
`TypeError: atob is not a function`, and because it is a plain `TypeError`
rather than a Temporal failure, the task is retried forever. The workflow shows
as Running with an ever-growing history of `WorkflowTaskFailed` events.

The call comes from the AI SDK's prompt conversion, which runs inside the
workflow sandbox. When it converts a tool result `file` part, it sniffs the
media type by decoding the base64 data with `atob`. The workflow isolate has no
`atob` or `btoa`.

`contrib/ai-sdk/src/load-polyfills.ts` already installs the Web-API globals the
AI SDK needs at module load (`Headers`, Web Streams, `structuredClone`), but not
`atob`/`btoa`, which are only reached on this codepath.

Stack trace from the `WorkflowTaskFailed` event:

```
TypeError: atob is not a function
    at convertBase64ToUint8Array (node_modules/@ai-sdk/provider-utils/src/uint8-utils.ts:8:23)
    at decodePrefix (node_modules/@ai-sdk/provider-utils/src/detect-media-type.ts:206:16)
    at detectMediaTypeBySignatures (node_modules/@ai-sdk/provider-utils/src/detect-media-type.ts:242:14)
    at detectMediaType (node_modules/@ai-sdk/provider-utils/src/detect-media-type.ts:310:9)
    at convertPartToLanguageModelPart (node_modules/ai/src/prompt/convert-to-language-model-prompt.ts:586:43)
    at node_modules/ai/src/prompt/convert-to-language-model-prompt.ts:633:32
    at Array.map (<anonymous>)
    at mapToolResultOutput (node_modules/ai/src/prompt/convert-to-language-model-prompt.ts:630:24)
    at node_modules/ai/src/prompt/convert-to-language-model-prompt.ts:384:26
    at Array.map (<anonymous>)
```

Expected: the image part is forwarded to the model, as it is when the same
`generateText` call runs outside Temporal.

Suggested fix: add `atob`/`btoa` to `installPolyfills()` in
`contrib/ai-sdk/src/load-polyfills.ts`, guarded like the existing entries.
Both are pure functions of their input, so they are safe in workflow code.

### Minimal Reproduction

```ts
// workflows.ts
import { temporalProvider } from "@temporalio/ai-sdk/workflow";
import { generateText, stepCountIs, tool } from "ai";
import { z } from "zod";

// 1x1 transparent PNG
const PNG =
  "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAAAABJRU5ErkJggg==";

export async function repro(): Promise<string> {
  const result = await generateText({
    model: temporalProvider.languageModel("<any model id>"),
    prompt: "Call the screenshot tool once, then describe what you see.",
    tools: {
      screenshot: tool({
        description: "Takes a screenshot",
        inputSchema: z.object({}),
        execute: async () => ({ data: PNG, mimeType: "image/png" }),
        toModelOutput: ({ output }) => ({
          type: "content",
          value: [{ type: "file", data: { type: "data", data: output.data }, mediaType: output.mimeType }],
        }),
      }),
    },
    stopWhen: stepCountIs(3),
  });
  return result.text;
}
```

Starting from the `hello-world` TypeScript sample with `@temporalio/ai-sdk`.

### Environment/Versions

- OS and processor: Linux x86_64
- Temporal Server: `temporalio/auto-setup:1.28` (also see the CLI dev server behave the same way)
- SDK: `@temporalio/worker` / `@temporalio/workflow` / `@temporalio/ai-sdk` 1.23.0
- AI SDK: `ai` 7.0.87, `@ai-sdk/provider-utils` 5.0.34
- Node.js v26.7.0, Worker running on Node (not Bun)
- Server running in Kubernetes (k3d), not built from source

brianstrauch:
## What was changed

- Added deterministic, Workflow-safe `atob` and `btoa` polyfills to the AI SDK preload.
- Preserved native globals when they are already available.
- Added an end-to-end regression test for image content returned by an AI tool, including a base64 round trip in the Workflow sandbox.

## Why?

The Vercel AI SDK decodes base64 data while detecting the media type of file tool results. Temporal's Workflow isolate did not provide `atob` or `btoa`, causing image tool results to repeatedly fail their Workflow Task.

## Checklist

1. Closes #2408
2. How was this tested:
   - `pnpm run build`
   - `pnpm -C contrib/ai-sdk exec ava lib/__tests__/test-ai-sdk.js --match='Image tool results are converted inside the workflow sandbox'`
3. Any docs updates needed? No.

### DIFF
--- a/contrib/ai-sdk/src/load-polyfills.ts
+++ b/contrib/ai-sdk/src/load-polyfills.ts
@@ -1,40 +1,11 @@
 import { Headers } from 'headers-polyfill';
 import structuredClonePolyfill from '@ungap/structured-clone';
 import * as webStreamsPolyfill from 'web-streams-polyfill';
-import { decodeBase64, encodeBase64 } from '@temporalio/workflow-streams/workflow';
-
-function atobPolyfill(input: string): string {
-  const encoded = String(input).replace(/[\t\n\f\r ]/g, '');
-  const paddingLength = encoded.match(/=+$/)?.[0].length ?? 0;
-  if (paddingLength > 0 && encoded.length % 4 !== 0) {
-    throw new TypeError('The string to be decoded is not correctly encoded.');
-  }
-
-  const bytes = decodeBase64(encoded);
-  let output = '';
-  for (const byte of bytes) {
-    output += String.fromCharCode(byte);
-  }
-  return output;
-}
-
-function btoaPolyfill(input: string): string {
-  const data = String(input);
-  const bytes = new Uint8Array(data.length);
-  for (let index = 0; index < data.length; index++) {
-    const byte = data.charCodeAt(index);
-    if (byte > 0xff) {
-      throw new TypeError('The string to be encoded contains characters outside of the Latin1 range.');
-    }
-    bytes[index] = byte;
-  }
-  return encodeBase64(bytes);
-}
 
 /**
  * Installs the Web-API globals that the AI SDK needs inside the workflow sandbox: `Headers`,
- * the Web Streams classes (`ReadableStream`, `WritableStream`, `TransformStream`, ...),
- * `structuredClone`, and the base64 helpers `atob` and `btoa`.
+ * the Web Streams classes (`ReadableStream`, `WritableStream`, `TransformStream`, ...), and
+ * `structuredClone`.
  *
  * Idempotent, and never overwrites globals that already exist — calling it outside the sandbox
  * (where Node.js provides all of these natively) is a no-op.
@@ -55,13 +26,5 @@ export function installPolyfills(): void {
 
   if (!('structuredClone' in globalThis)) {
     globalThis.structuredClone = structuredClonePolyfill as typeof globalThis.structuredClone;
-  }
-
-  if (typeof globalThis.atob === 'undefined') {
-    globalThis.atob = atobPolyfill;
-  }
-
-  if (typeof globalThis.btoa === 'undefined') {
-    globalThis.btoa = btoaPolyfill;
   }
 }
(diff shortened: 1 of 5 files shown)


contrib/ai-sdk/src/load-polyfills.ts
```diff
@@ -1,11 +1,40 @@
 import { Headers } from 'headers-polyfill';
```
patbeqo, line 1: Could `atob`/`btoa` be injected from `node:buffer` via `injectGlobals` in `vm-shared.ts` ?

brianstrauch: Gave it a try and it works great! Of course, now we'll need SDK team signoff.

patbeqo (approved):

packages/worker/src/workflow/vm-shared.ts
```diff
@@ -2,6 +2,7 @@ import v8 from 'node:v8';
 import vm from 'node:vm';
 import { AsyncLocalStorage as AsyncLocalStorageOriginal } from 'node:async_hooks';
 import assert from 'node:assert';
+import { atob, btoa } from 'node:buffer';
```
chris-olszewski, line 5: Note this only exposes `atob`/`btoa` as globals in the isolate so importing from `buffer`/`node:buffer` won't work. I think this is desired since we need to avoid `Buffer` from being used in the isolate IIRC.

Just calling this out in case AI SDK expects these to be available via a standard `node:buffer` import.

chris-olszewski (approved):
I do think we might need to come up with a way to allow users/plugins to configure globals that should be forwarded. I don't want to necessarily block this while we figure out what that would look like so approving this.

PR #2416, opened 2026-09-11
feat(google-adk-agents): support @google/adk 2.0

DABH:
## What was changed

Moves `@temporalio/google-adk-agents` to `@google/adk` `>=2.0.0 <2.1.0` (was `>=1.5.0 <1.6.0`) and makes the Workflow bundle load ADK 2.0:

- Pin the bare `@google/adk` specifier to ADK's web build (`dist/web/index_web.js`) with an exact `resolve.alias`. Which of ADK's three builds webpack picks otherwise depends on the consumer's browserslist config, and the node barrel reaches ADK's uninstalled optional peers. The web surface omits what a Workflow never runs: a2a, DB sessions, GCS artifacts, telemetry setup, `LocalEnvironment`, the agent registry, MCP.
- Redirect `crypto` / `node:crypto` requests issued from inside `@google/adk` to `src/crypto-shim.ts`. ADK 1.6 dropped the `Math.random()` UUID fallback, so its id generator now falls through to `node:crypto`, which the bundler aliases to an empty module, and the first `createEvent()` in a Workflow threw. The shim draws from a named workflow random stream (`getRandomStream`), so ADK's ids replay identically and do not consume the Workflow's own `Math.random()` sequence.
- Replace ADK's `dist/web/models/apigee_llm.js`, which does not parse as published (esbuild lowered an async generator and left `super.generateContentAsync` outside a method), with an inert class carrying the same registry pattern and `BaseLlm` brand. Both redirects are gated on the issuer, so only `@google/adk`'s own requests are rewritten.
- Look up `MCPToolset` lazily on the non-Workflow path (it is off the web surface but still declared by ADK's single typings file), read HTTP status and headers through the error's `cause` chain (ADK 2.0 wraps MCP session failures in one), add `adm-zip` to the ignore list, refresh the shim comments.
- `@modelcontextprotocol/sdk` becomes an optional peer, as it is in ADK itself.

## Why

ADK 2.0 reorganised its published builds and dropped the id fallback the sandbox relied on, so the 1.5 recipe no longer produces a loadable Workflow bundle. This PR is the bump plus the bundler work to make it load; support for what 2.0 actually added comes in the stacked PRs.

A Workflow started under the 1.5 plugin cannot be replayed by a Worker on this version: ADK's internals and its id generation both changed. The README and CHANGELOG say to drain in-flight Workflows first, or to run the two versions on separate task queues.

2.1.0 was published on 2026-09-15 and is deliberately outside the closed ceiling; the repo's 14-day `minimumReleaseAge` also rejects it until about 2026-09-29. It changes the web build, `transfer_to_agent`'s wire format and the confirmation binding, so a 2.1 pass is separate work.

## Testing

Package suite green in both sandbox modes, 86 tests with `REUSE_V8_CONTEXT` on and off. New: a `crypto-shim` E2E test that sends ADK ids through Activities across two Workflow Tasks and then replays the history with `Worker.runReplayHistory`; unit tests for the `@google/adk` web pin and for both ADK-scoped resolve redirects, including that a non-ADK issuer is left alone; and three for the `cause`-chain status and header lookup. `eslint`, `prettier --check` and `ts-prune` clean.

Stack: **1/4 bump (this, #2416)** → [2/4 graph runtime](https://github.com/temporalio/sdk-typescript/pull/2417) → [3/4 HITL](https://github.com/temporalio/sdk-typescript/pull/2418) → [4/4 MCP resources + model routing](https://github.com/temporalio/sdk-typescript/pull/2419)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ to docs, or any other relevant information.
 
 ### Breaking Changes
 
+- `@temporalio/google-adk-agents` now requires `@google/adk` `>=2.0.0 <2.1.0` (was `>=1.5.0 <1.6.0`).
+  ADK 2.0 changed its internals and its id generation, so a Workflow started under the 1.5 plugin
+  cannot be replayed by a Worker on this version: drain in-flight Workflows before upgrading, or
+  run the versions on separate task queues. `@modelcontextprotocol/sdk` is now an optional peer
+  dependency, needed only for `TemporalMCPToolset`.
 - `@temporalio/openai-agents` now requires `@openai/agents-core` and `@openai/agents-openai` `~0.14.3`. A project
   pinned to `0.13.x` or earlier hits a peer dependency conflict until it upgrades.
 - **Experimental**: External storage `StorageDriverSelector` now receives a
@@ -62,6 +67,9 @@ to docs, or any other relevant information.
 
 ### Added
 
+- `@temporalio/google-adk-agents` runs on Google ADK 2.0: the Workflow bundle now uses ADK's web
+  build, pinned regardless of the consumer's webpack target, and ADK's UUIDs are generated from a
+  named workflow random stream inside the sandbox, so its ids are replay-stable.
 - **Experimental**: `@temporalio/openai-agents` can run OpenAI Agents `SandboxAgent`s as Temporal Workflows. SandboxAgent
   operations are Activities; hosted tool credentials and sandbox environment values that reference allowlisted Worker
   environment variables are resolved on Worker so their values are not recorded in Workflow history.
--- a/contrib/google-adk-agents/README.md
+++ b/contrib/google-adk-agents/README.md
@@ -19,13 +19,22 @@ wrap an existing Temporal Activity with `activityAsTool`.
 npm install @temporalio/google-adk-agents @google/adk @google/genai
 ```
 
-The supported peer range is `@google/adk` `>=1.5.0 <1.6.0` and `@google/genai`
+The supported peer range is `@google/adk` `>=2.0.0 <2.1.0` and `@google/genai`
 `^2.9.0`. The ceiling is exact by design: the plugin's Workflow-sandbox shims are
 keyed to what ADK reaches at module load, so an ADK minor outside this range can
-break the Workflow bundle.
+break the Workflow bundle. `@modelcontextprotocol/sdk` is an optional peer: install
+it if you use `TemporalMCPToolset` (ADK 2.0 made it optional too).
 
 Provide Gemini credentials to the Worker as usual, for example with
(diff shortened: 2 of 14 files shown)


DABH (commented):
Ultrareview completed against the exact head. I found no regression in the ADK 2.0-specific changes: the web-bundle pinning/shims, deterministic UUID stream, Apigee replacement, optional MCP loading, cause-chain parsing, packaging, and migration guidance all held up.

One confirmed pre-existing cleanup issue remains:

**[P2] Preserve Activity cancellation before normalizing ADK failures.** All model/MCP Activity catches pass every error through `toApplicationFailure`, so an `AbortError` or `CancelledFailure` becomes a retryable `ApplicationFailure` before the worker can classify it as cancellation. I reproduced this at this head by running `adk-invokeModel` in the SDK Activity wrapper with a model that rejects with `AbortError` when its supplied signal aborts: after `Activity.cancel(...)`, the result was a failed, retryable `GoogleAdkModelError` with no canceled-failure info. This can turn cancellation into retries/additional external calls.

Please preserve cancellation before failure classification across the model and MCP catches (for example, rethrow the cancellation signal's reason when it is aborted), and cover a fake that rejects on abort. The current `SlowLlm` resolves and then yields on abort, so it masks this behavior. This predates the PR rather than being an ADK 2.0 regression, but this upgrade is a good point to close it.

There were no existing review threads or prior bot/human comments on this PR to adjudicate. Exact-head CI is green; the relevant local TypeScript checks and diff check also passed.

DABH:
Fixed in 4219b324. All four catches did run everything through `toApplicationFailure`, and the Worker reports an Activity as cancelled only when the error leaving it is a `CancelledFailure` or an `AbortError`, so a client that rejected on its abort signal had the cancel rewritten into a retryable `GoogleAdkModelError` and the attempt chain kept going. There is now a `rethrowIfCancelled(err, signal)` called first in the model, streaming, `listTools` and `callTool` catches; it raises the signal's `CancelledFailure` reason, so an Activity pause or reset still reports as such, and it uses the Worker's own `isAbortError` predicate so the two cannot disagree about what an abort is. New `cancellation.test.ts` covers it with an E2E that cancels a Workflow whose model rejects on abort and asserts the Activity ended cancelled on one attempt with no `ActivityTaskFailed`, `MockActivityEnvironment` cases for the model and MCP `callTool` catches, and a negative case showing an `AbortError` raised without a cancel is still a model failure. You were right that `SlowLlm` hid this: it resolves on abort and then yields, so the new `AbortingLlm` double rejects instead. The E2E needs `WAIT_CANCELLATION_COMPLETED` plus a `heartbeatTimeout` to see the outcome at all, since under `TRY_CANCEL` the Workflow closes before the Activity reports and without a heartbeat the Activity is never told about the cancel.

PR #2417, opened 2026-09-11
feat(google-adk-agents): run ADK 2.0 workflow graphs with Activity nodes

DABH:
## What was changed

ADK 2.0's workflow runtime (`Workflow`, `node()`, `JoinNode`, routing, dynamic nodes through `ctx.runNode`, node-as-tool, node retries and timeouts) is plain async code driven by session events, so it already runs inside a Temporal Workflow. This PR adds the piece it lacks and the failure semantics it needs.

- `activityNode(options)` in `src/nodes.ts` makes a registered Activity a graph node: a `FunctionNode` whose body is a `proxyActivities` call. The node's input is the Activity's argument unless `args` maps it, and the Activity's result is the node's output. A completed node is fast-forwarded on resume, as ADK does for every node, so a resumed run never schedules the Activity again. The node runs its own `timeout`: on the deadline it cancels the Activity's `CancellationScope`, waits for the cancellation to settle on the Activity's `cancellationType` terms, then throws ADK's `NodeTimeoutError`, so a retried attempt never overlaps a still-running one. An abort from the invocation or a failed sibling cancels the Activity the same way. Node names containing a dot are refused at construction, since ADK uses the dot as its node path separator.
- ADK's runtime errors (`NodeTimeoutError`, `NodeReportedError`, `NodeSchemaValidationError`, `IntentMismatchError`, `StateSchemaError`, `InvocationAbortedError`, `DynamicNodeFailError`) are plain `Error`s, and the SDK fails the Workflow _Task_ on an unexpected error, which retries forever. The absorbed-failure interceptor converts them into non-retryable `ApplicationFailure`s keyed by `ADK_RUNTIME_FAILURE_TYPES`, with the ADK error as the `cause`.
- Two of those carry something better than the wrapper, and that is what gets raised: a `NodeReportedError` reporting an absorbed `TemporalModel` call raises the recorded `ActivityFailure`, so the model failure's HTTP status survives; a `DynamicNodeFailError` over a Temporal failure raises that failure, so a cancelled Activity still ends the execution CANCELLED rather than FAILED.

## Testing

New `graph.test.ts` and `dynamic.test.ts` (25 cases): sequential Activity nodes, routing, fan-out and join, an Activity returning nothing, an `LlmAgent` task-mode node, a node timeout cancelling the Activity and waiting for the cancel before failing or retrying, ADK retry with jitter replayed from history, a model failure that a node retry recovers from, Activity failure propagation for both static and dynamic nodes, a cancelled Workflow, an absorbed model failure surfaced and recovered, a dotted node name refused, node callbacks, an `App` root, context compaction, and workflow-as-tool with Zod and genai schemas. Also graph spans in `telemetry.test.ts`, and a checked-in graph history replayed in `replay.test.ts` (`UPDATE_ADK_HISTORIES=1` re-records it). The package suite is 117 tests, green with `REUSE_V8_CONTEXT` both true and false.

Stack: [1/4 bump](https://github.com/temporalio/sdk-typescript/pull/2416) → **2/4 graph runtime (this, #2417)** → [3/4 HITL](https://github.com/temporalio/sdk-typescript/pull/2418) → [4/4 MCP resources + model routing](https://github.com/temporalio/sdk-typescript/pull/2419)

### DIFF
--- a/contrib/google-adk-agents/src/absorbed-failure.ts
+++ b/contrib/google-adk-agents/src/absorbed-failure.ts
@@ -9,10 +9,19 @@
  * is open surfaces there, and one recorded after a handler frame returned surfaces in the
  * main function's. One recorded after the main function returned never surfaces —
  * completing with unfinished handlers is a user error Temporal already warns about.
+ *
+ * The same interceptor also converts the errors ADK 2.0's workflow runtime *throws* —
+ * a node past its `timeout`, an output that failed its schema, an approval ADK refused
+ * to bind — from the plain `Error`s they are into non-retryable `ApplicationFailure`s.
+ * The Temporal SDK fails a Workflow only for a `TemporalFailure` (or a configured
+ * failure type); any other error fails the Workflow *Task*, which the server retries
+ * forever. These are expected runtime outcomes of a graph, not bugs, so they must end
+ * the execution (or reject the Update) with a typed failure instead.
  */
 
 import type { AsyncLocalStorage as ALS } from 'node:async_hooks';
 
+import { ApplicationFailure } from '@temporalio/common';
 import {
   AsyncLocalStorage,
   CancellationScope,
@@ -21,6 +30,8 @@ import {
   isCancellation,
   type WorkflowInterceptorsFactory,
 } from '@temporalio/workflow';
+
+import { ADK_RUNTIME_FAILURE_TYPES } from './error-types';
 
 // Held on the per-execution sandbox `globalThis` rather than in module scope: a
 // bundle can hold two copies of this module (the interceptor list registers its
@@ -112,6 +123,29 @@ function raiseAbsorbed(frame: Frame): void {
   if (frame.pending.length > 0) throw frame.pending[0];
 }
 
+/**
+ * The name of ADK's `NodeReportedError`: a graph node that ended without output because
+ * the agent it ran absorbed an error. When that error was a `TemporalModel` call the frame
+ * has it recorded, and the recording — an `ActivityFailure` carrying the model failure's
+ * status and cause chain — is the better thing to raise.
+ */
+const NODE_REPORTED_ERROR = 'NodeReportedError';
+
+/**
+ * Converts an ADK workflow-runtime error escaping a frame into the failure that should
+ * end the execution: the frame's recorded model failure when the error merely reports
+ * that a model call was absorbed, otherwise a non-retryable `ApplicationFailure` typed
+ * per {@link ADK_RUNTIME_FAILURE_TYPES} with the ADK error as its cause. Anything else —
+ * a `TemporalFailure`, a user's own error — is returned unchanged.
+ */
+function toWorkflowFailure(err: unknown, frame: Frame): unknown {
+  if (!(err instanceof Error)) return err;
+  const type = ADK_RUNTIME_FAILURE_TYPES[err.name];
+  if (type === undefined) return err;
+  if (err.name === NODE_REPORTED_ERROR && frame.pending.length > 0) return frame.pending[0];
+  return ApplicationFailure.create({ message: err.message, type, nonRetryable: true, cause: err });
+}
+
 async function surfaceAbsorbedFailure<T>(frame: Frame, next: () => Promise<T>): Promise<T> {
   let result: T;
   try {
@@ -125,7 +159,7 @@ async function surfaceAbsorbedFailure<T>(frame: Frame, next: () => Promise<T>): 
       const { main } = recorded();
       if (main !== undefined && main !== frame) raiseAbsorbed(main);
     }
-    throw err;
+    throw toWorkflowFailure(err, frame);
   } finally {
     frame.surfaced = true;
   }
--- a/contrib/google-adk-agents/src/nodes.ts
+++ b/contrib/google-adk-agents/src/nodes.ts
@@ -1,0 +1,172 @@
+/**
+ * A Temporal Activity as a node in an ADK 2.0 workflow graph.
+ *
+ * ADK 2.0's workflow runtime (`Workflow`, `node()`, `JoinNode`, `dynamicEntry`,
+ * …) is plain async code driven by session events, so it runs inside the
+ * Workflow sandbox as-is. {@link activityNode} is the one piece it lacks: a node
+ * whose body is a registered Temporal Activity, the same way `activityAsTool`
+ * exposes an Activity to the model.
+ *
+ * Failure semantics: a node that throws fails the whole `Runner.runAsync` (ADK
+ * rethrows node errors; only an agent's model call is absorbed), so a failed
+ * Activity fails the Workflow through the usual `ActivityFailure` and needs no
+ * absorbed-failure recording. ADK's own graph errors (a `NodeTimeoutError`, …)
+ * are plain `Error`s that the plugin converts as they leave the Workflow — see
+ * `absorbed-failure.ts`.
+ *
+ * IMPORTANT: this module is part of the Workflow-sandbox import graph. It must
+ * not import any worker-only module.
+ */
+
+import {
+  FunctionNode,
+  type BaseNode,
+  type FunctionNodeConfig,
+  type NodeContext,
+  type RetryConfig,
+  type SchemaLike,
+} from '@google/adk';
+import { ApplicationFailure } from '@temporalio/common';
+import { CancellationScope, inWorkflowContext, proxyActivities, type ActivityOptions } from '@temporalio/workflow';
+
+import { ACTIVITY_NODE_OUTSIDE_WORKFLOW_FAILURE_TYPE } from './error-types';
+import { activityOptionsFrom } from './model';
+
+/** Options for {@link activityNode}. */
+export interface ActivityNodeOptions<TInput = unknown> {
+  /**
+   * The registered Activity type to run — an `@activity`-style function
+   * registered on the worker (e.g. via the worker's `activities`). Also the
+   * node's name unless `nodeName` is set.
+   */
+  name: string;
+  /** The graph node's name, when it must differ from the Activity's (e.g. two nodes running one Activity). */
+  nodeName?: string;
+  /** The node's description, advertised when the node is used as a tool. */
+  description?: string;
+  /**
+   * Per-call Temporal Activity configuration (timeouts, retry, task queue,
+   * summary). Prefer `retry` here over ADK's `retryConfig`: an ADK retry
+   * re-runs the whole node on top of Temporal's own Activity retries.
+   */
+  activity?: ActivityOptions;
+  /**
+   * Maps the node's input to the Activity's argument list. Defaults to passing
+   * the input as the single argument. TypeScript has no runtime signature
+   * introspection, so — unlike Python's `activity_node` — there is no
+   * named-parameter binding; use the `NodeContext` for state-bound values:
+   * `(input, ctx) => [ctx.state.get('customerId'), input]`.
+   */
+  args?: (input: TInput, ctx: NodeContext) => unknown[];
+
+  // --- ADK node configuration, passed through to the node ---
+  /**
+   * Whether the node re-runs when a paused (human-in-the-loop) run resumes.
+   * Default `false`: a node that already produced its output is fast-forwarded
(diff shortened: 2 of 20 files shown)


contrib/google-adk-agents/src/nodes.ts
```diff
@@ -0,0 +1,178 @@
...
+    const run = activities[name]!;
+    const activityArgs = args ? args(input, ctx) : [input];
+    return (await underNodeAbort(ctx, () => run(...activityArgs))) as TOutput;
+  };
+
+  const config: FunctionNodeConfig = {
+    rerunOnResume: options.rerunOnResume ?? false,
+  };
+  if (description !== undefined) config.description = description;
+  if (options.waitForOutput !== undefined) config.waitForOutput = options.waitForOutput;
```
DABH: [P1] Do not expose `waitForOutput` until this node can actually fan in

ADK's scheduler gates fan-in on `node.requiresAllPredecessors`, but `FunctionNode` leaves that false; `waitForOutput` is consulted only after execution when output is missing. On this exact head, a diamond probe with two predecessors and `waitForOutput: true` invoked the underlying node twice with inputs `A` and `B`. For `activityNode`, that means two Activity side effects rather than one fan-in Activity.

Please implement a node subclass that opts into `requiresAllPredecessors`, or reject/remove this option until the pinned ADK can honor it.

contrib/google-adk-agents/src/nodes.ts
```diff
@@ -0,0 +1,178 @@
...
+    try {
+      scope.cancel();
+    } catch {
+      /* cancelling an already-cancelled scope, or a scope with nothing in it, is not an error worth surfacing here */
+    }
+  };
+  if (signal.aborted) onAbort();
+  else signal.addEventListener('abort', onAbort, { once: true });
+  try {
+    return await scope.run(body);
```
DABH: [P1] Await Activity cancellation before surfacing the node timeout

ADK races the node iterator against its deadline and calls `iterator.return()` without awaiting the unwind, so the pending `scope.run(body)`/Activity cancellation is detached. The existing timeout test passed while Core logged that the Activity completed after the Workflow had already ended. This bypasses the documented default `WAIT_CANCELLATION_COMPLETED`; with `retryConfig`, a replacement attempt can overlap the still-running timed-out Activity.

This needs an awaited upstream unwind or a plugin-owned timeout implementation. Otherwise, reject/limit `timeout` rather than claiming normal Temporal cancellation semantics.

contrib/google-adk-agents/src/absorbed-failure.ts
```diff
@@ -112,6 +123,45 @@ function raiseAbsorbed(frame: Frame): void {
   if (frame.pending.length > 0) throw frame.pending[0];
```
DABH: [P1] Clear a model failure that ADK consumes through a successful retry

`recordAbsorbedFailure` appends the first failed model call, but nothing removes it when ADK's node retry handles `NodeReportedError` and a later attempt succeeds. I reproduced a first model attempt failing with HTTP 400 and a second attempt succeeding; the Workflow still failed here with the stale first `GoogleAdkModelError.400`.

Please associate/claim the pending failure when ADK's retry path consumes it, so a successfully retried node does not fail the Workflow after the fact.

contrib/google-adk-agents/src/nodes.ts
```diff
@@ -0,0 +1,178 @@
...
+  if (description !== undefined) config.description = description;
+  if (options.waitForOutput !== undefined) config.waitForOutput = options.waitForOutput;
+  if (options.retryConfig !== undefined) config.retryConfig = options.retryConfig;
+  if (options.timeout !== undefined) config.timeout = options.timeout;
+  if (options.inputSchema !== undefined) config.inputSchema = options.inputSchema;
+  if (options.outputSchema !== undefined) config.outputSchema = options.outputSchema;
+  if (options.stateSchema !== undefined) config.stateSchema = options.stateSchema;
+  if (options.isolationScope !== undefined) config.isolationScope = options.isolationScope;
+
+  return new FunctionNode<TInput, TOutput>(nodeName ?? name, handler, config);
```
DABH: [P1] Validate or encode node names before defaulting to the Activity type

Temporal Activity type names may contain dots, while ADK uses `.` as its node-path delimiter. In a two-turn pause/resume probe, an Activity node named `payments.charge` with the default `rerunOnResume: false` ran twice: ADK recorded `probe.payments.charge`, then could not rehydrate it as a direct child on resume.

Please encode/validate path components, or require a path-safe `nodeName` whenever the Activity type contains ADK's delimiters. Otherwise HITL resume can duplicate Activity side effects despite the advertised default.

contrib/google-adk-agents/src/absorbed-failure.ts
```diff
@@ -112,6 +123,45 @@ function raiseAbsorbed(frame: Frame): void {
...
+  if (!(err instanceof Error)) return err;
+  const type = ADK_RUNTIME_FAILURE_TYPES[err.name];
+  if (type === undefined) return err;
+  if (err.name === NODE_REPORTED_ERROR && frame.pending.length > 0) return frame.pending[0];
+  if (err.name === DYNAMIC_NODE_FAIL_ERROR) {
+    // A Temporal failure the dynamic child raised is what must end the execution: a
+    // cancelled Activity has to end it CANCELLED rather than FAILED, and a failed one
+    // keeps the cause chain (status, retry state) that the wrapper would hide.
+    const wrapped: unknown = (err as { error?: unknown }).error;
+    if (wrapped instanceof TemporalFailure) return wrapped;
```
DABH: [P2] Recover the recorded model failure through `DynamicNodeFailError`

For `ctx.runNode(LlmAgent)`, ADK wraps the agent's `NodeReportedError` in `DynamicNodeFailError.error`. That wrapper is not itself a `TemporalFailure`, so this branch misses the recorded model failure and emits `GoogleAdkDynamicNodeFailError` instead. An exact-head HTTP-400 probe lost the original `ActivityFailure` and model status/cause chain.

Please recognize the nested `NodeReportedError` and surface the corresponding pending model failure, as the static-node path does.

contrib/google-adk-agents/src/nodes.ts
```diff
@@ -0,0 +1,178 @@
...
+    // `proxyActivities` returns a Proxy that materializes a stub for any name,
+    // so the indexed access is always defined; `noUncheckedIndexedAccess`
+    // widens the static type to `| undefined`, hence the assertion.
+    const run = activities[name]!;
+    const activityArgs = args ? args(input, ctx) : [input];
+    return (await underNodeAbort(ctx, () => run(...activityArgs))) as TOutput;
+  };
+
+  const config: FunctionNodeConfig = {
+    rerunOnResume: options.rerunOnResume ?? false,
```
DABH: [P2] Reject `rerunOnResume: true` while the pinned ADK ignores it

ADK fast-forwards a completed node from session history before consulting this flag. In a two-turn `RequestInput` probe, a completed node configured with `rerunOnResume: true` remained at one execution instead of running again.

Since this public option currently promises behavior the pinned runtime cannot provide, remove/reject `true` (or clearly mark it unsupported) until upstream ordering is fixed.

DABH (commented):
Ultrareview completed against the exact head. I found six reproducible behavioral gaps. Several originate in pinned ADK 2.0 internals, but they are exposed here as supported adapter options or break the adapter's documented Temporal guarantees, so callers need either an implementation workaround or an explicit rejection.

The package/dependency build and diff check passed. There were no existing review threads or prior bot/human comments to adjudicate.

contrib/google-adk-agents/src/nodes.ts
DABH: Agreed, and the option is gone in d4c8e8b. The flag is not a fan-in gate at all: `Workflow.handleCompletion` parks a node in WAITING when it ended with neither an output nor a route (the task-mode agent still waiting for `finish_task`), and fan-in is `BaseNode.requiresAllPredecessors`, which only `JoinNode` overrides. On an Activity node that makes it a no-op when the Activity returns a value and a hang when it returns nothing, so there is no setting worth exposing. The docs now point at `JoinNode` for fan-in (its input is the predecessor-name to output map) and say that an Activity returning nothing completes the node with an undefined output, which is covered by a new test.

contrib/google-adk-agents/src/nodes.ts
DABH: Removed in d4c8e8b, though for a slightly different reason than the one here. ADK never re-runs a node that completed, whatever the flag says: `Workflow.scheduleNode` fast-forwards a fast-forwardable prior run before it looks at `rerunOnResume`, and `ScheduleDynamicNode.runNode` says the same in its own comment. The flag only decides what happens to a node that paused for input last turn, and an Activity node never raises an interrupt, so neither value could ever change anything for it. Our doc promising a re-run was simply wrong; it now says completed nodes are always fast-forwarded and what `rerunOnResume` actually governs.

contrib/google-adk-agents/src/nodes.ts
DABH: Fixed in 1473025: `activityNode` now refuses a node name containing a dot at construction, with a non-retryable `GoogleAdkActivityNodeName` failure whose message suggests a path-safe replacement. Encoding it silently would have left the graph, the events and the traces carrying a name the caller never chose, so it fails loudly instead and `nodeName` is the escape hatch for a dotted Activity type. Two tests cover it: construction with `payments.charge` and no `nodeName` throws, and the same Activity under `nodeName: 'payments_charge'` runs once end to end. I did not add the cross-turn half of your repro here, because `eventsForCurrentRun` only reaches back over prior invocations that ended paused, so reproducing the double execution needs an interrupt and that lands in the HITL PR.

contrib/google-adk-agents/src/absorbed-failure.ts
DABH: Fixed in 2cd0504. `toWorkflowFailure` now walks `.error` down to the innermost error (nested dynamic runs wrap the carrier more than once) and decides there, which is the decision a static node would have produced: a Temporal failure is returned as it is, a `NodeReportedError` with a recording in the frame re-raises that recording, and anything else is typed from the innermost error with the carrier's own type as the fallback. New test runs an `LlmAgent` through `ctx.runNode` with a failing model and asserts the `GoogleAdkModelError.400` chain survives; it fails on the previous code.

contrib/google-adk-agents/src/absorbed-failure.ts
DABH: Real, fixed in a547b44 and 9b03163. A recording now carries the request's `adk_agent_name` label plus the `AbortSignal` ADK hands the model, which is `InvocationContext.abortSignal`; a later successful call matching both drops it. The signal is what makes this safe: a `Workflow` run mints one and gives every node the same object, so a node retry and a re-activation of the node share it, while a later turn does not, and a plain agent turn outside a graph has none at all and never clears anything. Keying on the agent alone (my first attempt) cleared a failed first turn when a second turn answered, which broke three of the existing absorbed-failure tests. Three tests now: a retried agent node ends the Workflow successfully, a swallowed failure by one agent still fails it when a different agent answers, and the existing non-retried cases are unchanged.

contrib/google-adk-agents/src/nodes.ts
DABH: Fixed in 34529e8. `activityNode` no longer declares `timeout` on the node; it runs the deadline itself as a durable timer, cancels the Activity's `CancellationScope` when the timer fires, awaits that cancellation on the Activity's own `cancellationType` terms, and only then throws ADK's `NodeTimeoutError`, so `retryConfig.exceptions` and `ADK_RUNTIME_FAILURE_TYPES` still match it and a retry cannot start on top of the Activity it replaces. An abort that is not the deadline still surfaces as the cancellation it is. Two tests, both failing on the old code: history shows `ActivityTaskCanceled` before `WorkflowExecutionFailed`, and with a `retryConfig` the second `ActivityTaskScheduled` comes after the first cancellation completed. One correction to the premise: an unset `cancellationType` is TRY_CANCEL on the wire, since that is the zero value of the core proto enum and the SDK never fills a default in, so the tests set `WAIT_CANCELLATION_COMPLETED` explicitly and the docs say which one waits.

PR #2418, opened 2026-09-11
feat(google-adk-agents): durable human-in-the-loop for ADK 2.0

DABH:
## What was changed

ADK 2.0 pauses a run by emitting a long-running function call (`adk_request_input` from a
`RequestInput` node or `requestInputTool`, `adk_request_confirmation` from a tool gated with
`requireConfirmation`) and resumes when a later user message answers it. The runner already runs
inside the Workflow, so the wait is ordinary Workflow code: expose what is pending from a Query,
take the answer through a Signal or Update, `condition()`, run the next `runAsync` turn. This PR
adds the wire format and the gates (AI-510):

- `pendingHitlRequests(events)` (`src/hitl.ts`) returns the pauses ADK still considers
  unanswered as `HitlRequest`s (the input and confirmation variants of ADK's `UserInputRequest`,
  plain JSON a Query can return), minus credential requests. Those stay unsupported:
  answering one puts a secret in an Update payload that history keeps, and the plugin's
  deterministic ids make ADK's OAuth2 `state` predictable.
- `hitlInputResponse` and `hitlConfirmationResponse` build the `functionResponse` parts ADK's resume
  path reads. The input builder applies the same unwrap and JSON parse as ADK's `unwrapResponse`
  before it hands the part over, and refuses a `'42'` or a `'"foo"'` the request's `responseSchema`
  does not accept as text rather than let ADK deliver it retyped. The refusal is a non-retryable
  `ApplicationFailure` (`GoogleAdkHitlResponseError`), so an Update handler that builds the part
  rejects the bad answer instead of failing the Workflow Task; the README pattern does exactly that.
- `requireConfirmation` on `activityAsTool` and `TemporalMCPToolset`, over a shared
  `src/confirmation.ts` that reproduces the hint and error strings of `FunctionTool.checkConfirmation`
  byte for byte. The Activity (or the `<name>-callTool` Activity) is not scheduled until the human
  approves, and a rejection returns ADK's rejection result to the model. Outside a Workflow the
  toolset wraps ADK's own MCP tools in the same gate, so the option never runs a call unapproved.

Both tools answer `checkRequireConfirmation`, which is what ADK binds an approval to. A gate a tool
body asks for at run time is refused with `confirmation_not_required`, and so is `SecurityPlugin`'s
CONFIRM outcome. That limit, a plain-text approval running the tool without ever answering the gate's
call, and `requestInputTool` reaching the model only as text are all pinned by tests and written down
in the README.

## Testing

Sixteen E2E cases in `hitl.test.ts` drive the real ADK runtime through a Workflow-side pause/answer
loop: a `RequestInput` node, a replay-stable default interrupt id, two pending requests with a
partial answer, an agent-raised input answered as text, a gated Activity tool approved and rejected
(asserting from history that nothing was scheduled first), a gated MCP tool, a plain-text approval,
the two refusals above, a refused answer rejecting its Update and the corrected one completing the
run, and a resumed dynamic node fast-forwarding its completed child. The stub server covers the gate
on the direct ADK path. Unit tests cover the builders; `replay.test.ts` replays a checked-in HITL history. Green in both sandbox modes,
plus eslint, prettier and ts-prune.

Stack: [1/4 bump](https://github.com/temporalio/sdk-typescript/pull/2416) → [2/4 graph runtime](https://github.com/temporalio/sdk-typescript/pull/2417) → 3/4 HITL (this, #2418) → [4/4 MCP resources + model routing](https://github.com/temporalio/sdk-typescript/pull/2419)

### DIFF
--- a/contrib/google-adk-agents/README.md
+++ b/contrib/google-adk-agents/README.md
@@ -162,6 +162,9 @@ connection params above and ADK's `MCPToolset` both open a new session per
 operation rather than reusing one. Holding a session open across operations is
 the factory's job — see `MCPToolsetFactory`.
 
+Gate the toolset's tools behind human approval with
+`requireConfirmation` — see [Durable human-in-the-loop](#durable-human-in-the-loop).
+
 ### Activities as tools
 
 Use `activityAsTool` to expose an existing Temporal Activity to the agent:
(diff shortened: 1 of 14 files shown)


contrib/google-adk-agents/src/mcp.ts
```diff
@@ -79,6 +93,15 @@ export interface TemporalMCPToolsetOptions {
...
   connectionParams?: MCPConnectionParams;
+  /**
+   * Gate the toolset's tools behind human approval, like ADK's
+   * `FunctionTool({ requireConfirmation })`. The `<name>-callTool` Activity is
+   * NOT scheduled on the pass that raises the confirmation request, nor after a
+   * rejection; it runs once the resumed turn carries the approval (see
+   * `hitlConfirmationResponse`). Only enforced on an `LlmAgent` turn — ADK
+   * 2.0's workflow `ToolNode` does not route through the confirmation path.
+   */
+  requireConfirmation?: MCPRequireConfirmation;
```
DABH, line 119: [P1] Enforce `requireConfirmation` in the supported direct-ADK path

This returns raw tools from ADK's `MCPToolset`, so `TemporalMCPToolsetOptions.requireConfirmation` is silently discarded outside a Workflow. I reproduced this exact head with the checked-in MCP stub: `requireConfirmation: true` returned an `echo` tool whose `checkRequireConfirmation(...)` was `false`, and invoking it immediately sent `tools/call` and returned the unapproved result.

Please wrap the returned tools with the same gate, or reject this option outside Workflow context. An approval option must not silently execute the call.

contrib/google-adk-agents/README.md
```diff
@@ -234,6 +237,111 @@ const runner = new InMemoryRunner({ agent: graph });
...
+
+export const pendingQuery = defineQuery<HitlRequest[]>('pending');
+export const respondUpdate = defineUpdate<void, [string, unknown]>('respond');
+
+// `graph` is the Workflow built in the section above; any RunnableRoot works.
+export async function reviewWorkflow(prompt: string): Promise<unknown> {
+  let pending: HitlRequest[] = [];
+  const answers = new Map<string, unknown>();
+  setHandler(pendingQuery, () => pending);
+  setHandler(respondUpdate, (interruptId, value) => void answers.set(interruptId, value));
```
DABH: [P1] Validate the answer before this Update commits it

The handler accepts and durably stores arbitrary `unknown`, while `hitlInputResponse` and ADK schema validation happen later in the Workflow body. For example, an input such as `'42'` without a string schema makes `hitlInputResponse` throw after the Update has succeeded. That error is outside the Update handler, so it repeatedly fails the Workflow Task; replay restores the same accepted value, and a caller cannot correct it through this loop.

Please build/validate the response in an Update validator or handler before mutating state, store only validated `Part` values, and cover an invalid-answer-then-correction flow.

contrib/google-adk-agents/src/hitl.ts
```diff
@@ -0,0 +1,191 @@
...
+ * same two steps as ADK's `unwrapResponse`
+ * (`workflow/utils/rehydration_utils.ts`): unwrap a response whose only key is
+ * `result`, then JSON-parse a string the declared schema does not accept.
+ */
+function assertNoJsonCoercion(request: HitlRequest, response: Record<string, unknown>): void {
+  if (Object.keys(response).length !== 1 || !(RESULT_KEY in response)) return;
+  const unwrapped = response[RESULT_KEY];
+  if (typeof unwrapped !== 'string' || acceptsString(request.responseSchema)) return;
+  const parsed = parseJsonIfPossible(unwrapped);
+  if (parsed === NOT_JSON || typeof parsed === 'string') return;
```
DABH: [P2] Reject JSON strings that parse to strings too

ADK's matching unwrap logic returns the parsed value, so `hitlInputResponse(request, '"foo"')` is delivered as `foo`, silently stripping the quotes. This branch accepts it even though the README promises that any string which reads as JSON is refused when the schema does not accept strings.

Once `acceptsString(...)` is false, only `NOT_JSON` should bypass the guard; a parsed string is still a coercion.

contrib/google-adk-agents/src/hitl.ts
```diff
@@ -0,0 +1,191 @@
...
+  type Event,
+  type UserInputRequest,
+} from '@google/adk';
+
+/**
+ * A pause awaiting a human answer: ADK's own {@link UserInputRequest}, plain
+ * JSON, so a Workflow Query can return it as-is. `kind` is `'input'` or
+ * `'confirmation'` here — see {@link pendingHitlRequests}.
+ */
+export type HitlRequest = UserInputRequest;
```
DABH: [P3] Exclude credential requests from the public `HitlRequest` type

Upstream `UserInputRequest` also includes the credential variant and its `authConfig`, while this API deliberately drops credential requests and documents `kind` as only `input | confirmation`. Keeping the wider alias contradicts the runtime contract and prevents consumers from writing an exhaustive switch over the supported cases.

Please export a credential-excluding union (ideally distinct input and confirmation request aliases used by their respective builders).

DABH (commented):
Ultrareview completed against the current head (the latest force-push was history-only; its tree is identical to the one probed). I found four issues, including one approval-gate bypass and one documented Update pattern that can trap a Workflow in repeated Workflow Task failures.

The core in-Workflow approval/rejection protocol otherwise matches ADK 2.0. There were no existing review threads or prior bot/human comments to adjudicate. The prior identical tree passed all checks; the refreshed CI run was still in progress when this review was submitted.

contrib/google-adk-agents/src/hitl.ts
DABH: Right, and ADK returns the parsed value whatever it parses to, so `'"foo"'` reached the node as `foo` with the quotes gone. Fixed in 6106f7dc: once the schema does not accept a string, only text `JSON.parse` rejects gets through, and the message now names the value ADK would deliver. The unit test covers the quoted-string case and keeps a non-JSON answer passing.

contrib/google-adk-agents/src/hitl.ts
DABH: Fixed in 1bb31eb3. ADK declares `UserInputRequest` as one interface with a union-typed `kind`, not a discriminated union, so `Extract` narrows to `never`; the aliases restate `kind` as its literal instead. `HitlInputRequest` and `HitlConfirmationRequest` are exported, `HitlRequest` is their union, and the builders take the narrow type, so narrowing on `kind` now picks the right one. The runtime `assertKind` stays, because a request that came back over a Query is plain JSON by then, and it still names credential requests explicitly.

contrib/google-adk-agents/src/mcp.ts
DABH: Confirmed and fixed in 015ae609. ADK's `MCPToolset` knows nothing about the option, so outside a Workflow the gate was simply not there. `GatedMCPTool` now wraps each tool ADK returns and applies the same `checkRequireConfirmation` plus `gateOnConfirmation` protocol, delegating the declaration and the call to the wrapped tool. ADK already applies `prefix` to the names it hands back (`mcp_toolset.ts` line 83), so the predicate sees the advertised name here as it does inside a Workflow. Two tests against the stub server cover it: gated, the first pass returns the "requires confirmation" result and the log still shows only `tools/list`, a rejection adds nothing either, and an approval produces exactly one `tools/call`; ungated, the tool runs as before.

contrib/google-adk-agents/README.md
DABH: Agreed, and the thrown type was wrong for that too. Fixed in a0895b25 in two parts. The builders now raise a non-retryable `ApplicationFailure` of a new exported `HITL_RESPONSE_FAILURE_TYPE` instead of a `TypeError`, because `doUpdateImpl` in `packages/workflow/src/internals.ts` rejects an update only for a `TemporalFailure` and sends anything else to `handleWorkflowFailure`. The README pattern and the `runWithHitl` fixture then build the `Part` inside the handler and store only validated ones, so an unknown interrupt id or a refused value rejects the Update and leaves the interrupt open. New E2E test: `'42'` for a request with no string schema rejects with that type, `42` then completes the run, and the history has no `WorkflowTaskFailed`.

PR #2419, opened 2026-09-11
feat(google-adk-agents): MCP resources and in-sandbox model routing

DABH:
## What was changed

- MCP resources. `TemporalMCPToolset.listResources()` / `readResource(name)` route through new
  `<name>-listResources` / `<name>-readResource` Activities. A read resolves the name to a URI and
  reads it over one MCP session; ADK's own `MCPToolset.readResource` opens two.
- `loadMcpResourceTool(toolset)`, the counterpart of ADK's two-phase `LoadMcpResourceTool`: same
  declaration, same result, same injected instruction, and `processLlmRequest` appends the requested
  resource contents to the next model request. The listing is memoized per tool instance, since every
  model turn would otherwise be an Activity and its history events; `refreshResourceList: true`
  restores ADK's behavior. A failed listing or read is logged and skipped, as in ADK, and the resource
  Activities retry three times by default (`activity.retry` overrides it), so a server that is down
  cannot hang the turn. A cancelled listing or read stays a cancellation rather than becoming a
  retryable MCP failure.
  `mockMCPToolset` and the stdio stub server grow resources.
- Model auto-routing. A new per-Workflow interceptor module re-registers ADK's built-in model
  patterns to a `TemporalModel` subclass, so vanilla ADK code with a raw model string
  (`model: 'gemini-2.5-flash'`) is durable with no change, on `TemporalModel`'s default Activity
  options. `GoogleAdkPluginOptions.autoRouteModels` (default `true`) is the kill switch.

## Why

`Gemini` and `ApigeeLlm` cannot run inside a Workflow: they call the network from the sandbox, and
the error names neither ADK nor the real mistake. ADK's `LLMRegistry` is a `Map` keyed by the
`RegExp` objects in each class's `supportedModels`, resolved first-registered-first, so spreading the
built-in classes' own arrays replaces their entries in place rather than appending duplicates behind
them. The registration is gated on `inWorkflowContext()`, so the Worker keeps the real classes:
`modelProvider` still resolves the same names there, and `TemporalModel`'s own fallback to
`LLMRegistry.newLlm` cannot recurse into itself.

## Testing

Resource list and read against the mock and the stdio stub server (session boundaries asserted), an
unknown resource, a toolset without the resource methods, the two-phase `load_mcp_resource` flow with
the listing memoized and refreshed, a server whose Activities fail reaching the skip path on the
default retry policy, and a hanging read cancelled through the Activity's signal. For routing: raw Gemini and
Apigee strings running through the model Activity, a registry probe over every built-in pattern with
auto-routing on and off, and `RoutedLlm` over two `TemporalModel`s.

Stack: [1/4 bump](https://github.com/temporalio/sdk-typescript/pull/2416) → [2/4 graph runtime](https://github.com/temporalio/sdk-typescript/pull/2417) → [3/4 HITL](https://github.com/temporalio/sdk-typescript/pull/2418) → 4/4 MCP resources + model routing (this, #2419)

### DIFF
--- a/contrib/google-adk-agents/src/activities.ts
+++ b/contrib/google-adk-agents/src/activities.ts
@@ -17,6 +17,7 @@ import {
   MCPToolset,
   isBaseToolset,
   type BaseLlm,
+  type BaseToolset,
   type Context as AdkToolContext,
   type LlmRequest,
   type LlmResponse,
@@ -28,9 +29,14 @@ import { ApplicationFailure } from '@temporalio/common';
 import { Context as ActivityContext } from '@temporalio/activity';
 import { WorkflowStreamClient } from '@temporalio/workflow-streams/client';
 
-import { MCP_ERROR_FAILURE_TYPE, MCP_TOOL_NOT_FOUND_FAILURE_TYPE, MODEL_ERROR_FAILURE_TYPE } from './error-types';
+import {
+  MCP_ERROR_FAILURE_TYPE,
+  MCP_RESOURCES_UNSUPPORTED_FAILURE_TYPE,
+  MCP_TOOL_NOT_FOUND_FAILURE_TYPE,
+  MODEL_ERROR_FAILURE_TYPE,
+} from './error-types';
 import type { InvokeModelArgs, InvokeModelStreamingArgs, ModelActivities, WireLlmRequest } from './model';
-import type { MCPCallToolArgs, MCPToolsetFactory } from './mcp';
+import type { MCPCallToolArgs, MCPReadResourceArgs, MCPResourceContents, MCPToolsetFactory } from './mcp';
 
 const DEFAULT_STREAM_BATCH_INTERVAL = '100 milliseconds';
 
@@ -99,7 +105,10 @@ export function createModelActivities(options: ModelActivitiesOptions = {}): Mod
   };
 }
 
-/** Builds the per-server `<name>-listTools` / `<name>-callTool` pairs. @internal */
+/**
+ * Builds the per-server `<name>-listTools` / `<name>-callTool` /
+ * `<name>-listResources` / `<name>-readResource` Activities. @internal
+ */
 export function createMCPActivities(
   toolsets: Record<string, MCPToolsetFactory> = {}
 ): Record<string, (args: never) => Promise<unknown>> {
@@ -108,6 +117,31 @@ export function createMCPActivities(
     Object.assign(activities, mcpActivitiesForName(name, factory));
   }
   return activities;
+}
+
+/** One live MCP session, as produced by ADK's `MCPSessionManager`. */
+type MCPSession = Awaited<ReturnType<MCPSessionManager['createSession']>>;
+
+/**
+ * The resource methods ADK 2.0 added to `MCPToolset`, duck-typed so a
+ * factory-supplied toolset (a real `MCPToolset`, or a test double) qualifies
+ * without a class check across two ADK copies.
+ */
+interface ResourceCapableToolset {
+  listResources(): Promise<string[]>;
+  readResource(name: string): Promise<MCPResourceContents[]>;
+}
+
+function resourceCapable(toolset: BaseToolset, name: string): ResourceCapableToolset {
+  const candidate = toolset as Partial<ResourceCapableToolset>;
+  if (typeof candidate.listResources === 'function' && typeof candidate.readResource === 'function') {
+    return candidate as ResourceCapableToolset;
+  }
+  throw ApplicationFailure.nonRetryable(
+    `The toolset registered for MCP server '${name}' cannot serve resources: it has no ` +
+      "'listResources' / 'readResource' methods. Return an MCPToolset (or MCPConnectionParams) from the factory.",
+    MCP_RESOURCES_UNSUPPORTED_FAILURE_TYPE
+  );
 }
 
 function mcpActivitiesForName(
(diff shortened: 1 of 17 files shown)


DABH (commented):
Ultrareview completed against the current head (the latest force-push was history-only; its tree is identical to the one probed). I found two issues at the MCP resource boundary. I also checked the automatic resource-routing registry against the actual ADK 2.0 implementation; that behavior held up. The full dependency-closure build, all 133 package tests, targeted ESLint, Prettier, and diff checks passed. There were no existing pre-review threads to adjudicate.

contrib/google-adk-agents/src/mcp.ts
```diff
@@ -267,3 +331,176 @@ class TemporalMCPTool extends BaseTool {
...
+
+/**
+ * The Temporal counterpart of ADK 2.0's `LoadMcpResourceTool`, byte-for-byte
+ * in what the model sees: same declaration, same result, same injected
+ * instruction. It is a two-phase tool. The model calls `load_mcp_resource`
+ * with resource names; on the *next* turn, `processLlmRequest` reads those
+ * resources — through the toolset's `<name>-readResource` Activity — and
+ * appends their contents to the request. Each turn it also lists the server's
+ * resources (memoized unless `refreshResourceList`) and tells the model about
+ * them. Like ADK, a failed list or read is logged and skipped rather than
```
DABH, line 440: [P1] Bound retries if failures are meant to be skipped

Both resource Activities use `activityOptionsFrom`, whose default provides only a one-minute start-to-close timeout. MCP errors without an HTTP status are retryable, and Temporal's default maximum attempts is unlimited, so a persistently unavailable server never reaches the catch/log/skip path here; it retries one-minute attempts forever and the model turn hangs. The failing-server test sets `maximumAttempts: 1`, which masks the default behavior.

Please give this helper a finite retry/schedule-to-close policy (or expose one specifically for its best-effort reads), or correct the contract/docs and test the default semantics explicitly.

contrib/google-adk-agents/src/activities.ts
```diff
@@ -164,18 +200,62 @@ function mcpActivitiesForName(
...
+            const resource = listed.resources.find((candidate) => candidate.name === args.name);
+            if (!resource) throw new Error(`Resource with name '${args.name}' not found.`);
+            if (!resource.uri) throw new Error(`Resource '${args.name}' has no URI.`);
+            const result = await session.readResource({ uri: resource.uri }, { signal: abortSignal });
+            return result.contents as MCPResourceContents[];
+          });
+        }
+        return await resourceCapable(produced, name).readResource(args.name);
+      } catch (err) {
+        throw toApplicationFailure(err, MCP_ERROR_FAILURE_TYPE);
```
DABH, line 278: [P2] Preserve resource Activity cancellation before normalizing the error

The MCP client uses the Activity cancellation signal, but this catch turns its cancellation rejection into a retryable `GoogleAdkMCPError`. On this exact tree, a hanging read with heartbeat delivery and `WAIT_CANCELLATION_COMPLETED` returned an `ActivityFailure` instead of cancellation. As a result, `loadMcpResourceTool`'s `isCancellation` guard also does not rethrow it and can swallow a child-scope cancellation.

Preserve `AbortError`/`CancelledFailure` (or rethrow the aborted signal's reason) before `toApplicationFailure`, and cover list/read cancellation. This is the resource-specific extension of the foundational cancellation issue noted on #2416.

contrib/google-adk-agents/src/mcp.ts
DABH: Agreed, and the failure mode is worse than slow: `toApplicationFailure` leaves a status-less MCP error retryable and Temporal's default is unlimited attempts, so against a server that is down the listing retried forever inside `processLlmRequest` and the turn never returned. The skip path the README promises was unreachable.

Fixed in 0ac43fbc. `listResources` and `readResource` now default `retry.maximumAttempts` to 3, merged field by field under the caller's `activity.retry`, so an explicit value (including `0` for unlimited) still wins. Tool calls keep the unlimited default, since their result is what the turn is waiting for. Documented in the MCP resources section of the README.

You were right that the test was masking it. It no longer sets `maximumAttempts`, only a 1ms backoff, so it runs the default and asserts each Activity ended on attempt 3 and failed while the turn still completed. With the fix reverted that test hangs instead of failing, which is the old behavior exactly.

contrib/google-adk-agents/src/activities.ts
DABH: Confirmed, and it bit twice over. The Worker only reports an Activity as cancelled when the error leaving it is a `CancelledFailure` or an `AbortError`, so classifying the abort as a retryable `GoogleAdkMCPError` both kept the attempt chain going and hid the cancel from `loadMcpResourceTool`: its `isCancellation` guard never fired, and a cancelled read was logged and skipped like any other failure.

Fixed in 2f6b5f51, on top of the `rethrowIfCancelled` helper that landed in #2416. Both resource catches now call it first, with the signal hoisted to the top of the Activity body the way the model and tool Activities do.

Three tests, each failing before the change: a read hanging on its abort signal, cancelled mid-flight, now ends the Activity cancelled on its first attempt with no Activity failure in history, and the Workflow ends cancelled instead of answering, which is the tool re-raising rather than skipping. Two `MockActivityEnvironment` checks cover `listResources` and `readResource` directly and assert a `CancelledFailure` rather than an `ApplicationFailure`; before the fix they got `ApplicationFailure`.

PR #2423, opened 2026-09-14
Simplify Google ADK integration install command

xumaple:
Modern npm installs the integration's required peer dependencies automatically. This simplifies the README command to install only `@temporalio/google-adk-agents`, while retaining the supported peer ranges below it for users resolving an existing dependency graph.


brianstrauch (approved):
Thanks!

PR #2424, opened 2026-09-14
feat!(extstore): add new concurrency controls that provide that cap per-message and total driver operations.

cconstable:
## What was changed

**Previously:**
- Message payloads were walked sequentially which meant external storage performed it's operations sequentially. 

**Now:**
- All payloads sites in a message are walked concurrently but inside a driver, operations are capped by `maxOperationsPerMessage` per message and the total number of operations drivers perform is capped "globally" (per instance of ExternalStorage) by `maxDriverOperations`. 
- To do this, we need drivers to cooperate so a new `context.limiter` is passed into drivers which they can use to take a permit for performing work. 
- The existing S3 and GCS drivers were updated. 
- If a driver doesn't use the limiter during an operation a warning is logged.
- defaults were added (8 per message, 64 global)

## Why?

- TS was lacking concurrency controls for external storage.
- https://github.com/temporalio/sdk-typescript/issues/2275

## Checklist

New tests were added.


packages/common/src/internal-non-workflow/external-storage-visitor.ts
```diff
@@ -26,12 +27,11 @@ export interface ExternalStorageStoreOptions {
   initialTarget?: StorageDriverTargetInfo;
   /** Derives new storage target from the current message. */
   deriveContext?: ContextDeriver<StoreTarget>;
-  /** Bounds concurrent transform calls across payload sites. Omit for sequential. */
-  limit?: ConcurrencyLimit;
   /** Aborts the walk and every in-flight driver call. */
   abortSignal?: AbortSignal;
   /** Collects metrics of the store operations performed during the walk. */
   metrics?: ExternalStorageMetricsAccumulator;
+  logger?: Logger;
```
cconstable, line 34: I don't think there is another way to thread the logger in right now.

packages/common/src/converter/extstore.ts
```diff
@@ -116,6 +138,46 @@ export type StorageDriverSelector = (context: StorageDriverSelectContext, payloa
 /** Default {@link ExternalStorage.payloadSizeThreshold}: 256 KiB. */
 const DEFAULT_PAYLOAD_SIZE_THRESHOLD = 256 * 1024;
 
+/** Default {@link ExternalStorageConcurrency.maxDriverOperations}. */
+const DEFAULT_MAX_DRIVER_OPERATIONS = 64;
+
+/** Default {@link ExternalStorageConcurrency.maxOperationsPerMessage}. */
+const DEFAULT_MAX_OPERATIONS_PER_MESSAGE = 8;
```
cconstable, line 145: These were arbitrarily chosen.

packages/common/src/converter/extstore.ts
```diff
@@ -52,6 +52,24 @@ export interface StorageDriverActivityInfo {
...
 export type StorageDriverTargetInfo = StorageDriverWorkflowInfo | StorageDriverActivityInfo;
 
+/**
+ * Bounds how many external storage operations all drivers managed by a single ExternalStorage
+ * instance may have in flight at once. This limit is cooperative which means driver
+ * implementations must use the provided context.limiter to acquire permits for each operation.
+ *
+ * This limit can be configured via {@link ExternalStorageConcurrency.maxDriverOperations}.
+ *
+ * n.b. taking permits from inside another permit can lead to deadlocks if the nesting exceeds
```
chris-olszewski, line 62: TIL about "n.b."

Might be worth switching to just "Note:" or something along those lines in case there are other devs out there who have been slacking in their latin lessons.

packages/common/src/converter/extstore.ts
```diff
@@ -52,6 +52,24 @@ export interface StorageDriverActivityInfo {
...
+ * the configured concurrency limit. It is up to the driver to determine what a single "operation"
+ * looks like.
+ *
+ * @experimental
+ */
+export interface StorageDriverLimiter<Item> {
+  /** Runs `operation` once a permit for `item` is available, releasing the permit when it settles. */
+  permit<T>(item: Item, operation: () => Promise<T>): Promise<T>;
```
chris-olszewski, line 70: I'm unsure I understand why `permit` takes an item. The argument is always ignored by the implementations. Looking at the RFC, I see that Python/Go were written taking it, but TS/Java omitted it.

chris-olszewski (commented):
This makes sense to me. 

My only question (and it might be a dumb one) is regarding the passing of the item to `permit`. Do we expect permit acquisition behavior to be different per payload/claim instances?

packages/common/src/converter/extstore.ts
cconstable: that's fair. old habit 😄

packages/common/src/converter/extstore.ts
cconstable: The original implementation used these to acquire permits. It is not needed anymore. Good catch.

chris-olszewski (approved):

packages/common/src/converter/extstore.ts
jmaeagle99: The intention was to possibly allow permitting based on size instead of or in addition to count. If we don't include this now, then it's likely we won't be able to do it in the future without making carve outs.

cconstable: Was just coming back to comment ^ @jmaeagle99 and I just spoke about this and it was noted in the RFC I just misattributed it to an older design. Adding this back now.

Issue #740, opened 2022-07-01
[Feature Request] Add `ChildWorkflowHandle.cancel()`


lorensr:
### Is your feature request related to a problem? Please describe.

Currently can only cancel a child by canceling a scope.

### Describe the solution you'd like

Would be convenient to be able to do `childHandle.cancel()`

Current handle API: https://typescript.temporal.io/api/interfaces/workflow.childworkflowhandle/

mjameswh:
Related [Slack](https://temporalio.slack.com/archives/C01DKSMU94L/p1683053747688419) discussion.

Two possible work arounds:
- Use `getExternalWorkflowHandle` to get a regular `WorkflowHandle` on the child workflow. Then use that handle to send the cancel command.
- Start the child in a cancellation scope, then cancel the scope.

PR #2429, opened 2026-09-15
feat: ID-first Event Groups and additional attach sites (experimental)

mjameswh:
Follow-up to #2209.

## What was changed

`createEventGroup` now takes a required ID and an optional `{ label }`. The ID is stored verbatim (no SHA-1 derivation); don't put secrets in it. The user-facing type is renamed `EventGroup` (was `EventGroupMarker`).

Direct attach now works on `upsertMemo`, `upsertSearchAttributes`, `patched` / `deprecatePatch`, workflow-side `signalWithOptions`, and `ExternalWorkflowHandle.cancel`. Variadic `signal(...)` still cannot take options, so tests cover both that path (scope only) and `signalWithOptions` (direct + scope), including the child-workflow handle.

`ChildWorkflowHandle.cancel()` is still missing (#740), so first-class child cancel with Event Groups is not in this PR.

## Why

Matches the revised Event Groups design: caller-provided IDs, optional labels, and attach sites that the 1.24.0 POC did not cover.

## Testing

History tests in `packages/test/src/test-event-groups.cloud-pending.ts`. Needs an Event Groups-capable server, e.g.:

```
TESTS_CLI_VERSION=v1.7.4-standalone-nexus-operations
```


Sushisource (approved):
Nice. Appreciate the docstring cleanup.

PR #2431, opened 2026-09-16
chore: upgrade to TS 7 and migrate to oxlint

chris-olszewski:
## What was changed
- Upgrade to TS 7
- Partially migrate from eslint to oxlint: there are 4 `import/*` lints that are fairly useful to keep around that aren't available in oxlint. We have a new "sidecar" project pinned to TS 6 so we can use eslint to perform these lints.

Commented on any source changes with the respective reason.

### Before
```
Benchmark 1: pnpm run lint:check
  Time (mean ± σ):     82.647 s ±  3.367 s    [User: 109.712 s, System: 9.114 s]
  Range (min … max):   78.234 s … 86.745 s    10 runs
```

### After
```
Benchmark 1: pnpm run lint:check
  Time (mean ± σ):     10.537 s ±  0.201 s    [User: 22.052 s, System: 1.990 s]
  Range (min … max):   10.223 s … 10.929 s    10 runs
```

## Checklist

1. Closes N/A

2. How was this tested:
CI

3. Any docs updates needed?
N/A


contrib/openai-agents/src/common/sandbox-activity-types.ts
```diff
@@ -205,7 +205,7 @@ export function bytesToBase64(bytes: Uint8Array): string {
   return result;
 }
 
-export function base64ToBytes(value: string): Uint8Array {
+export function base64ToBytes(value: string): Uint8Array<ArrayBuffer> {
```
chris-olszewski, line 208: This is actually due to a change from TS 5.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-7.html#typedarrays-are-now-generic-over-arraybufferlike

packages/common/package.json
```diff
@@ -14,7 +14,7 @@
   "dependencies": {
     "@temporalio/proto": "workspace:*",
     "long": "^5.3.2",
-    "ms": "3.0.0-canary.1",
+    "ms": "3.0.0-canary.202508261828",
```
chris-olszewski, line 17: Previous canary had an incomplete exports field in `package.json`, only matters with the `moduleResolution` change.

packages/meta/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "module": "ES2022",
```
chris-olszewski, line 4: Typedoc still uses TS 5 which doesn't support commonjs/bundler combination that is supported in TS 6/7. We can upgrade to TS6 and remove this override for docs, but that requires a major bump of our Typedoc that deserves its own PR.

packages/worker/src/utils.ts
```diff
@@ -8,7 +8,7 @@ export function toMB(bytes: number, fractionDigits = 2): string {
 }
 
 export function byteArrayToBuffer(array: Uint8Array): Buffer {
-  return Buffer.from(array, array.byteOffset, array.byteLength + array.byteOffset);
```
chris-olszewski, line 11: The new more accurate Buffer types in TS5.7+ now raise this as invalid use of `Buffer.from` since we're passing it a `Uint8Array`: https://nodejs.org/api/buffer.html#static-method-bufferfrombuffer

packages/worker/src/workflow/threaded-vm.ts
```diff
@@ -160,7 +160,11 @@ export class WorkerThreadClient {
       this.activeWorkflowCount--;
     }
     // Transfer ownership of activation buffer for zero-copy transfer
-    if (request.input.type === 'activate-workflow' && request.input.activation instanceof Uint8Array) {
+    if (
+      request.input.type === 'activate-workflow' &&
+      request.input.activation instanceof Uint8Array &&
+      request.input.activation.buffer instanceof ArrayBuffer
```
chris-olszewski, line 166: Fallout of `Uint8Array` now carrying the underlying storage in the type.

packages/test/src/test-interface-type-safety.ts
```diff
@@ -115,20 +115,20 @@ test('startUpdate and executeUpdate call signatures', async (t) => {
     await handle.startUpdate(nullaryUpdate, {});
     // @ts-expect-error: waitForStage required
     await handle.startUpdate(nullaryUpdate, { args: [] });
-    // @ts-expect-error: waitForStage must be ACCEPTED
     await handle.startUpdate(nullaryUpdate, {
+      // @ts-expect-error: waitForStage must be ACCEPTED
```
chris-olszewski, line 119: TS 6/7 now reports these errors on the object property instead of the entire call

packages/test/src/test-nexus-handler.ts
```diff
@@ -466,11 +466,12 @@ test('decodePayload maps non-retryable PayloadValidationError from converter to
   t.false(err?.retryable);
   t.is(err?.message, 'Invalid operation input');
   // The original failure, including its message, is retained as the cause.
-  t.true(err?.cause instanceof ApplicationFailure);
-  t.is(err?.cause, cause);
-  t.is((err?.cause as ApplicationFailure).type, PAYLOAD_VALIDATION_ERROR_TYPE);
-  t.true((err?.cause as ApplicationFailure).nonRetryable);
-  t.is((err?.cause as ApplicationFailure).message, 'Payload validation failed');
+  const actualCause = err?.cause as ApplicationFailure;
```
chris-olszewski, line 469: oxlint complained about the casting in the assertions, seemed reasonable.

tsconfig.base.json
```diff
@@ -4,7 +4,7 @@
     "composite": true,
     "incremental": true,
     "module": "commonjs",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
chris-olszewski, line 7: As mentioned in the PR this is the "leave me alone" alternative for upgrading to TS 6/7, but shouldn't be our final configuration: https://github.com/microsoft/TypeScript/pull/62320

PR #2432, opened 2026-09-16
Require waitForStage parameter for Nexus Workflow Updates

atol:
## What was changed
Expose and require the `waitForStage` parameter to be explicitly set to `ACCEPTED` for Nexus Workflow Updates, instead of being hardcoded internally.

## Why?
`waitForStage` is expected to be a required parameter because `ACCEPTED` is a non-obvious default that users should be aware of. This also aligns with the behaviour in the Go, Java, and .NET SDKs.

## How was this tested
Unit tests.


packages/nexus/src/index.ts
```diff
@@ -4,6 +4,8 @@
  * @module
  */
 
+export { WorkflowUpdateStage } from '@temporalio/client';
```
bergundy: Don't re-export please from top level modules. I have made that mistake when I started this SDK and regret it. It confused auto-completion and the reference documentation gets messed up.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -736,9 +747,18 @@ async function updateWorkflowOperation<Ret, Args extends any[]>(
     );
   }
 
+  // Guards the untyped caller: `waitForStage` is typed to only accept ACCEPTED, but a JavaScript
+  // caller (or an `as any` cast) can still reach here with another stage.
+  if (options?.waitForStage !== WorkflowUpdateStage.ACCEPTED) {
+    throw new nexus.HandlerError(
```
bergundy: This shouldn't be a HandlerError, those are for handlers to throw. Instead throw `TypeError`.

packages/nexus/src/index.ts
atol: Removed the export.

packages/nexus/src/workflow-helpers.ts
```diff
@@ -736,9 +747,18 @@ async function updateWorkflowOperation<Ret, Args extends any[]>(
...
 
+  // Guards the untyped caller: `waitForStage` is typed to only accept ACCEPTED, but a JavaScript
+  // caller (or an `as any` cast) can still reach here with another stage.
+  if (options?.waitForStage !== WorkflowUpdateStage.ACCEPTED) {
+    throw new TypeError(
+      `Only waitForStage '${WorkflowUpdateStage.ACCEPTED}' is supported, ` +
+        `got ${JSON.stringify(options?.waitForStage)}`
+    );
+  }
+
```
Sushisource, line 758: I'm not sure why we would do this... and this section demonstrates why. It's now possible to just break things if you go around the type system, in exchange for... what value? We can document ACCEPTED is used in the docstring.

bergundy: Traditionally we have relied on the type system. But I think there are exceptions. Not blocking AFAIC, deferring to current SDK maintainers.

bergundy (approved):
Approved but unsure whether @Sushisource's comment is blocking.

PR #2433, opened 2026-09-17
fix(create-project): parse tsconfig.json as JSONC to support comments and trailing commas

### LINKED ISSUE #2397: [Bug] create-project fails when tsconfig.json contains comments

Create the TypeScript AI SDK sample with the documented project initializer:

```sh
npx @temporalio/create@latest ./my-temporal-app
# Select: ai-sdk
```

### Describe the bug

`@temporalio/create@1.23.0` aborts before installing dependencies:

```text
Aborting installation.
Unexpected error. Please report it as a bug:
SyntaxError: Expected double-quoted property name in JSON at position 121 (line 6 column 5)
    at JSON.parse (<anonymous>)
    at updateNodeVersion (.../@temporalio/create/lib/helpers/install.js:43:39)
```

`updateNodeVersion()` discovers every `tsconfig.json` and parses each one with `JSON.parse()`. TypeScript configuration files use JSONC and may validly contain comments and trailing commas. The current `ai-sdk/tsconfig.json` contains a comment beginning at line 6, so strict JSON parsing fails.

The comment was introduced in temporalio/samples-typescript#497. Removing it locally allows installation to continue, but any commented TypeScript configuration can reproduce the underlying problem.

### Minimal reproduction

1. Run `npx @temporalio/create@latest ./my-temporal-app`.
2. Select the `ai-sdk` sample.
3. Observe the syntax error before `npm install` runs.

Equivalent direct reproduction:

```js
JSON.parse('{ // comment\n "extends": "@tsconfig/node22/tsconfig.json" }');
```

### Expected behavior

The initializer should accept TypeScript's JSONC syntax when reading `tsconfig.json`, including comments and trailing commas. Parsing of `package.json` should remain strict JSON.

### Proposed fix

Parse `tsconfig.json` as JSONC rather than strict JSON, and add a regression test covering comments
and a trailing comma.

Since `@temporalio/create` is published to npm, this means adding a runtime dependency to
`packages/create-project`. `jsonc-parser` looks like the right choice: it accepts exactly what `tsc`
accepts, whereas JSON5 is a looser superset that also allows unquoted keys, single-quoted strings
and `NaN`, none of which `tsc` will compile.

### Environment/Versions

- `@temporalio/create`: 1.23.0
- Sample: `ai-sdk`
- macOS

### Related history

- temporalio/samples-typescript#497 introduced the comment that exposes the problem.
- #1088 / #1089 added recursive discovery and updating of `tsconfig.json` files.
- temporalio/samples-typescript#251 / #1049 addressed missing `tsconfig.json` files, which is related code but a different failure.

OMD-123:
## Summary

This PR fixes issue #2397 where `@temporalio/create` fails when `tsconfig.json` contains comments or trailing commas.

## Problem

The `updateNodeVersion` function in `packages/create-project/src/helpers/install.ts` uses `JSON.parse()` to read `tsconfig.json` files. However, TypeScript configuration files use JSONC (JSON with Comments) format, which supports:
- Single-line comments (`// comment`)
- Multi-line comments (`/* comment */`)
- Trailing commas

When a sample project (like `ai-sdk`) has a `tsconfig.json` with comments, `JSON.parse()` throws a `SyntaxError`, causing the project initialization to fail before `npm install` runs.

## Solution

Replace `JSON.parse()` with `jsonc-parser.parse()` which correctly handles JSONC syntax. The `jsonc-parser` package is what `tsc` itself uses, so it accepts exactly what TypeScript accepts.

## Changes

1. **packages/create-project/package.json**: Added `jsonc-parser` dependency
2. **packages/create-project/src/helpers/install.ts**: Use `jsonc-parser.parse()` for `tsconfig.json` files
3. **packages/create-project/src/helpers/install.test.ts**: Added regression tests for:
   - Comments in tsconfig.json
   - Trailing commas in tsconfig.json

## Testing

- All existing tests pass
- New tests verify JSONC parsing works correctly
- Build succeeds
- Lint/format passes for create-project package

## Related

Closes #2397


chris-olszewski:
Addressed by #2398

PR #2444, opened 2026-09-21
ci: run two @temporalio/test files at a time on Node

DABH:
## What was changed

`scripts/ava-ci.ts` gains an `AVA_CONCURRENCY` override for ava's file concurrency, and the Node test step in CI sets it to 2 for the `@temporalio/test` group only. The package's own ava config stays at 1, so `pnpm test` locally and the Bun jobs are unchanged.

Stacked on #2443.

## Why

Once `@temporalio/test` runs alongside the other packages (#2443), it is the whole critical path of the test step, and its files mostly wait on the dev server and on timers. Measured at matching runner speed (compile step 148s vs 154s) with the parallel groups in place on both sides:

| job | one file at a time (#2443, run 35556668321) | two files (run 35556781438) |
|---|---|---|
| windows-x64 Node 20, `@temporalio/test` | 1088s | 777s |
| windows-x64 Node 20, whole test step | 1090s | 780s |
| linux-x64 Node 20, `@temporalio/test` | 859s | 649s |

On the Node jobs the suite passed 1100/1100 in every cell. On both Bun jobs it did not: Nexus operation cancellation tests timed out in `waitUntil` (`nexus-workflow-caller`, `nexus-temporal-operation`) and one `serialization-context` async completion returned `Not found`. So Bun stays at one file.

The earlier revision of this PR set `concurrency: 2` in `packages/test/package.json` for every runtime; its comment below compared the contended two-file number against the uncontended one-file baseline and drew the wrong conclusion, which is why it was briefly closed.

## Measured, and why this stays a draft

Two runs of this revision (35558471174, attempts 1 and 2), against #2443's runs at matching runner speed:

| job | #2443 tests step | this PR tests step | `@temporalio/test` alone, #2443 vs this |
|---|---|---|---|
| windows-x64 Node 20 | 1090s, 1102s | 613s (normal runner), 822s (slow runner) | 1088s vs 605s |
| linux-x64 Node 20 | 859s, 624s | 451s, 505s | 859s vs 450s |
| macos-arm Node 24 | 781s, 636s | 487s, 977s (3 failures) | 781s vs 485s |

The whole Windows Node 20 job came in at 13.5 minutes on a normal runner and 18.3 on a slow one, against 25 minutes for #2443 alone and 25.5 for the baseline on a good runner. So the saving is real and large.

The cost is contention flakes in timing-sensitive tests, and they are not Bun-only after all:

- `nexus-workflow-caller.cloud-unavailable` and `nexus-temporal-operation.cloud-unavailable` cancellation tests time out in `waitUntil`: on both Bun jobs in the first revision, and on macos-arm Node 24 in attempt 2 of this one (three tests, `@temporalio/test` 900s).
- `workflows › childAndNonCancellable` on windows-x64 Node 20, attempt 2: a Workflow Task hit the test's 400ms `Script execution timed out` sandbox deadline.
- `worker-deployment-versioning › ... ramping` on windows-x64 Node 20, attempt 1: `Routing config update still in progress` at `maxTimeout` (this one also flakes at one file at a time).

Those three would need their waits widened, or the Nexus cancellation files marked to run alone, before this is safe to merge. I have not made those changes here; this PR is the measurement and the list.


DABH:
Measured on run 35556781438 (this branch, so parallel groups plus two files at a time in `@temporalio/test`):

| job | `@temporalio/test` | baseline (one file at a time, run 35548300805) |
|---|---|---|
| windows-x64 Node 20 | 777s | 755s |
| linux-x64 Node 20 | 647s | 724s |
| linux-arm Bun | 595s, 5 failed | green |
| windows-x64 Bun | 628s, 2 failed | green |

The suite is mostly serial inside each file and shares one dev server, so a second file buys 11% on Linux and nothing on Windows, where the runner is already saturated by the parallel package group. The failures on both Bun jobs are Nexus operation cancellation tests timing out in `waitUntil` (`nexus-workflow-caller`, `nexus-temporal-operation`) plus one `serialization-context` async-completion `Not found`, which is what two files contending for the same dev server looks like. Not worth it. Closing; the numbers stay here for reference.

DABH:
Correction to the comment above: it compared the two-file suite under the parallel package group (777s) against the one-file suite with nothing else running (755s). The right comparison is against the one-file suite under the same parallel group, run 35556668321, where it took 1088s at the same runner speed. That is a 29% cut on Windows and 24% on Linux, so this is reworked as a Node-only setting; see the updated description.

DABH:
Closing: the saving is real but it trades the timeout for contention flakes in the Nexus cancellation tests, childAndNonCancellable and the ramping test. #2443 goes ahead on its own. The measurements stay here if we want to revisit after those tests get sturdier waits.

